PR #1, opened 2016-02-23
Improve SR startup: only run CheckVersion at startup

fernandog:
Proposed changes in this pull request:
- only run CheckVersion at startup


fernandog:
@labrys what do you think?

adaur:
What if you never restart SR? Would it ever update?

fernandog:
All threads will run but not T-0

If Trakt check is 1h, it sill run at T+1h and not T-0
We are only delaying then

PR #2, opened 2016-02-23
Make ep wanted only after show finish airs

fernandog:
image

```
2016-02-24 00:46:33 INFO     DAILYSEARCHER :: New episode American Crime Story - S01E04 - 100% Not Guilty airs today, setting to default episode status for this show: Wanted
2016-02-23 23:46:31 INFO     DAILYSEARCHER :: New episode Marvel's Agent Carter - S02E09 - A Little Song and Dance airs today, setting to default episode status for this show: Wanted
2016-02-23 23:46:31 INFO     DAILYSEARCHER :: New episode Marvel's Agent Carter - S02E08 - The Edge of Mystery airs today, setting to default episode status for this show: Wanted
2016-02-23 22:46:28 INFO     DAILYSEARCHER :: New episode The Flash (2014) - S02E15 - King Shark airs today, setting to default episode status for this show: Wanted
```

### DIFF
--- a/sickbeard/dailysearcher.py
+++ b/sickbeard/dailysearcher.py
@@ -28,6 +28,7 @@ from sickbeard import common
 from sickbeard import network_timezones
 from sickrage.show.Show import Show
 from sickrage.helper.exceptions import MultipleShowObjectsException
+from sickrage.helper.common import try_int
 
 
 class DailySearcher(object):  # pylint:disable=too-few-public-methods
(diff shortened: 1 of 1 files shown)


fernandog:
working as expected
@medariox @OmgImAlexis @labrys

labrys:
Approved!
I haven't tested in person but the logic looks good.  One thing that may need tested is on windows since dateutils handling of timezone is currently spotty but our work around is mostly solid and dateutil is nearly finished with its fixes.

labrys:
Also we might consider putting in a guard for an incorrect run time if we find it necessary.

fernandog:
we already do:
datetime.timedelta(minutes=try_int(show.runtime, 60))

if not runtime, use 60 instead

labrys:
I mean if run time = 600000 minutes for example.   Or run time = -300.  Those are still valid ints, but not really valid values.

fernandog:
I see. I think we can check if runtime <= 90/120 (like double episodes)

@duramato ideias ? what about talk shows or else?

duramato:
Just default to a runtine of 40 min since thats the average

PR #5, opened 2016-02-23
Avoid run 2 or more auto PP threads at the same time

fernandog:
if exiting thread is already running or too much files to PP or UNPACK taking too long

Use the same logic as DAILYSEARCHER
https://github.com/pyMedusa/SickRage/blob/develop/sickbeard/dailysearcher.py#L44-L45


fernandog:
to test this put a large RAR in PP folder that would take more than 10min

PR #9, opened 2016-02-24
Always write debug/db lines to log file

fernandog:
So when user submit an error, gist will always have DEBUG level even if user don't have sickbeard.DEBUG enabled

too much hassle to ask user to reproduce error again with debug enabled

Setting sickbeard.DEBUG will only prevent from show in the UI.
if user wants to see in UI. just need to enabled. Don't need restart

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -5241,8 +5241,6 @@ class ErrorLogs(WebRoot):
                 if match:
                     level = match.group(7)
                     logName = match.group(8)
-                    if not sickbeard.DEBUG and (level == 'DEBUG' or level == 'DB'):
-                        continue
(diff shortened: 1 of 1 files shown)


labrys:
@fernandog Can't include db logging... this would cause the logs to roll over too quickly and user wouldn't be able to post the issue.  Defaulting to debug being logged is not a bad idea, but there needs to be a way to disable it though as debug logging slows down python significantly.  I would be fine with having the disable a config only option (like db is now) so that by default debug logs are written until we have a more effective logging system and/or a more stable implementation that doesnt require so many debug logs.

fernandog:
There is a hidden config dbdebug that must be enabled. So no problem

labrys:
ok I was thinking that you were bypassing that to log that always here.

PR #10, opened 2016-02-24
Prevent logging "No results found" message

fernandog:
Not an actual error:
{"error":"No results found","error_code":20}

Prevents: https://gist.github.com/fernandog/cd67490ba247380ce4b5


medariox:
Approved!

PR #24, opened 2016-02-26
Angular

OmgImAlexis:
Proposed changes in this pull request:
- Remove `layout/main.mako`
- Add `index.mako` as default route, this will wrap all other routes and act as `layouts/main.mako`
- Add angular and the needed deps
- 404s now return `{ "status": 404 }` instead of rendering the 404.mako file
- `hometablesorter` can be added to any home layout table to have tablesorter load on it, e.g. `<table tablesorter ng-repeat="showList in showLists">`
- Redirects are changing from `return self.redirect('/home')` to 

    ```
    return {
        "status": 200,
        "redirect": {
            "ui-state": "home",
            "url": "/home"
        }
    }
    ```
    
    with `state` being the ui-state that needs to be loaded or `url` being the url it needs to redirect to. For external links it's best to use the `url` parameter and for internal links use ui-state as that means angular can handle it without needing to know the base-url.
- `anon-href="http://example.com"` should be used instead of `href="${anon_url('http://example.com')"`

##### TODO:
- [ ] Add config `POST` routes for saving
- [ ] Add inline indications when a field is changed (e.g. field gets tinted red)
- [ ] Add inline indications when an action is triggered (e.g. clicking save tints changed fields to green)
- [ ] Move all base routes to angular routes
- [ ] Convert routes to JSON and remove mako
    - [ ] Add meta values to index or new route
    - [ ] Convert templates to use angular
    - [ ] Add REST post methods to replace forms
- [ ] Move all menus to the main index def
- [ ] Table sorter
    - [ ] Fix `.nowrap` in tablesorter
    - [ ] Either change tablesorter into Angular service/directive or find a replacement 
- [ ] Remove unneeded whitespace
- [ ] Fix https://github.com/pymedusa/SickRage/issues/50


OmgImAlexis:
I need a Python dev to add in `/internal/FUNCTION_HERE` so we can have JSON for the angular routes. I need things like `myShowList` from home.mako as JSON so angular can bind it to the template.

p0psicles:
@OmgImAlexis can you describe the TODO comments, for how you want them as references for you in python, html/mako and js code?

For ex: for html:


js:
// @TODO: Angular

python: 
# @TODO: Angular

OmgImAlexis:
As long as it's a comment with `@TODO:` then the plugin I use for Atom will pick it up, same goes for `@NOTE:`.

Python's commenting is `#`, JS is `//` and HTML is `<!-- -->`.

p0psicles:
image

image

OmgImAlexis:
@labrys do you know how we can avoid needing a `json.loads` here? Some of the time it returns JSON and sometimes it's returns a string with JSON inside of it.

````
return json.loads(json.JSONEncoder().encode({
    "status": 200,
    "redirect": {
        "ui-state": "home",
        "url": "/home"
    }
}))
````

Hellowlol:
@OmgImAlexis all dicts could be returned as json with tornado so unless you have changed the default behaviour you should just pass it as dict.
````
return {
        "status": 200,
        "redirect": {
            "ui-state": "home",
            "url": "/home"
            }
        }
````

OmgImAlexis:
@Hellowlol that works a lot better, thanks for that.

p0psicles:
@OmgImAlexis i've been studying some angular. And l've looked at the changes you made so far. I'd like to help, but I think it's wise to have a chat first? What you think?

OmgImAlexis:
@p0psicles I'll try to get on IRC in about 10 hours from now.

Is there any reason for sorting it since the tablesorter plugin does it for us? If you want to look at the actual data seen by angular just goto the old route and it'll display JSON. I suggest if you're using Chrome to install [JSONView](https://chrome.google.com/webstore/detail/jsonview/chklaanhfefbnpoihckbnefhakgolnmc?hl=en) so it's easier to read but you should see something like [this](https://pastebin.com/raw/bFHUKYu3) notice the seasons are inside of an object meaning we can use `show.seasons[1].episodes` instead of having to use an array. This also means we don't end up with `seasons[0]` having nothing inside of it when there's no specials.

OmgImAlexis:
For some reason `$('.tablesorter').trigger('update');` needs to be called after the page is loaded for the tablesorter's filter to work, @Mottie any idea what's causing this?

Mottie:
If you are using the combined widgets file (`jquery.tablesorter.widgets.js`) then it might be because the UMD wrapper doesn't list tablesorter as a required module. There is an [open issue](https://github.com/Mottie/tablesorter/issues/855) about this with which I could use help.

And while I am happy you guys are using my fork of tablesorter, I recently found this [Angular UI Grid](https://github.com/angular-ui/ui-grid) package which might be a better fit for your project. It doesn't have any dependencies, and I think they did a really nice job on it.

OmgImAlexis:
@pymedusa/developers  Closing this as we're changing how the angular branch will be done.
I've discussed with a few members about converting page by page instead of a full redo like this as it became too big to handle.

PR #25, opened 2016-02-26
A number of fixes and changes for redirecting to manualSelect page and starting search, when no results.

p0psicles:
This should fix a number of issues like:
* Double spinner
* Anime shows with scene numbering no results found
* Doing manual searches for every manualSelect

TODO: Replace the page (jQuery.load(), when new results are found.


gui/slick/views/manualSelect.mako
```diff
@@ -166,56 +166,56 @@
...
-                    <td align="center">${provider_img.provider_type.title()}</td>
-                    <td align="center" class="col-status">Ignored</td>
-                    <td align="center" class="col-search" width="5%"><a class="epManualSearch" id="${str(show.indexerid)}x${season}x${episode}" name="${str(show.indexerid)}x${season}x${episode}" href="manualSnatchSelect?provider=${provider}&amp;rowid=${hItem["rowid"]}&show=${show.indexerid}&amp;season=${season}&amp;episode=${episode}"><img src="${srRoot}/images/download.png" width="16" height="16" alt="search" title="Download selected episode" /></a></td>
-                </tr>
-            % endfor
-        % endfor
-        </tbody>
-    </table>
-
+    <div id="container">
```
OmgImAlexis: These should all be 4 spaces, not tabs.

fernandog:
@p0psicles 
so if you manual search a show with 5 seasons, 10 episodes each. you will 50 episodes. if each episodes returns 5 results and have 10 providers you will have = 50*5*10 = 2500 results

we should fix the scene numbering properly
cache.db should have the correct season and episode (scene numbering)
can you confirm that cache.db has the correct scene numbering value or its the non-scene numbering?
If is wrong, problem is while adding to cache.db

@labrys

p0psicles:
@fernandog by default the url has been configured to use the season+ep query. So it's just an option, we can look into. Maybe in future, just to analyze why certain episodes are not found/snatches, you can use the option. Or maybe to download specials?

PR #44, opened 2016-02-27
Make URL column unique in providers table in cache database

fernandog:
Fix dupe items

Need erase cache.db or should do a ALTER TABLE do change existing cache.db to use UNIQUE?

@labrys @p0psicles @medariox @OmgImAlexis this PR fixed issue..... but we already have this in code but it seems its not working!

```python
            # add unique index to prevent further dupes from happening if one does not exist
            self.action("CREATE UNIQUE INDEX IF NOT EXISTS idx_url ON [" + providerName + "] (url)")
```

Seems code already remove dups. but only at boot:
```python
                 for cur_dupe in sql_results:
                     self.action("DELETE FROM [" + providerName + "] WHERE url = ?", [cur_dupe["url"]])
```

I know SQL but not an expert

### DIFF
--- a/sickbeard/tvcache.py
+++ b/sickbeard/tvcache.py
@@ -39,7 +39,7 @@ class CacheDBConnection(db.DBConnection):
         try:
             if not self.hasTable(providerName):
                 self.action(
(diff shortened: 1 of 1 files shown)


fernandog:
I think it worked. See first item: Before 161 seeds than 168

## First search
image

## After forced results 
image

fernandog:
will merge for now

PR #46, opened 2016-02-27
Fixes https://github.com/pymedusa/SickRage/issues/45

fernandog:
and revert last commit to tvcache.py

https://github.com/pymedusa/SickRage/issues/45


fernandog:
@p0psicles here it handles:

https://github.com/pymedusa/SickRage/blob/manual-search/sickrage/providers/GenericProvider.py#L458

PR #56, opened 2016-02-29
Limit subtitle results to search

medariox:
- Limits the subtitles that are being searched for on each run to 300 (can be adjusted).
- Subtitles are searched based on last time they were searched, meaning that newly added shows will always have precedence.
- Shows with air date <= 30 days, have a limit of 100 results (aka subtitles to search).
- Shows with air date > 30 days, have a limit of 200 results.

@pymedusa/owners @pymedusa/moderators 
What do you think about the limits? Any suggestions/feedback?


labrys:
:+1:  I think maybe increase the backlog limit slightly but we can start at these numbers and see how it does :)

labrys:
Approved!

PR #60, opened 2016-03-01
Upstream

labrys:
* miigotu	- Fix login issue with tvchaosUK, fixes need to double login  …	da35244
* miigotu	- Force mimetypes for commonly misconfigured types, ie: when a user has…  …	9efed22


labrys:
Approved!

PR #66, opened 2016-03-02
Added check on expired cookies for tvchaosuk.py

p0psicles:
Testing if we need this for snatching "old" expired sessions (read cookies), where authentication is required.

If this works, requests auth hook should be used, to create a generic solution.


p0psicles:
approved.
@duramato can you test this?

PR #69, opened 2016-03-02
[Manual Search] Remove unused variables

fernandog:
Remove unused variables not needed in: 
https://github.com/pyMedusa/SickRage/blob/manual-search/sickbeard/search.py#L94

@adaur @labrys @p0psicles can you please review with care this change?

correct way is checking with an IDE

### DIFF
--- a/sickbeard/search.py
+++ b/sickbeard/search.py
@@ -498,7 +498,7 @@ def searchProviders(show, episodes, manualSearch=False, downCurQuality=False, ma
         search_mode = curProvider.search_mode
 
         # Always search for episode when manually searching when in sponly
-        if search_mode == 'sponly' and (manualSearch is True or manualSelect is True):
+        if search_mode == 'sponly' and (manualSearch or manualSelect):
             search_mode = 'eponly'
 
         while True:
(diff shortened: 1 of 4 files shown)


sickbeard/search.py
```diff
@@ -498,7 +498,7 @@ def searchProviders(show, episodes, manualSearch=False, downCurQuality=False, ma
         search_mode = curProvider.search_mode
 
         # Always search for episode when manually searching when in sponly
-        if search_mode == 'sponly' and (manualSearch is True or manualSelect is True):
+        if search_mode == 'sponly' and (manualSearch or manualSelect):
```
p0psicles: Nothing really changed here? (manualSearch is True or manualSelect is True) == (manualSearch or manualSelect)

sickbeard/search_queue.py
```diff
@@ -154,9 +158,11 @@ def run(self):
 
                     # give the CPU a break
                     time.sleep(common.cpu_presets[sickbeard.CPU_PRESET])
-
+            # What is this for ?!
             generic_queue.QueueItem.finish(self)
```
fernandog: @labrys what this does? and why only this thread have it ?

p0psicles: I think i once added it to flag when A search was finished searching. But thats A long time ago.

PR #72, opened 2016-03-02
Proper fix for tntvillage

medariox:
@duramato works?


medariox:
@p0psicles 
While your idea is surely good and I agree this is the proper way to do it, checking the expiration date of the cookies is not always an acceptable solution. For example in the case of tntvillage, the cookies are set to expire after one year, but that is simply not the case and they will expire after a few days. That might be caused by a wrongly configured server, faulty software, etc. This might not be the case for Gazelle or TBDev based trackers (although they are often highly customized), under the assumption that the server is properly configured and the software functionality hasn't been compromised.

A quick look at tntvillage's cookies shows us this exact problem:
tnt-logged
These are the cookies while logged in. They show us an expiration of one year.

Here are the cookies after you get logged out (few days usually):
tnt-logged-out

As you can see we will need some kind of exceptions handling, for providers that are not so easy to handle by just checking the cookies validity.

p0psicles:
Then you could implement an exception for this prov using the custom_auth_check()

medariox:
@p0psicles 
Sure, just wanted to point out that checking the cookies validity might not always be the way to go. Did you check that this generic solution works for most trackers?

p0psicles:
No, i'm not saying I'll be able to solve all provider auth issues with one check. But this will give us much more flexibility for in the future. I do have to confess that this was @labrys idea. Haha. And the're might still be some issues with the way I implemented it tough.

PR #77, opened 2016-03-02
Added authentication hook to helpers.py

p0psicles:
This is purely a POC! Don't merge! Needs to be improved on.
I've put login_required param in GenericProvider.py. This should be Provider param. Need to work that out.


p0psicles:
@labrys Need to think of a way to get Provider params in get_auth_hook().

labrys:
@p0psicles this isn't how auth should work, will discuss with you in IRC when you get a chance.

PR #78, opened 2016-03-02
Providers

labrys:
* Fix rarbg rate limit
* Fix unknown qualities for tvcuk
 * appends `.hdtv.x264` to the title
* Fix WEB-DL and WEB-Rip detections
 * Optimize regex
 * Add WEB detection
 * Add DLMux detection to `tagger.py` using `in` operator instead of regex for optimization

### DIFF
--- a/sickbeard/providers/tvchaosuk.py
+++ b/sickbeard/providers/tvchaosuk.py
@@ -169,7 +169,7 @@ class TVChaosUKProvider(TorrentProvider):  # pylint: disable=too-many-instance-a
                             torrent_size = torrent.find_all('td')[labels.index('Size')].get_text(strip=True)
                             size = convert_size(torrent_size, units=units) or -1
 
-                            item = title, download_url, size, seeders, leechers
+                            item = title + '.hdtv.x264', download_url, size, seeders, leechers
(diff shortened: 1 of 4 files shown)


sickbeard/providers/tvchaosuk.py
```diff
@@ -169,7 +169,7 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
                             torrent_size = torrent.find_all('td')[labels.index('Size')].get_text(strip=True)
                             size = convert_size(torrent_size, units=units) or -1
 
-                            item = title, download_url, size, seeders, leechers
+                            item = title + '.hdtv.x264', download_url, size, seeders, leechers
```
fernandog: Can we assume that for all?

labrys: Per @duramato it's a safe assumption based on their labeling of releases.  If it's not an HD release, it won't have 720p or 1080p in the title so it won't parse as HD.  If it's BluRay or Web-dl then the HDTV will be ignored (as we test qualities from highest to lowest).  If its already labeled HDTV, double detection won't cause the check to fail.

PR #98, opened 2016-03-06
Update subliminal (4486c44) and legendastv

medariox:


fernandog:
@ratoaq2 we are in this new repo. we are not longer in "sickrage". almost all devs moved to this new org

ratoaq2:
Is there an IRC channel?

PR #103, opened 2016-03-06
Fix network timezone reporting

labrys:
* Make missing timezones reportable
* Make errors unique per network
* Suppress errors for network of None


labrys:
Approved!

PR #108, opened 2016-03-07
Add link in show title to go back to displayShow page

fernandog:
@OmgImAlexis 
should we don't change link color when a link is clicked?

### DIFF
--- a/gui/slick/views/manualSelect.mako
+++ b/gui/slick/views/manualSelect.mako
@@ -30,7 +30,7 @@
     <div class="clearfix"></div>
 
     <div id="showtitle" data-showname="${show.name}">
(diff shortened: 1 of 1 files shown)


OmgImAlexis:
@fernandog you can't just add a `href` attribute to any HTML element, you'd need to either wrap the `<h1>` in a `<a href=""></a>` or put it inside of the `<h1></h1>`. e.g.

````
 <h1 class="title" id="scene_exception_${show.indexerid}">
    <a href="${srRoot}/home/displayShow?show=${show.indexerid}">${show.name}</a>
</h1>
````

fernandog:
@OmgImAlexis what would be better? the title will become pink when clicked or with underscopes?

fernandog:
updated

PR #111, opened 2016-03-07
[Manual Search] Don't skip provider if table doesn't exist. Do a forced search

fernandog:
instead to create table, unique index and new columns seeders, leechers and size

This will fix the issue when fresh cache.db doesn't have results and it will still do a forced search instead of skip provider @medariox


p0psicles:
Please test this one first, and then check if this PR is still needed: https://github.com/pymedusa/SickRage/pull/112

PR #114, opened 2016-03-07
About time

labrys:
* Fix first aired hack
 * was preventing finding some shows on tvdb
* Speed up initial start
 * delay creating name cache until after browser starts
* Optimize webserve imports
* Fix broad exception for add show
 * use indexer_exception instead (may need additional excepts added)


labrys:
Approved!

PR #117, opened 2016-03-08
[Manual Search] Fix srroot for when reverse proxy is used

p0psicles:


p0psicles:
It is showing the commits for PR113.
Probably will need rebased after rebasing and merging of PR113

PR #118, opened 2016-03-08
Add CookieAuth

labrys:
* TODO:
 - [ ] Make login params update with user/pass changes
 - [ ] See if we need to validate values of pass_hash and member_id for TNTvillage
 - [ ] Clean up custom exceptions

### DIFF
--- a/sickbeard/providers/abnormal.py
+++ b/sickbeard/providers/abnormal.py
@@ -21,13 +21,14 @@
 from __future__ import unicode_literals
 
 import re
+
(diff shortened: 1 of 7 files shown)


labrys:
@p0psicles This is an example of a cookie based authentication implemented for TNTvillage
it's a blind commit, so needs checking before merging.

labrys:
don't have the time to work on this atm, will revisit in the future. closing but leaving branch as a reminder.

PR #131, opened 2016-03-10
Added some missing show attributes to displayShow json return

p0psicles:
I also had changed the html, to fix some of the errors, but I lost the changes while playing around.
But these webserve changes were needed.


OmgImAlexis:
Approved!

PR #135, opened 2016-03-10
Fix Trakt adding shows without year

fernandog:
thus creating another show folder. So user will have a new "Show1" while the existing "Show1 (2016)"

This happens because cache name creation runs (used to) together with trakt thread.
anyway I already fixed trakt thread delaying 5min after boot.
https://github.com/pymedusa/SickRage/commit/89f02fb23908937c51384e4af15a35298f1f2d12

but it could happen again when cache name creation fails


fernandog:
I could reproduce the error by applying this again: https://github.com/pymedusa/SickRage/pull/114/files

Logs:
https://gist.github.com/fernandog/456271f888c629613771

PR #146, opened 2016-03-15
Fix for the double season scene exception icons in displayShow.

p0psicles:
Fixes https://github.com/pymedusa/SickRage/issues/133

Rewritten the loops using jQuery each. Much better readability.


p0psicles:
@OmgImAlexis i'll need your approval on this one. If it is not to make dev before angular, that's also okay with me.

gui/slick/js/core.js
```diff
@@ -2453,37 +2453,38 @@ var SICKRAGE = {
...
-                                        'height': '16',
-                                        'width': '16',
-                                        'src': srRoot + '/images/xem.png',
-                                        'title': data.seasonExceptions[data.xemNumbering[indexerSeason]].join(', '),
-                                    }).appendTo('[data-season=' + indexerSeason + ']');
-                                }
+                $.each(data.seasonExceptions, function(season, nameExceptions) {
+                var foundInXem = false;
+                // Check if it is a season name exception, we don't handle the show name exceptions here
+                    if (season >= 1) {
```
OmgImAlexis: What about season 0?

p0psicles: I think you'r talking about the  "specials" ? Do we have a show that uses a season scene name exception for a special?

OmgImAlexis: I don't so but it'd be a good idea to add them if possible, some shows use stuff like "Behind the scenes Show xx - TITLE" as the name for the specials so this could mean we may actually be able to add in auto downloading of some shows. No?

p0psicles:
Yeah sure, I also can't think of a situation where it can hurt to add the 0 now. As the -1 if for show exception and everything from 0 would be season.

p0psicles:
Let's wait for the manual snatch merge, then kick this into dev, okay @OmgImAlexis ?

OmgImAlexis:
Sounds good.

p0psicles:
Fixed: https://github.com/pymedusa/SickRage/issues/133

PR #165, opened 2016-03-18
Fix logging

labrys:


sickbeard/properFinder.py
```diff
@@ -205,22 +220,22 @@ def _getProperList(self):  # pylint: disable=too-many-locals, too-many-branches,
                 old_release_group = (sql_results[0]['release_group'])
 
                 if -1 < old_version < bestResult.version:
-                    logger.log(u'Found new anime version {} to replace existing version {}'.format
-                               (str(bestResult.version), str(old_version)))
+                    logger.log(u'Found new anime version {new} to replace existing version {new}: {name}'.format
```
medariox: '''Found new anime version {new} to replace existing version {new}: {name}'''
One needs to be {old}

labrys: D'oh!

PR #172, opened 2016-03-18
Move some Manual Snatch code from mako to python

fernandog:
@OmgImAlexis you are welcome! =)


OmgImAlexis:
Good idea, thank you!

PR #174, opened 2016-03-19
Fixed manual searching for scene named anime shows.

p0psicles:
Somehow results = curProvider.cache.update_cache_manual_search(searchResults[curEp]) was always running with curEp= -1. As it should run the function with the episode number as a param.

With this fix it was getting all results for show Fairy Tails S2, for providers nyaatorrent and tokyotosho. As before it only found results from nzb providers.

We'll need to thoroughly test this will all different kinds of shows. I can also image that for some shows, you will get allot more results now.
@fernandog @labrys @duramato

### DIFF
--- a/sickbeard/search.py
+++ b/sickbeard/search.py
@@ -173,7 +173,8 @@ def snatchEpisode(result, endStatus=SNATCHED):  # pylint: disable=too-many-branc
 
         if curEpObj.status not in Quality.DOWNLOADED:
             try:
-                notifiers.notify_snatch("{} from {}".format(curEpObj._format_pattern('%SN - %Sx%0E - %EN - %QN'), result.provider.name))  # pylint: disable=protected-access
+                notifiers.notify_snatch("{} from {}".format(curEpObj._format_pattern('%SN - %Sx%0E - %EN - %QN'),
+                                                            result.provider.name))  # pylint: disable=protected-access
             except Exception:
                 # Without this, when notification fail, it crashes the snatch thread and SR will
                 # keep snatching until notification is sent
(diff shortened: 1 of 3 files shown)


sickbeard/search.py
```diff
@@ -565,7 +565,10 @@ def searchProviders(show, episodes, manualSearch=False, downCurQuality=False, ma
 
         # Update the cache if a manual search is being runned
         if manual_snatch:
-            results = curProvider.cache.update_cache_manual_search(searchResults[curEp])
+            results = False
```
labrys: Not needed

sickbeard/search.py
```diff
@@ -565,7 +565,10 @@ def searchProviders(show, episodes, manualSearch=False, downCurQuality=False, ma
...
+            for searched_episode in episodes:
+                results = curProvider.cache.update_cache_manual_search(searchResults[searched_episode.episode])
```
labrys: Should be a comprehension:
```python
results = [curProvider.cache.update_cache_manual_search(searchResults[searched_episode.episode]) for searched_episode in episodes]
```

sickbeard/search.py
```diff
@@ -565,7 +565,10 @@ def searchProviders(show, episodes, manualSearch=False, downCurQuality=False, ma
...
+
             if results:
                 # If we have at least a result from one provider, it's good enough to be marked as result
                 finalResults.append(results)
```
labrys: The existing code only appends the result for the last found.
With the change of results to a list comprehension this should become a list concat.  In addition the concat should be moved outside the `if` statement which will only be used to continue.

p0psicles: As discussed we're collecting a list of bools here? With what purpose? In line 732 is picked up again with finalResults += set(multiResults.values()). So then results are added to it?

I'll look further into this, if we can agree the solution for the caching is the correct one. I haven't really looked into the finalResults, so will do that first.

p0psicles:
@labrys can you approve this?

fernandog:
@p0psicles 
I gave one of the approves

sickbeard/search.py
```diff
@@ -470,6 +470,7 @@ def searchProviders(show, episodes, manualSearch=False, downCurQuality=False, ma
     """
     foundResults = {}
     finalResults = []
+    manual_snatch_results = []
```
fernandog: @p0psicles I remember I added this. maybe in one of the rebases it got lost

PR #179, opened 2016-03-20
Fix bs4 find_all

labrys:
Fix upstream typo in `beautifulsoup.find_all()` and convert all `find_all()` to the simplified form.


labrys:
Approved! Merging without required status - Coveralls is being a pita again.

PR #180, opened 2016-03-20
Using release name when searching for subtitles

ratoaq2:
This will add a new subliminal refiner that will make use of the original release name to "enrich" the video object information before searching for subtitles. That will allow users to choose whatever "renaming rules" they want without worrying about downloading the correct subtitles. It should increase the subtitle search success ratio as well, since the original release name might contain more info then the renamed video file.

I tested this on my local installation. Please, let me know if anything needs to be changed or someone wants to try it first.


medariox:
Approved

This looks really good, great job!

PR #194, opened 2016-03-22
Better search width and align

medariox:
After the change:
image

@pymedusa/contributors 
Looks good?

Some downsides of left alignment:
image

### DIFF
--- a/gui/slick/css/style.css
+++ b/gui/slick/css/style.css
@@ -1116,10 +1116,6 @@ span.snatched b {
     text-align: left !important;
 }
 
-.displayShowTable td.col-search {
-    text-align: center;
-}
-
 .sickbeardTable {
     table-layout: auto;
     width: 100%;
@@ -1226,7 +1222,7 @@ td.col-legend {
 
 th.col-search,
 td.col-search {
-  width: 46px;
+  width: 65px;
 }
 
 .showLegend {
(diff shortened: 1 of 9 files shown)


fernandog:
@medariox 
i vote por center alignment.
if you hide some columns it gets really ugly the left alignment

medariox:
I don't love the left alignment either. Both alignments have up- and downsides.

labrys:
Personally i think centered looks terrible.  I think it should be 3 sub columns within search so that icons always appear in same column.  Also if forced search icons are hidden, why aren't selective also hidden?  lastly i think that CC should be moved to the left side so it appears closer to the subtitles and so that the selective search arrow appears to point "off the page" instead of at the cc.

fernandog:
manual search/snatch shouldn't be hidden when status is skipped

labrys:
imo if one method of file search is present the other should also be present.

OmgImAlexis:
Why not just use `display:none;` or something similar and keep the icon there but just hide it from the user? That way they always stay in their spots and we don't need to add any extra columns.

labrys:
@OmgImAlexis thats what i was thinking but my sleep-deprived mind couldnt quite dredge it out :)

OmgImAlexis:
Also that icon may need to be changed as it is right now that looks like you're telling people that the search icon will do something called "CC" since it's pointing from one icon to the other.. :disappointed:

medariox:
I like the idea of just hiding the image and also agree that we need a better "manual snatch" icon. This PR was aiming to be a quick fix for the current ugliness, further improvements will need a new PR. Going to merge the wider search column without left alignment for now, hope that's okay.

PR #195, opened 2016-03-22
Add seeders and leechers information to notifiers

fernandog:
- [x] Create a global setting to enable/disable (hidden for now, no UI changes)
- [x] Change log messages to ".format"

### DIFF
--- a/sickbeard/search.py
+++ b/sickbeard/search.py
@@ -173,10 +173,15 @@ def snatchEpisode(result, endStatus=SNATCHED):  # pylint: disable=too-many-branc
 
         if curEpObj.status not in Quality.DOWNLOADED:
             try:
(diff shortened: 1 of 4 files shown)


sickbeard/search.py
```diff
@@ -173,7 +173,7 @@ def snatchEpisode(result, endStatus=SNATCHED):  # pylint: disable=too-many-branc
 
         if curEpObj.status not in Quality.DOWNLOADED:
             try:
-                if result.seeders is not -1 and result.leechers is not -1:
+                if all([sickbeard.SEEDERS_LEECHERS_IN_NOTIFY, result.seeders is not -1, result.leechers is not -1]):
```
labrys: what about if seeders or leechers is None

OmgImAlexis:
Approved!

PR #196, opened 2016-03-22
Fix babelfish bug that provented modern greek to be recognized as opensubtitles code

medariox:
Needs to be reported upstream.


medariox:
@pymedusa/contributors 

Need approval from a gui-dev.

PR #207, opened 2016-03-24
Migrated % string formating

fernandog:


sickbeard/postProcessor.py
```diff
@@ -1029,8 +1029,7 @@ def process(self):  # pylint: disable=too-many-return-statements, too-many-local
 
                 # If the file season (ep_obj.season) is bigger than the indexer season (max_season[0][0]), skip the file
                 if int(ep_obj.season) > int(max_season[0][0]):
-                    self._log(u"File has season %s, while the indexer is on season %s. The file may be incorrectly labeled or fake, aborting."
-                              % (str(ep_obj.season), str(max_season[0][0])))
+                    self._log(u"File has season {0!s}, while the indexer is on season {1!s}. The file may be incorrectly labeled or fake, aborting.".format(str(ep_obj.season), str(max_season[0][0])))
```
labrys: This is part of what I don't like about the automated fix.  All these `!s` format specifiers.  They should only be used in places like here to avoid the forced typecast to `str`
```pyton
"File has season {0!s}, while the indexer is on season {1!s}".format(
   ep_obj.season, max_season[0][0]
)
```

labrys:
What this type of auto-pr from QC is good for is to quickly highlight all the places where this type of change needs to be made so you can manually go through each one and make sure its done right.

OmgImAlexis:
Approved!

fernandog:
@labrys changed the {0!s} to {0}

sickbeard/browser.py
```diff
@@ -115,7 +115,7 @@ def foldersAtPath(path, includeParent=False, includeFiles=False):
     try:
         file_list = getFileList(path, includeFiles)
     except OSError as e:
-        logger.log('Unable to open %s: %s / %s' % (path, repr(e), str(e)), logger.WARNING)
+        logger.log('Unable to open {0}: {1} / {2}'.format(path, repr(e), str(e)), logger.WARNING)
```
labrys: Just removing all the `!s` isn't the correct fix either.  Lines like this should be
```python
logger.log('Unable to open {0}: {1!r} / {1!s}'.format(path, e), logger.WARNING)
```

sickbeard/helpers.py
```diff
@@ -962,7 +958,7 @@ def anon_url(*url):
     """
     Return a URL string consisting of the Anonymous redirect URL and an arbitrary number of values appended.
     """
-    return '' if None in url else '%s%s' % (sickbeard.ANON_REDIRECT, ''.join(str(s) for s in url))
+    return '' if None in url else '{0}{1}'.format(sickbeard.ANON_REDIRECT, ''.join(str(s) for s in url))
```
labrys: Lines like this though you can't do the `!s` change to eliminate the `str` or it will fail since it has to be a string to `.join`

sickbeard/helpers.py
```diff
@@ -1059,7 +1055,7 @@ def get_show(name, tryIndexers=False):
         if showObj and not fromCache:
             sickbeard.name_cache.addNameToCache(name, showObj.indexerid)
     except Exception as e:
-        logger.log(u"Error when attempting to find show: %s in SickRage. Error: %r " % (name, repr(e)), logger.DEBUG)
+        logger.log(u"Error when attempting to find show: {0} in SickRage. Error: {1!r} ".format(name, repr(e)), logger.DEBUG)
```
labrys: Lines like this have `!r` and `repr()` which is wrong.

sickbeard/webapi.py
```diff
@@ -420,12 +419,11 @@ def _check_param_type(self, value, name, arg_type):
...
             pass
         else:
-            logger.log(u'API :: Invalid param type: "%s" can not be checked. Ignoring it.' % str(arg_type), logger.ERROR)
+            logger.log(u'API :: Invalid param type: "{0}" can not be checked. Ignoring it.'.format(str(arg_type)), logger.ERROR)
 
         if error:
             # this is a real ApiError !!
-            raise ApiError(u'param "%s" with given value "%s" could not be parsed into "%s"'
-                           % (str(name), str(value), str(arg_type)))
+            raise ApiError(u'param "{0}" with given value "{1}" could not be parsed into "{2}"'.format(str(name), str(value), str(arg_type)))
```
labrys: All of these types of calls should be `!s` and the `str` calls should go away.

fernandog:
don't have time now to go over 100 files. Sorry. wont do it now. maybe when manual search and everything else is done

PR #215, opened 2016-03-25
Fix: We don't need to check for updates if we need the current installed version

p0psicles:

This will speedup loading the /config page allot!


fernandog:
Why not disabled "check for updates" setting?

p0psicles:
Because that option has nothing to do with showing the version.
This is what's causing the slow page load:
image

fernandog:
Approved

OmgImAlexis:
Approved!

PR #218, opened 2016-03-26
[Manual Search] Add provider setting 'Enable provider for Manual Searches'

fernandog:
It will only use "manual search" providers when you use "Force results" button. Because this providers don't get updated in DailySearcher as they are manual search only

So its a mix of @duramato and mine (fallback)

- Fixes thread name while doing forced search
Before:
 [c352080] Performing episode search for 
After:
[MoreThanTV] :: [c352080] Performing episode search for 

- Fixed "forced search when no results" using backlog enabled providers (in manual_snatch.py we used to use daily). Now we use "enable_manualsearch"

- Fixed: search for propers only in backlog enabled providers

- Initialize provider even if no query to execute

This changes UI but nothing that will break angular @OmgImAlexis

### DIFF
--- a/gui/slick/views/config_providers.mako
+++ b/gui/slick/views/config_providers.mako
@@ -168,6 +168,18 @@ $('#config-components').tabs();
                         </div>
                         % endif
 
+                        % if hasattr(curNewznabProvider, 'enable_manualsearch'):
+                        <div class="field-pair">
+                            <label for="${curNewznabProvider.get_id()}_enable_manualsearch">
+                                <span class="component-title">Enable provider for manual searches</span>
+                                <span class="component-desc">
+                                    <input type="checkbox" name="${curNewznabProvider.get_id()}_enable_manualsearch" id="${curNewznabProvider.get_id()}_enable_manualsearch" ${('', 'checked="checked"')[bool(curNewznabProvider.enable_manualsearch)]}/>
+                                    <p>enable provider to use in manual searches.</p>
+                                </span>
+                            </label>
+                        </div>
+                        % endif
+
                         % if hasattr(curNewznabProvider, 'enable_backlog'):
                         <div class="field-pair${(' hidden', '')[curNewznabProvider.supports_backlog]}">
                             <label for="${curNewznabProvider.get_id()}_enable_backlog">
@@ -254,6 +266,19 @@ $('#config-components').tabs();
                             </label>
                         </div>
                         % endif
+
+                        % if hasattr(curNzbProvider, 'enable_manualsearch'):
+                        <div class="field-pair">
+                            <label for="${curNzbProvider.get_id()}_enable_manualsearch">
+                                <span class="component-title">Enable provider for manual searches</span>
+                                <span class="component-desc">
+                                    <input type="checkbox" name="${curNzbProvider.get_id()}_enable_manualsearch" id="${curNzbProvider.get_id()}_enable_manualsearch" ${('', 'checked="checked"')[bool(curNzbProvider.enable_manualsearch)]}/>
+                                    <p>enable provider to use in manual searches.</p>
+                                </span>
+                            </label>
+                        </div>
+                        % endif
+
 
                         % if hasattr(curNzbProvider, 'enable_backlog'):
                         <div class="field-pair${(' hidden', '')[curNzbProvider.supports_backlog]}">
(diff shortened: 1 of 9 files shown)


OmgImAlexis:
Approved!

[Approved with PullApprove](https://pullapprove.com/pymedusa/SickRage/pull-request/218/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=pymedusa/SickRage)

fernandog:
@p0psycles @medariox @labrys can you test it too please?

sickbeard/webserve.py
```diff
@@ -5006,6 +5019,13 @@ def saveProviders(self, newznab_string='', torrentrss_string='', provider_order=
                 except Exception:
                     curNzbProvider.enable_daily = 0  # these exceptions are actually catching unselected checkboxes
 
+            if hasattr(curNzbProvider, 'enable_manualsearch'):
+                try:
+                    curNzbProvider.enable_manualsearch = config.checkbox_to_value(
+                        kwargs[curNzbProvider.get_id() + '_enable_manualsearch'])
+                except Exception:
```
labrys: These calls should be changed.  First of all caching `Exception` is generally a bad idea.  Secondly if you're already using a `try: ... except:` then catch `AttributeError` instead of checking with `hasattr`

sickbeard/search.py
```diff
@@ -487,8 +487,14 @@ def searchProviders(show, episodes, manualSearch=False, downCurQuality=False, ma
...
-    providers = [x for x in sickbeard.providers.sortedProviderList(sickbeard.RANDOMIZE_PROVIDERS)
-                 if x.is_active() and x.enable_backlog]
+    if manual_snatch:
+        logger.log("Using manual search providers")
+        providers = [x for x in sickbeard.providers.sortedProviderList(sickbeard.RANDOMIZE_PROVIDERS)
+                    if x.is_active() and x.enable_manualsearch]
+    else:
+        providers = [x for x in sickbeard.providers.sortedProviderList(sickbeard.RANDOMIZE_PROVIDERS)
+                    if x.is_active() and x.enable_backlog]
+
```
labrys: This section can be simplified and be made more useful with a simple change.
Add a method `.enabled_for(search_type)`
Then you can use the original code nearly unchanged:
```python
providers = [x for x in sickbeard.providers.sortedProviderList(sickbeard.RANDOMIZE_PROVIDERS)
                    if x.is_active() and x.enabled_for(cur_search_type)]
```
This has the advantage of always using the correct providers for a search type, regardless of how many search types you add.

sickbeard/manual_snatch.py
```diff
@@ -170,9 +171,11 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
     sql_return = []
     found_items = []
     provider_results = {'last_prov_updates': {}, 'error': {}, 'found_items': []}
+    original_thread_name = threading.currentThread().name
 
-    providers = [x for x in sickbeard.providers.sortedProviderList(sickbeard.RANDOMIZE_PROVIDERS) if x.is_active() and x.enable_daily]
+    providers = [x for x in sickbeard.providers.sortedProviderList(sickbeard.RANDOMIZE_PROVIDERS) if x.is_active() and x.enable_manualsearch]
     for curProvider in providers:
```
labrys: Fix camelCase to snake_case

sickbeard/manual_snatch.py
```diff
@@ -170,9 +171,11 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
...
+        threading.currentThread().name = original_thread_name + " :: [" + curProvider.name + "]"
```
labrys: `'{thread} :: [{provider}]'.format(thread=original_thread_name, provider=curProvider.name)`

fernandog:
@labrys the purpose of this PR its not rewrite code.

sickbeard/manual_snatch.py
fernandog: when we rewrite stuff we batch change this stuff to make consistant in all code

gui/slick/views/config_providers.mako
```diff
@@ -168,6 +168,18 @@ $('#config-components').tabs();
                         </div>
                         % endif
 
+                        % if hasattr(curNewznabProvider, 'enable_manualsearch'):
+                        <div class="field-pair">
+                            <label for="${curNewznabProvider.get_id()}_enable_manualsearch">
+                                <span class="component-title">Enable provider for manual searches</span>
+                                <span class="component-desc">
+                                    <input type="checkbox" name="${curNewznabProvider.get_id()}_enable_manualsearch" id="${curNewznabProvider.get_id()}_enable_manualsearch" ${('', 'checked="checked"')[bool(curNewznabProvider.enable_manualsearch)]}/>
```
labrys: These lines are absurdly long. And some of the method calls are being unneccessarily repeatedly (specifically `.get_id()`).  
If you add a single `<% cur_id = curNewznabProvider.get_id() %>` right after the for loop.  Then all the `curNewznabProvider.get_id()` can be replaced by `cur_id`.
This line then gets shortened to
`<input type="checkbox" name="${cur_id}_enable_manualsearch" id="${cur_id}_enable_manualsearch" ${('', 'checked="checked"')[bool(curNewznabProvider.enable_manualsearch)]}/>`

The second have can also be rewriiten (and more clearly) like this
`{'checked="checked"' if curNewznabProvider.enable_manualsearch else ''}`

So the final line would be
`<input type="checkbox" name="${cur_id}_enable_manualsearch" id="${cur_id}_enable_manualsearch" ${'checked="checked"' if curNewznabProvider.enable_manualsearch else ''}/>`

fernandog: again: a lot of big changes for this PR. maybe in the next one. Also dont want to change mako much

sickbeard/manual_snatch.py
labrys: Except if as you stated [here](https://github.com/pymedusa/SickRage/pull/207#issuecomment-201911631) "don't have time now to go over 100 files. Sorry. wont do it now. maybe when manual search and everything else is done".  Doing this when you are working on a section breaks this up into smaller jobs so it all doesn't have to be done at once.  Also as there is no consistency currently, and this is the consistent standard we are going for, there is no reason that this shouldnt be done.  Not at the start of the PR, but when getting it ready for merge.

fernandog:
please rewrite code so I can add features.

PR #222, opened 2016-03-27
[Manual Search] Add provider setting 'Enable manual searches'

fernandog:
It will only use "manual search" providers when you use "Force results" button. Because this providers don't get updated in DailySearcher as they are manual search only

So its a mix of @duramato and mine (fallback)

- Fixes thread name while doing forced search
Before:
 [c352080] Performing episode search for 
After:
[MoreThanTV] :: [c352080] Performing episode search for 

- Fixed "forced search when no results" using backlog enabled providers (in manual_snatch.py we used to use daily). Now we use "enable_manualsearch"

- Fixed: search for propers only in backlog enabled providers

- Initialize provider even if no query to execute

This changes UI but nothing that will break angular @OmgImAlexis

### DIFF
--- a/gui/slick/views/config_providers.mako
+++ b/gui/slick/views/config_providers.mako
@@ -168,6 +168,18 @@ $('#config-components').tabs();
                         </div>
                         % endif
 
+                        % if hasattr(curNewznabProvider, 'enable_manualsearch'):
+                        <div class="field-pair${(' hidden', '')[curNewznabProvider.supports_backlog]}">
+                            <label for="${curNewznabProvider.get_id()}_enable_manualsearch">
+                                <span class="component-title">Enable manual searches</span>
+                                <span class="component-desc">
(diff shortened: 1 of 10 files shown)


fernandog:
@labrys wont touch mako or webserve for now.

I don't want to break webserve checkbox. Want to test my code properly. Testing when changing too much stuff makes harder to debug

If this PR works ok then changes the try:except in webserve

labrys:
@fernandog testing in process

fernandog:
@labrys can you please commit your suggestion about loop providers  passing the type of active provider?

fernandog:
Maybe add to helpers this function?

labrys:
@fernandog sounds like a good idea, ill look at doing that later today.

fernandog:
@labrys please review. finished

labrys:
Approved

PR #233, opened 2016-03-29
Handle when hash or size are missing

duramato:

### DIFF
--- a/sickbeard/providers/torrentproject.py
+++ b/sickbeard/providers/torrentproject.py
@@ -91,16 +91,16 @@ class TorrentProjectProvider(TorrentProvider):  # pylint: disable=too-many-insta
 
                 results = []
                 for i in torrents:
-                    title = torrents[i]["title"]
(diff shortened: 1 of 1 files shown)


sickbeard/providers/torrentproject.py
```diff
@@ -101,6 +101,8 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
 
                     t_hash = torrents[i]["torrent_hash"]
                     torrent_size = torrents[i]["torrent_size"]
+                    if not all([t_hash, torrent_size]):
+                        continue
                     size = convert_size(torrent_size) or -1
```
fernandog: So are you discarding result because of that?
When no size it should set as -1
Maybe when no hash set as None

duramato: If there is no hash you cant download it.

medariox: But you can if there is no size, right? Right now it will skip it if there is a hash, but no torrent_size.

PR #235, opened 2016-03-29
Strip() in pre-script result

fernandog:
Log with AA (empty line):

```
2016-03-29 18:47:53 DEBUG    FINDSUBTITLES :: [fcb0d34] Could not search in legendastv provider. Discarding for now
AA
2016-03-29 18:47:53 DEBUG    FINDSUBTITLES :: [fcb0d34] Script result: Nothing to rename
2016-03-29 18:47:53 INFO     FINDSUBTITLES :: [fcb0d34] Executing command: ['python', '/home/osmc/scripts/rename_releases.py', '/media/SAMSUNG/media/downloaded/series/Show.S07E18.1080p.HDTV.X264-DIMENSION.mkv']
2016-03-29 18:47:53 INFO     FINDSUBTITLES :: [fcb0d34] Running subtitle pre-script: python /home/osmc/scripts/rename_releases.py
```

```
>>> import subprocess
>>> inner_cmd = ['python', '/home/osmc/scripts/rename_releases.py', '/media/SAMSUNG/media/downloaded/series/Show.S04E04.1080p.HDTV.x264-BRISK.mkv']
>>> process = subprocess.Popen(inner_cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
...                                        stderr=subprocess.STDOUT)
>>> out, _ = process.communicate()
>>> print out
Nothing to rename

>>> print out.strip()
Nothing to rename
>>>
```

My pre-script:
```
# coding=utf-8
#!/usr/bin/python

import os
import sys

filename = sys.argv[1]
if filename.rpartition('.')[2] in ('mkv', 'mp4', 'avi'):
    new_filename = filename.replace('VietHD', 'RARBG').replace('DRACULA','RARBG')
    if new_filename != filename:
        print("Renaming: {} para {}".format(filename, new_filename))
        os.rename(filename, new_filename)
    else:
        print("Nothing to rename")

```

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -884,7 +884,7 @@ def run_subs_scripts(scripts, script_args):
             process = subprocess.Popen(inner_cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
                                        stderr=subprocess.STDOUT, cwd=sickbeard.PROG_DIR)
             out, _ = process.communicate()  # @UnusedVariable
-            logger.log(u'Script result: {0}'.format(out), logger.DEBUG)
+            logger.log(u'Script result: {0}'.format(out.rstrip()), logger.DEBUG)
 
         except Exception as error:
             logger.log(u'Unable to run subtitles script: {0}'.format(ex(error)))


fernandog:
@labrys do you know why white space exist?

ratoaq2:
Or we could move away from communicate and use check_output, like suggested by your link [here](http://stackoverflow.com/questions/15374211/why-does-popen-communicate-return-bhi-n-instead-of-hi#comment21727184_15374211)

[Here](https://github.com/ratoaq2/library-scripts/blob/master/couchpotato/externalsubliminal/main.py#L58) is an example on how I use it on one of my scripts.

fernandog:
@ratoaq2 

>>> from subprocess import check_output, STDOUT
>>> output = check_output(inner_cmd, stderr=STDOUT)
>>> print output
Nothing to rename

>>>

medariox:
As I already tried to explain in IRC, there is nothing wrong with communicate().
The problem here is in the custom script. Is uses print, which aromatically adds a newline at the end.
If you want to avoid that, you can import 
```from __future__ import print_function```
and set an alternative end
```print('Nothing to rename', end='')```
in the custom script.
We shouldn't alter the output with strip() and I would say this isn't a bug at all.

fernandog:
@medariox so all users that will use the pre/extra scripts would need to do that? makes no sense to me

medariox:
@fernandog 
No, it's just for users that want to use print() and it's literally one import to add to the script.
What if I want a new line at the end of my script? I couldn't do it, because it will strip it every time.

PR #239, opened 2016-03-30
Replace messy disk_usage with existing method

medariox:
I extended the disk_usage method to allow optional output in Bytes instead of "human readable format". This allows us to use it instead of that mess we were using before.
As we use getDiskSpaceUsage everywhere else, I hope this won't cause issues with other OS - it would still be very nice if we could test it on different systems before merge, I tested it under Ubuntu 15 for now. Also fixes this bug: https://gist.github.com/medariox/cb9f1c9995ca15ee076888fea38a7abb

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -1574,38 +1574,12 @@ def verify_freespace(src, dest, oldfile=None):
 
     logger.log(u"Trying to determine free space on destination drive", logger.DEBUG)
 
-    if platform.system() == 'Windows':
-        import sys
-
-        def disk_usage(path):
-            _, total, free = ctypes.c_ulonglong(), ctypes.c_ulonglong(), ctypes.c_ulonglong()
-            if sys.version_info >= (3,) or isinstance(path, unicode):
-                method = ctypes.windll.kernel32.GetDiskFreeSpaceExW
-            else:
-                method = ctypes.windll.kernel32.GetDiskFreeSpaceExA
-            ret = method(path, ctypes.byref(_), ctypes.byref(total), ctypes.byref(free))
-            if ret == 0:
-                logger.log(u"Unable to determine free space, something went wrong", logger.WARNING)
-                raise ctypes.WinError()
-            return free.value
-
-    elif hasattr(os, 'statvfs'):  # POSIX
-        def disk_usage(path):
-            import subprocess
-            call = subprocess.Popen(["df", path], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            output = call.communicate()[0]
-            return output.split("\n")[1].split()[3]
-
-    else:
-        logger.log(u"Unable to determine free space on your OS")
-        return True
-
     if not ek(os.path.isfile, src):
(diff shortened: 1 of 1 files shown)


sickbeard/helpers.py
```diff
@@ -1574,38 +1574,12 @@ def verify_freespace(src, dest, oldfile=None):
...
-            output = call.communicate()[0]
-            return output.split("\n")[1].split()[3]
-
-    else:
-        logger.log(u"Unable to determine free space on your OS")
-        return True
-
     if not ek(os.path.isfile, src):
         logger.log("A path to a file is required for the source. {0} is not a file.".format(src), logger.WARNING)
         return True
```
medariox: Not sure why this is True, quite sure it should be False. Anyone has an idea?

ratoaq2: Reading the original method: if anything that happens that you cannot determine the available free space, it returns True (meaning: that should have enough space)

If it's not a file, you cannot know its size... so you assume that you'll have enough space.

Not that I like these kind of logic... there are situations that an exception should be thrown.

medariox: I really don't like the fact that we assume there is enough space every time something goes wrong. What happens if there really isn't enough space? Just because the file doesn't exist, there is space left? This just seems really badly designed to me.

OmgImAlexis: One time that this would need to not throw an exception even if it can't determine the space left is when using network drives, some setups don't work correctly with `statvfs` and if it were to always throw an exception you'd basically be unable to use network drives.

medariox: Question is why even use statvfs or that disk_usage method when we already have a method (getDiskSpaceUsage), which we already use almost all the time and apparently doesn't cause any issues? Just seems very strange to me that we used disk_usage in the first place.

fernandog:
@TagForce can you help us out understand ?

sickbeard/helpers.py
OmgImAlexis: Probably someone that hadn't worked on the project before added the code or something like that.

There's parts of the js where you can see that too as there's code where we already have functions for.

TagForce: Just use getDiskSpaceUsage... The verify_freespace method was a quick and dirty way to solve issues when files got corrupted because it would still copy even if there wasn't enough diskspace.

getDiskSpaceUsage wasn't there when the verify_freespace was done, so getDiskSpaceUsage should probably replace it in postProcessor.py.

verify_freespace is also only ever called once, during PostProcessing. Should be easy enough to replace it.

medariox: @TagForce 
Thanks for clearing that up. For now we will use getDiskSpaceUsage in verify_freespace and in the future we will try to get completely rid of it.

medariox:
IMO we should never relay on external commands when we have python modules that do exactly what we need. It's just very bad practice, makes the code much more errore prone and unreliable. Nonetheless we need to find out what is casing this issue. Maybe you guys can compare your OSX/lib versions and see what system is exactly affected?

PR #243, opened 2016-03-31
Store twitter consumer key and secret elsewhere

fernandog:
@labrys 
We also need to create a new Medusa application and not use 'SICKRAGETV' app created in twitter

### DIFF
--- a/sickbeard/notifiers/tweet.py
+++ b/sickbeard/notifiers/tweet.py
@@ -22,6 +22,7 @@ import sickbeard
 
 from sickbeard import logger, common
 from sickrage.helper.exceptions import ex
+from sickbeard import helpers
 
 # parse_qsl moved to urlparse module in v2.6
 try:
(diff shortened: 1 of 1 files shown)


fernandog:
@labrys or should we store in config.ini ?

OmgImAlexis:
@fernandog since we shouldn't be shipping with those keys we should either show insructions on how the user can add their own and just ship with no keys or use pymedusa.com as a proxy where their request goes, we add the keys to the request and then push it to twiter. I'd suggest the first way since that's what a lot of other applications do and it's quite simple.

fernandog:
@OmgImAlexis those consumer_* is from the SICKRAGETV twitter app

OmgImAlexis:
@fernandog yes I know that?

OmgImAlexis:
The changes you made don't make any sense and that's not what I was going for. That still leaves the keys online, I'd suggest we do as I first said and allow the user to add their own keys. They can easily make a new application on twitter and then enter their keys.

fernandog:
@OmgImAlexis can you join IRC?

fernandog:
@labrys did you registred pymedusa a gmail ?

OmgImAlexis:
@fernandog just add an var to the config.ini where it stores the keys and start with an empty string for both.

Get the user to follow [this](http://iag.me/socialmedia/how-to-create-a-twitter-app-in-8-easy-steps/) and use their own keys that way we don't have API keys floating around that're connected to a bunch of user's accounts. This is what a LOT of apps do since it's bad design to ship with keys.

PR #253, opened 2016-04-02
Add sleep before searching for subs

fernandog:
@ratoaq2 what do you think?

Now we are doing 1req/s (or more)

Looks like i didnt help:
```
2016-04-02 09:29:59 DEBUG    FINDSUBTITLES :: [d71fc3c] Could not search in legendastv provider. Discarding for now
2016-04-02 09:29:50 DEBUG    FINDSUBTITLES :: [d71fc3c] Could not search in legendastv provider. Discarding for now
2016-04-02 09:29:26 DEBUG    FINDSUBTITLES :: [d71fc3c] Could not search in legendastv provider. Discarding for now
2016-04-02 09:29:19 DEBUG    FINDSUBTITLES :: [d71fc3c] Could not search in legendastv provider. Discarding for now
2016-04-02 09:29:02 DEBUG    FINDSUBTITLES :: [d71fc3c] Could not search in legendastv provider. Discarding for now
```

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -26,6 +26,7 @@ import operator
 import traceback
 import subprocess
 import sickbeard
+import time
 from babelfish import Language, language_converters
 from subliminal import (compute_score, ProviderPool, provider_manager, refiner_manager, refine, region, save_subtitles,
                         scan_video)
@@ -306,6 +307,9 @@ def download_best_subs(video_path, subtitles_dir, release_name, languages, subti
 
         if sickbeard.SUBTITLES_PRE_SCRIPTS:
             run_subs_pre_scripts(video_path)
+
+        # give the CPU and providers a break
+        time.sleep(3)
 
         subtitles_list = pool.list_subtitles(video, languages)
         for provider in pool.providers:


ratoaq2:
You should try to reproduce the issue using subliminal:  command line or a small script. 
Then you should record the requests and responses and do the analysis.

fernandog:
@ratoaq2 I got the following issues:

https://gist.github.com/fernandog/f3c39eab8c54c1ea9a6070bd456aa427

https://gist.github.com/fernandog/98c5917a79030745d2debf128c0cfb17

So when provider is discarded in the first search, all following searches have the "discarded" message.

ratoaq2:
I just read the stacktrace. It seems that there's no `title` when searching for subtitles.
Is it possible that some process/method changes the `title` before calling subliminal?

After the exception, the provider is discarded for that provider pool. You'll only be able to use that provider again when a new instance of that provider pool is created (after cache expiration = 15 min)

ratoaq2:
For the second stacktrace, it's a subliminal bug for which you have already created an issue in subliminal's page:

https://github.com/Diaoul/subliminal/issues/602
https://github.com/Diaoul/subliminal/issues/609
https://github.com/Diaoul/subliminal/pull/614

ratoaq2:
And about the content of this PR, I don't think a sleep in that place is the correct way to go.
If at some point we figure it out that we should have a "sleep time" between each episode's subtitle search, that should be handled while looping through those episodes, not inside the "download subtitle" logic.

PR #268, opened 2016-04-04
reduced resolution and increased compression on default medusa poster

p0psicles:
We where loading 1,2M posters for each show, on first load.
Because the poster is not loaded as "static" content, it's loaded multiple times.

For ex: showPoster/?show=254559&which=poster_thumb loads the default poster for me.
If we can change this to for ex: images/default_poster.png, that would help installations with allot of shows that don't have poster loaded from indexer. But this is out of scope for the PR.

….png. Copied original poster to poster-large.png.


fernandog:
Approved

p0psicles:
@OmgImAlexis can you give me an approval and merge at the same time pls?

OmgImAlexis:
@p0psicles if we can look into what supports svg files we might be able to just use that instead of an actual image. That'd cut down on it even more.

p0psicles:
@OmgImAlexis: just checked http://caniuse.com/#feat=svg
But for example chrome 46 is not supported. And this is the version pushed by my company. Also IE8 will not be supported anymore. Although that's something I can live with.

OmgImAlexis:
Not sure what you were looking at but all Chrome's are supported from 4 and up. Only Chrome version 26 has issues and we could detect IE8 if need be and replace it with a png/jpg fall back.

p0psicles:
Ooh yes your right. I was looking at the wrong view.

PR #273, opened 2016-04-05
POC anidb indexer support

p0psicles:
## Do Not Merge!!!

This is just to get familiar with using anidb as an indexer.
At time of the first PR it isn't even functioning, as the http client requires a client and clientver registered, that apparently is banned?!

I've chosen to use lad1337's simpleanidb api client, as this seems to most feature complete. We still need to make some adjustments to it, to make it really usable. But let's play around in this branch, and then decide if we're really going to use this, then we can propose any upstream changes.

It's rebased on the tvmaze branch, as there i've made allot of changes that will be required for this feature.

What's working:
- [x] Searching of shows

What's not working:
Pretty much everything else needed for an indexer


sickbeard/indexers/indexer_config.py
```diff
@@ -59,6 +61,21 @@
...
+        'id': INDEXER_ANIDB,
+        'name': 'Anidb',
+        'module': Anidb_API,
+        'api_params': {
+            'language': 'en',
+            'useZip': True,
+            'session': helpers.make_session(cache_etags=False),
+        },
+        'icon': 'anidb.ico',
+        'scene_loc': 'https://api.pymedusa.com/scene_exceptions.json',
```
OmgImAlexis: I know this is POC but this should be using the corrected one otherwise it'll 404.

````
https://cdn.pymedusa.com/scene_exceptions/scene_exceptions.json
````

p0psicles:
Yeah, is already fixed locally through the rebase on tvmaze

p0psicles:
Biggest obatacle right now is that the search results doesnt include airdate. If we can fix that, it might have a chance of being Used as indexer. Still need to prevent http client from being banned.

labrys:
@p0psicles is it ok to delete this?

PR #278, opened 2016-04-05
[Manual Search] First call to snatchSelection always refreshed page with results

p0psicles:
The load() always handles the result, and replace the #container div with that. I've changed this to a $.getJson(), and only kicking the search to start. After this the container div is only refreshed when needed.

### DIFF
--- a/gui/slick/js/core.js
+++ b/gui/slick/js/core.js
@@ -2497,13 +2497,23 @@ var SICKRAGE = {
             });
         },
         snatchSelection: function() {
+            var spinner = $('#searchNotification');
+            var updateSpinner = function(spinner, message, showSpinner) {
+                if (showSpinner) {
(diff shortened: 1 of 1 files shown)


p0psicles:
@duramato @fernandog Can you test this please?

fernandog:
@p0psicles 
worked! thanks!
Can you make the spin wheel smaller to have the same height of the button?
Also disable the "force search" while search is running to avoid user click twice?

fernandog:
Also fixes: https://github.com/pymedusa/SickRage/issues/252

fernandog:
Approved

fernandog:
rebased

p0psicles:
Pls dont merge yet. Stil some weird behavior sometime. Need to test some more.

p0psicles:
@OmgImAlexis?

OmgImAlexis:
Looks fine to me.

PR #280, opened 2016-04-05
Added css classes for provider, size, date, to avoid ugly wrapping and reduce inline css

medariox:
This adds min-width to avoid wrapping of provider, size or date columns on smaller screens.


OmgImAlexis:
Approved!

PR #286, opened 2016-04-06
Fixes TypeError when cached_result is None in pickManualSearch

medariox:
Fixes https://gist.github.com/duramato/de804af989219f6a921500d1a8864636

@p0psicles 
Maybe a new error message would be more appropriate?


p0psicles:
Approved!

PR #314, opened 2016-04-08
DisplayShow: show required/ignored words excluding/including global words

fernandog:
@duramato @p0psicles @medariox


fernandog:
image

PR #315, opened 2016-04-08
Fixes dupe warnings and errors in log view

ratoaq2:
@duramato @fernandog 

This should get rid of the time duplication as well.


fernandog:
Approved

PR #317, opened 2016-04-08
Make scene exceptions failed as ERROR. Help debug wrong additions

fernandog:
...or when file not available at cdn

DO NO MERGE until cdn has the latest file

@OmgImAlexis your CDN for https://github.com/pymedusa/sickrage.github.io/blob/master/scene_exceptions/scene_exceptions.json is not using the latest file. How often does it update?

### DIFF
--- a/sickbeard/scene_exceptions.py
+++ b/sickbeard/scene_exceptions.py
@@ -23,7 +23,7 @@ import time
 import datetime
 import requests
 import threading
-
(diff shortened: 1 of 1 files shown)


OmgImAlexis:
@fernandog not sure what happened but the cdn was 1 commit ahead of master somehow so I've reset it. It should be up to date.

PR #353, opened 2016-04-09
Check if search threads like forceSearch, backLogsearch are running when manual snatching

p0psicles:
…ailySearch, properSearch, when snatching an episode using manual search. If a search is running, this is shown to the user, using an alert-danger.

Please don't merge this yet. I've done this in a rush, with minimal testing.
This is more to discuss the solution. And this can be used as a POC.

Note! Manual Search snatches are always queued behind backlog searches.

### DIFF
--- a/sickbeard/search_queue.py
+++ b/sickbeard/search_queue.py
@@ -89,13 +89,19 @@ class SearchQueue(generic_queue.GenericQueue):
 
     def is_backlog_in_progress(self):
         for cur_item in self.queue + [self.currentItem]:
-            if isinstance(cur_item, BacklogQueueItem):
+            if isinstance(cur_item, (BacklogQueueItem, FailedQueueItem)):
                 return True
         return False
 
     def is_dailysearch_in_progress(self):
         for cur_item in self.queue + [self.currentItem]:
             if isinstance(cur_item, DailySearchQueueItem):
+                return True
+        return False
+
+    def is_forced_search_in_progress(self):
+        for cur_item in self.queue + [self.currentItem]:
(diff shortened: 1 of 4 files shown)


sickbeard/search_queue.py
```diff
@@ -89,7 +89,7 @@ def is_manualsearch_in_progress(self):
 
     def is_backlog_in_progress(self):
         for cur_item in self.queue + [self.currentItem]:
-            if isinstance(cur_item, BacklogQueueItem):
+            if isinstance(cur_item, (BacklogQueueItem, FailedQueueItem)):
```
p0psicles: Need to check if we need this, and if this can cause regression?

sickbeard/search_queue.py
```diff
@@ -99,6 +99,12 @@ def is_dailysearch_in_progress(self):
                 return True
         return False
 
+    def is_forced_search_in_progress(self):
```
p0psicles: Don't know if we need this? Can we perform a force search, while doing manual search/snatch?

sickbeard/search_queue.py
fernandog: Isnt called anywhere? Even in manage searches page?

sickbeard/search_queue.py
fernandog: We can start a forced search then manual. User may have multiple tabs open also

sickbeard/search_queue.py
fernandog: Didnt understood why add failed here.

PR #361, opened 2016-04-11
[Manual Searc] Split ManualSearch and ForcedSearch threads

fernandog:
Fixed forced_search counting queue with manual search queue. and manualsnatch couting as manualsearch

with this:
- forced_search queue is only for forced searches
- manual_search queue is only for manual searches
- manual snatch doesn't count as search queue

### DIFF
--- a/sickbeard/manual_search.py
+++ b/sickbeard/manual_search.py
@@ -87,7 +87,7 @@ def getEpisodes(search_thread, searchstatus):
         logger.log(u'No Show Object found for show with indexerID: {}'.format(search_thread.show.indexerid), logger.ERROR)
         return results
 
-    if isinstance(search_thread, sickbeard.search_queue.ForcedSearchQueueItem):
+    if isinstance(search_thread, (sickbeard.search_queue.ManualSearchQueueItem, sickbeard.search_queue.ForcedSearchQueueItem)):
         results.append({
             'show': search_thread.show.indexerid,
             'episode': search_thread.segment.episode,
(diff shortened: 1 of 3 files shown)


p0psicles:
changing https://github.com/pymedusa/SickRage/blob/more_rename/sickbeard/search_queue.py#L180 to:
self.name = 'FORCEDSEARCH-' + str(show.indexerid) if not manual_search else 'MANUALSEARCH-' + str(show.indexerid), would rename the queueItem. That's probably enough to show it as a manualSearch in logs.

sickbeard/manual_search.py
```diff
@@ -136,11 +137,12 @@ def collectEpisodesFromSearchThread(show):
 
     # Finished Searches
     searchstatus = SEARCH_STATUS_FINISHED
-    for search_thread in sickbeard.search_queue.FORCED_SEARCH_HISTORY:
+    for search_thread in sickbeard.search_queue.MANUAL_SEARCH_HISTORY:
         if show and not str(search_thread.show.indexerid) == show:
             continue
 
-        if isinstance(search_thread, sickbeard.search_queue.ForcedSearchQueueItem):
+        if isinstance(search_thread, sickbeard.search_queue.ManualSearchQueueItem) or
```
p0psicles: can be passed as a tuple: if isinstance(search_thread, (sickbeard.search_queue.ManualSearchQueueItem, sickbeard.search_queue.ForcedSearchQueueItem))

sickbeard/search_queue.py
```diff
@@ -110,7 +112,7 @@ def queue_length(self):
                 length['forced_search'] += 1
             elif isinstance(cur_item, FailedQueueItem):
                 length['failed'] += 1
-            elif isinstance(cur_item, ManualSearchQueueItem):
+            elif isinstance(cur_item, ManualSearchQueueItem) or isinstance(cur_item, ManualSnatchQueueItem):
```
p0psicles: pass as tuple

fernandog:
@p0psicles done and working. did the tuple as you suggested.

p0psicles:
I'm not convinced you should create a separate type search queue item for something that's almost similar, just for convenience. If they are alike, we should treat them alike. This is just wasting code.

medariox:
@p0psicles 
I agree, we should make a more generic class and extend it as needed.

fernandog:
i don't agree they are alike. but if you all decide its not good close PR

PR #365, opened 2016-04-12
[Manual Search] Introduced attribute to be able to stop a provider search mid search

p0psicles:
This is the first step that introduces the attributes and a UI to manually stop a backlog search.

The next step will be to stop a backlog, when a forced search or manual search is triggered. We can re-queue the backlog behind the forced search.

I've also used the PR to fix some minor typo's. Like spaces and a grammatical error.

~~Note! I've used a list for the force_stop. This is because Bool, Int, Strings are all immutable. When changes these external to the QueueItem, a new object is created, and the reference is lost. You can test this by changing the self.force_stop = [False] in generic_queue to self.force_stop = False, and try to pick up the var in Search() class.~~

This is also the reason i've put searchProviders() within a class, because else it's not possible to keep a reference to the original object.

@medariox @labrys If you know a better of doing this, pls let me know. Passing Bool really didn't work. I've tested with id(force_stop) and the id's are different when using bools, but the same when using a mutable.


p0psicles:
I'm not going to rename all the variable names in providerSearch() in this PR. I'll put it on my todo.

p0psicles:
thnx @medariox!

p0psicles:
We're going to put this one on hold. Let's see if moving the forcedSearch to a new thread works.

labrys:
@p0psicles I think this one can be safely closed, no?

p0psicles:
Jip sure.

PR #370, opened 2016-04-12
[Manual Search] Highlight ignored/required words in manual search results

fernandog:
@duramato please test the logic! not only UI.


fernandog:
@p0psicles how to add a icon or paint the cell as red (maybe better because of space) when we set a variable "ignored=True" or "required="True" ?

maybe red for ignored and blue for required

@duramato

duramato:
Background is already blue , so green

duramato:
image

PR #373, opened 2016-04-12
Subtitles to use standard logging API

ratoaq2:
- Subtitles to use standard logging API.
- Added video cache invalidation after executing pre/extra scripts.


fernandog:
@medariox

medariox:
Approved

PR #385, opened 2016-04-14
Initial rewrite of loadIMDbInfo()

medariox:
This is the first step to a better tv.py, reducing complexity, making the code cleaner and more maintainable.

* Adresses an issue mentioned here https://github.com/pymedusa/SickRage/issues/379. Shows with multiple IMDb IDs are now supported, only the first ID is used.
* Updates IMDbPY to newest version.

Big thanks to @labrys


labrys:
Approved!

Issue #318, opened 2016-04-08
[APP SUBMITTED]: 'ascii' codec can't encode character u'\xea' in position 93: ordinal not in range(128)


duramato:
### INFO
Python Version: **2.7.6 (default, Jun 22 2015, 18:00:18) [GCC 4.8.2]**
Operating System: **Linux-3.19.0-56-generic-i686-with-Ubuntu-14.04-trusty**
Locale: None
Branch: **develop**
Commit: PyMedusa/SickRage@d47493d2dac23359d47da5a3fb781921296bb49f
No Log available with ERRORS:
### ERROR
```
FINDSUBTITLES :: [d47493d] Traceback (most recent call last):
  File "/home/home/SickRage/sickbeard/subtitles.py", line 338, in download_best_subs
    save_subtitles(video, found_subtitles, directory=subtitles_dir, single=not sickbeard.SUBTITLES_MULTI)
  File "/home/home/SickRage/lib/subliminal/core.py", line 695, in save_subtitles
    with io.open(subtitle_path, 'wb') as f:
UnicodeEncodeError: 'ascii' codec can't encode character u'\xea' in position 93: ordinal not in range(128)
Traceback (most recent call last):
  File "/home/home/SickRage/sickbeard/subtitles.py", line 338, in download_best_subs
    save_subtitles(video, found_subtitles, directory=subtitles_dir, single=not sickbeard.SUBTITLES_MULTI)
  File "/home/home/SickRage/lib/subliminal/core.py", line 695, in save_subtitles
    with io.open(subtitle_path, 'wb') as f:
UnicodeEncodeError: 'ascii' codec can't encode character u'\xea' in position 93: ordinal not in range(128)
```
---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

fernandog:
@duramato new more logs with provider and etc. have subliminal logs enabled?

duramato:
Yes but dont have this part now, they have rolled over.
Why didnt it attach them?

fernandog:
https://github.com/pymedusa/SickRage/commit/784a614f0befef01041356d47c6eeef4b9e40b2d

duramato:
### INFO
Python Version: **2.7.6 (default, Jun 22 2015, 18:00:18) [GCC 4.8.2]**
Operating System: **Linux-3.19.0-56-generic-i686-with-Ubuntu-14.04-trusty**
Locale: None
Branch: **develop**
[...]

PR #390, opened 2016-04-15
Fix unicode issues during post processing.

labrys:
Fix unicode issues during post processing.

Addresses #318 as well.

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -71,7 +71,7 @@ PROVIDER_URLS = {
 
 
 def sorted_service_list():
-    """Returns an list of subliminal providers (it's not sorted, but the order matters!)
+    """Return an list of subliminal providers (it's not sorted, but the order matters!).
 
     Each item in the list is a dict containing:
         name: str: provider name
@@ -106,7 +106,7 @@ def sorted_service_list():
 
 
 def enabled_service_list():
-    """Returns an ordered list of enabled and valid subliminal provider names
+    """Return an ordered list of enabled and valid subliminal provider names.
 
     :return: list of provider names
     :rtype: list of str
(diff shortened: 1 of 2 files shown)


labrys:
This needs testing, particularly with shows that contain non-ascii characters in the title such as `ê` or `ō`.

medariox:
Getting this:
```2016-04-15 11:48:39 INFO     Thread-14 :: [316d27c] Error occurred when downloading subtitles for: /media/Data/Media/TV Shows/My_show/S04/My_show.S04E17.HDTV-GROUP.mp4
2016-04-15 11:48:39 DEBUG    Thread-14 :: [316d27c] Exception: 'ascii' codec can't encode character u'\xe8' in position 41: ordinal not in range(128)```

To reproduce: my locale is None and my subs folder has a non-unicode char.

File: sickbeard/subtitles.py [hunk 1]
```diff
@@ -531,8 +531,8 @@ def get_video(video_path, subtitles_dir=None, subtitles=True, embedded_subtitles
         return video
 
     try:
-        video_path = encode(video_path)
-        subtitles_dir = encode(subtitles_dir or get_subtitles_dir(video_path))
+        video_path = unicode(video_path)
+        subtitles_dir = unicode(subtitles_dir or get_subtitles_dir(video_path))
```
medariox: Both video_path and subtitles_dir will need to be str before being passed to subliminal, as subliminal doesn't have proper unicode support.

labrys: Then we need to do a type-check because the errors occur with latin characters when they are attempted to be encoded more than once.

fernandog:
@labrys do you know why the error is like that? utf-8/unicode issues too?

```
[57584ce] Error while trying to retrieve KODI API version for ***: u'error [Errno 111] Conex\xe3o recusada'
```

labrys:
@fernandog its printing the string with %r.  e.g. the representation of the object.

PR #399, opened 2016-04-15
[Manual Search] Show lastest status from episode in snatchSelection

fernandog:
- [x] Collapse button text doesn't change to "Hide". Its always "Show history"
- [X] Hide season manual search when unaired episodes

@p0psicles can you review please? i'm not expert

image

.
Related to this comment:

https://github.com/pymedusa/SickRage/issues/392#issuecomment-210617288

### DIFF
--- a/gui/slick/views/displayShow.mako
+++ b/gui/slick/views/displayShow.mako
@@ -353,7 +353,9 @@
             <th class="row-seasonheader displayShowTable" colspan="13" style="vertical-align: bottom; width: auto;">
                 <h3 style="display: inline;"><a name="season-${epResult["season"]}"></a>${("Specials", "Season " + str(epResult["season"]))[int(epResult["season"]) > 0]}
                 <!-- @TODO: port the season scene exceptions to angular -->
+                % if not any([i for i in sql_results if epResult['season'] == i['season'] and int(i['status']) == 1]):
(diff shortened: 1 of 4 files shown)


sickbeard/webserve.py
```diff
@@ -1634,6 +1634,19 @@ def titler(x):
...
 
+        episode_status = None
+        release_name = None
+        try:
+            main_db_con = db.DBConnection()
+            episode_status_result = main_db_con.action("SELECT action, resource from history where showid = ? and season = ? and episode = ? order by date DESC LIMIT 1 ",  
+                                                             [indexerid, season, episode])
+            if episode_status_result:
+                for item in episode_status_result:
+                    episode_status = int(item['action'])
```
fernandog: I had to do a loop because doing episode_status_result[0]['action'] didn't worked

gui/slick/views/displayShow.mako
```diff
@@ -353,7 +353,9 @@
             <th class="row-seasonheader displayShowTable" colspan="13" style="vertical-align: bottom; width: auto;">
                 <h3 style="display: inline;"><a name="season-${epResult["season"]}"></a>${("Specials", "Season " + str(epResult["season"]))[int(epResult["season"]) > 0]}
                 <!-- @TODO: port the season scene exceptions to angular -->
+                % if not any([i for i in sql_results if epResult['season'] == i['season'] and int(i['status']) == 1]):
```
p0psicles: Maybe this is the best option we have. But did wanted to note that this does not guarantee that the season has finished. It only says, that medusa indexed unaired episodes. What if the season hasn't been finished yet, but the indexer also doesn't show information on new eps?

fernandog: There is now way for us to know. We depend on indexer. I think this would happen only in rare cases

PR #407, opened 2016-04-17
Remove override auto PP setting when 'postpone if no subs'

fernandog:
@NicoLeOca


NicoLeOca:
@fernandog 
great! thank you

PR #433, opened 2016-04-20
Adding missing methods to show_queue

fernandog:


p0psicles:
I don't agree on this route.
If someone wants to remove a show, it should not be hindert.

labrys:
I agree with @p0psicles.  It should be removed, and the queue item should fail when it is processed.

OmgImAlexis:
I agree with @p0psicles and @labrys, the reason I removed the show that caused this in the first place was I noticed it wasn't the correct one and it then started filling my torrent client with every single episode even though I didn't want it.

If it was to wait until the search queue, etc. finished then I'd have to go and cleanup after it even though I had removed the show.

fernandog:
@labrys @p0psicles changed the code. Now it only adds the missing methods and dont call `_isInQueue` directly

labrys:
@fernandog personally I think it would be cleaner to just handle the exception, thats why we raise exceptions to begin with.  Also we wouldn't check to see if its in the remove queue, because the item may be out of the remove queue before another queue has finished getting to that item in the queue.  We could maintain an in-memory list of removed shows and check if its in that show during the exception handling that way we know if we should re-raise or ignore.

PR #448, opened 2016-04-21
Total rewritte of limetorrents to use page parsing , add trusted down…

duramato:
…loads only option, automated way of updating seeders/leechers info while searching plus no more unrelated results


fernandog:
Approved

PR #450, opened 2016-04-22
Made sure only anime marked shows are using this provider

p0psicles:

### DIFF
--- a/sickbeard/providers/anizb.py
+++ b/sickbeard/providers/anizb.py
@@ -32,9 +32,9 @@ class Anizb(NZBProvider):  # pylint: disable=too-many-instance-attributes
         # Provider Init
         NZBProvider.__init__(self, "Anizb")
 
-        # Credentials
         self.public = True
-        self.confirmed = True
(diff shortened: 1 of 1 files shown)


sickbeard/providers/anizb.py
```diff
@@ -32,8 +32,9 @@ def __init__(self):
         # Provider Init
         NZBProvider.__init__(self, "Anizb")
 
-        # Credentials
         self.public = True
+        self.supports_absolute_numbering = True
+        self.anime_only = True
         self.confirmed = True
```
p0psicles: Honestly I don't know what self.confirmed is used for. It got copied from another prov.

fernandog: Its like "Trusted" torrents/users from KAT.

duramato: confirmed its intended for verified torrents, ranked for internal torrents of the site

p0psicles: Ok then i'll remove it

PR #477, opened 2016-04-25
Add check for no results for limetorrents

medariox:


medariox:
@fernandog 
Can you rebase it please?

PR #482, opened 2016-04-26
Only show video name instead of full path for subs logging

medariox:

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -315,7 +315,7 @@ def download_best_subs(video_path, subtitles_dir, release_name, languages, subti
                 logger.debug(u'Could not search in %s provider. Discarding for now', provider)
 
         if not subtitles_list:
-            logger.info(u'No subtitles found for %s', video_path)
+            logger.info(u'No subtitles found for %s', os.path.basename(video_path))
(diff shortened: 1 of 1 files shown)


sickbeard/subtitles.py
```diff
@@ -315,7 +315,7 @@ def download_best_subs(video_path, subtitles_dir, release_name, languages, subti
                 logger.debug(u'Could not search in %s provider. Discarding for now', provider)
 
         if not subtitles_list:
-            logger.info(u'No subtitles found for %s', video_path)
+            logger.info(u'No subtitles found for %s', os.path.basename(video_path))
```
labrys: Couldn't this return incorrect information such as the season folder?

fernandog:
Approved

sickbeard/subtitles.py
medariox: Wouldn't know why, video_path is always the full path to the video the subtitle is searched for.

labrys: OK.  I assumed `video_path` could also be the full path to the containing folder, not including the video file. (i.e. `/series/season 01/` instead of `/series/season 01/show s01e01 title.mkv`)

PR #486, opened 2016-04-26
Fix RSS on Lime

duramato:
@labrys

### DIFF
--- a/sickbeard/providers/limetorrents.py
+++ b/sickbeard/providers/limetorrents.py
@@ -125,6 +125,7 @@ class LimeTorrentsProvider(TorrentProvider):  # pylint: disable=too-many-instanc
                                 'size': size,
                                 'seeders': seeders,
                                 'leechers': leechers,
+                                'pubdate': None,
                                 'hash': torrent_hash or ''
                             }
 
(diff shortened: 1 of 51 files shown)


sickbeard/providers/limetorrents.py
```diff
@@ -62,86 +62,99 @@ def __init__(self):
...
         results = []
         for mode in search_strings:
-            items = []
             logger.log('Search Mode: {0}'.format(mode), logger.DEBUG)
             for search_string in search_strings[mode]:
                 search_url = self.urls['rss'] if mode == 'RSS' else self.urls['search'].format(query=search_string)
-                data = self.get_url(search_url, returns='text')
+                if mode == 'RSS':
+                    for x in range(1, 4):
+                        data = self.get_url('https://www.limetorrents.cc/browse-torrents/TV-shows/date/{x}/'.format(x=x))
```
medariox: Why don't you use self.urls['rss'] instead?

sickbeard/providers/limetorrents.py
```diff
@@ -47,7 +47,7 @@ def __init__(self):
             'index': self.url,
             'update': urljoin(self.url, '/post/updatestats.php'),
             'search': urljoin(self.url, '/search/tv/{query}/'),
-            'rss': urljoin(self.url, '/browse-torrents/TV-shows/date/')
+            'rss': 'https://www.limetorrents.cc/browse-torrents/TV-shows/date/'
```
medariox: Why don't you use self.url instead?

PR #487, opened 2016-04-26
Use unicode everywhere in list_associated_files() to fix PP issues

medariox:
Works perfectly fine for me so far.

@duramato @duramato @fernandog 
Would appreciate some testing.


fernandog:
Testing this

labrys:
And apparently @duramato should be checking this twice :)

fernandog:
Approved

medariox:
@labrys 
About things not do do... https://github.com/pymedusa/SickRage/blob/develop/SickBeard.py#L188 
That explains quite a few things, we should work on getting rid of that hack asap.

medariox:
Will be included in the unicode branch when the time comes.

PR #490, opened 2016-04-27
Add lib simpleanidb

p0psicles:
This lib has support for anidb.net's http api, as also the http://anidb.net/api/anime-titles.xml.gz for the show search. It's a step in the direction for building an anidb indexer, with the biggest limitation that the xml does not have the airdates. Making it not ideal for indexer searches.

For now it will be used for retrieving the anidb.net recommended lists.

### DIFF
--- a/lib/simpleanidb/models.py
+++ b/lib/simpleanidb/models.py
@@ -16,7 +16,9 @@ class Anime(object):  # pylint: disable=too-many-instance-attributes
         self.episodes = {}
         self.picture = None
         self.rating_permanent = None
+        self.count_permanent = None
         self.rating_temporary = None
+        self.count_temporary = None
         self.rating_review = None
         self.categories = []
         self.tags = []
@@ -24,7 +26,7 @@ class Anime(object):  # pylint: disable=too-many-instance-attributes
         self.end_date = None
         self.description = None
 
-        if xml:
+        if len(xml):
             self.fill_from_xml(xml)
 
         self._loaded = False
@@ -72,8 +74,10 @@ class Anime(object):  # pylint: disable=too-many-instance-attributes
         if xml.find("ratings") is not None:
             if xml.find("ratings").find("permanent") is not None:
                 self.rating_permanent = xml.find("ratings").find("permanent").text
+                self.count_permanent = xml.find("ratings").find("permanent").get('count', 0)
             if xml.find("ratings").find("temporary") is not None:
                 self.rating_temporary = xml.find("ratings").find("temporary").text
+                self.count_temporary = xml.find("ratings").find("temporary").get('count', 0)
             if xml.find("ratings").find("review") is not None:
                 self.rating_review = xml.find("ratings").find("review").text
         if xml.find("categories") is not None:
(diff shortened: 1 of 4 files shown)


lib/simpleanidb/models.py
```diff
@@ -0,0 +1,230 @@
...
+            self.episodes = {e.number: e for e in self.all_episodes if e.type == 1}
+        if xml.find("picture") is not None:
+            self.picture = Picture(self, xml.find("picture"))
+        if xml.find("ratings") is not None:
+            if xml.find("ratings").find("permanent") is not None:
+                self.rating_permanent = xml.find("ratings").find("permanent").text
+            if xml.find("ratings").find("temporary") is not None:
+                self.rating_temporary = xml.find("ratings").find("temporary").text
+            if xml.find("ratings").find("review") is not None:
+                self.rating_review = xml.find("ratings").find("review").text
```
p0psicles: Hmm, number of votes is in xml, but not parsed. As we use those on the recommended lists, maybe i'll need to add those.

labrys:
Was this only on hold for the release?  If so then we can add it to #605 for v0.1.1

p0psicles:
Yes, it was also still in progress. As I was testing it in combi with the recommended lists. That's also where I noticed it didnt support the nr. of votes. That's still something I need to fix. But 0.1.1 is fine. Isn't that much work anymore

lib/simpleanidb/models.py
```diff
@@ -0,0 +1,234 @@
...
+        self.synonyms = [t for t in self.titles if t.type == "synonym"]
+        if xml.find("episodes") is not None:
+            self.all_episodes = sorted([Episode(self, n) for n in xml.find("episodes")])
+            self.episodes = {e.number: e for e in self.all_episodes if e.type == 1}
+        if xml.find("picture") is not None:
+            self.picture = Picture(self, xml.find("picture"))
+        if xml.find("ratings") is not None:
+            if xml.find("ratings").find("permanent") is not None:
+                self.rating_permanent = xml.find("ratings").find("permanent").text
+                self.count_permanent = xml.find("ratings").find("permanent").get('count', 0)
```
p0psicles: I've added the number of votes, that where originally missing from the lib.

lib/simpleanidb/models.py
```diff
@@ -16,15 +16,17 @@ def __init__(self, anidb, aid, auto_load=True, xml=None):
...
+        self.count_temporary = None
         self.rating_review = None
         self.categories = []
         self.tags = []
         self.start_date = None
         self.end_date = None
         self.description = None
 
-        if xml:
+        if len(xml):
```
medariox: Why this change?

p0psicles: Else it will give this warning: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  if xml:

Quick google recommends to use an explict len() or not None. Using len(), means the xml object has to have a minimum of one child. As the xml object should always have children, len() is also a valid check.

medariox: Oh, it's something Element specific, interesting. Thanks for clearing that up.

medariox:
Approved

I don't see any problem with it. Is this still meant to be for the 0.1.1 release?

p0psicles:
Approved

PR #494, opened 2016-04-27
Hide sensitive info from the logs

duramato:
Like ips/domains , api keys and tokens

### DIFF
--- a/gui/slick/views/config_general.mako
+++ b/gui/slick/views/config_general.mako
@@ -4,7 +4,7 @@
     import locale
     import sickbeard
     from sickbeard.common import SKIPPED, WANTED, UNAIRED, ARCHIVED, IGNORED, SNATCHED, SNATCHED_PROPER, SNATCHED_BEST, FAILED
-    from sickbeard.common import Quality, qualityPresets, statusStrings, qualityPresetStrings, cpu_presets
+    from sickbeard.common import Quality, qualityPresets, statusStrings, qualityPresetStrings, cpu_presets, privacy_levels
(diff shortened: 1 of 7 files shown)


fernandog:
Approved

labrys:
While I definitely understand a desire (and need) to censor these items I think excessive censoring can make some things difficult to debug.  Until we fix logging in general, how about a Paranoia setting?  It would replace the `True` value for the censor argument.

fernandog:
@duramato @labrys added setting

duramato:
needs to be added to all `censor_log`s

labrys:
Sorry kinda forgot about this one, I'll try to find the time to test before this weekend.

fernandog:
Not working.
I used kodi notifications and still host appear in log

duramato:
The setting you made it disabled by default

fernandog:
so? I enabled to test

fernandog:
@duramato you should close it as no one seems interested

labrys:
@fernandog No just some people like to put some thought into their code instead of rushing things.

labrys:
A setting of `Normal` provides the same filters as previously.  `Low` filters only passwords/api keys/etc.  `Disabled` provides no filtering at all.  `High` adds the additional filters such as hosts.  Nothing is currently set to `Max`.  `Stupid` filters every single attribute from the config file.  @duramato I think you'll get a kick out of the `Absurd` level :grinning:

labrys:
Approved.  Privacy levels are low, normal, high.

labrys:
QC issues are pre-existing and will be addressed as part of the webserver and config file changes

PR #497, opened 2016-04-27
Fix season search in tvcuk

duramato:

### DIFF
--- a/sickbeard/providers/tvchaosuk.py
+++ b/sickbeard/providers/tvchaosuk.py
@@ -1,7 +1,5 @@
 # coding=utf-8
 #
-
-#
 # This file is part of SickRage.
 #
 # SickRage is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


sickbeard/providers/tvchaosuk.py
```diff
@@ -102,7 +100,7 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
             for search_string in search_strings[mode]:
 
                 if mode == 'Season':
-                    search_string = re.sub(ur'(.*)Season', ur'\1Series', search_string)
+                    search_string = re.sub(ur'(.*)S0?', ur'\1Series', search_string)
```
fernandog: @duramato can you add comments with examples of releases that matches it ?

what abou releases like Show1.Season.6...... ?

duramato: There are no Season 6, just Series 6.
They are all like that or at least 90%, making the 10% despicable

labrys: What about seasons > 9? And series with a year for a season?
How about `ur'\bs\d{1,4}\b'` instead?  It will match any Sx through Sxxxx. Also since it doesn't have the `.* ` it will match multi-season too.

duramato: Would it bring any advantage? The search string ever has multiple seasons? 
With the current one it will only replace `S` to `Series` and if its a season shorter than 9 `S0` to Series becouse in the site all seasons less than 10 dont have a 0 and if its searched with it no results will be returned
Also there arent any seasons with year from my research in the site

labrys:
@duramato If this tested good i think its ready for rebase and merge.

fernandog:
Approved

labrys:
Approved

labrys:
Attempted to kickstart QC by closing and re-opening with no luck. Forcing merge since QC is failing to report status.

PR #501, opened 2016-04-28
Add Feature Improvements to Recommended Lists

p0psicles:
Goal of this PR are: 
* Rewrite the libtrakt because it had sickbeard global settings in it, and other imports, that shouldn't be part of a lib.
* Create a generic structure for adding recommended lists from all kinds of sources. Currently we have one page for IMDB and one page for Trakt lists. If we would add a new source, that would mean another page, while they are all very similar. The structure is located in /sickrage/show/recommendations. Also a new page will be introduced dedicated for recommended lists. I've named it http://localhost:8081/addRecommended/ for now.
* Add anidb recommended lists (i've picked a slightly modified version of lad1337 simpleanidb lib for this). This version will be maintained in the medusa orgainization until (if ever) lad accepts it into upstream.
* A generic show options menu, for when a recommended show is added.

As for the libtrakt rewrite the following i've used the following principles:
* No new functionality should be introduced.
* The libtrakt lib should really be used like a lib. Meaning for ex. no sickbeard.TRAKT_ACCESS_TOKEN in the lib.
* The template for trakt, should be able to be reused for other sources like anidb recommended shows or imdb watchlists.

changes:
* Introduced a generic template. This one should look the same as addShows_recommended.mako
* Lot's of PEP8 changes
* notifier changes to adapt to the libtrakt changes
* traktChecker changes, to adapt to the libtrakt changes.
* changes to core.js, to make sure shows can be added from trakt, through the addShowByID webserve function. That function has been changed in webserve, to accept default show options params.
* Added is_anime attribute with check. Uses simpleanidb to map tvdbid to aid.
* Added a Force Sync button in trakt notifications config for triggering a trakt sync.
* Trakt shows automatically added, are now added with the "cleaned" showname and show directory of the indexer, that it's added with.

TODO:
- [x]  Fix some of the libtrakt exception docstrings. Some are to detailed, or just plain wrong. @labrys ?
- [x]  test
- [x]  Change double to single quotes
- [x] Fix mapping of simpleanidb object to recommendedList
- [x] Add icon for the anidb hotanime recommended list
- [x] Reminder: Make sure traktChecker doesn't start immediatly after startup! The delay has been removed for dev purposses.
- [x] Fix trakt syncing when option "Remove Episodes From Collection" is enabled.
- [x] Trakt is not properly refreshing tokens
- [x] Add lib AppDirs to medusa

FUTURE:
- [ ] Add more anidb lists, with a select box. For ex. add randomized anime list.

Some screenies:
Anidb:
image

Imdb:
image

addRecommended page:
image

image


p0psicles:
https://github.com/pymedusa/SickRage/issues/150

gui/slick/js/core.js
```diff
@@ -3176,6 +3178,98 @@ var SICKRAGE = {
...
+                        'default_status': $('#statusSelect').val(),
+                        'any_qualities': anyQualArray.join(','),
+                        'best_qualities': bestQualArray.join(','),
+                        'default_flatten_folders': $('#flatten_folders').prop('checked'),
+                        'subtitles': $('#subtitles').prop('checked'),
+                        'anime': anime,
+                        'scene': $('#scene').prop('checked'),
+                        'default_status_after': $('#statusSelectAfter').val(),
+                    });
+                    return false;
```
OmgImAlexis: This was made redundant by using `e.preventDefault();` at the top of the function.

sickbeard/webserve.py
```diff
@@ -77,9 +75,18 @@
...
-from sickrage.helper.common import (
-    episode_num, sanitize_filename, try_int, enabled_providers,
-)
+from sickbeard.imdbPopular import imdb_popular
+from sickbeard.helpers import get_showname_from_indexer
+from sickrage.show.recommendations.trakt import TraktPopular
+
+from dateutil import tz
+from unrar2 import RarFile
+import adba
```
OmgImAlexis: This is already imported on LN#31

p0psicles:
@OmgImAlexis thx for the feedback. But you don't already have to spend allot of time on this. As allot might still change. I'm also planning to create generic structure that get's a particular list, but always returns it as json. This way, we'll end up with one page (html) that loads it's lists/data dynamically.

Then it also won't be allot of work to port to angular.

Thraxis:
I haven't had a lot of free time to do any coding recently but i do have one code suggestion that could be implemented in this.  With the current anidb integration you can check if a show is an Anime or not by using the tvdb id.  

When I get home I'll grab my POC code and add the small section here.   It basically a part 1 of 2 for the trakt / imdb lists to automatically mark anime when added,  part 2 of the idea was to add multiple default config options where you would have different defaults for regular TV vs Anime.

From memory the basic code piece was as follows:
```
anime = adba.Anime(None, tvdbid=show['show']['ids']['tvdb'])
```
This was wrapped in a try / except and if it passed would add the anidb id to the show['show']['id'] list and in the template if that id existed would in my POC print the word "Anime" next to the add button.

Thraxis:
@p0psicles 
Below is the code block I used in webserve.py to add the anidb id in the trakt lists

```
                    try:
                        anime = adba.Anime(None,tvdbid=show['show']['ids']['tvdb'])
                    except Exception as e:
                        pass
                    else:
                        show['show']['ids']['anidb'] = anime.aid
```

And this was the code in the trendingShows.mako to show the Anime tag:

```
                % if 'anidb' in cur_show['show']['ids']:
                <a href="#" class="btn btn-xs">Anime</a>
                % endif
```

As I said it was a basic POC that I never got far in.

p0psicles:
Cool. Thnx Thraxis. I will thing on it. Don't know right away if or where we can use it, because i'd like to implement the recommended lists as generic as possible.
If you ever feel like picking it up again, feel free to join our irc! #pymedusa.

p0psicles:
7dc2d2a fixes https://github.com/pymedusa/SickRage/issues/518

sickbeard/helpers.py
```diff
@@ -1695,36 +1696,44 @@ def getTVDBFromID(indexer_id, indexer):  # pylint:disable=too-many-return-statem
...
         url = "http://api.tvmaze.com/shows/%s" % indexer_id
         data = getURL(url, session=session, returns='json')
         if data is None:
             return tvdb_id
         tvdb_id = data['externals']['thetvdb']
         return tvdb_id
+
+    # If indexer is IMDB and we've still not returned a tvdb_id, let's try to use tvmaze's api, to get the tvdbid
+    if indexer == 'IMDB':
+        url = 'http://api.tvmaze.com/lookup/shows?imdb={indexer_id}'.format(indexer_id=indexer_id)
```
p0psicles: @labrys what you think of this?

Hellowlol:
While you at it can you remove recommendations on shows that user already has?

p0psicles:
Now I've changed the "add" button to make it move you to your existing show, What are the opinions on this? Leave it like is, or don't show, when show is already in list?

labrys:
@p0psicles I think shows you have in your selection should now show upon visiting the page.. i think shows that you add while there should just be marked as "selected" so you can "unselect" to not have them added in case you change your mind

p0psicles:
With the latest additions to the simpleanidb lib, it is now also possible to show if a recommended show, is an anime. I've used the adivce of @Thraxis, to check the anidb-lists.xml provided by ScudLee.

Only difference is that I'm using the simpleanidb lib, and have adjusted that, to also be able to search the xml, by tvdbid.

Looks now like this. But can be changed obviously:
image

labrys:
Initial impressions:  I'm not sure I care for the separate page for recommendations, i think they should all be on the same page, however the current add shows page needs cleaned up imo.  

The anidb uses the IMDb logo.

Some of the wording is a little awkward..  e.g. Add From Recommended.. maybe Add Recommended Shows or Show Recommendations.

labrys:
When loading IMDB lists:
```
22:22:44 DEBUG::Thread-24 :: [8d4754a] Requested download url https://walter.trakt.us/images/shows/000/106/722/posters/thumb/f71803bbf4.jpg returned status code is 403: Forbidden
```

labrys:
All lists have following text:
`If you don't want to use the default show options, you can change them here! Leaving it as it is, the show will be added as an anime anyhow.`

Which means either its broken in that its trying to add all of them as anime, or just needs to be fixed in the UI.

Also I really don't care for the user flow of adding like this, configuring settings before you've even selected shows seems wrong.  IMO it should follow the same user flow as add existing but instead of getting the list of shows from directories it should get it from your current selection.  Also there is no way to add shows with different settings short of re-visiting the page multiple times for each diff setting.

labrys:
Shows already in list aren't hidden.

labrys:
When a show gets added the `being added` button text wraps the button to the next line changing the size of the poster div breaking the flow of the UI.

p0psicles:
@labrys like discussed, i'm going to make the following changes:
* Make it visible when shows have already been added. I'm going to use the css property grayscale().
* imdb button should also change to `in list` for already added shows
* Change when a show has been added button label `being added` to `added`. Prevents it from being wrapped to new line.
* The anidb uses the IMDb logo.
* Add from recommended -> Add Recommended Shows
* Fix anidb used imdb logo. I need to check where this is the case.
* Changed the options label text to: `Recommended shows will added using your default options. Use this option if you want to change the options for that show.`

labrys:
My idea for a single page UI would be something like this:

ui

p0psicles:
Yeah @labrys problem with that, is that you expect it to be a static list. But what if we add 4 more list providers. Or 10 more. Then each time you have to rethink your design. That's why I wanted to move it to a separate page.

labrys:
@p0psicles No I'm thinking an icon-based view similar to the posters layout for shows.  It dynamically fills the recommend show categories with the icon for the recommenders instead of a banner style layout.

p0psicles:
I've added the update to the simpleanidb lib here. But as it's a lib, i'd rather merged in the other PR. It will be removed here after rebase.

labrys:
Still not fixed for Trakt:
```
Leaving it as it is, the show will be added as an anime anyhow.
```

labrys:
Colors wrong on poster info:

screenshot_4

labrys:
AniDB lists don't load:
```
Fetching of Recommender Data failed. Are you online? Exception:

HTTPConnectionPool(host='api.anidb.net', port=9001): Max retries exceeded with url: /httpapi?clientver=100&client=adbahttp&request=hotanime&protover=1 (Caused by NewConnectionError(': Failed to establish a new connection: [Errno 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond',))
```

labrys:
I don't think the grayscale effect does enough to distinguish a show that's already in your database, particularly if the show poster is already black and white.  Perhaps an overlay pill/banner/checkmark or changing the color of the "in list" pill, or a colored border for the poster, or just removing them from the list as was done previously.

p0psicles:
Okay what you think of this? 
image

labrys:
I might opt for centered, maybe not quite so large, and possibly slightly translucent, but it definitely sticks out that its already been added.  From this point I think its just a style opinion since it fulfills the requirement.  If anyone else has any opinions on the look-and-feel for it, I think now would be the time to weigh in.

p0psicles:
Latest changes should fix:
* Adding the green check, i've made it slightly trancelucent now, like @labrys suggested.
* Fix the dark them, white text
* Fix for the trakt page. This is actually still a separate page for the wrapper. The covers themselves, are loaded through the default recommendedShows.mako template. So now that also shows the same text.

labrys:
I'll review again when I get home.  As for Travis failing, it looks like the `appdirs` dependency for `simpleanidb` isn't included.  On that note, I'm against using `appdirs` for determining cache directory for various reasons including that we already have a cache directory and it keeps Medusa from being a "portable" app.  This may also cause issues if running multiple instances on a single server.

p0psicles:
Like also mentioned on irc. It's true that appdirs as a dependency. But as this is a lib, there is nothing wrong with using the users app data directory. But, if we want to use a local cache dir. We could ofcourse pass that as a paramater. That's possible.

On a sidenote, i noticed that next to appdirs it's also using: tempfile and getpass. Should we also include these as libs?

labrys:
IMO, in the interest of making Medusa a portable app, this should pass our cache directory as a default.  Regardless if this is a library or not, I don't see any advantage to using the user directory for caching the data except in the case of running multiple instances of Medusa and even then that could potentially lead to issues with simultaneous file access.

labrys:
Oh, and as for getpass and tempfile, those are part of the Python standard library.

p0psicles:
I've switched to using the medusa cache dir. I now don't see any need to include the appsdir lib, as we don't make use of it. But that's is ofcourse debatable.

lib/libtrakt/exceptions.py
```diff
@@ -1,8 +1,43 @@
-class traktException(Exception):
-    pass
+from requests.exceptions import (ConnectionError, Timeout, TooManyRedirects)
 
-class traktAuthException(traktException):
-    pass
 
-class traktServerBusy(traktException):
-    pass
+class TraktException(Exception):
```
labrys: Exceptions should be called Errors (assuming they are an error) and not Exception.

lib/libtrakt/exceptions.py
```diff
@@ -1,8 +1,43 @@
...
+    """A Generic Trakt Exception"""
+
+
+class TraktAuthException(TraktException):
```
labrys: While the base class makes sense to prefix with Trakt, prefixing the remainder makes it unecessarily verbose.  Verbosity that is easily replicated through import statemtents. e.g.:
```python
import trakt
trakt.exceptions.AuthException

from trakt.exceptions import AuthException as TraktAuthException
```

lib/libtrakt/exceptions.py
```diff
@@ -1,8 +1,43 @@
...
+    """A Generic Trakt Authentication Exception"""
+
+
+class TraktServerBusy(TraktException):
+    """A Generic Trakt Server Busy Exception"""
```
labrys: Not all the doc strings need to be "A Generic...".  For example this is a specific exception for the Trakt server being busy.

lib/libtrakt/trakt.py
```diff
@@ -1,128 +1,178 @@
...
+import time
+import json
+import logging
 import requests
 import certifi
-import json
-import sickbeard
-import time
-from sickbeard import logger
+from exceptions import (TraktConnectionException, TraktMissingTokenException, TraktAuthException,
```
labrys: Use absolute or relative import syntax for python 3 compatibility, e.g.:
```python
# absolute imports
from libtrakt import exceptions
from libtrakt.exceptions import TraktException

# relative imports
from . import exceptions
from .exceptions import TraktException
```

lib/libtrakt/__init__.py
```diff
@@ -1 +0,0 @@
-from trakt import TraktAPI
```
labrys: Can we rename the library?  libtrakt is obviously a library so the lib prefix is just added noise.

lib/libtrakt/exceptions.py
```diff
@@ -1,8 +1,43 @@
...
+    """A Generic Trakt Timeout Exception, inherited from TraktIOError and IOError"""
+
+
+class TraktUnavailableException(TraktException):
+    """A Generic Trakt Unavailable Exception,
+    possibly raised when Trakt is reachable but is showing an unavailable response code.
+    Possibly raised on in 500 series response codes"""
+
+
+class TraktResourceNotExistException(TraktException):
```
labrys: `class ResourceUnavailable(TraktError)`

lib/libtrakt/exceptions.py
```diff
@@ -1,8 +1,43 @@
...
+
+
+class TraktTimeoutException(TraktIOError, Timeout):
+    """A Generic Trakt Timeout Exception, inherited from TraktIOError and IOError"""
+
+
+class TraktUnavailableException(TraktException):
+    """A Generic Trakt Unavailable Exception,
+    possibly raised when Trakt is reachable but is showing an unavailable response code.
+    Possibly raised on in 500 series response codes"""
```
labrys: 'Possibly raised on ~~in~~ 500 series response codes'

fernandog:
@p0psicles 

```
File "/home/osmc/SickRage/sickbeard/notifiers/trakt.py", line 28, in <module>
    from traktor import TraktApi
ImportError: No module named traktor
```

fernandog:
@p0psicles can you rebase please?

sickbeard/traktChecker.py
```diff
@@ -210,136 +213,142 @@ def remove_episode_trakt_collection(self, filter_show=None):
...
 
             main_db_con = db.DBConnection()
-            sql_selection = 'select tv_shows.indexer, tv_shows.startyear, showid, show_name, season, ' \
-                            'episode from tv_episodes,tv_shows where tv_shows.indexer_id = tv_episodes.showid ' \
-                            'and tv_episodes.status in (' + ','.join([str(x) for x in Quality.DOWNLOADED + Quality.ARCHIVED]) + ')'
-            episodes = main_db_con.select(sql_selection)
+            selection_status = ['?' for _ in Quality.DOWNLOADED + Quality.ARCHIVED]
+            sql_selection = b'select tv_shows.indexer, tv_shows.startyear, showid, show_name, season, ' \
+                            b'episode from tv_episodes,tv_shows where tv_shows.indexer_id = tv_episodes.showid ' \
+                            b"and tv_episodes.status in ({0}) and tv_episodes.location <> ''".format(','.join(selection_status))
```
p0psicles: The tv_episodes.location <> '', fixes the add/remove loop.

gui/slick/views/addShows_newShow.mako
```diff
@@ -7,7 +7,6 @@
 <script type="text/javascript" src="${srRoot}/js/qualityChooser.js?${sbPID}"></script>
 <script type="text/javascript" src="${srRoot}/js/addShowOptions.js?${sbPID}"></script>
 <script type="text/javascript" src="${srRoot}/js/rootDirs.js?${sbPID}"></script>
-<script type="text/javascript" src="${srRoot}/js/blackwhite.js?${sbPID}"></script>
```
Thraxis: @p0psicles 

I don't think you meant to remove this.   This is the code for the black / white list for anime on the Add New Show page, not recommended.

p0psicles:
Ok, still some flake8 errors to fix. But in the mean time, think it's finished. Still room for allot of speed improvements, but as i'm going to create local cache tables for the recomm. lists, it doesn't make much sense to look into that now.

@pymedusa/developers please test.

p0psicles:
Note: switching between trakt recommended lists doesn't change title.

ratoaq2:
Approved

[Approved with PullApprove](https://pullapprove.com/pymedusa/SickRage/pull-request/501/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=pymedusa/SickRage)

PR #511, opened 2016-04-30
Add log to help debug 'unable to match..." cases

fernandog:
This helped find this wrong regex matching:

https://github.com/pymedusa/SickRage/pull/510

@duramato

### DIFF
--- a/sickbeard/name_parser/parser.py
+++ b/sickbeard/name_parser/parser.py
@@ -463,13 +463,13 @@ class NameParser(object):
         final_result.quality = self._combine_results(file_name_result, dir_name_result, 'quality')
 
         if not final_result.show:
-            raise InvalidShowException("Unable to match {0} to a show in your database. Parser result: {1}".format
-                                       (name.encode(sickbeard.SYS_ENCODING, 'xmlcharrefreplace'), file_name_result or dir_name_result))
+            raise InvalidShowException(u"Unable to match {0} to a show in your database. Parser result: {1}".format(
+                                         name, file_name_result or dir_name_result))
 
         # if there's no useful info in it then raise an exception
         if final_result.season_number is None and not final_result.episode_numbers and final_result.air_date is None and not final_result.ab_episode_numbers and not final_result.series_name:
-            raise InvalidNameException("Unable to parse {0} to a valid episode. Parser result: {1}".format
-                                       (name.encode(sickbeard.SYS_ENCODING, 'xmlcharrefreplace'), file_name_result or dir_name_result))
+            raise InvalidNameException(u"Unable to parse {0} to a valid episode. Parser result: {1}".format(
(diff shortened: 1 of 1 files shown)


sickbeard/name_parser/parser.py
```diff
@@ -463,13 +463,13 @@ def parse(self, name, cache_result=True):
         final_result.quality = self._combine_results(file_name_result, dir_name_result, 'quality')
 
         if not final_result.show:
-            raise InvalidShowException("Unable to match {} to a show in your database".format
-                                       (name.encode(sickbeard.SYS_ENCODING, 'xmlcharrefreplace')))
+            raise InvalidShowException("Unable to match {0} to a show in your database. Parser result: {1}".format
+                                       (name.encode(sickbeard.SYS_ENCODING, 'xmlcharrefreplace'), file_name_result or dir_name_result))
```
labrys: Instead of encoding to sys encoding this should probably decode and be a unicode string (same with the other).

fernandog: This change should be in another PR right?
Because if if breaks we will revert the scope of this PR

labrys: No it should be another commit in this PR.  And then it would ofc require additional tests, if it fails then this commit would be reverted leaving the original commit.

fernandog: Until the PP encoding issues is really fixed we shoulnt change any encoding stuff.

This PR is done. Its just a informative PR

labrys: This encoding isn't related to the PP encoding issues.  The idea is it's safer to log unicode messages so making the string unicode and skipping the encode is less prone to errors because python itself does the unicode coercion.  e.g. 
```python
u"Unable to match {0} to a show in your database. Parser result: {1}".format(
    name, file_name_result or dir_name_result)
```

sickbeard/name_parser/parser.py
```diff
@@ -463,13 +463,13 @@ def parse(self, name, cache_result=True):
         final_result.quality = self._combine_results(file_name_result, dir_name_result, 'quality')
 
         if not final_result.show:
-            raise InvalidShowException("Unable to match {0} to a show in your database. Parser result: {1}".format
-                                       (name.encode(sickbeard.SYS_ENCODING, 'xmlcharrefreplace'), file_name_result or dir_name_result))
+            raise InvalidShowException(u"Unable to match {0} to a show in your database. Parser result: {1}".format(
+                                         name, file_name_result or dir_name_result)
```
fernandog: @labrys

fernandog:
@duramato any locale issues with that commit in pubdate2 branch?

duramato:
all good

duramato:
Approved

PR #515, opened 2016-04-30
Support for non US "season", meaning AU, UK , etc

duramato:
Adds support for `The Katering Show Series 2 Episode 1 Red Ramen` (UK,AU), while maintaining support for `Show Name Season 1 Episode 2 Ep Name` (US)  alike namings

### DIFF
--- a/sickbeard/name_parser/regexes.py
+++ b/sickbeard/name_parser/regexes.py
@@ -1,5 +1,4 @@
 # coding=utf-8
-
 # Author: Nic Wolfe <nic@wolfeden.ca>
 # URL: http://code.google.com/p/sickbeard/
 #
@@ -125,7 +124,7 @@ normal_regexes = [
      # Show Name Season 1 Episode 2 Ep Name
      r'''
      ^(?P<series_name>.+?)[. _-]+                # Show Name and separator
-     season[. _-]+                               # season and separator
+     (season|series)[. _-]+                      # season and separator
      (?P<season_num>\d+)[. _-]+                  # 1
      episode[. _-]+                              # episode and separator
      (?P<ep_num>\d+)[. _-]+                      # 02 and separator


duramato:
TBMerged after https://github.com/pymedusa/SickRage/pull/510

p0psicles:
I think that for each regex change we should add Unit test. @Labrys what you think?

sickbeard/name_parser/regexes.py
```diff
@@ -116,7 +115,7 @@
     ('stupid',
      # tpz-abc102
      r'''
-     (?P<release_group>.+?)-\w+?[\. ]?           # tpz-abc
+     (?P<release_group>.+?)(?<!WEB)-\w+?[\. ]?   # tpz-abc
```
fernandog: @duramato this fix was already merged right?

duramato: yes i use it as a source for this PR

fernandog:
its just a new name for "season"

labrys:
I'm always for more unit testing, however in this case I agree with @fernandog that this is a fairly minor change to represent how seasons are named for a specific UK-based provider.  This kind of change isn't easily represented in a unit test.

duramato:
@labrys this isnt just specific to a tracker :P , all english countries non US name it like this series, at least the UK and AU.
That is releases "native" from those countries, so if a tracker is from one of those it will follow it too, linguistics! :)

labrys:
@duramato I'm aware its a localization difference, however to my knowledge the only currently supported provider that uses this convention is TVCUK.

labrys:
Also one minor change to the regex, it should be a non-matching group as per the original regex.

labrys:
On a related note do we want to support `Season` in other languages as well? For example `Seizoen`.

OmgImAlexis:
@labrys as the new show page states "This will only affect the language of the retrieved metadata file contents and episode filenames. This DOES NOT allow Medusa to download non-english TV episodes!"

Maybe we should open a FR for supporting other languages instead people needing to use the require/ignore list to get other languages downloaded?

duramato:
@labrys depends on the number of languages relases they use. We could support and would make users "happy" i dont know any other, but if any of you know tell me so i can add or any one else. If you notice i only added to the verbouse regex becouse usually they either are S01E01 ( should be ok regatdless of language) or with words Season/Series/Seizoen, the verbouse.
Funny enough this release isnt from tvcuk , tvcuk releases name the pack series 1 but episodes are named as usual "S01e01"

p0psicles:
Approved

labrys:
@p0psicles you merged this, but what about this being a non-matching as per the original regex?

fernandog:
https://regex101.com/r/iQ8pR9/1

duramato:
Ill change it to non matching in a few

labrys:
@fernandog You're missing the point.
Current regex https://regex101.com/r/iQ8pR9/1:
```
series_name	[0-9]	`Show.Name`
2.	[10-16]	`Season`
season_num	[17-18]	`1`
ep_num	[27-28]	`2`
extra_info	[29-36]	`Ep.Name`
```
Non-matching group version https://regex101.com/r/iQ8pR9/4:
```
series_name	[0-9]	`Show.Name`
season_num	[17-18]	`1`
ep_num	[27-28]	`2`
extra_info	[29-36]	`Ep.Name`
```
It removes the unnamed 2nd group, making the returned match tuple unchanged from the previous implementation.

fernandog:
Afaik parser only read named groups

labrys:
You're probably right, but without being certain, this is a safer bet.  In addition, there's no need to capture the information, so no reason not to make it non-capturing.  Finally if we did choose to make it a capture group, it should be done as a named group to remain consistent with the other capturing groups in the regex.

p0psicles:
Hmm sorry i totally do not understand why this is still a long discussion. Okay, I fucked up! Didn't notice the matching-group. But now changing it to: ^(?P<series_name>.+?)[. _-]+(?:season|series)[. _-]+(?P<season_num>\d+)[. _-]+episode[. _-]+(?P<ep_num>\d+)[. _-]+(?P<extra_info>.+)$ will correct it right?

As far as I understand, the current PR won't break anything, but it's not consistent. Let's push another commit, and move on?

fernandog:
Parser only read named groups as I said
https://github.com/pymedusa/SickRage/blob/develop/sickbeard/name_parser/parser.py#L125-L182

p0psicles:
I agree with fernandog, if this were really in conflict with standards & guidelines or it would have a functional or non-functional impact, we should discuss it, but as it's not described anywhere, let's move on.

PR #516, opened 2016-04-30
Remove useless dupe code from tvcuk

duramato:
TBMerged after https://github.com/pymedusa/SickRage/pull/497


fernandog:
Added in the other open PR as a new commit

PR #522, opened 2016-05-02
Cache is already being created at startup

fernandog:
Issue from old repo. This was added when the name cache creation was removed from SR startup. Then it was reverted and "someone" forgot to remove this line

this was fully tested by me and @duramato in `pubdate2` branch for the past 4 days


duramato:
Approved

Issue #495, opened 2016-04-27
[ FEATURE ] Add websockets


OmgImAlexis:
If websockets were added then we could use them to update the page as new changes come along instead of having to rely on polling the server for new changes. This would also fix the `get_messages` spam that we currently have.

p0psicles:
Do you have some examples? Or possibly docs for implementing it in python? I found this https://pypi.python.org/pypi/websockets, but that's min python 3.3

OmgImAlexis:
Have a look at the docs for tornado, from what I can see there's already an implementation built in, you just need to use it like we already do with normal http requests.

www.tornadoweb.org/en/stable/websocket.html

p0psicles:
Ooh cool! I'll see if I can replace the messages with this as a POC

p0psicles:
How should we create the websocket route? Ive got something working using /ws/(.*). But maybe you also want to make this part of the versioned API?

OmgImAlexis:
Maybe we just use `/ws/` and then in the data we add in what it's for.

Like this and then we can just check if the message we get from the server has the event we need.
````
{
    "event": "notification",
    "data": {
        "title": "Test",
        "body": "This is a test notification!"
    }
}
````

**Comment by labrys** [comment]
@OmgImAlexis That sounds good.  If in the future we find that its needed we can extend this to additional routes if needed, possibly even mirroring the APIv2 endpoints for route-specific websockets.

**Comment by p0psicles** [comment]
hmm just discovered that the ws is opened blocking other requests. Meaning it's running under the Tornado thread. But looping in the open() method like shown below, will not work. Maybe create a new thread specific for checking the ui messages?

```
class WebSocketUIHandler(WebSocketHandler):
    def check_origin(self, origin):
        return True

    def open(self, *args, **kwargs):
        clients.append(self)
        WebSocketHandler.open(self, *args, **kwargs)
        while True:
            messages = {}
            cur_notification_num = 1
            for cur_notification in notifications.get_notifications(self.request.remote_ip):
                messages['notification-' + str(cur_notification_num)] = {'title': cur_notification.title,
                                                                         'message': cur_notification.message,
                                                                         'type': cur_notification.type}
                cur_notification_num += 1
            if messages:
                self.write_message(messages)
            time.sleep(1)

    def on_message(self, message):
        print(u"Received: " + message)

    def data_received(self, chunk):
        WebSocketHandler.data_received(self, chunk)

    def on_close(self):
        clients.remove(self)
[...]

PR #526, opened 2016-05-03
POC: First working implementation of websockets.

p0psicles:
This is a working POC, for replacing the current REST polling mechanisme we use for the Pnotifications, with Websockets. For more info please refer to FR: #495.

Currently it's just a POC and based on #432 , for discussing the actual implementation.
Replaced the Rest call used for the notification, with the websockets.

### DIFF
--- a/static/js/ajax-notifications.js
+++ b/static/js/ajax-notifications.js
@@ -44,11 +44,17 @@ function wsCheckNotifications() {
             displayPNotify('info', '', msg);
         }
     };
+
+    ws.onerror = function(evt) {
+        console.log('Error connecting to websocket. Please check your network connecton: ' + evt);
+        displayPNotify('notice', 'Error connecting to websocket.', 'Please check your network connection. ' +
+            'If you are using a reverse proxy, please take a look at our wiki for config examples.');
+    };
 }
 
 $(document).ready(function() {
     wsCheckNotifications();
     if (test) {
-        displayPNotify('notice', 'test', 'test<br><i class="test-class">hello <b>world</b></i><ul><li>item 1</li><li>item 2</li></ul>');
+        displayPNotify('error', 'test', 'test<br><i class="test-class">hello <b>world</b></i><ul><li>item 1</li><li>item 2</li></ul>');
     }
 });
--- a/static/js/ajax-notifications.js
+++ b/static/js/ajax-notifications.js
@@ -27,7 +27,8 @@ function displayPNotify(type, title, message) {
 }
 
 function wsCheckNotifications() {
-    var ws = new WebSocket('ws://' + window.location.hostname + ':' + window.location.port + '/ws' + WSMessageUrl);
+    var proto = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
+    var ws = new WebSocket(proto + '//' + window.location.hostname + ':' + window.location.port + '/ws' + WSMessageUrl);
     ws.onmessage = function(evt) {
         var msg;
         try {
--- a/static/js/ajax-notifications.js
+++ b/static/js/ajax-notifications.js
@@ -45,8 +45,9 @@ function wsCheckNotifications() {
         }
     };
 
-    ws.onerror = function(evt) {
-        console.log('Error connecting to websocket. Please check your network connecton: ' + evt);
+    ws.onerror = function() {
+        log.warn('Error connecting to websocket. Please check your network connection. ' +
+            'If you are using a reverse proxy, please take a look at our wiki for config examples.');
         displayPNotify('notice', 'Error connecting to websocket.', 'Please check your network connection. ' +
             'If you are using a reverse proxy, please take a look at our wiki for config examples.');
     };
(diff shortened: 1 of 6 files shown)


sickrage/ws/MedusaWebSocketHandler.py
```diff
@@ -0,0 +1,48 @@
...
+class WebSocketUIHandler(WebSocketHandler):
+    def check_origin(self, origin):
+        return True
+
+    def open(self, *args, **kwargs):
+        clients.append(self)
+        WebSocketHandler.open(self, *args, **kwargs)
+
+    def on_message(self, message):
+        print(u"Received: " + message)
```
p0psicles: I know i know, just for debugging currently

sickrage/ws/MedusaWebSocketHandler.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def push_to_web_socket(msg):
+    for client in clients:
+        client.write_message(msg)
+
+
+class WebSocketUIHandler(WebSocketHandler):
+    def check_origin(self, origin):
+        return True
```
p0psicles: This should be changed, bringing in some security

OmgImAlexis: This will be fine since we don't accept anything server side. This is really only a XSS issue and since we only broadcast to the client there's nothing that XSS could cause.

sickrage/ws/MedusaWebSocketHandler.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+import json
+import time
+from tornado.websocket import WebSocketHandler
+
+clients = []
+
+
+def push_to_web_socket(msg):
```
p0psicles: This is the function that can be included and used to write to the ws. In this case i'm sending a notification. But you could already start sending other jsons.

sickbeard/ui.py
```diff
@@ -39,7 +41,13 @@ def message(self, title, message=''):
         title: The title of the notification
         message: The message portion of the notification
         """
-        self._messages.append(Notification(title, message, MESSAGE))
+        #self._messages.append(Notification(title, message, MESSAGE))
+        new_notification = Notification(title, message, MESSAGE)
+
+        push_to_web_socket(json.dumps({'event': 'notification',
```
p0psicles: Here it is sending the json. Previously it was queueing up the notifications, and keeping track of the clients with there source ip, to make sure if there were two clients, client one did not remove the notification from the queue, and as a result making sure client two could not get the notification anymore.

I believe by echoing to all clients, we don't need this logic anymore?!

p0psicles:
@pymedusa/developers As we've decided that removing the messages requests is high prio, and scheduled for the 0.1.1 milestone, i'd like everyone to start looking at the implementation, and give feedback. Any is welcome.

OmgImAlexis:
@fernandog why didn't you update the branch since most of those errors have been fixed already?

OmgImAlexis:
@p0psicles are you able to fix this PR to have only the WS stuff and remove all the extra apiv2 commit, etc.
It also needs to be rebased on the remove-ek and the move to static PRs which should both be in develop in a day or two.

p0psicles:
Yeah will be working on this next

fernandog:
@p0psicles rebased

original here:  fernandog/Medusa/commits/apiv2-websockets

fernandog:
@p0psicles need to test this PR?

p0psicles:
No it needs to be migrated to https://github.com/pymedusa/frontend.
@OmgImAlexis and i decided, to add it there.

OmgImAlexis:
I think for now we should be able to merge this in as the frontend will also need the backend code that this adds.

fernandog:
@p0psicles merge then?

p0psicles:
It's a poc.havent tested this in 4 months.  Its nowhere good enough for production

static/js/ajax-notifications.js
```diff
@@ -26,13 +26,13 @@ function displayPNotify(type, title, message) {
...
-function WSCheckNotifications() {
+function wsCheckNotifications() {
     var ws = new WebSocket('ws://' + window.location.hostname + ':' + window.location.port + '/ws' + WSMessageUrl);
-    ws.onmessage = function (evt) {
+    ws.onmessage = function(evt) {
         var msg;
         try {
             msg = JSON.parse(evt.data);
-        } catch(e) {
+        } catch (err) {
```
fernandog: @OmgImAlexis shouldn't we `log.error(err);` here?

OmgImAlexis: No and this isn't `err` this is `e`. err should return a new error object where as this is an exception object. There's not actually any errors as this is used to check if the string if a json encoded string or just a normal string.

If xo is giving an error just add this to the right of it.
`// eslint-disable-line-only ADD_ERROR_NAME_HERE`

OmgImAlexis: @fernandog Can you revert this line's change?

OmgImAlexis:
@pymedusa/developers @pymedusa/administrators @pymedusa/contributors 

Can you guys please try this so it can be merged as it's been open for almost a year now.

static/js/ajax-notifications.js
fernandog: Sure

OmgImAlexis:
@p0psicles can we please merge this into dev soon?

PR #537, opened 2016-05-04
[Manual Search] Add 'proper' to default sorting

fernandog:
Because PROPERs are better among same qualities results

Client side sorting wont be added soon (table sorter). And this only updates the default sorting (seeders and quality)

@duramato @labrys @p0psicles @medariox  if no reason against I will merge in the end of day

### DIFF
--- a/sickbeard/manual_search.py
+++ b/sickbeard/manual_search.py
@@ -202,8 +202,12 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
 
             common_sql = "SELECT rowid, ? as 'provider_type', ? as 'provider_image', \
                           ? as 'provider', ? as 'provider_id', ? 'provider_minseed', ? 'provider_minleech', \
-                          name, season, episodes, indexerid, url, time, (select max(time) \
-                          from '{provider_id}') as lastupdate, \
+                          name, season, episodes, indexerid, url, time, \
+                          CASE WHEN LOWER(name) LIKE '%proper%' THEN 1 \
+                               WHEN LOWER(name) LIKE '%real%' THEN 1 \
+                               WHEN LOWER(name) LIKE '%repack%' THEN 1 \
+                                ELSE 0 END AS 'is_proper', \
+                          (select max(time) from '{provider_id}') as lastupdate, \
                           quality, release_group, version, seeders, leechers, size, time \
                           FROM '{provider_id}' WHERE indexerid = ?".format(provider_id=cur_provider.get_id())
             additional_sql = " AND episodes LIKE ? AND season = ?"
@@ -246,7 +250,7 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
         time.sleep(cpu_presets[sickbeard.CPU_PRESET])
     else:
         # Sort the list of found items
-        found_items = sorted(found_items, key=lambda k: (try_int(k['quality']), try_int(k['seeders'])), reverse=True)
+        found_items = sorted(found_items, key=lambda k: (try_int(k['quality']), try_int(k['is_proper']), try_int(k['seeders'])), reverse=True)
         # Make unknown qualities at the botton
         found_items = [d for d in found_items if try_int(d['quality']) < 32768] + [d for d in found_items if try_int(d['quality']) == 32768]
         provider_results['found_items'] = found_items


medariox:
I think that giving Proper results priority in the ordering makes sense, as those should be normally preferred to non-proper releases. Don't like the "Proper" tag next to the result though, I think the manual search layout is already quite packed and it's starting to get a bit messy imo.

labrys:
@medariox If we implemented a table with sorting/filtering/etc as per the other tables then it would still be appropriate to have the column.  The user could just choose to disable the columns they aren't interested in.  For example I use NZBs so the seeds, peers, and type columns are useless to me.

fernandog:
I removed tag. Now it only sorts by proper.

p0psicles:
Imo, you should make the choice, to put sorting logic in server or client. As you already know we've decided to implement sorting client side, you should really keep it to a minimum server side. Thats just plain architecture.

fernandog:
@p0psicles Im only adding to the existing default sort. If later user want to sort using something else, its his choice. And we wont have table sorter soon.

sickbeard/manual_search.py
```diff
@@ -246,7 +250,7 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
         time.sleep(cpu_presets[sickbeard.CPU_PRESET])
     else:
         # Sort the list of found items
-        found_items = sorted(found_items, key=lambda k: (try_int(k['quality']), try_int(k['seeders'])), reverse=True)
+        found_items = sorted(found_items, key=lambda k: (try_int(k['quality']), try_int(k['is_proper']), try_int(k['seeders'])), reverse=True)
         # Make unknown qualities at the botton
         found_items = [d for d in found_items if try_int(d['quality']) < 32768] + [d for d in found_items if try_int(d['quality']) == 32768]
```
p0psicles: And this, is a lot of sorting logic, for something that should be really light. Not to mention the sql that's also become quite large. You really want to move as much processing from server to client as possible. Allot of the sql and lambda soring, can be done fine in client.

sickbeard/manual_search.py
```diff
@@ -246,7 +250,7 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
         time.sleep(cpu_presets[sickbeard.CPU_PRESET])
     else:
         # Sort the list of found items
-        found_items = sorted(found_items, key=lambda k: (try_int(k['quality']), try_int(k['seeders'])), reverse=True)
+        found_items = sorted(found_items, key=lambda k: (try_int(k['quality']), try_int(k['is_proper']), try_int(k['seeders'])), reverse=True)
```
p0psicles: This

sickbeard/manual_search.py
fernandog: No way to do this kind of sort with table sort. This was decided a long time ago. This is the default sorting when loading page. So why so many discussions in a simple PR?

p0psicles:
It's not that i'm against the PR. I'm just sharing my opinion on it, from an architectural pov. Personal I think it has a very low prio, as the server side sorting, isn't important IMHO.

fernandog:
@p0psicles 
only because its low priority i cant be approved/merged?

fernandog:
Approved

fernandog:
@p0psicles did you read my last comment?

sickbeard/manual_search.py
labrys: This kind of sorting is fine to do server side instead of client side if the sorting is static, but if it is done server-side it shouldn't be done with a lambda, it should be done during the SQL select.

p0psicles:
I'm going to close this one, as whe are moving on in the new PR. It does already have soring by Proper. Altough the feature is still debatable, because it's not a cache db field. And doing queries where we need to do string parsing, is not something we should want.

fernandog:
At least proper is in the sort. That was what I wanted. Doesnt matter which PR.

p0psicles:
yeah, but also in the other PR, the solution is ugly. If we decide proper is a valid search result attribute, we should add it as a db field. So it's still debatable if we should allow it to be filtered like its now

labrys:
@p0psicles I agree with you, if we plan on filtering by proper, we should really make it a separate db field, for reliability, consistency, and efficiency.

Issue #392, opened 2016-04-15
[ The road to RC ] Todo's


p0psicles:
This list will be used as a roadmap towards RC and later the merge to Master.
The list is meant for finishing already started features or bugfixes.

### General
- [x] Bump Database version

### Manual Search
- [x] Episode snatched with manual search doesn't change its status to snatched in show overview  https://github.com/pymedusa/SickRage/pull/451
- [x] Result snatches should be queued in a separate thread, as opposed to currently being queued in the searchQueueScheduler thread. #395 
- [x] A forced search (also used by manual search) should be queued in a separate thread, as opposed to currently being queued in the searchQueueScheduler thread. #366 
- [x] The snatchSelection page should display snatch/download/failed download history, like how it's displayed in the displayShow page.
- [X] Disable season search if season has still unaired episodes
- [X] Add season pack manual searching #389

### General bugs
- [x] ~~Fix URL limit (I guess that's about  https://github.com/pymedusa/SickRage/pull/434)~~
- [x] Queue show removals, as to wait for all threads to finish and make sure it can be removed safe #244 #567 
- [x] Fix unicode issues during post processing #390 
- [x] Fix tmdb certificate error https://github.com/pymedusa/SickRage/pull/145/commits/a6a504a12353f67bacf9ab42ad6d7210e2c13530
- [x] Not saving NZB cats
- [x] Check IMDB for showinfo results in keyErrors like KeyError: 'runtimes' #379 #391 #385
- [x] Updated ConfigObj does not properly save non-ascii values. (Revert for RC until we can finish the ConfigObj validators) @medariox can you verify this is fixed now?

### Subliminal
- [x] Limit the amount of "2016-04-15 10:18:22 DEBUG    FINDSUBTITLES :: [1f87e70] Found subtitle" messages. Found subtitle isn't really "informative". @ratoaq2 
- [x] Improve performance or throttle subtitleFinder thread cpu usage. As a search is bringing my dualcore Atom on it knees. Wonder what it will do to a low end arm device. https://github.com/pymedusa/SickRage/pull/472

@pymedusa/contributors

fernandog:
@p0psicles 

"The snatchSelection page should display snatch/download/failed download history, like how it's displayed in the displayShow page" - with this, page will take a LONG TIME run. For each result, we will need to query full history of user

labrys:
You don't need to query the full history, just the history for that particular episode.

fernandog:
[...]

PR #539, opened 2016-05-05
Fix pymedusa/SickRage/issues/533

fernandog:
now populateNewznabSection is only called in the tab "Configure custom newznab providers" and only for the provider user selected in the select object.

Today if user refreshes provider list it calls populateNewznabSection for all nzb providers.

@p0psicles

### DIFF
--- a/gui/slick/js/configProviders.js
+++ b/gui/slick/js/configProviders.js
@@ -73,7 +73,6 @@ $(document).ready(function(){
         newznabProviders[id] = newData;
 
         $('#editANewznabProvider').addOption(id, name);
-        $(this).populateNewznabSection();
 
         if ($('#provider_order_list > #'+id).length === 0 && showProvider !== false) {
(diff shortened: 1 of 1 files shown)


p0psicles:
Only downside of this, is that the user needs to refresh the page, before he can modify the categories. But IMO i think that's an acceptable trade-off. I didn't notice anything strange. But because the code is messy, it's hard to assess the impact and risks when doing changes.

p0psicles:
Approved

labrys:
I'd like to try it out to see how the reloading affects the user-flow before I weigh in on it.  If we do decide to go with a reload, a UI note should be included.

fernandog:
@labrys Which reload?  When you save medusa will always open providers list. Its always been like that. Nothing chaged.

@p0psicles I didnt had to refresh page. Didnt understood.

p0psicles:
no @fernandog it's just as your saying. But when you added a new provider, in the past, you could set the categories before saving it. Now you have to save it (which will make it to refresh the page, like  always) after which you can change the categories. So yes, there is a slight change to user flow. But IMO it's really minimal, and acceptable, because I know that we're rewriting the providers page in a few weeks anyway.

fernandog:
@p0psicles I remember we had a button to refresh categories. I was a long time since i used those tabs

p0psicles:
#392

PR #541, opened 2016-05-05
Use cached video only when keyword args match

ratoaq2:
- Just after a download finishes, the initial subtitle search doesn't have release name and the video is cached with incomplete info (e.g. without release group). When the "FINDSUBTITLE" thread takes place, the incomplete cached video might prevent the subtitle download until it expires, so subtitles are only downloaded after 24h.
- This issue is noticeable when using "episode renaming rules" that doesn't contain "quality" information in the filename (release group, screen size, etc)

- This fix will keep the video cache key as the video path. But the cached video will be invalidated/discarded if any keyword argument changes: in this case the video is scanned again and re-created (the same process as before). I've been testing this for about 10 days now, and it works like a charm.

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -553,10 +553,12 @@ def get_video(video_path, subtitles_dir=None, subtitles=True, embedded_subtitles
     :rtype: subliminal.video.Video
     """
     key = video_key.format(video_path=video_path)
-    video = region.get(key, expiration_time=VIDEO_EXPIRATION_TIME)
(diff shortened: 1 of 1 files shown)


fernandog:
@ratoaq2 im testing this in my pubdate2 branch that im also testing others PR and the PR itself

medariox:
Approved

Code looks good to me and I have been running this without any issues so far.

labrys:
Approved

Same here.

p0psicles:

PR #547, opened 2016-05-06
Fix deluge sending ratio -1 when it should disable stop_at_ratio setting

fernandog:
Fix from old repo: SickRage/sickrage-issues/issues/506

-1 is to seed forever so we should disable that stop_at_ratio setting
blank is default client ratio, so we also shouldn't set ratio

image

@ratoaq2 will test when he can

Transmission does the same:
https://github.com/pymedusa/SickRage/blob/old_issue506/sickbeard/clients/transmission_client.py#L108-L109

### DIFF
--- a/sickbeard/clients/deluge_client.py
+++ b/sickbeard/clients/deluge_client.py
@@ -167,12 +167,17 @@ class DelugeAPI(GenericClient):
         if result.ratio:
             ratio = result.ratio
 
-        if ratio:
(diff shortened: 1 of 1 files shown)


fernandog:
@ratoaq2 do you have any ETA to test this? thanks!

ratoaq2:
Tested and it works just fine :+1:

labrys:
Approved

PR #548, opened 2016-05-06
Do not log warning when no results exist in RARBG

duramato:
No need to show warning when show doesnt exist in rarbg , same as done for erro code 20 `No results found` when a show exists in the site already.
error code 14 is preety much the same but its becouse the show isnt even in the site, so no results at all `Cant find thetvdb in database. Are you sure this thetvdb exists? 14 `

### DIFF
--- a/sickbeard/providers/rarbg.py
+++ b/sickbeard/providers/rarbg.py
@@ -1,8 +1,6 @@
 # coding=utf-8
 # Author: Dustyn Gibson <miigotu@gmail.com>
 #
-
-#
 # This file is part of SickRage.
 #
 # SickRage is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


sickbeard/providers/rarbg.py
```diff
@@ -136,7 +134,7 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
                 # Don't log when {"error":"No results found","error_code":20}
                 # List of errors: https://github.com/rarbg/torrentapi/issues/1#issuecomment-114763312
                 if error:
-                    if try_int(error_code) != 20:
+                    if try_int(error_code) != 20 or try_int(error_code) != 14:
```
p0psicles: Ok Let me rephrase this.. You want to raise a warning for All error codes, but not 20 and 14? In that case just do a: not in (20, 14)

fernandog:
Approved

labrys:
Approved

[Approved with PullApprove](https://pullapprove.com/pymedusa/SickRage/pull-request/548/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=pymedusa/SickRage)

PR #553, opened 2016-05-07
Hide TooManyRequests from RARBG

fernandog:
As requested @medariox 

Discussion:
`PROPER thread doesn't have a scheduler`
https://github.com/pymedusa/SickRage/issues/534


medariox:
I think it would be a better idea to start re-thinking how the proper-finder works, if we really need it to run in parallel or if it can be implemented in a different way.

labrys:
I'm not a fan of hiding warnings unless you are raising them intentionally (e.g. like the timeout during force updates of limetorrent info).

medariox:
Let's continue the discussion here, before changing the warning: https://github.com/pymedusa/SickRage/issues/534

PR #556, opened 2016-05-08
Remove delay from subtitles as it slows down not only subs, but all t…

labrys:
- [ ] Always make sure the PR has a clear description of why,how, and what your trying to fix. The same goes for improvements or new features. Also if you refering to an issue, make sure a summary is available in the PR.

…hreads and UI.

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -26,11 +26,9 @@ import re
 import sickbeard
 import subprocess
 import traceback
-import time
 
 from babelfish import Language, language_converters
 from dogpile.cache.api import NO_VALUE
-from sickbeard.common import cpu_presets
 from sickrage.helper.common import dateTimeFormat, episode_num, subtitle_extensions
 from sickrage.helper.exceptions import ex
 from sickrage.show.Show import Show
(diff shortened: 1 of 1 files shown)


fernandog:
@labrys 
How come it slow UI if it only adds a delay before searching for subs? What UI has to do with it?
Its this code is a problem why not remove from ALL files in Medusa? This is default code to CPU Throttling

labrys:
@fernandog I'm not going to get into all the intricacies of threading, GIL, blocking, etc.  Suffice it to say that any time the subtitle queue is started all of medusa slows to a crawl.  Running a go-kart (i.e. low powered arm, etc) like a race car, of course there would be issues, but that doesn't mean that you should cripple a racecar to make it run like a go-kart.  There are much better solutions to time.sleep, but as you are fond of saying, that's beyond the scope of this PR.

p0psicles:
Hi Labrys. I'm also not sure where this comes from? As I am aware we're only blocking the FINDSUBTITLES thread. As far as I know, there is no direct interaction with this thread from UI.

Please, we really need some more info, as we're removing the times.sleep that's used as a throttling technique for many other threads, like for ex. also the SearchSchedulers. If this is really a big issue, they also should be removed there.

labrys:
@p0psicles hop on IRC so we can discuss this.

p0psicles:
To summarize the discussion. It seems that the general use of the time.sleep() in threads, has a bigger then wanted impact on overall system performance, while it should only impact the thread where it's implemented. The exact reason is still not clear (at least for me).

We could say that on a low end device like (rasp pi or other arm devices), the sleep is really needed, else a subtitle search will result in a total unresponsive system.

On higher end systems, the sleep will result in overall bad performance, also in UI.

I don't think we should implement a "quick fix" that will result in regression. Therefore I propose to close this one, and continue the discussion in an Issue.

As a workaround it has been discussed to add a new CPU throttling profile, which will disable the sleep all together? Am I correct? If so, that should be further discussed in the opened Issue.

p0psicles:
As "solved" in a separate PR, i'm closing this one. If this is still valid, we can reopen.

PR #566, opened 2016-05-09
Fix scene name unicode save error

p0psicles:
How to reproduce the issue:
1. Run medusa on windows. I've used Windows 7
2. Make sure used locale is x.cp1252
3. Add Naruto Shippuuden
4. Now you can open the show in displayShow and editShow just fine.
5. Go to editShow and save the show, you should be redirected to displayShow
6. You'r getting a mako 500 error page.

image

- [ ] Always make sure the PR has a clear description of why,how, and what your trying to fix. The same goes for improvements or new features. Also if you refering to an issue, make sure a summary is available in the PR.

### DIFF
--- a/sickbeard/scene_exceptions.py
+++ b/sickbeard/scene_exceptions.py
@@ -273,9 +273,9 @@ def update_scene_exceptions(indexer_id, scene_exceptions, season=-1):
     # A change has been made to the scene exception list. Let's clear the cache, to make this visible
     if indexer_id in exceptionsCache:
         exceptionsCache[indexer_id] = {}
-        exceptionsCache[indexer_id][season] = scene_exceptions
+        exceptionsCache[indexer_id][season] = [se.decode('utf-8', 'ignore') for se in scene_exceptions]
 
-    for cur_exception in scene_exceptions:
+    for cur_exception in [se.decode('utf-8', 'ignore') for se in scene_exceptions]:
         cache_db_con.action("INSERT INTO scene_exceptions (indexer_id, show_name, season) VALUES (?,?,?)",
                             [indexer_id, cur_exception, season])
 
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -205,6 +205,10 @@ class PageTemplate(MakoTemplate):
             kwargs['title'] = '500'
             kwargs['header'] = 'Mako Error'
             kwargs['backtrace'] = RichTraceback()
+            for (filename, lineno, function, line) in kwargs['backtrace'].traceback:
+                print("File %s, line %s, in %s" % (filename, lineno, function))
+                print(line, "\n")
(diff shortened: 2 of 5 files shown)


sickbeard/scene_exceptions.py
```diff
@@ -273,9 +273,9 @@ def update_scene_exceptions(indexer_id, scene_exceptions, season=-1):
     # A change has been made to the scene exception list. Let's clear the cache, to make this visible
     if indexer_id in exceptionsCache:
         exceptionsCache[indexer_id] = {}
-        exceptionsCache[indexer_id][season] = scene_exceptions
+        exceptionsCache[indexer_id][season] = [se.decode('utf-8', 'ignore') for se in scene_exceptions]
```
p0psicles: @labrys @medariox your both the unicode guru's here. Is this the way to solve this? It seems to work for me now? Still have to test on other linux with utf8.

sickbeard/webserve.py
```diff
@@ -205,6 +205,10 @@ def render(self, *args, **kwargs):
             kwargs['title'] = '500'
             kwargs['header'] = 'Mako Error'
             kwargs['backtrace'] = RichTraceback()
+            for (filename, lineno, function, line) in kwargs['backtrace'].traceback:
+                print("File %s, line %s, in %s" % (filename, lineno, function))
+                print(line, "\n")
+            print("%s: %s" % (str(kwargs['backtrace'].error.__class__.__name__), kwargs['backtrace'].error))
```
p0psicles: OOps, these where just for debugging purposes. I'll remove them.

p0psicles:
I've PR'd his one towards fix-bwl. When that's merged, i'm going to create a new one for dev. In the mean time, we can discuss here.

fernandog:
https://github.com/pymedusa/SickRage/pull/557

sickbeard/scene_exceptions.py
medariox: The question is: why is scene_exceptions a list of encoded items and not decoded items (like it should be)? It would be best to decode scene_exceptions at the "source". About decode('utf-8', 'ignore') I would personally prefer decode('utf-8', 'replace'), but - until we finish the transition to unicode everywhere - there is nothing wrong about it.

p0psicles: It should be unicode encoded right? Else you can't display the chinese chars right?
I personally think it's an issue with tornado, it is processing the post by default as encoded in str(). While it should be encoded in unicode.

medariox: Yes, I would say the problem is that when you add an exception that contains chars that your locale doesn't support (in your case cp1252) the [join() that takes place in displayShow](https://github.com/pymedusa/SickRage/blob/master/gui/slick/views/displayShow.mako#L186) tries to decode them with the default locale, but fails.

sickbeard/webserve.py
p0psicles: @labrys i'd like to keep these, but then as debug loggers, for mako debugging. What you think?

PR #567, opened 2016-05-10
Added a global list that's updated with shows that have been deleted.

p0psicles:

Added a global list that's updated with shows that have been (or in the process) of being deleted. The list can be used in combination with a TV.TVShow() object, to check if it has been recently deleted.

Fixed the warning in manual search, where the /getManualSearchStatus was trying to still retrieve snatch/download status for a show, that has been removed. It's now only showing, when it can't find a show, that is also not recently removed. This removes the warning, reported in: https://github.com/pymedusa/SickRage/issues/244

I've reproduced and tested through:
1. Opening a show
2. Opening a second tab of the show
3. Rappidly forced search a couple of episodes in tab 1
4. Then removed the show in tab 1
5. Tab 2 is still open, and previously would generate the warning. Now it won't anymore.

I didn't notice any other errors. But I didn't test in combination with daily or backlog search.


p0psicles:
@OmgImAlexis as issue owner, can you please also verify this resolves your issue?

labrys:
@p0psicles I would suggest a set instead of a list for this since a show should only exist uniquely in the set to keep the list from having duplicates (e.g. a user clicks the delete multiple times)

p0psicles:
What's the impact of the show being there multiple times?

labrys:
@p0psicles None that immediately come to mind.  It's more of a logic thing, for example if in the future we modify the code to remove a show from the deleted-list if the show gets re-added, we don't have to search for multiple occurrences in the list.

p0psicles:
Yes fair argument. I'll change it.

p0psicles:
@labrys also this one?

sickbeard/tv.py
```diff
@@ -166,9 +166,7 @@ def network_logo_name(self):
 
     @property
     def is_recently_deleted(self):
```
labrys: I probably wouldn't even bother with this.

p0psicles: Nah, but it makes for readable code. As now your 'forced' to use the TVobject, as opposed to do the check on a random indexerid. Also in the future if you want to do a more advanced check, or make changes to the way you check if a show has been deleted, you only need to change here.

p0psicles:
Approved

p0psicles:

PR #569, opened 2016-05-10
"Fix" propers being downloaded in dupe from torrentz becouse title wo…

duramato:
…uld differ

**You might ask:**
_Why replace with a - after the codec. _Scene releases always "end" with codec followed  by the group, so "...x264 KILLERS" turns to "...x264-KILLERS"
_Why not add more codecs?_ these are the only one released by the Scene

As talked @fernandog

### DIFF
--- a/sickbeard/providers/torrentz.py
+++ b/sickbeard/providers/torrentz.py
@@ -1,8 +1,6 @@
 # coding=utf-8
 # Author: Dustyn Gibson <miigotu@gmail.com>
 #
-
-#
 # This file is part of SickRage.
 #
 # SickRage is free software: you can redistribute it and/or modify
@@ -90,7 +88,7 @@ class TorrentzProvider(TorrentProvider):  # pylint: disable=too-many-instance-at
                             if item.category and 'tv' not in item.category.get_text(strip=True):
                                 continue
 
-                            title = item.title.text.rsplit(' ', 1)[0].replace(' ', '.')
+                            title = item.title.text.replace('x264 ', 'x264-').replace('h264 ', 'h264-').replace('h 264 ', 'h264-').replace('xvid ', 'xvid-').replace(' ', '.')
                             t_hash = item.guid.text.rsplit('/', 1)[-1]
 
                             if not all([title, t_hash]):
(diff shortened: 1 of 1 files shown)


fernandog:
Approved

[Approved with PullApprove](https://pullapprove.com/pymedusa/SickRage/pull-request/569/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=pymedusa/SickRage)

sickbeard/providers/torrentz.py
```diff
@@ -90,7 +90,7 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
                             if item.category and 'tv' not in item.category.get_text(strip=True):
                                 continue
 
-                            title = item.title.text.rsplit(' ', 1)[0].replace(' ', '.')
+                            title = item.title.text.rsplit(' ', 1)[0].replace('x264 ', 'x264-').replace('h264 ', 'h264-').replace('h 264 ', 'h264-').replace('xvid ', 'xvid-').replace(' ', '.')
```
labrys: This should be a re.sub instead.  Either reuse regex `re.compile` from the tagger or use `r'(?:[xh][ .]?26[45]|xvid)'`

duramato: Isn't regex "heavier"?

duramato: And also we only want to replace scene releases,  we can't be sure others follow the same naming of "..codec-group"

labrys: Regex is heavier than a single replace, but after repeated replaces and the fact that replace is case sensitive, and regex's would match better, its probably a better choice.  

Is your second point related to the regex recommendation?

duramato: Yes because we only can be sure of scene releases to replace, currently  they only release  in x264, h264 and some xvid, but not for the tv-264 category  ofc. Everything  else if exists it's an internal and aren't nuked/propered
Plus what I said before

labrys: Ok, so then that's not an argument against the regex, per-se, but that the regex should be changed to something like `r'([xh][ .]?264|xvid)'`

p0psicles:
Approved

labrys:
@fernandog nice.  @duramato is detecting a space after required? the regex can be modified to take it into account, however i don't really think it is likely to be necessary.

p0psicles:

PR #570, opened 2016-05-11
Fix pymedusa/SickRage/issues/540

fernandog:
Using the same "solution" found in home.mako

TODO:
- [x] Wrapping first text
- [x] Align text with object
- [x] Text doesn't fit object (also it seems its not inheriting font - guessing)

@p0psicles @OmgImAlexis 

### DIFF
--- a/gui/slick/css/style.css
+++ b/gui/slick/css/style.css
@@ -3495,3 +3495,8 @@ viewLog.mako
     }
 }
 
+/* Because the H1 has a margin of 20px withing the column col-sm-2 of 20px we need
+to configure the same for the columns which are used for the filter inputs.*/
+.log-filter {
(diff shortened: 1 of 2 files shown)


p0psicles:
@fernandog can you check again for me? Don't expect it to be the same as it was. Because i'm using a grid now, we can't stack the input's as a train after one another. But as one of the benifits of using grids, The behavior is much more predictable on different screen sizes. If you do want it like it was, that's possible, but will require allot of custom css.

p0psicles:
Full Screen:
image

From about 1052 px:
image

From 768px:
image

fernandog:
Approved

p0psicles:
Approved

p0psicles:

PR #571, opened 2016-05-11
Spring cleaning

labrys:

I know this seems like a large PR, but the majority of the changes are fairly minor and have been grouped into similar commits to make review easier.

The majority of this PR is just simple style changes to `sickbeard/clients`:
* Fix triple-quoted strings
* Refactor non-pythonic names
* Use single-quoted strings
* Standardize string formatting
* Use unicode literals
* Use b'' for SQL
* Remove unused variables and code
* Clean up logging
* Clean up comments
* Prettify literal indentation
* Fix line too long
* Simplify url init
* Minimize global usage in code

The meat of the PR is the changes below:
* Fix too broad exceptions
 * Catch `RequestException` instead of `Exception`
* Clean up conditional/exceptional statements
 * Mostly a style change to conditionals and exceptional statements.  Included as part of the "meat" since it touches application logic
* Streamline return statements
 * Same as above
* `UTorrentAPI._request' does not accept all arguments from the overridden method.
 * This adds the `cookies` arg with a default value of `None`.  If `cookies` are passed to `_request`, it will log a debug message with the value passed.
* Fix `UTorrentAPI._request()` arguments not matching
* Fix `RTorrentAPI._get_auth` invalidating authentication every call.
* `RTorrentAPI.auth` is set to `None` each time `_get_auth()` is called.
 * This is likely unintended, as the next line returns `self.auth` if it's not `None`.  This moves the invalidation of `self.auth` to `test_authentication()` to make sure the authentication test is actually performed.
* Add missing import
  * Mostly optimizing imports, but adds the missing `import traceback` in `generic.py`
* Add missing method used in DSM client
  * `handle_requests_exception` from `sickbeard/helpers.py` did not exist
* Streamlined `handle_requests_exception`
  * Simplify requests exception handler by catching RequestException instead of all the individual sub-classes.

### DIFF
--- a/sickbeard/clients/deluged_client.py
+++ b/sickbeard/clients/deluged_client.py
@@ -37,10 +37,6 @@ class DelugeDAPI(GenericClient):
         return self.drpc
 
     def _add_torrent_uri(self, result):
-        # label = sickbeard.TORRENT_LABEL
-        # if result.show.is_anime:
-        #     label = sickbeard.TORRENT_LABEL_ANIME
-
         options = {
             'add_paused': sickbeard.TORRENT_PAUSED
         }
@@ -55,10 +51,6 @@ class DelugeDAPI(GenericClient):
         return remote_torrent
 
     def _add_torrent_file(self, result):
-        # label = sickbeard.TORRENT_LABEL
-        # if result.show.is_anime:
-        #     label = sickbeard.TORRENT_LABEL_ANIME
-
         if not result.content:
             result.content = {}
             return None
@@ -150,7 +142,6 @@ class DelugeRPC(object):
         return True
 
     def add_torrent_magnet(self, torrent, options, torrent_hash):
-        torrent_id = False
         try:
             self.connect()
             torrent_id = self.client.core.add_torrent_magnet(torrent, options).get()  # pylint:disable=no-member
@@ -165,7 +156,6 @@ class DelugeRPC(object):
         return torrent_id
 
     def add_torrent_file(self, filename, torrent, options, torrent_hash):
-        torrent_id = False
         try:
             self.connect()
             torrent_id = self.client.core.add_torrent_file(filename, b64encode(torrent), options).get()  # pylint:disable=no-member
--- a/sickbeard/clients/deluged_client.py
+++ b/sickbeard/clients/deluged_client.py
@@ -5,6 +5,8 @@
 # This client script allows connection to Deluge Daemon directly, completely
 # circumventing the requirement to use the WebUI.
 
+from __future__ import unicode_literals
+
 from base64 import b64encode
 
 import sickbeard
@@ -81,7 +83,7 @@ class DelugeDAPI(GenericClient):
         if result.show.is_anime:
             label = sickbeard.TORRENT_LABEL_ANIME.lower()
         if ' ' in label:
-            logger.log(u'{name}: Invalid label. Label must not contain a space'.format
+            logger.log('{name}: Invalid label. Label must not contain a space'.format
                        (name=self.name), logger.ERROR)
             return False
 
@@ -241,7 +243,7 @@ class DelugeRPC(object):
     def _check_torrent(self, torrent_hash):
         torrent_id = self.client.core.get_torrent_status(torrent_hash, {}).get()  # pylint:disable=no-member
         if torrent_id['hash']:
-            logger.log(u'DelugeD: Torrent already exists in Deluge', logger.DEBUG)
+            logger.log('DelugeD: Torrent already exists in Deluge', logger.DEBUG)
             return torrent_hash
         return False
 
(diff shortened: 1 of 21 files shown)


sickbeard/clients/deluged_client.py
```diff
@@ -143,48 +119,48 @@ def test(self):
...
     def add_torrent_file(self, filename, torrent, options, torrent_hash):
-        torrent_id = False
         try:
             self.connect()
             torrent_id = self.client.core.add_torrent_file(filename, b64encode(torrent), options).get()  # pylint:disable=no-member
             if not torrent_id:
                 torrent_id = self._check_torrent(torrent_hash)
         except Exception:
             return False
+        else:
```
p0psicles: placing the else with return here, it will never hit finaly. Is that the intention?

labrys: Finally is ALWAYS executed regardless of any returns, exceptions caught, exceptions missed, etc.

p0psicles: yeah but doubt, that's still the case if you return from the function

p0psicles: Just tested with a snippet. Your right!

labrys: It is... a simple example:
```python
>>> def test(var):
	try:
		var['apple']
	except KeyError as msg:
		print 'KeyError:', msg
	else:
		print 'else'
	finally:
		print 'Finally!'

>>> test(1)
Finally!

  File "<pyshell#35>", line 1, in <module>
    test(1)
  File "<pyshell#34>", line 3, in test
    var['apple']
TypeError: 'int' object has no attribute '__getitem__'
>>> test({'cake': 'is a lie'})
KeyError: 'apple'
Finally!
>>> test({'apple': 'pie is good!'})
else
Finally!
>>> 
```

labrys: Finally clauses are what make context managers (like open()) work.

sickbeard/clients/generic.py
```diff
@@ -1,16 +1,20 @@
...
 import re
 import time
 from hashlib import sha1
 from base64 import b16encode, b32decode
+import traceback
 
-import sickbeard
-from sickbeard import logger, helpers, db
-from bencode import bencode, bdecode
+from six.moves.http_cookiejar import CookieJar
```
p0psicles: Your switching libs? Any reason why?

sickbeard/clients/qbittorrent_client.py
```diff
@@ -65,44 +71,51 @@ def _get_auth(self):
...
 
     def _set_torrent_pause(self, result):
-
-        self.url = self.host + 'command/resume'
-        if sickbeard.TORRENT_PAUSED:
-            self.url = self.host + 'command/pause'
-
-        data = {'hash': result.hash}
+        self.url = '{host}command/{state}'.format(host=self.host,
+                                                  state='pause' if sickbeard.TORRENT_PAUSED else 'resume')
```
p0psicles: I know i seems to look more pythonic. But I can't say it increases the readability. I'm really for keeping it simple. Personally i would have choosen for using format, but keeping the conditions.

sickbeard/clients/rtorrent_client.py
```diff
@@ -32,13 +34,11 @@
...
 
 
-class rTorrentAPI(GenericClient):  # pylint: disable=invalid-name
+class RTorrentAPI(GenericClient):  # pylint: disable=invalid-name
     def __init__(self, host=None, username=None, password=None):
-        super(rTorrentAPI, self).__init__(u'rTorrent', host, username, password)
+        super(RTorrentAPI, self).__init__('rTorrent', host, username, password)
 
     def _get_auth(self):
-        self.auth = None
```
p0psicles: nice catch

p0psicles:
I've reviewed and couldn't find any strange things. I've removed the test label, as I vote for not testing this with every client that has been touched. I've tested it with transmission, and that worked fine.

sickbeard/clients/generic.py
labrys: `six` is a python compatibility library.  It's because some standard libs have changed name between python 2 and python 3.  This is part of making Medusa python 3 compliant.

labrys:
I've also tested with qbittorrent and utorrent.

sickbeard/clients/qbittorrent_client.py
labrys: If you'd prefer I can change it back to a conditional.

labrys:
Approved

sickbeard/clients/qbittorrent_client.py
p0psicles: No don't need to. Not that important. But I do think we should draw a line somewhere, making writing readable and expressive code more important then writing compact code. That way new contributors will have less trouble reading through our code.

p0psicles:
Approved

sickbeard/clients/qbittorrent_client.py
labrys: IMO, that's just as readable.  But that's entirely opinion based.  The advantage to conditionals would have been ease of adding additional states, however i don't think there will be any
new states to add their, so I opted for this way.

labrys:
QC not reporting status - merging.

PR #578, opened 2016-05-12
Removed the lambda, and replaced them with SQL order by. Also sorted by Proper.

p0psicles:

Note! This will probably break stuff as is!!

Had to do allot of strange stuff in SQL, and even had to change UNKOWN value to -1 (just for sorting). So please use this to discuss the alternatives.

### DIFF
--- a/sickbeard/manual_search.py
+++ b/sickbeard/manual_search.py
@@ -238,7 +238,7 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
     # Check if we have the combined sql strings
     if combined_sql_q:
         sql_prepend = b"SELECT * FROM ("
-        sql_append = b") ORDER BY CASE quality WHEN '{quality_unknown}' THEN -1 ELSE quality END DESC, " \
+        sql_append = b") ORDER BY CASE quality WHEN '{quality_unknown}' THEN -1 ELSE CAST(quality as DECIMAL) END DESC, " \
                      b" seeders DESC, isproper DESC".format(quality_unknown=Quality.UNKNOWN)
 
         # Add all results
--- a/sickbeard/manual_search.py
+++ b/sickbeard/manual_search.py
@@ -22,6 +22,7 @@
 import time
 import sickbeard
 import threading
+import json
 from sickbeard import search_queue
 from sickbeard.common import Quality, Overview, statusStrings, cpu_presets
 from sickbeard import logger, db
@@ -233,7 +234,7 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
 
             # Get the last updated cache items timestamp
             last_update = main_db_con.select(b"select max(time) as lastupdate from '{provider_id}'".format(provider_id=cur_provider.get_id()))
-            provider_results['last_prov_updates'][cur_provider.get_id()] = last_update[0]['lastupdate']
+            provider_results['last_prov_updates'][cur_provider.get_id()] = str(last_update[0]['lastupdate'])
 
     # Check if we have the combined sql strings
     if combined_sql_q:
@@ -269,4 +270,6 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
     # Remove provider from thread name before return results
     threading.currentThread().name = original_thread_name
 
+    # Sanitize the last_prov_updates key
+    provider_results['last_prov_updates'] = json.dumps(provider_results['last_prov_updates'])
     return provider_results
--- a/sickbeard/manual_search.py
+++ b/sickbeard/manual_search.py
@@ -189,15 +189,19 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
     show_obj = Show.find(sickbeard.showList, int(show))
 
     main_db_con = db.DBConnection('cache.db')
-    found_items = []
+
     provider_results = {'last_prov_updates': {}, 'error': {}, 'found_items': []}
     original_thread_name = threading.currentThread().name
 
+    combined_sql = []
+    combined_sql_params = []
+    sql_return = []
+
     for cur_provider in enabled_providers('manualsearch'):
         threading.currentThread().name = '{thread} :: [{provider}]'.format(thread=original_thread_name, provider=cur_provider.name)
-        sql_return = []
+
         # Let's check if this provider table already exists
-        table_exists = main_db_con.select("SELECT name FROM sqlite_master WHERE type='table' AND name=?", [cur_provider.get_id()])
+        table_exists = main_db_con.select(b"SELECT name FROM sqlite_master WHERE type='table' AND name=?", [cur_provider.get_id()])
         columns = [i[1] for i in main_db_con.select("PRAGMA table_info('{0}')".format(cur_provider.get_id()))] if table_exists else []
         minseed = int(cur_provider.minseed) if hasattr(cur_provider, 'minseed') else -1
         minleech = int(cur_provider.minleech) if hasattr(cur_provider, 'minleech') else -1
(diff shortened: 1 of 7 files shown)


p0psicles:
@labrys as poc alternative for the lambda sorting. The unkown is also bugging me, as that's the highest value quality. I just changed it -boldly- to -1 in common.py. But, that's just to showcase the solution. The UNION is pretty much the only way I can think of to handle this for 100% in db. Maybe we should accept that this will get to complex, and handle it in application?

What do you think?

labrys:
There is no problem with combining SQL with application logic, the problem is in how it was done.  Anything that can be done in SQL should be as you will see better performance.  As a matter of simplicity I would probably make two simple SQL queries and join them in python.  First would select all where Quality != Unknown.  Second would be identical except select all where quality == unknown.  Then append the unknown results to bottom of the list.

labrys:
From #537: 
```python
# Sort the list of found items
found_items = sorted(found_items, key=lambda k: (try_int(k['quality']), try_int(k['is_proper']), try_int(k['seeders'])), reverse=True)
# Make unknown qualities at the botton
found_items = [d for d in found_items if try_int(d['quality']) < 32768] + [d for d in found_items if try_int(d['quality']) == 32768]
provider_results['found_items'] = found_items
```

`found_items = sorted(found_items, key=lambda k: (try_int(k['quality']), try_int(k['is_proper']), try_int(k['seeders'])), reverse=True)` applies the try_int method to the keys to be sorted by which isn't required if the sort is done in SQL.  Then uses a lambda to accomplish the sort, also not required if the sort is done in SQL.

`found_items = [d for d in found_items if try_int(d['quality']) < 32768] + [d for d in found_items if try_int(d['quality']) == 32768]` iterates through the list 2 times, generating 2 additional lists and applying the try_int function yet another 2 times.

`provider_results['found_items'] = found_items` makes yet another copy of the list in `provider_results`

So by my count 5 try_int calls and a lambda call per item, 3 list iterations, 4 list creations.

instead (pseudocode):
```python
sql_result_1 = sorted_query_without_unknown_quality()
sql_result_2 = sorted_query_with_only_unknown_quality()
provider_results['found_items'] = sql_result_1 + sql_result_2
```

p0psicles:
Yeah, I get the idea. I'll modify this PR, to use the separate query for UNKNOWN. I'll make sure the proper sorting is also in it, although it's a little bit messy of sql code, because I need to check for the keywords 'proper', 'repack' or 'real'.

fernandog:
@p0psicles `try_int` was used because `quality` INT fails to convert when quality value was not integer. Field is `NUMERIC`

I can't find the traceback I had when that was added. Commit was fixed when we rebased manual search branch

labrys:
@fernandog When the sort is done in SQL it doesn't need a `try_int` is my point, thus eliminating the 5 `try_int` calls required in python.

labrys:
@p0psicles in the future I think we should add a column PROPER, so that way we don't have to validate whether its a proper repeatedly and the logic to determine if a result is a proper can be done a single time.  For example any of the "Real Housewives.." would be considered propers as well as "Proper Manors" or several other cases.

p0psicles:
Ok @labrys, think this is what you meant. Also there is still a known error. Refreshing in between provider searches, doesn't work anymore. Need to look into that.

@fernandog can you compare this one in performance with the existing implementation?

sickbeard/webserve.py
```diff
@@ -205,6 +205,9 @@ def render(self, *args, **kwargs):
             kwargs['title'] = '500'
             kwargs['header'] = 'Mako Error'
             kwargs['backtrace'] = RichTraceback()
+            for (filename, lineno, function, line) in kwargs['backtrace'].traceback:
+                logger.log(u'File %s, line %s, in %s' % (filename, lineno, function), logger.DEBUG)
+            logger.log(u'%s: %s' % (str(kwargs['backtrace'].error.__class__.__name__), kwargs['backtrace'].error))
```
p0psicles: This came with another PR, i was working on. But would strongly advice to keep this in.

labrys:
@p0psicles I'd need to look a little closer, and I'm getting ready to leave right now so just glanced at it briefly.  There are a couple minor changes I'd make, but essentially that's on that the right track from what I saw.

fernandog:
Load time: 0.5602s / Mako: 0.3623s
Load time: 0.4521s / Mako: 0.2901s
Load time: 0.4552s / Mako: 0.2864s
Load time: 0.4925s / Mako: 0.2940s
develop:
Load time: 0.4766s / Mako: 0.2822s
Load time: 0.4781s / Mako: 0.2887s
Load time: 0.4744s / Mako: 0.2786s 
Load time: 0.4290s / Mako: 0.2876s

removing min and max values and taking average:
average sql:  292
average develop: 288.1

p0psicles:
Tnx @fernandog. Thats roughly the same. I Suspect that Most of the performance gains are lost because of the Double query because of the UNKNOWN quality and the instr it Does for the proper.

labrys:
@p0psicles and lack of appropriate indexes.  But also time isn't the only consideration, SQL is slower because it accesses hard drive, but also should have less cpu impact and smaller memory footprint.  Also does the unknown need to be at the bottom for manual search?  IMO it might be better for it to be at the top for manual since someone might want to look through it for qualities that just weren't properly detected by the quality detection.

medariox:
> Also does the unknown need to be at the bottom for manual search? IMO it might be better for it to be at the top for manual since someone might want to look through it for qualities that just weren't properly detected by the quality detection.

I disagree with that. The vast majority of "unkown" results are just very badly tagged releases from unkown P2P groups or individuals. Most of them are just junk (e.g. fake) or have extremely poor quality - as they are almost always crap, they should remain where they belong, at the bottom of the list.

labrys:
@medariox Fair point.  Well I'm not certain that SQLite supports it but we could do an ORDER BY CASE

```
ORDER BY 
  CASE ID
    WHEN unkown_quality THEN -1
    ELSE quality
  END
```

p0psicles:
@labrys is that as a comment on @medariox? About placing them at the bottom? Or a way to more efficiently place them at the top?

labrys:
ORDER BY CASE would be a way to dynamically assign a value to unknown qualities to place them at the bottom of the list.  When quality is unknown it substitutes the value -1 for ordering which would place it at the bottom of descending list.

p0psicles:
That works, but then it tries to translate to quality_string = -1, and that doesnt exist. Should we just add -1 as additional quality? Don't know if it can break anything.

labrys:
Hmm, it shouldn't change what's returned, as you should still be selecting Quality not the result of the CASE, the case should just be for ordering.  I'd have to think about it, as currently I believe -1 is supposed to represent an invalid quality or something like that.

p0psicles:
Oow sorry. It didnt work in Order by. It only worked in select

labrys:
Hmm, well it should work for ORDER BY as well.  I'm at work right now and can't really look at it atm.

p0psicles:
Never mind, I got it working with the CASE in the order by. Currently i'm moving the last provider update timestamp out of the query, into a separate one. Once that's finished. It should be good.

labrys:
@p0psicles yeah and would probably be worth it to add a proper column to db, but diff PR for that, unless you feel like rolling them into a single PR.

p0psicles:
@labrys nah, let's finish this PR, then do the db column in the next.

sickbeard/manual_search.py
```diff
@@ -183,52 +183,65 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
...
+    # Check if we have the combined sql strings
+    if combined_sql_q:
+        sql_prepend = b"SELECT * FROM ("
+        sql_append = b") ORDER BY CASE quality WHEN '{quality_unknown}' THEN -1 ELSE quality END DESC, " \
+                     b" seeders DESC, isproper DESC".format(quality_unknown=Quality.UNKNOWN)
+
+        # Add all results
+        sql_total += main_db_con.select(b'{0} {1} {2}'.
+                                        format(sql_prepend, ' UNION ALL '.join(combined_sql_q), sql_append),
+                                        combined_sql_params)
```
p0psicles: This is where everything is combined, and the query is run

sickbeard/manual_search.py
```diff
@@ -189,52 +189,65 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
...
+                common_sql += additional_sql
+                add_params += ["%|{0}|%".format(sql_episode), season]
+
+            # Add the created sql, to lists, that are used down below to perform one big UNIONED query
+            combined_sql_q.append(common_sql)
+            combined_sql_params += add_params
+
+        # Get the last updated cache items timestamp
+        last_update = main_db_con.select(b"select max(time) as lastupdate from '{provider_id}'".format(provider_id=cur_provider.get_id()))
+        provider_results['last_prov_updates'][cur_provider.get_id()] = last_update[0]['lastupdate']
```
fernandog: @p0psicles need to be inside the "if table_exists". Its outside

fernandog:
@p0psicles quality sort is wrong:

image

fernandog:
Approved

medariox:
@p0psicles 
Getting this error -sometimes- with manual search and fresh cache.db:
```
  File "/home/orangepi/.sickrage/sickbeard/webserve.py", line 322, in async_call
    result = function(**kwargs)
  File "/home/orangepi/.sickrage/sickbeard/webserve.py", line 1502, in manualSearchCheckCache
    last_prov_updates = json.loads(last_prov_updates.replace("'", '"'))
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 384, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
```

p0psicles:
Tnx med. Was afraid that bug was still in there. Will Look into it.

sickbeard/webserve.py
```diff
@@ -1502,7 +1502,7 @@ def manualSearchCheckCache(self, show, season, episode, manual_search_type, **kw
         if kwargs:
             # Only need the first provided dict
             last_prov_updates = kwargs.iteritems().next()[0]
-            last_prov_updates = json.loads(last_prov_updates.replace("'", '"'))
+            last_prov_updates = json.loads(last_prov_updates.replace("'", '"')) if last_prov_updates else {}
         else:
```
fernandog: @p0psicles what do you think?

p0psicles: Fort want to know what the root cause is. Also, might to Reverence this issue in @Labrys webserve PR. Because the issue starts because the json arg is not processen correctly. So it's also a partly Tornado implementation issue.

fernandog: feel free to remove commit then

p0psicles:
I made a mistake by returning a non-valid json (single quotes). Because of that I implemented the single quote replace. Now fixed it at the source.

@medariox and @fernandog could you test again?

p0psicles:
Issue with manual searching shows, that are not giving results. The search keeps looping.

p0psicles:
Without the last commit, it was having (regression) issues with refreshing with fresh db. Things to look out for testing are:
- No more errors when starting fresh (empty)
- No more errors when starting with a 'sickrage' db
- No looping when no results are found
- Results should be refreshed with each provider search
- Proper ordering (quality, seeders, propers). Lookout for torrent/nzb mixed results
- Status updating when snatching a show, through the refresh
- Status updating when snatching/downloading in displayShow
- Season pack downloads should update status for all episodes in displayShow
- Failed season pack downloads should also update status to failed for all eps in pack

sickbeard/manual_search.py
```diff
@@ -189,52 +190,65 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
...
+
+            # Get the last updated cache items timestamp
+            last_update = main_db_con.select(b"select max(time) as lastupdate from '{provider_id}'".format(provider_id=cur_provider.get_id()))
+            provider_results['last_prov_updates'][cur_provider.get_id()] = last_update[0]['lastupdate'] if last_update[0]['lastupdate'] else 0
+
+    # Check if we have the combined sql strings
+    if combined_sql_q:
+        sql_prepend = b"SELECT * FROM ("
+        sql_append = b") ORDER BY CASE quality WHEN '{quality_unknown}' THEN -1 ELSE CAST(quality as DECIMAL) END DESC, " \
+                     b" seeders DESC, isproper DESC".format(quality_unknown=Quality.UNKNOWN)
```
fernandog: @p0psicles 
shouldn't be `quality desc, proper desc, seeders desc` ?

my old PR: `key=lambda k: (try_int(k['quality']), try_int(k['is_proper']), try_int(k['seeders']))`

@duramato

fernandog:
Who else is testing this? @duramato and me are testing in `pubdate2`

@p0psicles If no one else is testing it should be merged to develop so all can be "forced" to test

p0psicles:
Approved

PR #579, opened 2016-05-12
Add more non_release_groups filter

fernandog:
@duramato @labrys


fernandog:
@pymedusa/owners @pymedusa/developers 
Still waiting review....

sickbeard/helpers.py
```diff
@@ -114,65 +114,81 @@ def remove_non_release_groups(name, clean_proper=False):
...
+        'www.cpasbien.pw',
+        'www.newpct1.com',
+        'www.scenetime.com',
+        'www.tensiontorrent.com',
+        'www.torrentday.com',
+        'www.usabit.com',
+        'youtubetomp3.uno'
+    }
+
+    non_groups_re = r'(({delimiter}+)({groups})({delimiter}*)$)|(^({delimiter}*)({groups})({delimiter}+))'.format(
```
ratoaq2: Here's an example with the generated regex: https://regex101.com/r/vY9sM0/2

The idea is:
- `delimiter` 1 or more times + non group name + `delimiter 0 or more times` + `end`
- `start` + `delimiter` 0 or more times + non group name + `delimiter 1 or more times`

sickbeard/helpers.py
```diff
@@ -114,65 +114,81 @@ def remove_non_release_groups(name, clean_proper=False):
     # Check your database for funky release_names and add them here, to improve failed handling, archiving, and history.
     # select release_name from tv_episodes WHERE LENGTH(release_name);
     # [eSc], [SSG], [GWC] are valid release groups for non-anime
+
+    # All lowercase and alphabetically sorted
+    groups = {
+        '-20-40',
```
fernandog: @ratoaq2 don't you need to escape this and `chamele0n$` and `simple@blur`

sickbeard/helpers.py
```diff
@@ -114,65 +114,81 @@ def remove_non_release_groups(name, clean_proper=False):
...
+        '1044',
+        'agusiq',
+        'androidtwou',
+        'brassetv',
+        'bt',
+        'buymore',
+        'chamele0n$',
```
ratoaq2: $ needs to be removed hete

sickbeard/helpers.py
fernandog: also dont forget to use "python" in `regex101`

sickbeard/helpers.py
```diff
@@ -114,65 +114,81 @@ def remove_non_release_groups(name, clean_proper=False):
     # Check your database for funky release_names and add them here, to improve failed handling, archiving, and history.
     # select release_name from tv_episodes WHERE LENGTH(release_name);
     # [eSc], [SSG], [GWC] are valid release groups for non-anime
+
+    # All lowercase and alphabetically sorted
+    groups = {
+        '-20-40',
+        '1044',
+        'agusiq',
+        'androidtwou',
```
duramato: extra u at the end?

sickbeard/helpers.py
ratoaq2: No... all entries in this list will be escaped later with `re.escape`
See below

sickbeard/helpers.py
ratoaq2: Don't know. It was like that before

fernandog: Castle 2009 S07E18 HDTV XviD-FUM[AndroidTwoU]

sickbeard/helpers.py
fernandog: so it's n not working: u'SiMPLE@BluRG'

i"m running this to test: https://gist.github.com/fernandog/96b63b2e860ebce95657

sickbeard/helpers.py
labrys: Instead of inserting groups into the regex you should just make a capturing group to capture anything between the delimiters, then check and see if the captured item is in the group list, if so then replace the entire tag.

ratoaq2: Yesterday I looked at some real cases and also tried this approach. There are some gotchas like non_group names with delimiter characters  (e.g.: `. `) that can make it much harder. 

I would like to take a look on a good way to handle this during my free time. But I would also like to have a lot o real examples in order to have them covered by tests. Could you guys provide me your cache.db file so I can have a big data set. I'll be on irc later today

labrys: @ratoaq2 You are absolutely correct, this single regex would not catch 100% of the cases, however a handful of similar regex's (or possibly extending this one, although I'd lean towards a couple of separate regex's instead of a single "mega" regex) should be fairly reliable.  Also the advantage here is it could just as easily provide release group detection, not just non-release group parsing.  

If the number of gotchas for the regex's is relatively few, it might be easier to do:
```
for gotcha in problem_non_groups:
     title = title.replace(gotcha, '')
```
Although I'm not as fond of that approach because of possible unintended consequences when adding new non-groups.

fernandog: @duramato can you send your cache.db to @ratoaq2 ? You have the largest dataset

fernandog:
@ratoaq2 are you still working on this?

@duramato did you send your cache.db to him?

ratoaq2:
I didn't get any cache.db file yet (I wasn't online so often as well).

I did look at this for a couple of hours and there are some issues that, in my opinion, should be tackled first.

### Some of the issues
There are some expressions that should be handled in the name parsing logic and not in this method (like [this](https://github.com/pymedusa/SickRage/blob/00717def9506d9b1303bc7524d5e060f91e646b3/sickbeard/helpers.py#L188) ). This regex might remove information about resolution and coded that should be extracted and used during the name parsing.
There are some words like English, Italian that if removed, you might affect the Episode title or the series name when running this method against a file name.
There are other issues that I cannot recall now, but most of them are related to a wrong name parsing, or removing parts that shouldn't be removed.

### Proposal:
- Close this PR. If there are missing "non-release words", then update the current list in the develop branch without trying to update the logic/expressions
- Split the issue in 2 parts as follow:

### Name Parsing
This is the old discussion to leverage the guessit 2 usage and combine with some regex to handle cases where guessit cannot help. This should be started with a data set where we have real release names that represents each different scenario that medusa handles and have a unit test against this data set. Then we can start the name parser with guessit. Guessit might handle most of the cases and the ones that it doesn't, we can fallback to existing regexes for it.

### Release Group sanitization
After solving the parsing issue, we have all information already splitted into different fields. And the release group might need to be sanitized. Then we can just enhance this clear_non_release_words method to properly sanitize the release group. There will be no issues with resolution, codec, language names in the release_group field since this was already handled by the parsing logic. Then it would be cleaner and safer to create a set of regexes to sanitize it.

PR #581, opened 2016-05-13
Use guessit for proper detection

labrys:

Use guessit to detect propers instead of regex.


p0psicles:
Yes this is what i'm talking about! This is really going to be helpful when adding the proper field to db.I think then we start using guessit everywhere, where we do something with the proper detection. And then gradually start using guessit for the rest.

p0psicles:
Approved

PR #583, opened 2016-05-14
Add space in torrentz regex

fernandog:
@duramato @labrys @p0psicles 

I still prefer replace. Easier to read.

Adding this space will work? Cause using re.sub it appends and not replace.

### DIFF
--- a/sickbeard/providers/torrentz.py
+++ b/sickbeard/providers/torrentz.py
@@ -90,7 +90,7 @@ class TorrentzProvider(TorrentProvider):  # pylint: disable=too-many-instance-at
 
                             title_raw = item.title.text
                             # Add "-" after codec and add missing "."
-                            title = re.sub(r'([xh][ .]?264|xvid)', r'\1-', title_raw).replace(' ','.') if title_raw else ''
+                            title = re.sub(r'([xh][ .]?264|xvid)( )', r'\1-', title_raw).replace(' ','.') if title_raw else ''
                             t_hash = item.guid.text.rsplit('/', 1)[-1]
 
                             if not all([title, t_hash]):


sickbeard/providers/torrentz.py
```diff
@@ -90,7 +90,7 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
 
                             title_raw = item.title.text
                             # Add "-" after codec and add missing "."
-                            title = re.sub(r'([xh][ .]?264|xvid)', r'\1-', title_raw).replace(' ','.') if title_raw else ''
+                            title = re.sub(r'([xh][ .]?264|xvid) ', r'\1-', title_raw).replace(' ','.') if title_raw else ''
```
p0psicles: For whitespace, i usually use \s. As that's a little better readable. It does work a bit different, because that also detects tabs, and this won't. In stead of \s you can also use [:blank:]. But both also catches tab and newline.

p0psicles:
I can't really detect what this PR is for? Why is this needed?

duramato:
Regex was/is wrong, turns "x264 " in to x264- " becouse its taking the previous group "x264" which is what it matches and not the space , and adding the dash (-)

labrys:
@duramato That's what I thought, which is why I mentioned it.

sickbeard/providers/torrentz.py
labrys: I think the chance of it matching a newline or a tab in a title is fairly low, however the regex could make the space only match a little more obvious using: `r'([xh][ .]?264|xvid)( )'`
(or `r'([xh][ .]?264|xvid)(?: )'` if you prefer).

labrys:
Approved

PR #586, opened 2016-05-14
Sanitize release name in release refiner

ratoaq2:
- MINOR FIX: Release name might come from release file (not only from DB), therefore it needs to be sanitized by calling `remove_non_release_groups`


fernandog:
Approved

p0psicles:
Approved

Issue #587, opened 2016-05-15
Subs extra scripts is broken (wrong arguments)


ratoaq2:
This commit https://github.com/pymedusa/SickRage/commit/d0c5e9c89fc9076e5733f79a7734503867d130a7

changed the extra scripts arguments to an array where the 3rd element is another array. Look the logs below:

```
2016-05-14 15:28:51 INFO     FINDSUBTITLES :: [24ddc49] Running subtitle pre-script: /usr/bin/python /scripts/library-scripts/sickrage/cleanit.py
2016-05-14 15:28:51 INFO     FINDSUBTITLES :: [24ddc49] Executing command: ['/usr/bin/python', '/scripts/library-scripts/sickrage/cleanit.py', "[u'/library/series/The 100/Season 03/The 100 - S03E15 - Perverse Instantiation Part One.mkv', u'/library/series/The 100/Season 03/The 100 - S03E15 - Perverse Instantiation Part One.pt-BR.srt', u'pob', u'The 100', 3, 15, u'Perverse Instantiation (1)', 268592]"]
2016-05-14 15:28:52 DEBUG    FINDSUBTITLES :: [24ddc49] Script result: Traceback (most recent call last):
  File "/scripts/library-scripts/sickrage/cleanit.py", line 21, in <module>
    subtitle_path = sys.argv[2]
IndexError: list index out of range
```

Note the list/array:
```
['/usr/bin/python', '/scripts/library-scripts/sickrage/cleanit.py', "[u'/library/series/The 100/Season 03/The 100 - S03E15 - Perverse Instantiation Part One.mkv', u'/library/series/The 100/Season 03/The 100 - S03E15 - Perverse Instantiation Part One.pt-BR.srt', u'pob', u'The 100', 3, 15, u'Perverse Instantiation (1)', 268592]"]
```

On another note, this log message is shown only on debug level (maybe it should be changed to a higher level)

p0psicles:
Is this fixed now?

p0psicles:
@ratoaq2 can this issue be closed?

ratoaq2:
Yes. Closed

PR #588, opened 2016-05-15
Fix subs scripts #587

ratoaq2:
This fixes the subs scripts passing the correct arguments to the scripts, as described in #587


fernandog:
Approved

medariox:
Approved

PR #585, opened 2016-05-14
Modularize webserver

labrys:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Always make sure the PR has a clear description of why,how, and what your trying to fix. The same goes for improvements or new features. Also if you refering to an issue, make sure a summary is available in the PR.
- [x] Read [contribution guide](https://github.com/PyMedusa/SickRage/blob/master/.github/CONTRIBUTING.md)

This begins the process of modularizing the web server and api

PR #594, opened 2016-05-19
Modularize web server

labrys:

First step in modularizing web server.  Replaces #585
>By splitting up the server/api into separate files it makes it easier to focus on relevant changes in the affected code. It also makes the structure a little bit clearer and makes it a little less intimidating to make changes. Also, since people should take ownership of the code they touch, making sure it is all up to standards, it is much easier to do that when you aren't dealing with a single mega-file.


labrys:
The QC errors are not new issues, they are existing issues in the touched files.  Splitting this up will make it easier to address these issues while limiting the consequences.

p0psicles:
Yeah I know, QC will be really usefull, when all code is up2date. Until then we can't help to ignore them from time2time.

sickbeard/server/web/config/notifications.py
```diff
@@ -258,7 +258,7 @@ def saveNotifications(self, use_kodi=None, kodi_always_on=None, kodi_notify_onsn
 
         sickbeard.save_config()
 
-        if len(results) > 0:
+        if results:
```
p0psicles: Where is results getting it's value from?

labrys: Nearest I could tell, nowhere (its assigned an empty list that does not appear to be used) so this code may be able to be removed.  However, I figured it would be premature to remove it until thats confirmed, so I have left that for later optimization.

sickbeard/server/web/config/providers.py
```diff
@@ -547,7 +547,7 @@ def saveProviders(self, newznab_string='', torrentrss_string='', provider_order=
 
         sickbeard.save_config()
 
-        if len(results) > 0:
+        if results:
```
p0psicles: Same here, where does results get it's value?

labrys: Same answer.  Nearest i can tell nowhere.  Leaving it for now to look into it later.

labrys: My assumption is they copy-pasta'd code and never implemented it.

sickbeard/server/core.py
```diff
@@ -49,7 +52,7 @@ def __init__(self, options=None, io_loop=None):
         # api root
         if not sickbeard.API_KEY:
             sickbeard.API_KEY = generateApiKey()
-        self.options['api_root'] = r'%s/api/%s' % (sickbeard.WEB_ROOT, sickbeard.API_KEY)
+        self.options['api_root'] = '{root}/api/{key}'.format(root=sickbeard.WEB_ROOT, key=sickbeard.API_KEY)
```
p0psicles: No need for r''?

sickbeard/server/web/config/providers.py
```diff
@@ -0,0 +1,558 @@
...
+
+            return '|'.join([providerDict[name].get_id(), providerDict[name].configStr()])
+
+        else:
+            newProvider = rsstorrent.TorrentRssProvider(name, url, cookies, titleTAG)
+            sickbeard.torrentRssProviderList.append(newProvider)
+            return '|'.join([newProvider.get_id(), newProvider.configStr()])
+
+    @staticmethod
+    def deleteTorrentRssProvider(id):
```
p0psicles: change the buildin?

sickbeard/server/web/config/subtitles.py
```diff
@@ -0,0 +1,96 @@
...
+        # Reset provider pool so next time we use the newest settings
+        subtitles.get_provider_pool.invalidate()
+
+        if results:
+            for x in results:
+                logger.log(x, logger.ERROR)
+            ui.notifications.error('Error(s) Saving Configuration',
+                                   '<br>\n'.join(results))
+        else:
+            ui.notifications.message('Configuration Saved', ek(os.path.join, sickbeard.CONFIG_FILE))
```
p0psicles: Just out of curiosity, wasn't it that we didn't need ek() anymore, because of the future unicode_literals?

sickbeard/server/web/core/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+            username = sickbeard.WEB_USERNAME
+            password = sickbeard.WEB_PASSWORD
+
+            if (self.get_argument('u', None) == username or not username) and \
+                    (self.get_argument('p', None) == password or not password):
+                api_key = sickbeard.API_KEY
+
+            self.finish({'success': api_key is not None, 'api_key': api_key})
+        except Exception:
+            logger.log('Failed doing key request: {error}' .format(error=traceback.format_exc()), logger.ERROR)
```
p0psicles: space after }'

sickbeard/server/web/core/base.py
```diff
@@ -0,0 +1,471 @@
...
+        if sickbeard.NEWS_UNREAD:
+            self.arguments['newsBadge'] = ' <span class="badge">{news}</span>'.format(news=sickbeard.NEWS_UNREAD)
+
+        numCombined = error_count + warning_count + sickbeard.NEWS_UNREAD
+        if numCombined:
+            if error_count:
+                self.arguments['toolsBadgeClass'] = ' btn-danger'
+            elif warning_count:
+                self.arguments['toolsBadgeClass'] = ' btn-warning'
+            self.arguments['toolsBadge'] = ' <span class="badge{type}">{number}</span>'.format(
```
p0psicles: I've noticed we break the format in three different ways:
1. </span>'.
format(...)
2. </span>'.format
(...)
3. </span>'.format(
...)

Think we should choose one?

sickbeard/server/web/core/calendar.py
```diff
@@ -0,0 +1,121 @@
+# coding=utf-8
+
+"""
```
p0psicles: bookmark

sickbeard/server/core.py
labrys: good catch, that was a typo.

sickbeard/server/web/config/subtitles.py
labrys: No, ek is still needed to handle locale issues.

sickbeard/server/web/core/authentication.py
labrys: Doesn't actually make a difference, but good catch because it wasn't intended.

sickbeard/server/web/core/base.py
labrys: When the line length isn't excessive I prefer 1.
2. can't be used here since the format isn't inside a function (or else a `\` would be required to wrap to the next line).
In some cases 3 leads to a wrapped indent that doesn't neatly line up with the previous line.

I generally prefer them in the order listed, myself.

sickbeard/server/web/config/providers.py
labrys: I haven't yet, but I saw that when I was going through, also good catch.

p0psicles:
Do we have any Unit tests with References to the old locations?

labrys:
Approved

PR #616, opened 2016-05-24
Show RARBG error code next to error message

fernandog:
and Show "Too many requests per second!" as INFO
@duramato

### DIFF
--- a/sickbeard/providers/rarbg.py
+++ b/sickbeard/providers/rarbg.py
@@ -130,11 +130,16 @@ class RarbgProvider(TorrentProvider):  # pylint: disable=too-many-instance-attri
                     continue
 
                 error = data.get("error")
-                error_code = data.get("error_code")
+                error_code = try_int(data.get("error_code"))
                 # Don't log when {"error":"No results found","error_code":20}
                 # List of errors: https://github.com/rarbg/torrentapi/issues/1#issuecomment-114763312
                 if error:
-                    if try_int(error_code) not in (14, 20):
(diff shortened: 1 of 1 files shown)


p0psicles:
I say go, if you can also get an approval of @medariox or @labrys merge it.

p0psicles:
Approved

sickbeard/providers/rarbg.py
```diff
@@ -130,11 +130,16 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
                     continue
 
                 error = data.get("error")
-                error_code = data.get("error_code")
+                error_code = try_int(data.get("error_code"))
```
labrys: Does this need to be converted to int?  If its a string compare it as a string.

p0psicles: @fernandog can you check if removing the try_int works? The other should then also be changed to if error_code not in ('20', '14'). but that should not be a problem.

fernandog: ......

labrys:
Approved

p0psicles:
@pymedusa/developers anyone who can test this with rarbg?

labrys:
@p0psicles tested good here:
```
14:03:26 INFO::SEARCHQUEUE-BACKLOG-121361 :: [Rarbg] :: [43b42f6] Too many requests per second. Maximum requests allowed are 1req/2sec Please try again later!. Error code: 5
```

PR #617, opened 2016-05-24
Don't call clear_non_release_groups for non media files

fernandog:
@ratoaq2 also you weren't passing full path creating error:

`Couldn't remove non release groups from video file. Error: error 2 : The system cannot find the file specified`

@medariox

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -665,11 +665,12 @@ def delete_unwanted_subtitles(dirpath, filename):
             logger.info(u"Couldn't delete subtitle: %s. Error: %s", filename, ex(error))
 
 
-def clear_non_release_groups(filepath):
+def clear_non_release_groups(filename, filepath):
     """Remove non release groups from the name of the given file path.
 
     It also renames/moves the file to the path
 
+    :param filename: the file name
     :param filepath: the file path
     :type filepath: str
     :return: the new file path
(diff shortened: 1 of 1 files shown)


medariox:
os.path.join() is used later. Where do you get this error?

fernandog:
In function `clear_non_release_groups`

medariox:
https://github.com/pymedusa/SickRage/blob/d93ebc45225eecd5c4b7e63420d308ce88d0ef10/sickbeard/subtitles.py#L726-L733

```
filename = clear_non_release_groups(os.path.join(root, filename))

if processTV.subtitles_enabled(filename) is False:
                    logger.debug(u'Subtitle disabled for show: %s', filename)
                    continue

video_path = os.path.join(root, filename)
release_name = os.path.splitext(filename)[0]
```

IIrc processTV.subtitles_enabled() needs just the file name, not the full path.

Join is repeated here: video_path = os.path.join(root, filename)
Also look at os.path.splitext(filename)[0], are you sure that filename can be safely a path, instead of a file name?

fernandog:
Will say again: Look a the function code. It does a os.rename. without path = error.

medariox:
It is obviously not a proper solution if it adds other issues.

fernandog:
Np. Let's wait @ratoaq2 proper fix

fernandog:
@medariox will this be merged? i've tested since your commits and no issues

medariox:
Approved

fernandog:
@pymedusa/developers someone merge please

p0psicles:
Approved

PR #623, opened 2016-05-25
Reduce release parsing

p0psicles:
I've started this as a solution alternative for the reducing daily search result parsing, that fernandog came up with based on pubdate.

This has a little different approach. It uses a single table, that has the 5 last search results for each provider. Every search starts checking the results from newest to oldest. For every result, the URL is compared to that in the cache db. When it starts getting matches, it assumes, it has processed these results before. And stops.

Basically this PR exists out of the following elements:
* Code Cleanup like the %s -> format
* A unicode bug I noticed when using some anime providers. Like tokyotosho. It crashes on the RSS feed, and then stops processing all together. I've wrapped this in a try/catch for now, to make sure it continues. But does need to be fixed in a proper way.
* cache.db changes (added table)
* Changes to tvcache.py. (no changes required to the individual providers! As of now)


sickrage/providers/GenericProvider.py
```diff
@@ -518,3 +519,18 @@ def _make_url(self, result):
 
     def _verify_download(self, file_name=None):  # pylint: disable=unused-argument,no-self-use
         return True
+
+    @property
+    def recent_results(self):
+        if sickbeard.provider_recent_results.get(self.get_id()):
+            return sickbeard.provider_recent_results.get(self.get_id())[::-1]
```
p0psicles: Reverse the order, making sure that when looping the (newest/latest added) items are on top

sickrage/providers/GenericProvider.py
```diff
@@ -518,3 +519,18 @@ def _make_url(self, result):
...
+        else:
+            return []
+
+    @recent_results.setter
+    def recent_results(self, items):
+        if not sickbeard.provider_recent_results.get(self.get_id()):
+            sickbeard.provider_recent_results.update({self.get_id(): []})
+        sickbeard.provider_recent_results[self.get_id()] += items
+        if items:
+            del sickbeard.provider_recent_results[self.get_id()][:len(sickbeard.provider_recent_results[self.get_id()]) - self.max_recent_items]
```
p0psicles: This will cut of the the first n items. For ex. thelist = [1,2,3,4,5,6,7], del thelist[:len(thelist) - 5], will translate to: del thelist[:2] will result in [3,4,5,6,7]. Making sure there are always 5 items in the list.

sickbeard/tvcache.py
```diff
@@ -224,30 +244,28 @@ def _parseItem(self, item):
...
         self._checkItemAuth(title, url)
 
         if title and url:
             title = self._translateTitle(title)
             url = self._translateLinkURL(url)
 
-            # logger.log(u"Attempting to add item to cache: " + title, logger.DEBUG)
-            return self._addCacheEntry(title, url, seeders, leechers, size, pubdate, hash)
+            # Placed the self._get_hash(item) inline, because hash is a buildin. Could cause issues.
+            return self._addCacheEntry(title, url, seeders, leechers, size, pubdate, self._get_hash(item))
```
p0psicles: Possible issue, therefor removed the buildin hash.

sickbeard/tvcache.py
```diff
@@ -171,37 +170,58 @@ def updateCache(self):
...
+                        continue
+
+                cache_db_con = self._get_db()
                 if cl:
                     cache_db_con.mass_action(cl)
 
+                # finished processing, let's save the newest x (index) items and store these in cache with a max of 5 
+                # (overwritable per provider, throug hthe max_recent_items attribute.
+                self.provider.recent_results = data['entries'][0:min(index, self.provider.max_recent_items)]
+
```
p0psicles: This part is the meat of the PR. Rest is pretty much cleaning and some bug fixes.

sickbeard/tvcache.py
```diff
@@ -32,98 +32,97 @@
...
-        #    cache_db_con = self._getDB()
-        #    cache_db_con.action("DELETE FROM [" + self.providerID + "] WHERE 1")
 
     def _clearCacheItem(self):
-        cache_db_con = self._getDB()
+        cache_db_con = self._get_db()
         today = int(time.mktime(datetime.datetime.today().timetuple()))
         # Keep item in cache for 7 days
-        cache_db_con.action("DELETE FROM [" + self.providerID + "] WHERE time > ? ", [today + 7*86400]) # 86400 POSIX day (exact value)
+        cache_db_con.action('DELETE FROM [{0}] WHERE time < ? '.format(self.provider_id), [today - 7 * 86400])  # 86400 POSIX day (exact value)
```
p0psicles: Note that with this change, items from the cache tables will really be cleaned after 7 days!!! Before this, this was never happening.

sickbeard/__init__.py
```diff
@@ -657,7 +659,8 @@ def initialize(consoleLogging=True):  # pylint: disable=too-many-locals, too-man
             AUTOPOSTPROCESSER_FREQUENCY, SHOWUPDATE_HOUR, \
             ANIME_DEFAULT, NAMING_ANIME, ANIMESUPPORT, USE_ANIDB, ANIDB_USERNAME, ANIDB_PASSWORD, ANIDB_USE_MYLIST, \
             ANIME_SPLIT_HOME, SCENE_DEFAULT, DOWNLOAD_URL, BACKLOG_DAYS, GIT_USERNAME, GIT_PASSWORD, \
-            DEVELOPER, gh, DISPLAY_ALL_SEASONS, SSL_VERIFY, NEWS_LAST_READ, NEWS_LATEST, SOCKET_TIMEOUT, RECENTLY_DELETED
+            DEVELOPER, gh, DISPLAY_ALL_SEASONS, SSL_VERIFY, NEWS_LAST_READ, NEWS_LATEST, SOCKET_TIMEOUT, RECENTLY_DELETED, \
+            provider_recent_results
```
p0psicles: Used this global to keep track

sickrage/providers/GenericProvider.py
medariox: What if the provider returns eg. only 2 results?

p0psicles: There should be always at least 5 items int the list. Then 2 will be added, making it 7. Then del list[:2] will result in removing 2 from the top.

sickrage/providers/GenericProvider.py
labrys:
```python
    @recent_results.setter
    def recent_results(self, items):
        results = items + self.recent_items
        if len(results) > self.max_recent_items:
            results = results[:self.max_recent_items]
        provider_recent_results.update({self.get_id(): results})
```

labrys:
Reposting here so it's not lost in all the line-notes.
```python
@recent_results.setter
def recent_results(self, items):
    results = items + self.recent_items
    if len(results) > self.max_recent_items:
        results = results[:self.max_recent_items]
    provider_recent_results.update({self.get_id(): results})
```
This way you build up to `max_recent_items`  in the cache.  There's no need to reverse the list or anything like that.  If you extend the new items with the old then they are sorted by age desc, thus you can just grab the first X results.

sickbeard/tvcache.py
labrys: This shouldn't be done here, but in the provider code, right at the end of the for loop that iterates through the result rows.

sickbeard/tvcache.py
labrys: Good catch, however imo i don't care for enforcing a cache age of 7 days, particularly since we use the cache for manual search.  If anything this should be a configurable setting to allow the user how many days to trim, if the user wants to trim.

p0psicles:
@labrys Is there anything against reversing the list?

sickbeard/__init__.py
labrys: This doesn't need to be a global.

labrys:
Why reverse the list if there's no need?

sickrage/providers/GenericProvider.py
```diff
@@ -523,14 +523,19 @@ def _verify_download(self, file_name=None):  # pylint: disable=unused-argument,n
     @property
     def recent_results(self):
         if sickbeard.provider_recent_results.get(self.get_id()):
-            return sickbeard.provider_recent_results.get(self.get_id())[::-1]
+            return sickbeard.provider_recent_results.get(self.get_id())
```
labrys: With a `get()` you don't need an if statement, just pass the default value, `[]` in this case, as the second arg to `get()`

sickrage/providers/GenericProvider.py
```diff
@@ -523,14 +523,19 @@ def _verify_download(self, file_name=None):  # pylint: disable=unused-argument,n
...
+        if not recent_results.get(self.get_id()):
+            recent_results.update({self.get_id(): []})
         if items:
-            del sickbeard.provider_recent_results[self.get_id()][:len(sickbeard.provider_recent_results[self.get_id()]) - self.max_recent_items]
+            add_to_list = []
+            for item in items:
+                if item['link'] not in {cache_item['link'] for cache_item in recent_results[self.get_id()]}:
+                    add_to_list += [item]
+            recent_results[self.get_id()] = add_to_list + recent_results[self.get_id()]
+            recent_results[self.get_id()][:self.max_recent_items]
```
labrys: I'm not sure what you're trying to do here as this line has no effect.  I assume you're trying to do this:
```python
recent_results[self.get_id()] = add_to_list + recent_results[self.get_id()]
recent_results[self.get_id()] = recent_results[self.get_id()][:self.max_recent_items]
```
Creating a temp variable to store the list would probably be clearer
```python
results = add_to_list + recent_results[self.get_id()]
recent_results[self.get_id()] = results[:self.max_recent_items]
```

sickbeard/tvcache.py
```diff
@@ -178,7 +178,7 @@ def updateCache(self):
                 cl = []
                 index = 0
                 for index, item in enumerate(data['entries'] or []):
-                    if recent_results and item.get('link').strip() in [cached_item['link'].strip() for cached_item in recent_results]:
+                    if recent_results and item['link'] in {cache_item['link'] for cache_item in recent_results}:
```
labrys: no need to check if recent_results.  Recent results is a list and if its empty `{cache_item['link'] for cache_item in recent_results}` would be an empty list anyway so the test would already eval false.

sickbeard/tvcache.py
```diff
@@ -171,37 +170,58 @@ def updateCache(self):
...
 
-                cl = []
-                for item in data['entries'] or []:
-                    ci = self._parseItem(item)
-                    if ci is not None:
-                        cl.append(ci)
+                # get last 5 rss cache results
+                recent_results = self.provider.recent_results
+                found_recent_results = 0  # A counter that keeps track of the number of items that have been found in cache
+                stop_at = 3  # Configuration as an error margin, to stop at. The lower the number, the faster it will stop parsing items
```
labrys: Consider making `stop_at` a provider variable in case we want to make it vary by provider.

sickbeard/tvcache.py
```diff
@@ -171,37 +170,58 @@ def updateCache(self):
...
+
+                    if found_recent_results >= stop_at:
+                        logger.log(u'Hit the old cached items, not parsing any more for: {0}'.format
+                                   (self.provider_id), logger.ERROR)
+                        break
+                    try:
+                        ci = self._parseItem(item)
+                        if ci is not None:
+                            cl.append(ci)
+                    except UnicodeDecodeError, e:
```
labrys: `except UncidodeDecodeError as e:`
The `exc, e` style is deprecated and no longer exists in Python 3.

p0psicles: Yeah I put it there to do some debugging on the unicode exception. Forgot to remov it. I'll put a debug there.

sickbeard/tvcache.py
p0psicles: Good idea. Was also considering that.

sickbeard/tvcache.py
p0psicles: I agree. If others agree, Lets implement this. @fernandog, do you see time to implement this?

sickbeard/__init__.py
p0psicles: Np. Will change that

sickbeard/tvcache.py
fernandog: No. Sorry. As I said a couple of weeks ago I wont have much time to code anymore.

sickbeard/tvcache.py
p0psicles: That is debatable. As it's a tradeof with provider managability. Personally im not convinced. Other opinion?

sickbeard/tvcache.py
labrys: @p0psicles for this end of it:
```python
if sickbeard.trim_cache:
    days_to_keep = sickbeard.max_cache_age * 86400  # 86400 POSIX day (exact value)
    cache_db_con.action('DELETE FROM [{0}] WHERE time < ? '.format(self.provider_id), [today - days_to_keep])  
```
And then we'd need to add those to attributes to the UI/Config file

p0psicles: Global or per provider?

labrys: I would think global, we could go provider, but i doubt the user really would care to go that granular.

p0psicles: Sure. If nobody beats me to it, il look at. But shouldnt be part of this PR.

labrys: np, but if you're going to leave the fix in, could you comment it out or something so when it gets merged it doesn't trim peoples cache until we have a setting?

p0psicles: Nah, i'll just make sure that other PR is merged before this. And this gets rebased on it.

labrys: @p0psicles #634

sickbeard/__init__.py
p0psicles: Fixed together with a rebase on latest dev

p0psicles:
I'm still in doubt if we should use the margin? Might as well just bail out on first match? Can you guys think of any situation where you need the margin?

fernandog:
@p0psicles are you matching newest provider rss item with newsted from recent_results? if you do otherwise it will take longer right to find the match?

p0psicles:
Yes i am

fernandog:
about the first match question, I agree it can stop

duramato:
@p0psicles usenet and public torrent indexers get multiple uploads with the same title

labrys:
@fernandog I agree the per provider configuration can come later if necessary, but I think making the setting be stored in the provider now would be both fine.  I agree the margin will likely not be necessary.
@duramato Unless I'm mistaken we are comparing the link not the title so they should all be unique.

p0psicles:
Yes we are comparing purely link.
I will change it to a "margin" of 1. So will stop at first release, when it matches.

p0psicles:
@labrys @medariox as everything has been merged now anyway, i'm thinking of dropping this branch. And start with a new one, of dev. What you guys think?

PR #628, opened 2016-05-26
Refactor POSTPROCESSER to POSTPROCESSOR

labrys:

Refactors all the misspelled `postprocessEr` references to `postprocessOr`.
The only consideration not yet addressed by this PR is the change to the spelling in config.ini.  Do we want to just ignore the previous setting or migrate it to the new setting?  If we migrate it we will need to retain a reference to the old mis-spelled setting for an extended period, whereas right now it would just require the user to update the setting after migrating.  If we choose to ignore the change, then this PR is concluded and ready for review.


labrys:
Approved

PR #638, opened 2016-05-29
Fixed cache_retention granularity.

p0psicles:

* Should be cast to int, as we want to step by day, not tenth of day.

@labrys ?
I've changed it because else in config you get to see 30,0. While stepping did change, saving did not.

### DIFF
--- a/sickbeard/server/web/config/search.py
+++ b/sickbeard/server/web/config/search.py
@@ -69,7 +69,7 @@ class ConfigSearch(Config):
         sickbeard.BACKLOG_DAYS = try_int(backlog_days, 7)
 
         sickbeard.CACHE_TRIMMING = config.checkbox_to_value(cache_trimming)
-        sickbeard.MAX_CACHE_AGE = float(max_cache_age)
+        sickbeard.MAX_CACHE_AGE = try_int(max_cache_age)
 
         sickbeard.USE_NZBS = config.checkbox_to_value(use_nzbs)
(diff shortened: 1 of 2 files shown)


sickbeard/server/web/config/search.py
```diff
@@ -69,7 +69,7 @@ def saveSearch(self, use_nzbs=None, use_torrents=None, nzb_dir=None, sab_usernam
         sickbeard.BACKLOG_DAYS = try_int(backlog_days, 7)
 
         sickbeard.CACHE_TRIMMING = config.checkbox_to_value(cache_trimming)
-        sickbeard.MAX_CACHE_AGE = float(max_cache_age)
+        sickbeard.MAX_CACHE_AGE = try_int(max_cache_age)
```
medariox: Should a non-int default to 0 or 1?

labrys: 0 since in cache trim i do `if days:` so that it would skip the trim instead of trimming all but 1 day.

labrys: Also as this is set by the UI, it should be `int`, `try_int` is unnecessary.

p0psicles: hmm, then we should change those everywhere. As everywhere try_int is used. Also for ex. BACKLOG_DAYS and USENET_RETENTION.

p0psicles:
Approved

PR #648, opened 2016-05-31
Add Pubdate parsing

fernandog:
also some provider fixes

@pymedusa/developers


p0psicles:
I'm going to close this, because as discussed. Because of the following reasons, this PR won't provide much added value, compared to the changes it does to the different providers.:
* Allot of new code to providers. We should try to limit the amount of parse code. It's not bad that new parsing logic is added, but it should have a really good reason.
* Parsing of date from values like: "5 min ago", or "1 day ago". As also some proper parsing from providers that do provide a decent formatted date.
* Maintenance of code. What if a provider makes a small change to the date format.
* Relatively large lib added. For a feature that isn't bringing allot of value, at least now.
* Currently no use for the pubdate. It's being parsed, also taking up CPU and MEM and disk, but that's it.

Maybe in the future we do really need something like this. But then we should first make sure we've cleaned up the providers.

If there are any fixes we need, please PR these as fixes.

fernandog:
1) Lot of new code? Lol 2 lines for parsing
2) Only lime had '5min ago'
3) lib takes care if provider changes date format
4) value is subjective. Imo it has a lot of value. Also a new column in manual search result. Read economics value concept
5) cpu? Lol. Wont comment this bs

You should really rewrite the reasons
Glad im steping out of all this ego bs.

PR #659, opened 2016-06-03
Playing with guessit 2

ratoaq2:
- **NameParser**: added a new flag `use_guessit` that enables guessit name parser (default `true` on this branch). Setting it to false will keep the current parser without any change.
- **GuessitNameParser**: This should be the new API and it should be really small and straightforward. Right now, it plugs in our own `rules` and it calls `guessit.guess(..)` with the correct parameters and adapt the response property names to the same ones `NameParseResult` has. I'm planning to move the `expected_titles` and `expected_groups` out of this file to a yaml file. But this list should always be short and simple.
- **Rebulk Rules**: This is where our magic happens. The main idea is to manipulate `matches` after guessit executed its work and before returning the guessed `matches`. With rebulk API, everything that we do is to apply our rules. I wrote some instructions and guidelines to always have each rule simple and maintainable. Some rules are fixes/workarounds for guessit bugs. I have flagged them with each issue url, so we can remove then when guessit is fixed.
- **Tests**: I have created a simple unit test that executes against an yaml file. For each regex in regexes.py and each rule in our rules.py, I added an entry in this yaml file, so they will work will regression tests. There are positive and negative test cases.

And the current status (as of 21.06.2016 20:40 GMT+2):
- Total = 47125
- Success = 45692 (96.96%)
- Discarded = 666 (1.41%)
- Mismatch: episode_numbers = 70 (0.15%)
- Mismatch: air_date = 40 (0.08%)
- Mismatch: series_name = 42 (0.09%)
- Mismatch: proper = 140 (0.29%)
- Mismatch: episode_numbers_range = 2 (0.005%)
- Mismatch: season_number_range = 0 (0%)
- Mismatch ab_episode_numbers = 13 (0.03%)
- Mismatch season_number = 28 (0.06%)
- Mismatch release_group = 432 (0.92%)

### DIFF
--- a/sickbeard/name_parser/parser.py
+++ b/sickbeard/name_parser/parser.py
@@ -26,11 +26,16 @@ import sickbeard
 from sickbeard.name_parser import regexes
 
 from sickbeard import logger, helpers, scene_numbering, common, scene_exceptions, db
+from sickbeard.name_parser.guessit_parser import parser as guessit_parser
 from sickrage.helper.common import remove_extension
 from sickrage.helper.encoding import ek
 from sickrage.helper.exceptions import ex
 from sickbeard.helpers import remove_non_release_groups
 import dateutil
+
+
+pre_cleanup_re = re.compile(r'(\.vol\d+\+\d+)?((\.par2\b)|(\.nzb\b)|(\.mkv\b)).*((\d+\.of\.\d+)|(\(\d+/\d+\))).*$',
+                            flags=re.IGNORECASE)
 
 
 class NameParser(object):
@@ -39,20 +44,24 @@ class NameParser(object):
     ANIME_REGEX = 2
 
     def __init__(self, file_name=True, showObj=None, tryIndexers=False,  # pylint: disable=too-many-arguments
-                 naming_pattern=False, parse_method=None):
+                 naming_pattern=False, parse_method=None, use_guessit=True):
 
         self.file_name = file_name
         self.showObj = showObj
         self.tryIndexers = tryIndexers
 
         self.naming_pattern = naming_pattern
+        self.use_guessit = use_guessit
 
         if (self.showObj and not self.showObj.is_anime) or parse_method == 'normal':
             self._compile_regexes(self.NORMAL_REGEX)
+            self.show_type = 'regular'
         elif (self.showObj and self.showObj.is_anime) or parse_method == 'anime':
             self._compile_regexes(self.ANIME_REGEX)
+            self.show_type = 'anime'
         else:
             self._compile_regexes(self.ALL_REGEX)
+            self.show_type = None
 
     @staticmethod
     def clean_series_name(series_name):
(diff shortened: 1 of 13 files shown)


sickbeard/name_parser/parser.py
```diff
@@ -568,6 +619,10 @@ def is_anime(self):
             return True
         return False
 
+    @property
+    def is_proper(self):
+        return re.search(r'(^|[\. _-])(proper|repack)([\. _-]|$)', self.extra_info, re.I) is not None if self.extra_info else False
```
duramato: Should just check if "if guess.get('proper_count', 0) > 0" , outherwise it doesn't detect `ReRip`  and `Real`

p0psicles:
While fixing the history page, I noticed guessit is really using allot of resources. We should be wary of that. In this case that would mean, limit the amount of parsed items to a minimum and prevent, parsing when not absolutely needed.

sickbeard/name_parser/parser.py
ratoaq2: This logic was just moved from one place to another. This is the current parser logic (not the guessit parser). I'm not changing/fixing the current parser

ratoaq2:
The performance is one of the biggest topics to be evaluated.
guessit shouldn't be used to parse/present data for the user since this might impact the user experience as you mentioned.
And for background operations, we will need to evaluate what's the current state

ratoaq2:
It's a good moment to test it and bring some feedback, so we can see if there's still something big to be handled. Otherwise I'll just fix the existing tests, QC and add an useful and easy to maintain test case against some really important scenarios.

Depending on the first test results, we could add an option to switch guessit on/off

labrys:
@ratoaq2 merged into my branch and testing

sickbeard/name_parser/parser.py
```diff
@@ -22,37 +22,44 @@
...
 from sickbeard import logger, helpers, scene_numbering, common, scene_exceptions, db
+from sickbeard.name_parser.guessit_parser import parser as guessit_parser
 from sickrage.helper.common import remove_extension
 from sickrage.helper.encoding import ek
 from sickrage.helper.exceptions import ex
 from sickbeard.helpers import remove_non_release_groups
 import dateutil
 
+pre_cleanup_re = re.compile(r'(\.vol\d+\+\d+)?((\.par2\b)|(\.nzb\b)|(\.mkv\b)).*((\d+\.of\.\d+)|(\(\d+/\d+\))).*$',
+                            flags=re.IGNORECASE)
```
p0psicles: Not really important, but need two linefeeds here for PEP8

sickbeard/name_parser/parser.py
```diff
@@ -22,37 +22,44 @@
...
 
 class NameParser(object):
     ALL_REGEX = 0
```
p0psicles: While your at it, and your the one who knows most about the parsing class, would you be willing to do some docstringing? Like here for the class, and for the class methods like the _parse_string() and parse() methods.

sickbeard/name_parser/parser.py
ratoaq2: Sure. I'll still revise de code style for the new files and the modifications I made on this file

sickbeard/name_parser/parser.py
ratoaq2: I'll do that, since there's a lot more other than parsing in this class

p0psicles:
I've been testing some. But got some unicode errors on anime shows. So that got me a little sidetracked. I'll test it some more when I get to it.

ratoaq2:
Can you provide the logs of this errors? Or the release names? 
This is valuable feedback mm 👍

p0psicles:
I think we hit an issue with animes that use season scene naming, like Assassination Classroom and Fariy Tail. Could you test it with these?

It's finding results for those shows and parses them, but parses them with an absolute ep number, while in the provided gist, I didn't look for the absolute number 19, but I searched for the episode 19 for S2, with absolute number 41.

I've tried with enabling and disabling "Scene numbering", both the same result.

https://gist.github.com/p0psicles/1af10d8b510025e46441dfb5370183e4
image

PR #660, opened 2016-06-04
History remove guessit

p0psicles:

Seriously guys, really think things through before your going to put these kind of heavy method calls (in a loop) in the template. As for now, these guessit calls really need to be removed, to make this page workable again.

### DIFF
--- a/gui/slick/views/history.mako
+++ b/gui/slick/views/history.mako
@@ -72,7 +72,7 @@
                     <% isoDate = datetime.strptime(str(hItem.date), History.date_format).isoformat('T') %>
                     <time datetime="${isoDate}" class="date">${airDate}</time>
                 </td>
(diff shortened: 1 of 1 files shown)


medariox:
Approved

Tested and working fine. I wonder when this was added, no wonder history is slow as hell.

duramato:
The history slowness was post this change, this was changed when we changed proper detection to guess-it

medariox:
Oh yes, found it: https://github.com/pymedusa/SickRage/pull/581/files
Maybe we want to bring back the old proper detection at least?

p0psicles:
Yeah sure. You can just add it here.

labrys:
Actually this being a heavy method was thought about.  I had said the proper way to do it would be to store the proper detection in the database but it was merged as is and without a TODO to go back to it, it kinda fell off the radar and got forgotten about.

labrys:
Approve

PR #674, opened 2016-06-07
Added provider AnimeBytes

p0psicles:

Added backlog, haven't tested RSS yet.

* Needed to manually create the release names.
* Because there isn't any Season pack support for anime shows, i've added the Regex to the end of the anime regexes.
* Added a UnicodeDecodeError catch, because that's going wrong in NameParser() and where going to phase that out anyway.

Todo:
- [ ] Make sure it's not "accidentally" used for episode searches
- [ ] Check that I don't add items from results that where already added before.
- [ ] I don't know if that should be part of this PR, but maybe try to fix the RC for the UnicodeDecodeError exception.
- [ ] Test


p0psicles:
Ab torrent urls change. So about every day you get a new url, meaning a double, tripple, ... searchresult. Need to think about that. Ideas?

labrys:
Do the hashes change?

p0psicles:
I don't know anything about torrents. So what do you mean with do the hashes change? Ik can share examples?

duramato:
@labrys The urls don't have hashes, just torrent ids and user passkeys most likely in this format `SITE/torrents.php?action=download&id=73616&authkey=PASSKEY&torrent_pass=OTHERPASSKEY

p0psicles:
Yes, and the id or another value, change after some time.

duramato:
Most likely the passkey as the id is what id's the torrent on the tracker

p0psicles:
It's something like this: https://animebytes.tv/torrents.php?action=download&id=111945&authkey=01DFSFEFWDFSDFSFSFSF&torrent_pass=DFASDWEDSDSS.

authkey changes.

medariox:
@p0psicles 
Let the user add the authkey and save the url without it, then recreate the full url on snatch?

p0psicles:
The only thing I can think of, is to connect back to AB when you try to snatch, to get the current URL. And just accept the double/tripple.. results in cache. Or set a short cache retention for this provider, making it drop results older then one day. Buy as these url's change frequently, you'l have to rescrape the torrent url.

sickrage/providers/GenericProvider.py
```diff
@@ -180,7 +180,10 @@ def find_search_results(self, show, episodes, search_mode, forced_search=False,
             try:
                 parse_result = NameParser(parse_method=('normal', 'anime')[show.is_anime]).parse(title)
             except (InvalidNameException, InvalidShowException) as error:
-                logger.log(u"{}".format(error), logger.DEBUG)
+                logger.log(u"{error}".format(error=error), logger.DEBUG)
+                continue
+            except UnicodeDecodeError as error:
```
medariox: Do we still need this?

p0psicles: Good point. Lets remove it!

p0psicles:
Did some more tests, searing, parsing and release creation goes well. But still medusa has some bugs, where it's not able to handle season packs / mulieps packs, icw scene namings. As AB is for 90% season packs / multieps / show packs, I think it will not always be as effective.

sickbeard/providers/__init__.py
```diff
@@ -36,7 +36,7 @@
     'xthor', 'abnormal', 'scenetime', 'btdigg', 'transmitthenet', 'tvchaosuk',
     'torrentproject', 'extratorrent', 'bitcannon', 'torrentz', 'pretome', 'gftracker',
     'hdspace', 'newpct', 'elitetorrent', 'bitsnoop', 'danishbits', 'hd4free', 'limetorrents',
-    'norbits', 'ilovetorrents', 'sceneelite', 'anizb', 'bithdtv', 'zooqle'
+    'norbits', 'ilovetorrents', 'sceneelite', 'anizb', 'bithdtv', 'zooqle', 'animebytes',
```
medariox: Maybe remove the last comma for consistency.

p0psicles:
Also rebased on dev

medariox:
Approved

PR #688, opened 2016-06-15
Unicode mass edit

labrys:
OS: Windows 10
Python: 2.7.11
Medusa: Current dev

When using mass edit to change directories containing non-ascii can cause UnicodeError.  This fixes the issue by removing an unnecessary `.decode()`.

Also fixes minor typo.


labrys:
@medariox

medariox:
Approved

PR #701, opened 2016-06-20
Fix HDSpace and SceneTime, improve Zooqle

medariox:


sickbeard/providers/scenetime.py
```diff
@@ -87,7 +85,7 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
                     logger.log('Search string: {search}'.format
                                (search=search_string), logger.DEBUG)
 
-                search_url = self.urls['search'] % (quote(search_string), self.categories)
+                search_url = self.urls['search'].format(search_string, self.categories)
```
labrys: I think this still needs to be quoted because its being passed as part of the URL.  Really the cats and search string should be converted to params to pass instead and then the quote would be unnecessary.

medariox: This has been tested and it is working as expected. It was supposed to be a quick fix for the provider, as it didn't work at all. If you can wait until tomorrow, I can change the strings to params, although the provider would need a full rewrite.

labrys: I know params would be a bigger change, but I think leaving the quoting in would be a good idea until the param change as it would only be an issue on occasion and thus may not have been encountered during testing.

medariox:
Added search params, should be ready now.

sickbeard/providers/hdspace.py
```diff
@@ -180,15 +179,15 @@ def login(self):
...
+            'submit': 'Confirm',
         }
 
         response = self.get_url(self.urls['login'], post_data=login_params, returns='text')
         if not response:
             logger.log('Unable to connect to provider', logger.WARNING)
             return False
 
-        if re.search('Password Incorrect', response):
+        if re.search('Automatic log off after 15 minutes inactivity', response):
```
p0psicles: Is there a specific reason this is one through regex? If it's a simple string, could as well be 'string' in response right?

p0psicles:
Approved

sickbeard/providers/hdspace.py
medariox: We use regex here in all providers. Although, as you said, we could probably just use "in response" instead in most of them. For now I would keep the regex, for a matter of consistency.

PR #706, opened 2016-06-22
Improve daily search (RSS) min_time / search_interval configuration

p0psicles:

@fernandog would this also work? I still need to do more tests, to really reproduce the issue. But if a 10s margin, would resolve it, why not?

### DIFF
--- a/sickbeard/tvcache.py
+++ b/sickbeard/tvcache.py
@@ -330,7 +330,7 @@ class TVCache(object):
 
     def shouldUpdate(self):
         # if we've updated recently then skip the update
-        if datetime.datetime.today() - self.lastUpdate < datetime.timedelta(minutes=self.minTime):
+        if datetime.datetime.today() - self.lastUpdate < datetime.timedelta(minutes=self.minTime, seconds=-10):
             logger.log(u'Last update was too soon, using old cache: {0}. Updated less then {1} minutes ago.'.format
                        (self.lastUpdate, self.minTime), logger.DEBUG)
             return False
--- a/sickbeard/tvcache.py
+++ b/sickbeard/tvcache.py
@@ -330,7 +330,7 @@ class TVCache(object):
 
     def shouldUpdate(self):
         # if we've updated recently then skip the update
-        if datetime.datetime.today() - self.lastUpdate < datetime.timedelta(minutes=self.minTime, seconds=-10):
+        if int((datetime.datetime.today() - self.lastUpdate).total_seconds() / 60) <= self.minTime:
             logger.log(u'Last update was too soon, using old cache: {0}. Updated less then {1} minutes ago.'.format
                        (self.lastUpdate, self.minTime), logger.DEBUG)
             return False


p0psicles:
Keeping this stashed for the time beeing:
```python:
  def shouldUpdate(self):
          # if we've updated recently then skip the update
 -        if datetime.datetime.today() - self.lastUpdate < datetime.timedelta(minutes=self.minTime):
 +        # Remove seconds and milliseconds because mintime = "10:00" so we compare only minutes
 +        update_diff = datetime.datetime.today().replace(second=0, microsecond=0) - self.lastUpdate.replace(second=0, microsecond=0)
 +        if update_diff < datetime.timedelta(minutes=self.minTime):
              logger.log(u"Last update was too soon, using old cache: " + str(self.lastUpdate) + ". Updated less then " + str(self.minTime) + " minutes ago", logger.DEBUG)
              return False

**Comment by fernandog** [comment]
to reproduce simple make daily to 10 and monitor a provider that has min_time = 10
the same can happen with 30min and 30min. But the issue will be worst as cache will be update in 60 min only

**Comment by p0psicles** [comment]
Just tested and, I can't exactly reproduce it. I've reconfigured anizb to do the daily each 10 min. But it as you can see in the log, it's doing them sometimes after 40 min and sometime after 80 min. Buy maybe i'm missing something.

```log:
2016-06-21 17:54:39 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Anizb] :: [881d122] Search mode: RSS
2016-06-21 17:14:39 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Anizb] :: [881d122] Search mode: RSS
2016-06-21 16:34:39 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Anizb] :: [881d122] Search mode: RSS
2016-06-21 15:54:39 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Anizb] :: [881d122] Search mode: RSS
2016-06-21 15:14:38 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Anizb] :: [881d122] Search mode: RSS
2016-06-21 14:34:37 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Anizb] :: [881d122] Search mode: RSS
2016-06-21 13:54:37 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Anizb] :: [881d122] Search mode: RSS
2016-06-21 13:14:34 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Anizb] :: [881d122] Search mode: RSS

**Comment by fernandog** [comment]
@p0psicles min_time for anizb is 10min and dailysearcher frequency is 10min?

**Comment by p0psicles** [comment]
jip. Well mintime for anizb is not set:
```python:
        # Cache
        self.cache = tvcache.TVCache(self)
```

And search interval is set to 10min.

duramato:
then its in the default, 20min. You can set a custom time

fernandog:
@p0psicles if default is 20, then you have the issue. is not updating in 20min but in 40min

fernandog:
@p0psicles i didn't test the commit I just did. I only evaluated logic.

medariox:
@duramato @fernandog 
Why do you think that min_time default is 20 min? It looks like it's 10 min to me: https://github.com/pymedusa/SickRage/blob/d0c1d270f08bc1b84183d81a1ea5b1d70cacc426/sickbeard/tvcache.py#L105

duramato:
Idk I said from the top of my head,  think it used to be 20

fernandog:
@p0psicles here my commit test. 

Note it didn't updated because of one second:
`2016-06-22 12:27:46 Last update was too soon, using old cache: 2016-06-22 12:22:47. Updated less then 5 minutes ago`

If user has a lot of shows, he will have a lot of logs like this: "Seeing if we need anything from".
More shows, more time it will take to check if something is needed. Then it will check if is time to update cache. IMO fewer wanted shows, issue is more frequent.

So we will still need the commit from pops or another solution

```
2016-06-22 12:22:46 INFO     SEARCHQUEUE-DAILY-SEARCH :: Beginning daily search for new episodes
2016-06-22 12:22:46 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: Seeing if we need anything from *********** (FIRST)
2016-06-22 12:22:46 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: Seeing if we need anything from *********** (LAST)
2016-06-22 12:22:48 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: Search mode: RSS
2016-06-22 12:22:47 INFO     SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: Setting LastUpdate to: 2016-06-22 12:22:47.403279

2016-06-22 12:27:45 INFO     SEARCHQUEUE-DAILY-SEARCH :: Beginning daily search for new episodes
2016-06-22 12:27:45 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: Seeing if we need anything from *********** (FIRST)
2016-06-22 12:27:46 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: Seeing if we need anything from *********** (LAST)
2016-06-22 12:27:46 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: Last update was too soon, using old cache: 2016-06-22 12:22:47. Updated less then 5 minutes ago



2016-06-22 12:37:46 INFO     SEARCHQUEUE-DAILY-SEARCH :: Beginning daily search for new episodes
2016-06-22 12:37:46 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: Seeing if we need anything from *********** (FIRST)
2016-06-22 12:37:46 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: Seeing if we need anything from *********** (LAST)
2016-06-22 12:37:46 INFO     SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: Setting LastUpdate to: 2016-06-22 12:37:46.711870
2016-06-22 12:37:47 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: Search mode: RSS
```

labrys:
I don't see what the point is?  It's min time not scheduled search time.  The purpose isn't to do it on an exact schedule.  It hasn't met the min time, sure by only one second, but it still hasn't met the minimum time.

ratoaq2:
Isn't this similar to https://github.com/pymedusa/SickRage/blob/master/sickbeard/subtitles.py#L862

The time resolution should be minute

fernandog:
@ratoaq2 yes.

OliRM:
I'm the owner of anizb. Not sure how much help it will be, but if you want me to implement something specific, or clarify anything about the (garbage) API that it has running, let me know.

p0psicles:
@labrys from a developer perspective I understand what your saying, and I even can't argue with it. But as a user i'd to have some indication as to what the schedule is. The min-time gives the indication that, that can be used as a schedule. Okay, when that's 10 min, doing the rss after 20 isn't that of an issue. But when a user sets it back to 60, I doubt that, that user knows it "might" use an interval of 120.

labrys:
At an interval of 60 it shouldn't ever use an interval of 120 minutes unless we have hard-coded the provider min-time to 60 minutes.  And the purpose of provider min time is to not hit the providers more frequently than necessary, otherwise it would be a user-configurable option per provider.

ratoaq2:
This seems a similar issue that I had with subtitles. At that time, I had the schedule to run every 8 hours and it was happening only after 16h. The issue is that the schedule executes precisely after 8h, but the last_timestamp was set a bit before, and the time slot was always missed by 1 or 2 seconds... Sometimes just a few ms. 

Then I changed the time resolution to minutes instead of comparing the full timestamp (seconds and milliseconds). So the difference between delta and search interval should always be bigger than 60 seconds

labrys:
@ratoaq2 would you mind hopping in IRC rq?

sickbeard/tvcache.py
```diff
@@ -179,14 +179,19 @@ def updateCache(self):
...
+            # as I showed in logs
+            self.setLastUpdate()
+
             data = self._getRSSData()
             if self._checkAuth(data):
                 # clear cache
                 self._clearCache()
 
-                # set updated
-                self.setLastUpdate()
```
p0psicles: Don't know @fernandog. I think theres a reason you mark a search as 'updated' after you know for sure you can authenticate. You maybe could move it up the _clearCache, but that's about it.

p0psicles:
Talked with labrys about it. These came out of it.
1. The goal of the feature min_time. That is there solely to make sure providers are not hit (In case of the 10 min.) more often then 10 minutes. Meaning if we check 10 minutes after the last search was registered in db, a new search is not allowed. After 10 minutes and 1 s, it is. This works as intended.

2. How users interpreted the min_time solution. That's maybe something we can fix easily by documenting the feature on the wiki (if not already done)

As an improvement I see two thing we can do:
1. Make the min_time a little bit more forgiving. We could do this al @ratoaq2 adviced, by checking by the minute, instead of using the timedelta. As that compares by the second. Or use a margin, like I proposed earlier.
2. Visualize the effective interval.

For now I'm going to change this to "Feature Request". If we can agree really fast on an improvement i'll keep the PR open. Else it should be candidate to be closed, and moved to a PR.

fernandog:
@p0psicles I can't find any reason for keeping where its is today. Can you?

LastUpdate = last time we query provider. So it should be as soon we do the first query and not after we parse all items, don't you agree? 

Otherwise the min_time vs lastUpdate logic is wrong.

p0psicles:
Like ive tried to explain, min_time and lastUpdate variables doet have anything to do with each other. Min_time is al safety restriction. Doesnt mean we cant improve for user.

fernandog:
@p0psicles 
ofc they are related. min_time is checked as soon you start daily (daily is supposed to get new items - respecting min_time)

if daily is set to 20min and min_time (ProviderA) is 10 min then first daily won't update cache for ProviderA, Second daily will update (20min), third won't update (30min), fourth will update (40min)

Isn't the expected behavior? 

with this issue, the second update (20min) might not update the ProviderA cache and it will be only updated in the fourth run (40min)

just check your logs

labrys:
If you want to avoid this possibility simply choose non-resonant frequencies.

p0psicles:
@pymedusa/developers Quick vote guys. I propose to make the change to compare by minute, like in the last commit. Pls thumbs up/down, to finish this!

labrys:
I vote no because imo this solves nothing.  An actual scheduler with polling would be a proper fix.  It breaks the provider min time (granted just by up to 60 seconds, but it still violates an established limit that is there for a reason, and is bad practice).  The 60 second hysteresis which this adds could be applied in virtually the same way by the user adding one minute to their daily search time.  At most this causes a single skipped period for a provider.   Also it does not prevent a similar situation from happening at the 61 second mark or the 1 second mark if the users daily search time is 1 minute off from a min-search interval.

fernandog:
Yes.
as I said before: if daily runs every 10min and min_time is 10 it should always update cache. setLastUpdate is in the wrong point of code.

labrys:
@fernandog setLastUpdate should occur after its successfully completed as it is since you don't want to set a last update time if it didn't complete successfully.  The search time set, however, should be the time of the search not the time when setLastUpdate is called which is what I believe you mean.  Even this could lead to similar racing conditions tho without a proper scheduler but it would reduce the likelihood somewhat.

fernandog:
As I said before Min time is interval between provider queries. That was the concept when developed. It should be set in the FIST hit/query. So last fisrt query and new query should be at least "min_time" minutes

Im done talking about this. My vote is yes

medariox:
Do you guys realize that this changes makes it even _less_ precise?
Let's say 356s passed. 356/60 = 5.9333 min, but int() would make it 5 min, although almost 6 min passed.
After reading the whole PR, I still don't know what the point of this is.

fernandog:
@p0psicles why not use min_time with timedelta instead of INT.
it will be explicit and we can control the values.

p0psicles:
Not enough support for a change.

fernandog:
3x1? lol. also waiting medariox and rato

PR #712, opened 2016-06-24
Make BTN more decent

medariox:

Provider was a mess, now it's more decent (aim is not a full rewrite).

### DIFF
--- a/sickbeard/providers/btn.py
+++ b/sickbeard/providers/btn.py
@@ -83,57 +83,56 @@ class BTNProvider(TorrentProvider):
             items = []
             logger.log('Search mode: {0}'.format(mode), logger.DEBUG)
 
-            for search_string in search_strings[mode]:
+            if mode != 'RSS':
+                search_params = self._episode_search_params(ep_obj)
+                logger.log('Search string: {search}'.format
+                           (search=search_params), logger.DEBUG)
 
-                if mode != 'RSS':
-                    search_params = self._episode_search_params(ep_obj)
-                    logger.log('Search string: {search}'.format
-                               (search=search_params), logger.DEBUG)
+            if mode == 'Season':
+                search_params = self._season_search_params(ep_obj)
 
-                if mode == 'Season':
-                    search_params = self._season_search_params(ep_obj)
+            response = self._api_call(self.api_key, search_params)
+            if not response:
+                logger.log('No data returned from provider', logger.DEBUG)
+                return results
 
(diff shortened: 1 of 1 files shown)


sickbeard/providers/btn.py
```diff
@@ -73,78 +73,89 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint:disable=too-many
...
-        if age:
-            search_params['age'] = '<=' + str(int(age))
+        for mode in search_strings:
+            items = []
+            logger.log('Search mode: {0}'.format(mode), logger.DEBUG)
 
-        if search_strings:
-            search_params.update(search_strings)
-            logger.log('Search string: {0}'.format(search_strings), logger.DEBUG)
+            for search_string in search_strings[mode]:
```
p0psicles: If search_string is really not needed, we can do: for _ in search_strings[mode].
But I don't really see a reason to loop over the search_strings? Then might better remove it all together, to prevent future confusion?

sickbeard/providers/btn.py
```diff
@@ -73,78 +73,89 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint:disable=too-many
...
         if 'api-error' in parsed_json:
             logger.log('Incorrect authentication credentials: {0}'.format(parsed_json['api-error']), logger.DEBUG)
-            raise AuthException('Your authentication credentials for {0} are missing,'
-                                ' check your config.'.format(self.name))
+            return False
 
         return True
 
-    def _get_title_and_url(self, parsed_json):
+    def _get_title_url(self, parsed_json):
```
p0psicles: They also have a point here. _get_title_and_url is inherited (or in this case overwritten). But _get_title_url is not. So this could be a @classmethod or @staticmethod.

medariox: Yes, they should all be static methods. I will change them asap.

sickbeard/providers/btn.py
medariox: For how the provider is currently written, search_string is indeed not needed, as all needed strings are added in the methods. That's why I will remove it for now. Still, when we get access to this provider, a full rewrite should be aimed - including proper standardization.

p0psicles:
Approved

[Approved with PullApprove](https://pullapprove.com/pymedusa/SickRage/pull-request/712/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=pymedusa/SickRage)

PR #714, opened 2016-06-26
Guessit name parser implementation

ratoaq2:
- **NameParser**: flag `use_guessit` that enables guessit name parser (default `true` on this branch). Setting it to false will keep the current parser without any change.
- **GuessitNameParser**: This should be the new API and it should be really small and straightforward. Right now, it plugs in our own `rules` and it calls `guessit.guess(..)` with the correct parameters and adapt the response property names to the same ones `NameParseResult` has. I'm planning to move the `expected_titles` and `expected_groups` out of this file to a yaml file. But this list should always be short and simple. Another idea is to use the shows library to collect all shows that has numbers (since these are the problematic ones) and add them to `expected_titles`. Add all shows names is a valid option as well.
- **rules/properties.py**: Additional property matches that guessit doesn't support (SATRip, BDMux, etc)
- **rules/rulespy**: Manipulate `matches`. I wrote some instructions and guidelines to always have each rule simple and maintainable. Some rules are fixes/workarounds for guessit bugs. I have flagged them with each issue url, so we can remove then when guessit is fixed.
- **Tests**: I have created a simple unit test that executes against an yaml file. For each regex in regexes.py and each rule in our rules.py, I added an entry in this yaml file, so they will work will regression tests. There are positive and negative test cases.


tests/guessit_tests.py
```diff
@@ -0,0 +1,116 @@
+# coding=utf-8
+"""Guessit name parser tests."""
+
+from __future__ import unicode_literals
+
+import os
+import unittest
+from io import open
```
p0psicles: you can do from io import open as io_open

labrys: io.open() is an alias to open() so just use open()

ratoaq2: I've changed this after the import `unicode_literals`, my initial though was that I wasn't loading the file with the correct encoding

I took a look how guessit was loading the yaml file, then I tried the same:
https://github.com/guessit-io/guessit/blob/master/guessit/test/test_yml.py#L6

But now, I already know the real issue: https://github.com/guessit-io/guessit/issues/326

and I'll just drop this part as it's not needed.

labrys:
Approved

PR #724, opened 2016-06-30
Use guessit for Proper detection in search.py, lint a bit (to decency)

medariox:


sickbeard/search.py
```diff
@@ -111,7 +108,7 @@ def snatchEpisode(result, endStatus=SNATCHED):  # pylint: disable=too-many-branc
         for curEp in result.episodes:
             if datetime.date.today() - curEp.airdate <= datetime.timedelta(days=7):
                 result.priority = 1
-    if re.search(r'(^|[\. _-])(proper|repack)([\. _-]|$)', result.name, re.I) is not None:
+    if guessit(result.name).get('proper_count'):
```
medariox: This is the actual change.

sickbeard/name_parser/parser.py
```diff
@@ -570,6 +575,10 @@ def is_anime(self):
             return True
         return False
 
+    @property
```
p0psicles: There is not really a reason to create a property for this. i think you can just choose what you want to use. proper or is_proper. My preference is: is_proper, as that indicates, it is a bool.

medariox:
@labrys 
That's not possible, at least not with guessit. The guessit match doesn't hold that kind of information. It might be a useful feature request for upstream though.

labrys:
```python
>>> name = 'Show S01E02 - real proper repack'
>>> guess = guessit(name, dict(advanced=True, implicit=True))
>>> for each in x.get('other'): each.raw
'real'
'proper'
'repack'
```

medariox:
I obviously don't have enough time to look into all the guessit options. Will open a new PR for the icons.

PR #730, opened 2016-07-05
Lint search.py, fix 2 bugs

medariox:

This also includes following fixes:
- [Unique seasons fix](https://github.com/SickRage/SickRage/pull/2027). Thanks to @guusdk
- [SQL syntax fix](https://github.com/SickRage/SickRage/pull/2029). Thanks to @dgholz

### DIFF
--- a/sickbeard/search.py
+++ b/sickbeard/search.py
@@ -1,22 +1,22 @@
 # coding=utf-8
 # Author: Nic Wolfe <nic@wolfeden.ca>
-
+#
 # Git: https://github.com/PyMedusa/SickRage.git
 #
-# This file is part of SickRage.
+# This file is part of Medusa.
 #
-# SickRage is free software: you can redistribute it and/or modify
+# Medusa is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
 #
-# SickRage is distributed in the hope that it will be useful,
+# Medusa is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


sickbeard/search.py
```diff
@@ -41,20 +38,22 @@
...
 from sickrage.providers.GenericProvider import GenericProvider
 from sickrage.helper.common import enabled_providers
 
+from socket import timeout as SocketTimeout
+
 
 def _downloadResult(result):
     """
-    Downloads a result to the appropriate black hole folder.
+    Download a result to the appropriate black hole folder.
```
ratoaq2: Is there a point in removing `s` and `es` from all the conjugated verbs?

`( The method / It ) Downloads a result to the appropriate black hole folder.`

I'm just asking since I saw @labrys doing the same... I'm a java guy and when working with java, all APIs (including the official java API) uses the verbs in the 3rd person as you can see it [here](http://docs.oracle.com/javase/8/docs/api/java/util/Collections.html)

medariox: Yes, it's part of [PEP 257 -- Docstring Conventions](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).
Although not explicitly mentioned, first line of multi-line docstrings should be in imperative mood and end with a point.
As you can also see it from the example in PEP 257:
```
def complex(real=0.0, imag=0.0):
    """Form a complex number.

    Keyword arguments:
    real -- the real part (default 0.0)
    imag -- the imaginary part (default 0.0)
    """
    if imag == 0.0 and real == 0.0:
        return complex_zero
    ...
``` 
[Other examples](http://dolphm.com/pep257-good-python-docstrings-by-example/).

Personally, I prefer the 3rd person as well, but I think PEP should have priority over personal preferences.

ratoaq2: Good to know. I'll make sure my code follows the standard 👍

labrys:
Approved

PR #731, opened 2016-07-05
Add confirmed|ranked icon next to provider and tooltip

fernandog:
@p0psicles complement to your old PR adding the icons

Sure icon can be changed.
Icons suggestion?

image


fernandog:
@labrys waiting your idea to this

p0psicles:
IMHO I don't really like it. It gets way to cramped there with all the icons there. There also all mixed together. Like you can devide them into two groups. The provider options, and provider types. The d,b,m icons are options. They can be changed by user. And the lock and new 'confirmed' icon are providers attributes. They can't be changed by user.

I also don't have a better option now. I think we should have a UX'r redesign it.

fernandog:
@p0psicles the "trusted|confirmed" are also options. User can choose if he want to download from trusted|confirmed uploaders only.

fernandog:
@thraxis ideas to improve this?

PR #732, opened 2016-07-05
Fix encoding userid/password before logging in to SMTP server

fernandog:
Fix from @avdleeuw thanks!

@medariox


medariox:
Approved

PR #742, opened 2016-07-07
Unwanted scene release to DEBUG instead of WARNING

fernandog:
I forget to change back this one when I rewrote ignored/required/... words


labrys:
Approved

PR #746, opened 2016-07-09
TVShow and TVEpisode cleanup

ratoaq2:
- Single quotes
- PEP8: Snake case (properties and methods)
- PEP8: Line length to 120 chars
- PEP8: Shadowing names from outer scopes
- Reducing visibility of some methods
- PEP257 fixes


labrys:
Approved

PR #759, opened 2016-07-13
Add 'sub' and 'idx' to default allowed extensions

fernandog:


duramato:
Why add .sub and not any of the other subtitle formats that exist out there? 
And now that i see it , why are `srr` and `sfv` even there?

fernandog:
@duramato legacy from sickbeard

fernandog:
@ratoaq2 @medariox thoughts?

medariox:
Approved

PR #761, opened 2016-07-13
Logger improvements (Lazy logging)

ratoaq2:
Lazy logging is here :+1: 

- PEP257 and PEP8 fixes.
- Add support for new string format style: for current custom logger and standard logging (lazy logging)
- There's no code change on other classes to use it. This PR just adds support for it.

How to use:
```
import logging

logger.info('Really {param} message', param='nice')
logger.log(logging.DEBUG, 'Hello {world}', world='rato')
logger.debug('{what} idea', what='No')
```

### DIFF
--- a/sickbeard/logger.py
+++ b/sickbeard/logger.py
@@ -17,10 +17,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with SickRage. If not, see <http://www.gnu.org/licenses/>.
-
-"""
-Custom Logger for SickRage
-"""
+"""Custom Logger."""
 
 from __future__ import unicode_literals
 
@@ -97,11 +94,16 @@ level_mapping = {
 
 
 class ContextFilter(logging.Filter):
-    """
-    This is a filter which injects contextual information into the log, in our case: commit hash
-    """
+    """This is a filter which injects contextual information into the log, in our case: commit hash."""
 
     def filter(self, record):
+        """Filter to add commit hash to log record, adjust log level and to add exception traceback for errors.
+
+        :param record:
+        :type record: logging.LogRecord
+        :return:
+        :rtype: bool
+        """
         cur_commit_hash = sickbeard.CUR_COMMIT_HASH
         record.curhash = cur_commit_hash[:7] if cur_commit_hash and len(cur_commit_hash) > 6 else ''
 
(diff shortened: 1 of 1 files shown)


p0psicles:
Been testing some. But what would be the best way of passing multiple parameters?
```python
logger.warning('Start method {name}, test logging second var: {var2}', name='10', var2='3')
```

ratoaq2:
Maybe @labrys could give some input here.

I was reading on how python 3 handles logging and I found [this](https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles):

> Logging calls (debug(), info() etc.) only take positional parameters for the actual logging message itself
> There would no changing this while preserving backward compatibility, since all logging calls which are out there in existing code will be using %-format strings.

So, the default logging behavior will not change due to backwards compatibility.

If you want to use new style they propose to use like this:
```
__ = BraceMessage

print(__('Message with coordinates: ({point.x:.2f}, {point.y:.2f})', point=p))
print(__('Message with $num $what', num=2, what='placeholders'))
````

So using that in logging would be:
```
logger.info(__('Message with {num} {what}', num=2, what='placeholders'))
```

My implementation was just based on the very same idea, but trying to remove the ugly/verbose part from it and making that transparent for the developer.

This is a decision that needs to be agreed among all of us and we need to stick with that. Otherwise we'll only waste time. I'm in favor of having a clear defined standard.
And among the options, I vote for:
```
import logging
logger.info('{param1} {param2}', param1='nice', param2='catch')
```

Thraxis:
Going to presume there is also a logger.warning

ratoaq2:
The logger API from standard python logging module hasn't changed at all. 

This PR only makes sure that we can use them with this new format style

p0psicles:
@ratoaq2, as a test i'm going to use this with the trakt recommended lists PR. So i'm going to rewrite all log lines in that PR. If that all works, it's tested from my POV, and should be gould to merge.

Thraxis:
With this change would it make more sense  to hold off on making changes to log lines to use .format() and wait until this is added and make the change to the new method?

#778, #799

p0psicles:
@Thraxis you got to start somewhere right? Also if it's little effort to rewrite the log to format, we can do it. But we shouldn't refactor large modules for now.

p0psicles:
Approved

fernandog:
rebased

PR #766, opened 2016-07-15
Fix routes and location called from logger

p0psicles:

* Fixed route, deleteShow was renamed to delete_show. That's okay for variables, but with routes we should be extra careful.
* Fixed bug where location is called as an attribute, but the attribute itself is verifying the location. Which does not work, if the file/location has been deleted from disk.

Fix for: https://github.com/pymedusa/SickRage/pull/746
And issue: https://github.com/pymedusa/SickRage/issues/765

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1205,10 +1205,10 @@ class TVShow(TVObject):
                 else:
                     ek(shutil.rmtree, self.location)
 
-                logger.log(u'%s show folder %s' % (('Deleted', 'Trashed')[sickbeard.TRASH_REMOVE_SHOW], self.location))
+                logger.log(u'%s show folder %s' % (('Deleted', 'Trashed')[sickbeard.TRASH_REMOVE_SHOW], self.raw_location))
 
             except ShowDirectoryNotFoundException:
-                logger.log(u'Show folder does not exist, no need to %s %s' % (action, self.location), logger.WARNING)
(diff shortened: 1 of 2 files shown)


p0psicles:
Approved

sickbeard/tv.py
```diff
@@ -1205,10 +1205,10 @@ def delete_show(self, full=False):
...
                     ek(shutil.rmtree, self.location)
 
-                logger.log(u'%s show folder %s' % (('Deleted', 'Trashed')[sickbeard.TRASH_REMOVE_SHOW], self.location))
+                logger.log(u'%s show folder %s' % (('Deleted', 'Trashed')[sickbeard.TRASH_REMOVE_SHOW], self.raw_location))
 
             except ShowDirectoryNotFoundException:
-                logger.log(u'Show folder does not exist, no need to %s %s' % (action, self.location), logger.WARNING)
+                logger.log(u'Show folder does not exist, no need to %s %s' % (action, self.raw_location), logger.WARNING)
             except OSError as e:
                 logger.log(u'Unable to %s %s: %s / %s' % (action, self.location, repr(e), str(e)), logger.WARNING)
```
p0psicles: we should do this one also!

medariox:
Approved

sickbeard/tv.py
ratoaq2: Yeah... It will be more resilient. 
The show.location is the only property that has a validation. Unfortunately I missed that part were it was valid at the beginning of the method and it becomes invalid during its execution. I'll do a sanity check tonight again on all places were we use it.

PR #769, opened 2016-07-15
Dirty flag fix

### LINKED ISSUE #768: Preview Rename Issue
### Before submitting your issue:

Enable debug logging in SickRage settings, reproduce the error (be sure to disable after the bug is fixed)

files before rename: http://www.fshare.it/image.jpeg?i=721468588hb
files then manually checked in shows folder and they are renamed when using preview rename. but:
db after rename: http://www.fshare.it/image.jpeg?i=se1468588m4

(had to rename another show so the images above shows the wrong show, but same result both on the page and in db.. old name persists, but file is renamed in folder)

Branch/Commit: develop width dbdebug=1 in config.ini
OS: Windows 8
What you did: tried to rename files
What happened: files were renamed but not updated in database
What you expected: that database would update with new names
Logs: https://0bin.net/paste/QGobOqbhwpjeVoiY#BOYrJHm-wUL3Ldym4+gFz9MUyBBT+jWHvzP2Er4FgXQ

ratoaq2:

This should fix #768 
- location changes should make object dirty; and dirty flag shouldn't be resetted when changed value evaluates to false.

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1205,10 +1205,12 @@ class TVShow(TVObject):
                 else:
                     ek(shutil.rmtree, self.location)
 
-                logger.log(u'%s show folder %s' % (('Deleted', 'Trashed')[sickbeard.TRASH_REMOVE_SHOW], self.raw_location))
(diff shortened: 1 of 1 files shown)


fernandog:
Approved

PR #773, opened 2016-07-16
Fixing guessit multiple formats

ratoaq2:
This is a guessit rule until https://github.com/guessit-io/guessit/issues/327 is fixed.

This rule will detect multiple formats and will keep the last format if the last format is surrounded by screen_size, video_codec and release_group

The episode_title is not handled/fixed since it doesn't cause issues for us.

Fixes https://github.com/pymedusa/SickRage/issues/772


fernandog:
Approved

PR #145, opened 2016-03-14
Added Tvmaze

p0psicles:
This PR is going to add Tvmaze indexer support.

As of yet it is in progress. But should already be "usable-ish".

Part of the PR (at this time) is cachecontrol upgrade. But as this is a lib upgrade, this can just as well be PR'd into Dev as a separate PR.

Proposed changes in this PR:
- Full TVmaze indexer support
- Supports TVmaze show, season, episode and casts indexing
- ?

TODO:
- [x] Code cleanup (especially pytvmaze_api.
- [x] _parseActors support, so it can be used like: t['show name]['_actors']. Don't know where it's used, yet.

~~\- [ ] Separate the cachecontrol lib changes into commit.~~
- [x] remove the cachecontrol lib changes.
- [x] change pytvmaze_api to tvmaze_api as i've merged the pytvmaze api with the existing tvdb_api structure. So eventually there won't be allot of pytvmaze api code left.
- [x] create a better tvmaze icon. I just resized the official icon to 16x16. Maybe someone can do a better job?
- [x] Test and probably fix the show updating
- [x] Test and probably fix metadata creation
- [x] Added indexer 2 indexer mapping, when adding new shows. Tvmaze shows are now mapped to tvdbid in indexer_mapping table when a tvdb_id is provided.
- [x] Fix errors regarding scene_name_exceptions
- [x] Missing genre and network
- [x] Add episode summaries
- [x] Fix the unit tests
- [x] Think of something for scene showname/season/episode searches. Anime doesn't work atm. Implemented a mapping to tvdb (made it configurable through the indexer_config) https://github.com/pymedusa/SickRage/pull/145/commits/02520ba0f208709624be728f8c3580da2ddfd9a5
- [ ] Implement the mapping also within the medusa scene exceptions

PR #779, opened 2016-07-17
Replace legacy tvdb_api with tvdb api v2

p0psicles:
### The goals of this PR are:
1. Restructure the indexer code, creating a more modular setup, where it's less work to add new indexers compared to how it is currently setup.

2. Replace the 'legacy' thetvdb indexer code with the api v2. https://api.thetvdb.com/swagger.
All tough allot the added api code is now in this PR, it will be added as a separate PR, as the api, should be fully independent from the indexer's code. This probably the biggest change from this PR. As up until now, api and indexer where pretty much the same.

A followup will be adding the tvmaze indexer. As is waiting in #145 

### To accomplish this the following has to be done:
- [x] Make sure the swagger client has all the features as the legacy client.
- [x] Separate the indexer code into a lib part and medusa indexer code. Currently they're tangled.
- [x] Add the v2 code as a new indexer next to the legacy indexer. The legacy indexer code will be replaced. But for testing purposses, they should work fine next to each other. The ID used will be '4' for now, as 3 has been reserved by tvmaze and '2' is used by tvrage.
- [x] When tested, remove the legacy indexer code, and make sure everything works fine, including the scene names / season scene names / scene episode mappings.

### Log:
This is a list to document what needs to be taken care of when implementing a new indexer. I'll update this as I move along.

- [x] Show search by name
- [x] Show search by id, indexing the show attributes, seasons, episodes, banner, etc. `Is working, except for the banner.`
- [x] Episode summaries
- [x] Add metadata creation
- [x] Add banner and poster support if available
- [x] Add actors information (Used for metadata creation)
- [x] Remove all references to tvdb legacy indexer. The api, might also be called directly.
- [x] Add multilingual support, (seach shows, get show data, get episode data and get images in the locale configured).
- [x] Add timezone and network information
- [ ] Add indexer mappings (when indexer providers indexer ids to other indexers)
- [x] Add the indexer icon
- [ ] Lot's of cleanup of the tvdbv2 indexer files (pep8, snake_case, unicode_literals)
- [ ] Add new unit tests @ratoaq2 could really need some help here.
- [x] Implement a new method for show updating, as we cannot use the old tvdb xml's anymore. This will need allot of testing and refinement!

### Tvdb Apiv2 Lib related
- [x] Make sure the api returns unicode. Test by searching: 'guide to pretty'. Currently i've adjusted the mapping for the Series modal. Need to also check for others, like episode, actors, etc.
- [x] Maybe replace urllib3 with requests. But the way it has been setup, that might not be that easy. Not needed, as for urllib3, we dont need to install any additional libs.
- [ ] Add searching with apostophes '.

### showUpdates related
More info: https://github.com/pymedusa/Medusa/wiki/Show-indexer-updating
For this the following needs to be implemented:
- [x] Support for force full update
- [x] Update only seasons that reached their next_update.
- [x] Add new shows, should set the next_updates
- [x] Episodes that start a new season, should update the cache indexer_update table with the next_update.
@fernandog, did you have some other use cases/flows?

### Screenies
v2 api is already giving more results back than the old one.
image

v1:
image

Update 01/10: From the list of unchecked bullets, it looks like this PR is far from finished. There is still allot of work/reviewing to be done. But this one is already at like 80%. With the major features in it.

### Legacy, places where TVDB constant is used. Need to check  these. I'm creating a note for it, because if were moving to multi indexer in the future, these need to go.
* /medusa/databases/main_db.py
* /medusa/providers/torrent/json/rarbg.py
* /medusa/scene_exceptions.py
* /medusa/server/web/home/handler.py
* /medusa/tv.py
* /tests/conftest.py


p0psicles:
It's not totally working right now. It's hanging on saving the poster/banner, as it's not getting a full url from the tvdb api. But it is adding to db, so it's somewhat testable.

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -0,0 +1,877 @@
...
+
+import cgi
+import getpass
+import logging
+import os
+import re
+import tempfile
+import time
+import warnings
+from six import next, iteritems
```
p0psicles: hmm okay, how should I resolve this? import it like: from six import next as next_six? As i'd like to use a python2/3 compatible method.

medusa/indexers/tvmaze/tvmaze_api.py
```diff
@@ -0,0 +1,816 @@
...
+            if key not in self.shows:
+                self._getShowData(key, self.config['language'], True)
+            return self.shows[key]
+
+        key = str(key).lower()
+        self.config['searchterm'] = key
+        selected_series = self._getSeries(key)
+        if isinstance(selected_series, dict):
+            selected_series = [selected_series]
+        [[self._setShowData(show['id'], k, v) for k, v in show.items()] for show in selected_series]
```
p0psicles: I think I can resolv this with _ = [[self._setShowData..., but would that be a proper solution?

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -0,0 +1,782 @@
...
+        auth_api = AuthenticationApi(unauthenticated_client)
+        access_token = auth_api.login_post(authentication_string)
+        auth_client = ApiClient(api_base_url, 'Authorization', 'Bearer ' + access_token.token)
+        self.search_api = SearchApi(auth_client)
+        self.series_api = SeriesApi(auth_client)
+
+    def _parse_show_data(self, indexer_data, parsing_into_key='series'):  # pylint: disable=no-self-use
+        """ These are the fields we'd like to see for a show search"""
+
+        name_map = {
```
p0psicles: This method has been removed/replaced.

OmgImAlexis:
@p0psicles does the new API also allow searching for things like `mr` as well as `mr.`?

p0psicles:
@OmgImAlexis if your referring to mr. robot. No, you still need to search for mr. robot. mr robot, won't work.

medusa/indexers/indexer_ui.py
```diff
@@ -0,0 +1,131 @@
+# coding=utf-8
+# Author: p0psicles
+#
+# This file is part of SickRage.
```
OmgImAlexis: The whole header needs to be replaced with the Medusa version.

p0psicles: Yeah, probably more of them need to be fixed. Tnx!

p0psicles:
@pymedusa/developers pls, start testing this over time. As I need to test the new updating.
I've also made some changes to the unicode stuff. So i'm curious how this will work on different setups.

fernandog:
@p0psicles rebase please so we can merge this first after release

p0psicles:
Rebased.

fernandog:
Can you fix so we can merge?

PR #786, opened 2016-07-18
 Clean up any empty season folders after deletion of associated files

fernandog:
For some old shows I had a lot of empty season folders: Season 1, Season 2, etc...

@ratoaq2 as I said the ideia

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1330,6 +1330,10 @@ class TVShow(TVObject):
                                 logger.log(u'Could not delete associated file: {0}. Error: {1}'.format
                                            (related_file, e), logger.WARNING)
 
+        # clean up any empty season folders after deletion of associated files
+        for sub_dir in ek(os.listdir, self.location):
+            helpers.delete_empty_folders(ek(os.path.join, self.location, sub_dir), self.location)
+
         if sql_l:
             main_db_con = db.DBConnection()
             main_db_con.mass_action(sql_l)


fernandog:
@labrys so what do you think?

labrys:
I'd like some clarification first.  How did you end up with the empty season folders?  If they are a result of a Medusa action, I would say cleaning up the folders should be done (but only post that action of course).  If this is just to clear empty folders as a library clean-up action post some external process then I would say that this is not necessary and should not be included.  Some users may have a use-case for leaving empty season folders.

fernandog:
@labrys as in the PR name:
@p0psicles 
"Clean up any empty season folders **after deletion of associated files**"

when I watch the episode and delete it, medusa will remove any leftovers associated files. so "Season 1" folder (example) will be empty

fernandog:
@p0psicles @labrys ?

sickbeard/tv.py
```diff
@@ -1313,6 +1313,10 @@ def refresh_dir(self):
                                 logger.log(u"Could not delete associate file: {0}. Error: {1}".format(related_file, e),
                                            logger.WARNING)
 
+        # clean up any empty season folders after deletion of associated files
+        for sub_dir in ek(os.listdir, self.location):
+            helpers.delete_empty_folders(ek(os.path.join, self.location, sub_dir), self.location)
```
fernandog: @medariox here its ok. using `ek`

PR #787, opened 2016-07-18
Show's Library: stop download other wanted languages when we already downloaded one

fernandog:
This only works if show is in the Medusa library (after PP)

For users that have wanted 'english' and 'german' (for example) as subtitles configured for a show. But is satisfied with either. The user could decide that as one of those is downloaded, it doesn't need to keep on searching for the other. Doesn't want to archive subs.

similar to old 'archive first match'

@ratoaq2 what do you think?


sickbeard/__init__.py
```diff
@@ -661,7 +662,7 @@ def initialize(consoleLogging=True):  # pylint: disable=too-many-locals, too-man
...
             POSTER_SORTBY, POSTER_SORTDIR, HISTORY_LIMIT, CREATE_MISSING_SHOW_DIRS, ADD_SHOWS_WO_DIR, \
             METADATA_WDTV, METADATA_TIVO, METADATA_MEDE8ER, IGNORE_WORDS, PREFERRED_WORDS, UNDESIRED_WORDS, TRACKERS_LIST, IGNORED_SUBS_LIST, REQUIRE_WORDS, CALENDAR_UNPROTECTED, CALENDAR_ICONS, NO_RESTART, IGNORE_UND_SUBS, \
-            USE_SUBTITLES, SUBTITLES_LANGUAGES, SUBTITLES_DIR, SUBTITLES_SERVICES_LIST, SUBTITLES_SERVICES_ENABLED, SUBTITLES_HISTORY, SUBTITLES_FINDER_FREQUENCY, SUBTITLES_MULTI, SUBTITLES_DOWNLOAD_IN_PP, SUBTITLES_KEEP_ONLY_WANTED, EMBEDDED_SUBTITLES_ALL, SUBTITLES_EXTRA_SCRIPTS, SUBTITLES_PRE_SCRIPTS, SUBTITLES_PERFECT_MATCH, subtitlesFinderScheduler, EMBEDDED_SUBTITLES_UNKNOWN_LANG, \
+            USE_SUBTITLES, SUBTITLES_LANGUAGES, SUBTITLES_DIR, SUBTITLES_SERVICES_LIST, SUBTITLES_SERVICES_ENABLED, SUBTITLES_HISTORY, SUBTITLES_FINDER_FREQUENCY, SUBTITLES_MULTI, SUBTITLES_DOWNLOAD_IN_PP, SUBTITLES_KEEP_ONLY_WANTED, EMBEDDED_SUBTITLES_ALL, SUBTITLES_EXTRA_SCRIPTS, SUBTITLES_PRE_SCRIPTS, SUBTITLES_PERFECT_MATCH, subtitlesFinderScheduler, EMBEDDED_SUBTITLES_UNKNOWN_LANG, SUBTITLES_STOP_AT_FIRST, \
```
p0psicles: Can we as part of this PR cut this long in half? Should have been done earlier.

p0psicles:
@pymedusa/developers. Is this really something we want/need? Pr itself looks okay. But I don't know where the request is coming from?

fernandog:
@p0psicles i have two languages enabled, PT-PT and PT-BR - any one of those are good to me

p0psicles:
Ok cool, then I understand

sickbeard/subtitles.py
```diff
@@ -849,6 +849,10 @@ def dhm(td):
                              ep_to_sub['show_name'], ep_num)
                 continue
 
+            if sickbeard.SUBTITLES_STOP_AT_FIRST and ep_to_sub['subtitles']:
```
p0psicles: @ratoaq2 i'd like your opinion on this solution?

fernandog: just to explain: if setting is enabled and database have subs, we should stop searching for more.

ratoaq2: It was hard to understand, but now I get it.

If you want a show in 2 or more languages, you could post pone the pp until one of them are available. Then you can post process it and watch the show. No need to find more subs.

That's the use case. Uncommon, I think... but it's still an use case.

fernandog: remember this has nothing to do with postpone

ratoaq2: Yeah, I just used it as an example.

After you have 1 sub, stop looking for more, since you're fine.

ratoaq2:
Approved

labrys:
@fernandog Make sure you add PRs to a milestone before you merge it.

PR #788, opened 2016-07-19
Split parsing

labrys:

Separate the search and parsing functions in an effort to continue standardization of providers and enhancing of search functionality.  This also subcategorizes the providers by type.


p0psicles:
@labrys whats your plan of getting this in? As this can't be reviewed.
Only thing we can do, is make sure it's tested with the providers we have.

@duramato you can maybe give this a go for a week or so? Think that's the only way.
I'll remove the needs review, ast that's pretty much impossible.

fernandog:

sickbeard/helpers.py
```diff
@@ -1413,58 +1414,42 @@ def getURL(url, post_data=None, params=None, headers=None,  # pylint:disable=too
...
 
-        if not resp.ok:
-            logger.log(u"Requested getURL %s returned status code is %s: %s"
-                       % (url, resp.status_code, http_code_description(resp.status_code)), logger.DEBUG)
-            return None
+    if not resp.ok:
+        logger.log(u'Requested url {url} returned status code {status}: {desc}'.format
+                   (url=url, status=resp.status_code, desc=http_code_description(resp.status_code)), logger.DEBUG)
 
-    except (SocketTimeout, TypeError) as e:
```
fernandog: @medariox @labrys sockets timeout its not part of requests exception

fernandog:
need to change all logs from helpers.py

sickbeard/helpers.py
```diff
@@ -1429,10 +1428,8 @@ def getURL(url, post_data=None, params=None, headers=None,  # pylint:disable=too
 
     try:
         resp.raise_for_status()
-    except requests.RequestException as e:
+    except requests.exceptions.RequestException as e:
```
fernandog: @medariox shouldn't we add both? `requests.RequestException` and `requests.exceptions` ?

medariox: I think just requests.exceptions.RequestException should be fine.

sickbeard/providers/__init__.py
```diff
@@ -107,4 +106,5 @@ def getProviderClass(provider_id):
         for x in sickbeard.providerList + sickbeard.newznabProviderList + sickbeard.torrentRssProviderList
         if x
     }
-    return providers[provider_id]
+    if providers:
+        return providers.values()[0]
```
fernandog: @medariox this change is making all providers in hstory appear with NZB image.

Also why getProviderClass used to return a list and now returns a dict?!
this change will need to be one in all over Medusa code!

medariox: Should be fixed now.
getProviderClass never returned a list, it always returned the first element of the list. So that's just the same, it's the first value of the dict. Since it makes very little sense to make a list/dict and return only the first element, I made it just directly return the first object now.

ratoaq2:
Approved

PR #789, opened 2016-07-19
Add lib apps dir, needed for lib simpleanidb

p0psicles:

Prerequisite for #501

### DIFF
--- a/lib/appdirs.py
+++ b/lib/appdirs.py
@@ -1,0 +1,346 @@
+#!/usr/bin/env python
+# Copyright (c) 2005-2010 ActiveState Software Inc.
+
+"""Utilities for determining application-specific dirs.
+
+See <http://github.com/ActiveState/appdirs> for details and usage.
+"""
+# Dev Notes:
+# - MSDN on where to store app data files:
+#   http://support.microsoft.com/default.aspx?scid=kb;en-us;310294#XSLTH3194121123120121120120
+# - Mac OS X: http://developer.apple.com/documentation/MacOSX/Conceptual/BPFileSystem/index.html
+# - XDG spec for Un*x: http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
+
+__version_info__ = (1, 2, 0)
+__version__ = '.'.join(map(str, __version_info__))
+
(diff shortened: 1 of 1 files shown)


p0psicles:
I'd like someone to make sure he has not installed the package appdirs through pip, and then checkout and try to get the anidb recommended list.

lib/appdirs/__init__.py
```diff
@@ -0,0 +1 @@
+from .appdirs import user_cache_dir
```
labrys: Normally a library like this would just be added as 'lib/appdirs.py' instead of in its own subfolder and adding an `init`.

p0psicles: Is it okay if I leave it like this? Think it's a little cleaner.

labrys: Up to you but personally I prefer to mimic site-packages.

ratoaq2: +1 to have it as a subfolder 👍

labrys: Even though it doesnt look as clean, the reason i prefer to mimic site packages is it makes updating the libraries easier (create a clean venv, update all packages and copy relevant files/folders to libs).  It makes utilization of it no different than if it was pip-installed.  It requires no maintenance/updates when the package changes.

For example, right now you have only `user_cache_dir` exposed.  If someone sees we have the lib installed they may want to leverage `user_log_dir`.  Currently you can't do that cleanly.  The same applies to attempting to catch `AppDirsError`.  For example you can do the following:
```python
from appdirs.appidrs import AppDirsError
```
...but then its implemented differently than if it was pip installed.

ratoaq2: Very good point. To manage and to upgrade packages in our lib would be straightforward

p0psicles:
Okay sure, i'll change it.

medariox:
Approved

medariox:
Why not version 1.4.0? Not compatible? https://github.com/ActiveState/appdirs/releases

p0psicles:
Think it is.  That's just not the version I got installed through pip.

PR #793, opened 2016-07-20
Fix XEM scene exception

fernandog:
http://thexem.de/map/allNames?origin=tvdb&seasonNumbers=1 

returns: {"result":"success","data":[],"message":""}

and travis:
```python
    scene_helpers_tests.py line 85 in setUp
      scene_exceptions.retrieve_exceptions()
    /home/travis/build/pymedusa/SickRage/sickbeard/scene_exceptions.py line 226 in retrieve_exceptions
      _xem_exceptions_fetcher()
    /home/travis/build/pymedusa/SickRage/sickbeard/scene_exceptions.py line 320 in _xem_exceptions_fetcher
      for indexerid, names in iteritems(parsedJSON['data']):
    /home/travis/build/pymedusa/SickRage/lib/six.py line 599 in iteritems
      return d.iteritems(**kw)
   AttributeError: 'list' object has no attribute 'iteritems'
```

### DIFF
--- a/sickbeard/scene_exceptions.py
+++ b/sickbeard/scene_exceptions.py
@@ -317,6 +317,11 @@ def _xem_exceptions_fetcher():
             if parsedJSON['result'] == 'failure':
                 continue
 
+            if not parsedJSON['data']: 
+                logger.log(u"No data returned from XEM when checking scene exceptions. Update failed for {}".format
+                            (sickbeard.indexerApi(indexer).name), logger.DEBUG)
+                continue
+
             for indexerid, names in iteritems(parsedJSON['data']):
                 try:
                     xem_exception_dict[int(indexerid)] = names
--- a/sickbeard/scene_exceptions.py
+++ b/sickbeard/scene_exceptions.py
@@ -173,7 +173,7 @@ def get_scene_exception_by_name_multiple(show_name):
                 cur_exception_name.lower(),
                 sickbeard.helpers.sanitizeSceneName(cur_exception_name).lower().replace('.', ' ')):
 
-            logger.log(u"Scene exception lookup got indexer id {}, using that".format
+            logger.log(u"Scene exception lookup got indexer id {0}, using that".format
                        (cur_indexer_id), logger.DEBUG)
 
             out.append((cur_indexer_id, int(cur_exception["season"])))
@@ -197,7 +197,7 @@ def retrieve_exceptions():  # pylint:disable=too-many-locals, too-many-branches
 
     if do_refresh:
         loc = sickbeard.indexerApi(INDEXER_TVDB).config['scene_loc']
-        logger.log(u"Checking for scene exception updates from {}".format(loc))
+        logger.log(u"Checking for scene exception updates from {0}".format(loc))
 
         try:
             jdata = helpers.getURL(loc, session=sickbeard.indexerApi(INDEXER_TVDB).session, returns='json')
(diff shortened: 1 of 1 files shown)


sickbeard/scene_exceptions.py
```diff
@@ -317,6 +317,11 @@ def _xem_exceptions_fetcher():
             if parsedJSON['result'] == 'failure':
                 continue
 
+            if not parsedJSON['data']: 
+                logger.log(u"No data returned from XEM when checking scene exceptions. Update failed for {}".format
+                            (sickbeard.indexerApi(indexer).name), logger.DEBUG)
```
p0psicles: Use {0}  or named

sickbeard/scene_exceptions.py
```diff
@@ -320,7 +320,7 @@ def _xem_exceptions_fetcher():
             if not parsedJSON['data']: 
                 logger.log(u"No data returned from XEM when checking scene exceptions. Update failed for {0}".format
                             (sickbeard.indexerApi(indexer).name), logger.DEBUG)
-                continue
+                break
```
p0psicles: I think this part is here, so we can use other indexers. If you break here, and for the first indexer no scene exceptions are returned, it won't look for others. Maybe that's something we want? But it's not in line with the existing functionality.

fernandog: I agree with you but @labrys said to break

```
[20:11:02] <Labrys> and allNames should always return a value as i expected
[20:11:22] <Labrys> so imo it should break if no value because it means xem is down
```

sickbeard/scene_exceptions.py
```diff
@@ -303,26 +303,31 @@ def _anidb_exceptions_fetcher():
 def _xem_exceptions_fetcher():
     if shouldRefresh('xem'):
         for indexer in sickbeard.indexerApi().indexers:
-            logger.log(u"Checking for XEM scene exception updates for {}".format
+            logger.log(u"Checking for XEM scene exception updates for {0}".format
                        (sickbeard.indexerApi(indexer).name))
 
             url = "http://thexem.de/map/allNames?origin={}&seasonNumbers=1".format(sickbeard.indexerApi(indexer).config['xem_origin'])
```
p0psicles: missed the {}

sickbeard/scene_exceptions.py
```diff
@@ -303,26 +303,31 @@ def _anidb_exceptions_fetcher():
...
 
             parsedJSON = helpers.getURL(url, session=xem_session, timeout=90, returns='json')
             if not parsedJSON:
-                logger.log(u"Check scene exceptions update failed for {}, Unable to get URL: {}".format
+                logger.log(u"Check scene exceptions update failed for {0}, Unable to get URL: {1}".format
```
p0psicles: It's not a big file. Can you cleanup the double quotes?

sickbeard/scene_exceptions.py
medariox: So why aren't we breaking on:
`if parsedJSON['result'] == 'failure':`
for example?

fernandog:
@medariox @p0psicles need review in the single quotes commit

sickbeard/scene_exceptions.py
p0psicles: Also: {"result":"success","data":[],"message":""}.. looks too me dat it did not get a result for that search. I would at least try the others. This does not look like an error that you get when the service is down? @labrys your thoughts?

fernandog: @p0psicles yes, the website was down when I got that response

p0psicles: But that doesn't mean, you can't get the same result, when it just doesn't have results. You have to be careful with assumptions you make.

fernandog: @p0psicles @labrys 
so what we should do?

fernandog:
rebased and updated to `continue`

fernandog:
@medariox can you review please?

PR #796, opened 2016-07-21
Add Kat to broken providers

fernandog:
@labrys Cherry pick to master?


labrys:
Approved

labrys:
I'm wondering if maybe we should move this to the github.io site as well to avoid having to push updates to master when providers get broken/fixed.

PR #798, opened 2016-07-21
Concept version of a CONTRIBUTING.md

p0psicles:
## First concept version of CONTRIBUTING.md

1. If you want to make a change, like different titles, paragraph order, etc. Then please let me know here, or in irc. And i'll change it.
2. If you want to add a chapter. For example, some more in depth Javascript guidelines, please add them yourself.
3. If you think any grammatical improvements can be made, please make them!

I'll keep this open for 1 week.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,45 @@
+# Contributing
+
+## A Guide to Contributing to Medusa
+So you think you can contribute to Medusa? We gladly have you on board! There are some pointers we'd like you give you before you start hacking into the code first. Contributions are always welcome. But if they're not using our guidelines, we can't allow that new cool feature or critical bugfix. The developers of this project have agreed on following some rules, that should improve the overall quality.
+
+## Fork-and-pull Git workflow
+In general, we follow the "fork-and-pull" Git workflow.
+1. Fork the repo on GitHub
+1. Clone the project to your own machine
+1. Push your work back up to your fork
+1. Commit changes to your own branch
+1. Submit a Pull request so that we can review your changes.
+
+**Be sure to rebase on top of the latest from "upstream" before making a pull request!**
+
+## A multilingual application
+Medusa makes use of python, javascript, css, and html (Mako templated) code. As you might understand it is important that we follow the guidelines for each of these languages. To help you get started we provide you with some basic guidelines per language.
+
+## Python Style guidelines
+We follow the PEP8 styleguide. Please visit the official https://www.python.org/dev/peps/pep-0008/ styleguide.
+
+Additionally we follow these rules:
+
+### Logging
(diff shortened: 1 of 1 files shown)


ratoaq2:
Should we add PEP257 as well? 
Talking about docstrings, I like to have the docstrings with the parameter types and return types, since they help a lot my IDE helping myself. 
What are the opinions from you all?

medariox:
My suggestion would be that we try to standardize our code as much as possible by forcing new PRs to pass specific conventions.

In my opinion, these should be mandatory:
- [pep8](https://pypi.python.org/pypi/pep8)
- [pyflakes](https://pypi.python.org/pypi/pyflakes)
- [pydocstyle](https://pypi.python.org/pypi/pydocstyle) (PEP257)
- Import-order style: [cryptography](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete.py) or [google](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_google.py)
- Max line length: 119

That's my opinion. What do you think?

ratoaq2:
Take a look in this proposal:
#836

OmgImAlexis:
I'd love to convert to the [JS style guide used at Airbnb](https://github.com/airbnb/javascript) but there's still a LOT of changes that'll need to be made before that's possible. I'll finally have internet back in the next 14 days so I'll be able to start work on this if everyone is okay with that. I'd also like to know officially what versions of browsers we support so I can start removing old js code that's mainly for outdated browsers.

One of first things that comes into mind is using `const` and `let` vs `var` which is only [supported by modern browsers](http://caniuse.com/const) this would mean we'd need to drop support for IE10 which I'm not sure if we can do. I need to have a look at the logs we've got from cdn.pymedusa.com first to see how many of the users are using anything below IE11.

fernandog:
Approved

CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+When you want to format a string and it's not for the purpose of a log message,
+please use the str.format() method. The following formats are allowed:
+```python
+'Did not find any results for show {0} on provider {1}'.format(show, provider)
+# or
+'Did not find any results for show {show} on provider {provider}'.format(show=show, provider=provider)
+```
+
+## Javascript guidelines
+We follow Google's Javascript guide: https://google.github.io/styleguide/javascriptguide.xml
```
OmgImAlexis: Is there a reason Google's was chosen over AirBNB's?

OmgImAlexis:
For the JS devs I've also been looking into https://github.com/sindresorhus/xo instead of jshint and the current style/lint guide we use.

CONTRIBUTING.md
p0psicles: No.  Just want to get starting with something.  And as where going towards Angular,  googles guidelines seemed logic. If you wanna take care of the js guidelines. Great! 

Where now on slack with for the devs channel.  Come to IRC for the invite.

PR #802, opened 2016-07-21
Replace cpasbien regex with more generic one

medariox:


ratoaq2:
This isn't used anymore with guessit_parser enabled.

ratoaq2:
And yes, there's a lot of places in the code that's still using it, but we need to get rid of it piece by piece when phasing out the legacy parser.

medariox:
@ratoaq2 
Good to know. Would it be possible to add the Cpasbien rule to guessit then? It's quite annoying and most french torrents include it (try some popular show with Extratorrent for example).

PR #819, opened 2016-07-25
Remove old detection of propers

fernandog:
@pymedusa/developers 

History.mako is also working

```
2016-07-25 12:43:36 DEBUG    SNATCHQUEUE-MANUALSNATCH-95021 :: [42fd472] Found proper tags for The.Middle.S04E08.PROPER.HDTV.x264-2HD.mp4. Snatching as PROPER
2016-07-25 12:43:36 INFO     SNATCHQUEUE-MANUALSNATCH-95021 :: [42fd472] Downloading The.Middle.S04E08.PROPER.HDTV.x264-2HD.mp4 with 1 seeders and 0 leechers from TorrentProject
2016-07-25 12:43:36 INFO     SNATCHQUEUE-MANUALSNATCH-95021 :: [42fd472] Beginning to manual snatch release: The.Middle.S04E08.PROPER.HDTV.x264-2HD.mp4
```

### DIFF
--- a/sickbeard/name_parser/parser.py
+++ b/sickbeard/name_parser/parser.py
@@ -658,10 +658,6 @@ class ParseResult(object):  # pylint: disable=too-many-instance-attributes
             return True
         return False
 
-    @property
-    def is_proper(self):
-        return re.search(r'(^|[\. _-])(proper|repack)([\. _-]|$)', self.extra_info, re.I) is not None if self.extra_info else False
-
 
 class NameParserCache(object):
     _previous_parsed = {}
--- a/gui/slick/views/history.mako
+++ b/gui/slick/views/history.mako
@@ -65,7 +65,7 @@
                     <% isoDate = datetime.strptime(str(hItem.date), History.date_format).isoformat('T') %>
                     <time datetime="${isoDate}" class="date">${airDate}</time>
                 </td>
-                <td class="tvShow"><a href="${srRoot}/home/displayShow?show=${hItem.show_id}#S${hItem.season}E${hItem.episode}">${hItem.show_name} - ${"S%02i" % int(hItem.season)}${"E%02i" % int(hItem.episode)} ${('', '<span class="quality Proper">Proper</span>')[any(x in hItem.resource.lower() for x in ['proper', 'repack', 'real', 'rerip'])]}</a></td>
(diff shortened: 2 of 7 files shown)


ratoaq2:
Should the same be applied here https://github.com/pymedusa/SickRage/blob/92d3153d1a8bf800a3185cbab7548587d9fa269b/sickbeard/search.py#L291 ?

duramato:
We should add something like there is on cache, a proper_tags that way you don't need to have checks/guessit or something else, just call guessit once and store it and its more accurate.
IMO

fernandog:
@duramato but don't you agree the when we snatch a proper status is SNATCHED_PROPER, so we don't need to call anything else.

Also history algorithm does a lof of "groupings".

sickbeard/name_parser/parser.py
```diff
@@ -660,7 +660,7 @@ def is_anime(self):
 
     @property
     def is_proper(self):
-        return re.search(r'(^|[\. _-])(proper|repack)([\. _-]|$)', self.extra_info, re.I) is not None if self.extra_info else False
+        return True if self.proper_tags else False
```
fernandog: @ratoaq2 should we do this?

fernandog:
```
    /home/travis/build/pymedusa/SickRage/sickbeard/postProcessor.py line 1111 in process
      if cur_ep.proper_tags:
   AttributeError: 'TVEpisode' object has no attribute 'proper_tags'
```

sickbeard/name_parser/parser.py
```diff
@@ -658,11 +658,6 @@ def is_anime(self):
             return True
         return False
 
-    @property
-    def is_proper(self):
-        # Part of TVEpisode object and it's inserted in 'tv_episodes' table, 'is_proper' column
```
fernandog: @medariox so from where is_proper will get its value ?

as I said in IRC:
```
    /home/travis/build/pymedusa/SickRage/sickbeard/postProcessor.py line 1111 in process
      if cur_ep.proper_tags:
   AttributeError: 'TVEpisode' object has no attribute 'proper_tags'
```

TVEpisode is created using tv_episodes table
and to populate column "is_proper" that line in parser is needed

PR #825, opened 2016-07-26
Fix adding new show

fernandog:
@medariox @Thraxis cant find out from where this come from....

### DIFF
--- a/gui/slick/js/core.js
+++ b/gui/slick/js/core.js
@@ -3353,7 +3353,6 @@ var SICKRAGE = {
                     alert('You must choose a show to continue');
                     return false;
                 }
-                generateBlackWhiteList();
                 $('#addShowForm').submit();
             });
 


Thraxis:
That was supposed to be building the black and white list for anime release groups so you could set those when adding an Anime show.

fernandog:
So confirm removal or need to add more code?

Thraxis:
All the code for it should already be there in gui/slick/js/blackwhite.js

The call is in gui/slick/views/addShows_newShow.mako

Thraxis:
I'm not seeing any problem that code in the current dev.

Below is the Black / Whitelist selector on the third step of adding a new show for "Highschool of the Dead" shortly after selecting "Is Anime" the list populates, of you have the anidb settings enabled in your config.

image

fernandog:
But if anime is disabled, it will never populate right? I still cant add

Thraxis:
Is this on the develop branch?  or another branch?  Is there an error generated?

fernandog:
Im using my branch with some PR addes like yours

referenceError: generateBlackWhiteList is not defined

Thraxis:
The function is defined in : https://github.com/pymedusa/SickRage/blob/master/gui/slick/js/blackwhite.js

The call to that JS file is in gui/slick/views/addShows_newShow.mako

https://github.com/pymedusa/SickRage/blob/19383871bc74d30648dbc44c239fd98e7830675a/gui/slick/views/addShows_newShow.mako#L10

Thraxis:
Looks like it comes from the Recommended Shows update PR (#501) will see if p0psicles will chime in there about it.

PR #827, opened 2016-07-28
Performs VACUUM on cache.db

fernandog:
My cached used to be 9M. After vacuum its now 350kb


p0psicles:
Hold off on this pr.  There are some risks using this. Like when not using explicit primary keys

fernandog:
@p0psicles we already do that in sickbeard.db

fernandog:
@p0psicles ?

PR #837, opened 2016-07-30
Moved old webapi to api/v1

OmgImAlexis:
For now I've moved the `core.py` for the webapi along with allowing `/api` and `/api/v1` to work as a single route. Eventually `/api` should be deprecated as the way it was written doesn't allow for REST routes and forces the API key to be added in the url path.


fernandog:
@OmgImAlexis concluded? in progress?

duramato:
Does this break backwards compability? i noticed you changed an api route to include v1

OmgImAlexis:
@duramato it shouldn't since it allows `/api/{key}` as well as `/api/v1/{key}`.

fernandog:
@OmgImAlexis third party apps will need to change to /api/v1/{key} or v1 is the default one?

OmgImAlexis:
Both are currently active. I'd suggest we update as much of the docs to point to `/api/v1/{key}` as the new correct path as `/api/{key}` will be removed eventually as we transition to `/api/v2`.

For now calling either URL will work exactly the same, eventually once `/api/{key}` is removed the `/api` endpoint will simply return a status if the apiKey is valid and nothing else.

OmgImAlexis:
@fernandog can this be merged?

fernandog:
@p0psicles? @medariox?  Anything against?

medariox:
Approved

OmgImAlexis:
@medariox yes, I was waiting on this being merged first though since I wanted the directory structure to be corrected before I go adding it in. I'm also working on the [RAML code](http://raml.org/) for the two APIs

PR #841, opened 2016-07-31
Rever old logic to hide special episode in season search

fernandog:
Another labrys issue. More time wasted

This it not the same!!
% if manual_search_type != 'season' and 'E00' not in hItem["name"]:


medariox:
Approved

PR #849, opened 2016-08-04
Force update when local branch is ahead

fernandog:
@pymedusa/developers 

we recently had to remove one commit from develop which caused a "Local branch is ahead of develop. Automatic update not possible" to develop users

this will affect only develop and master branches.

### DIFF
--- a/gui/slick/views/config_general.mako
+++ b/gui/slick/views/config_general.mako
@@ -10,6 +10,7 @@
     from sickbeard import metadata
     from sickbeard.metadata.generic import GenericMetadata
     from sickbeard.helpers import anon_url
+    gh_branch = sickbeard.versionCheckScheduler.action.list_remote_branches()
 %>
 <%block name="content">
 % if not header is UNDEFINED:
@@ -667,7 +668,6 @@
                                 <span class="component-title">Branch version:</span>
                                 <span class="component-desc">
                                     <select id="branchVersion" class="form-control form-control-inline input-sm pull-left">
-                                    <% gh_branch = sickbeard.versionCheckScheduler.action.list_remote_branches() %>
                                     % if gh_branch:
                                         % for cur_branch in gh_branch:
                                             % if sickbeard.GIT_USERNAME and sickbeard.GIT_PASSWORD and sickbeard.DEVELOPER == 1:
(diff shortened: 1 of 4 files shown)


sickbeard/versionChecker.py
```diff
@@ -597,8 +598,7 @@ def set_newest_text(self):
 
             newest_text = 'There is a <a href="' + url + '" onclick="window.open(this.href); return false;">newer version available</a> '
             newest_text += " (you're " + str(self._num_commits_behind) + " commit"
-            if self._num_commits_behind > 1:
-                newest_text += 's'
```
fernandog: we already check if `self._num_commits_behind` is > so this will always be true

medariox: This adds an 's' to 'commit' if we are more than 1 commit behind.

fernandog: ahh I see now

ratoaq2:
Is there a reason to have it only for master and develop?

fernandog:
@ratoaq2 if someone is using a testing branch or working on another branch and haven't pushed yet, he wil lose the commits if this force updated it not only for develop|master

ratoaq2:
Valid scenario. But there's another valid one:

- 1 'production' installation running a specific branch other than master or develop (for testing purposes, most of the time)
- 1 local dev environment

The local dev environment should never reset and force update, but the 'production' installation should.

sickbeard/versionChecker.py
```diff
@@ -625,6 +626,12 @@ def need_update(self):
             if self._num_commits_behind > 0:
                 return True
 
+            # Force update if GIT_RESET enabled and commits ahead
+            if self._num_commits_ahead > 0 and sickbeard.GIT_RESET \
+               and sickbeard.BRANCH in ['develop', 'master']:
```
medariox: I'd keep this as one line, since it's < 120 chars.

fernandog:
@ratoaq2 how to implement your scenario then?

gui/slick/views/config_general.mako
```diff
@@ -730,7 +730,7 @@
                                 </span>
                             </label>
                         </div>
-                        <div class="field-pair" hidden>
+                        <div class="field-pair">
                             <label for="git_reset">
```
fernandog: git_reset option was hidden

fernandog:
@ratoaq2 can you test please?

sickbeard/versionChecker.py
```diff
@@ -583,11 +583,7 @@ def set_newest_text(self):
         # if we're up to date then don't set this
         sickbeard.NEWEST_VERSION_STRING = None
 
-        if self._num_commits_ahead:
-            logger.log(u"Local branch is ahead of " + self.branch + ". Automatic update not possible.", logger.WARNING)
-            newest_text = "Local branch is ahead of " + self.branch + ". Automatic update not possible."
-
-        elif self._num_commits_behind > 0:
+        if self._num_commits_behind > 0 or (sickbeard.GIT_RESET and sickbeard.BRANCH in sickbeard.GIT_RESET_BRANCHES):
```
ratoaq2: If there's no branches in the list, it should accept as well

```
or (sickbeard.GIT_RESET and (not sickbeard.GIT_RESET_BRANCHES or sickbeard.BRANCH in sickbeard.GIT_RESET_BRANCHES)) 
```

The same logic some lines below (probably a function would fit: def should_reset)

fernandog: Did you try the Ui? Use must select branches in a pre-populated  select

ratoaq2:
Approved

PR #864, opened 2016-08-08
Fix pymedusa/SickRage/issues/856

fernandog:
Fix https://github.com/pymedusa/SickRage/issues/856

@medariox are you sure?


medariox:
Should be good now.

medariox:
Approved

PR #866, opened 2016-08-09
Improving tests and adding more subtitles tests

ratoaq2:
- [x] Adding more subtitles test
- [x] Refactoring new pytests to be more readable (using fixtures)
- [x] Fix for a subtle subtitle bug: When the downloaded subtitle has no content it should not be computed as downloaded.

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -249,66 +249,27 @@ def download_subtitles(tv_episode, video_path=None, subtitles=True, embedded_sub
     episode_name = tv_episode.name
     show_indexerid = tv_episode.show.indexerid
     status = tv_episode.status
-
-    subtitles_dir = get_subtitles_dir(video_path)
-    found_subtitles = download_best_subs(tv_episode, video_path, subtitles_dir,
-                                         subtitles=subtitles, embedded_subtitles=embedded_subtitles)
-
-    for subtitle in found_subtitles:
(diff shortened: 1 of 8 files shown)


sickbeard/subtitles.py
```diff
@@ -249,66 +249,27 @@ def download_subtitles(tv_episode, video_path=None, subtitles=True, embedded_sub
...
-                                   episode=episode, episode_name=episode_name, show_indexerid=show_indexerid)
-        if sickbeard.SUBTITLES_HISTORY:
-            logger.debug(u'history.logSubtitle %s, %s', subtitle.provider_name, subtitle.language.opensubtitles)
-            history.logSubtitle(show_indexerid, season, episode, status, subtitle)
-
-    return sorted({subtitle.language.opensubtitles for subtitle in found_subtitles})
-
-
-def download_best_subs(tv_episode, video_path, subtitles_dir, subtitles=True, embedded_subtitles=True):
-    """Download the best subtitle for the given video.
```
fernandog: @ratoaq2 is this ok remove all this?

ratoaq2:
We used to have two methods,  download_subtitles and  download_best_subs. After the minor refactoring for the tvepisode refiner, one method was just calling the other + this loop. Both were looping through the subs, so I  just moved this to one loop. I  didn't remove any code logic,  just merged in one loop. And now I'm also using the subs returned by save_subtitles, that's more accurate and it will make sure that the subtitles are saved

medariox:
Approved

PR #880, opened 2016-08-12
Get broken providers from sickrage.io and force disable provider if is broken

fernandog:
@medariox @ratoaq2 @p0psicles 

I'm done with features until next release

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -1570,5 +1570,4 @@ def get_broken_providers():
     if not response:
         logger.debug('Unable to update broken providers')
         return []
-
-    return list(response)
+    return response
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -1561,3 +1561,18 @@ def canonical_name(obj, fmt='{key}:{value}', separator='|', ignore_list=frozense
     """
     guess = obj if isinstance(obj, dict) else guessit.guessit(obj)
     return str(separator.join([fmt.format(key=k, value=v) for k, v in guess.items() if k not in ignore_list]))
+
+
+def get_broken_providers():
+    """Get broken providers from cdn.pymedusa.com."""
+    url = 'https://cdn.pymedusa.com/providers/broken_providers.txt'
+    response = getURL(url, session=make_session(), returns='response')
+
+    if not response or not response.text:
+        logger.debug('Unable to update broken providers')
+        return []
+
+    broken_providers = []
+    for line in response.text.splitlines():
+        broken_providers.append(line.strip())
+    return broken_providers
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -1565,14 +1565,10 @@ def canonical_name(obj, fmt='{key}:{value}', separator='|', ignore_list=frozense
 
 def get_broken_providers():
     """Get broken providers from cdn.pymedusa.com."""
-    url = 'https://cdn.pymedusa.com/providers/broken_providers.txt'
-    response = getURL(url, session=make_session(), returns='response')
-
-    if not response or not response.text:
+    url = 'https://cdn.pymedusa.com/providers/broken_providers.json'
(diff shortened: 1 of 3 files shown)


fernandog:
@medariox @ratoaq2 @p0psicles 
Its saving as 

`broken_providers = "set(['kickasstorrents', 'torrentz', 'btdigg'])"`

anyone sees another way to do this?

fernandog:
@medariox can you review this please?

sickbeard/helpers.py
```diff
@@ -1561,3 +1561,18 @@ def canonical_name(obj, fmt='{key}:{value}', separator='|', ignore_list=frozense
     """
     guess = obj if isinstance(obj, dict) else guessit.guessit(obj)
     return str(separator.join([fmt.format(key=k, value=v) for k, v in guess.items() if k not in ignore_list]))
+
+
+def get_broken_providers():
+    """Get broken providers from cdn.pymedusa.com."""
+    url = 'https://cdn.pymedusa.com/providers/broken_providers.txt'
```
p0psicles: I would really recommend to use json here.

fernandog: I tried but as we store as list I prefered this way.

p0psicles: That's what I'm saying.  Switch it to store it as json.  We also do the same for the scene exception. Will be less error prone and more future proof.

fernandog: We dont store anything in config as dict

fernandog: @p0psicles i stored in the config so if some how user can't connect to cdn, at least he has the "last" broken providers

If you see a better way to do this, feel free to change it.

sickbeard/helpers.py
```diff
@@ -1565,14 +1565,10 @@ def canonical_name(obj, fmt='{key}:{value}', separator='|', ignore_list=frozense
...
+    response = getURL(url, session=make_session(), returns='json')
+    if not response:
         logger.debug('Unable to update broken providers')
         return []
 
-    broken_providers = []
-    for line in response.text.splitlines():
-        broken_providers.append(line.strip())
-    return broken_providers
+    return list(response)
```
fernandog: @ratoaq2 @p0psicles can't make it work.

17:07:57 DEBUG::EVENT-QUEUE :: [f3ebd83] response: {}
17:07:57 DEBUG::EVENT-QUEUE :: [f3ebd83] Unable to update broken providers

sickbeard/helpers.py
fernandog: @p0psicles done. its json now

sickbeard/helpers.py
```diff
@@ -1561,3 +1561,14 @@ def canonical_name(obj, fmt='{key}:{value}', separator='|', ignore_list=frozense
...
+
+def get_broken_providers():
+    """Get broken providers from cdn.pymedusa.com."""
+    url = 'https://cdn.pymedusa.com/providers/broken_providers.json'
+    response = getURL(url, session=make_session(), returns='json')
+    if not response:
+        logger.debug('Unable to update broken providers')
+        return []
+
+    return list(response)
```
ratoaq2: Quick question: isn't the response already a list here?

fernandog: no, getUrl is returning 'json'

fernandog: @ratoaq2 you are right. sorry: `DEBUG::EVENT-QUEUE :: [f3ebd83] response type: <type 'list'>
`

sickbeard/helpers.py
OmgImAlexis: @fernandog is there a need for that since if they can't connect to the cdn they probably don't have internet so they can't connect to the providers anyway?

fernandog: i already got some errors connection to cdn and my internet was ok. 
its just to have a "offline" broken providers

ratoaq2:
Approved

PR #885, opened 2016-08-13
Avoid list remote branches every /config/general page load

fernandog:
From Load time: 2.4544 to Load time: 0.0308s

list remote branches if update is disabled.
if update hasn't run yet (global var is empty), then mako will call list_remote_branches. Next load it will get the value from global var

@OmgImAlexis 

@ratoaq2 @medariox 
image


OmgImAlexis:
@fernandog can we merge this? Looks fine to me.

fernandog:
@OmgImAlexis we still need to add the "force update" button right?

@medariox @ratoaq2 any of you can review my commits please?

OmgImAlexis:
In `core.js` just after the click event handler for `#branchCheckout'` add something like this.

```
$('#branchForceUpdate').on('click', function() {
                $.getJSON(srRoot + '/home/branchForceUpdate', function(data) {
                    $('#branchVersion').empty();
                    data.branches.forEach(function(branch) {
                        append('<option value="' + branch + '" ' + (data.currentBranch === branch ? 'selected="slected"' : '') + '>' + branch + '</option>');
                    });
                });
            });
```

`config_general.mako` add this on LN674 `<input class="btn btn-inline" style="margin-left: 6px;" type="button" id="branchForceUpdate" value="Update Branch List">`.

Then add the Python into the handler file.
This is just an example, no clue how you need to actually write it but basically we do the update and then return the branches.

```
    def branchForceUpdate(self):
        SourceUpdateManager.list_remote_branches()
        return {
            'currentBranch': branch,
            'branches': ['master', 'develop']
        }
```


Also the checks need to be added to this python similar to how is in the mako for the normal select.
```
% for cur_branch in gh_branch:
                                            % if sickbeard.GIT_USERNAME and sickbeard.GIT_PASSWORD and sickbeard.DEVELOPER == 1:
                                                <option value="${cur_branch}" ${'selected="selected"' if sickbeard.BRANCH == cur_branch else ''}>${cur_branch}</option>
                                            % elif sickbeard.GIT_USERNAME and sickbeard.GIT_PASSWORD and cur_branch in ['master', 'develop']:
                                                <option value="${cur_branch}" ${'selected="selected"' if sickbeard.BRANCH == cur_branch else ''}>${cur_branch}</option>
                                            % elif cur_branch == 'master':
                                                <option value="${cur_branch}" ${'selected="selected"' if sickbeard.BRANCH == cur_branch else ''}>${cur_branch}</option>
                                            % endif
                                        % endfor
```

fernandog:
@OmgImAlexis i prefer you do it, can you?

The python command is: `sickbeard.versionCheckScheduler.action.list_remote_branches()`

sickbeard/server/web/home/handler.py
```diff
@@ -654,6 +654,12 @@ def branchCheckout(self, branch):
             ui.notifications.message('Already on branch: ', branch)
             return self.redirect('/{page}/'.format(page=sickbeard.DEFAULT_PAGE))
 
+    def branchForceUpdate(self):
+        return {
+            'currentBranch': sickbeard.BRANCH,
+            'branches': sickbeard.versionCheckScheduler.action.list_remote_branches()
+        }
+
```
fernandog: @OmgImAlexis like this?

fernandog:
@OmgImAlexis need to align better the button. can you do it ?

fernandog:
@OmgImAlexis it only adds to the select when page is full refreshed?

OmgImAlexis:
@fernandog it should only add it it to the list after the button is pressed as it'll hit the backend and get the new branches.

fernandog:
@OmgImAlexis ok i got that but it will update the "select" without need to full refresh page?

OmgImAlexis:
Yes it will, that's the point of the js you added. It gets the new branches and then replaces the whole select with the new ones.

gui/slick/js/core.js
```diff
@@ -298,6 +298,20 @@ var SICKRAGE = {
...
+                $.getJSON(srRoot + '/home/branchForceUpdate', function(data) {
+                    data.resetBranches.forEach(function(branch) {
+                        $('#git_reset_branches').append('<option value="' + branch + '" selected="selected" >' + branch + '</option>');
+                    });
+                    data.branches.forEach(function(branch) {
+                        $('#git_reset_branches').append('<option value="' + branch + '" >' + branch + '</option>');
+                    });
+                    $(this).prop('disabled', false);
+                });
+            });
```
fernandog: @OmgImAlexis I manage to make JS to put resetBranches on top and not reset branches in the botton

gui/slick/js/core.js
```diff
@@ -298,6 +298,20 @@ var SICKRAGE = {
...
+                $(this).attr('disabled', 'disabled');
+                $('#git_reset_branches').empty();
+                $.getJSON(srRoot + '/home/branchForceUpdate', function(data) {
+                    data.resetBranches.forEach(function(branch) {
+                        $('#git_reset_branches').append('<option value="' + branch + '" selected="selected" >' + branch + '</option>');
+                    });
+                    data.branches.forEach(function(branch) {
+                        $('#git_reset_branches').append('<option value="' + branch + '" >' + branch + '</option>');
+                    });
+                    $(this).prop('disabled', false);
```
fernandog: @OmgImAlexis its not enabling it again.

Issue #810, opened 2016-07-24
General guessit induced/related issues


fernandog:
TODO:
- [x] [Wrong season](https://github.com/pymedusa/Medusa/issues/810#issuecomment-254490731)
- [x] [Wrong release group](https://github.com/pymedusa/Medusa/issues/810#issuecomment-255558111)
- [x] [Wrong ep title and air date](https://github.com/pymedusa/Medusa/issues/810#issuecomment-255858187)
- [x] [Wrong release group](https://github.com/pymedusa/Medusa/issues/810#issuecomment-256492077)
- [x] [Wrong release group](https://github.com/pymedusa/Medusa/issues/810#issuecomment-257167849)
- [x] [Scene release numbering not detected ](https://github.com/pymedusa/Medusa/issues/810#issuecomment-258586604)
- [ ] https://github.com/pymedusa/Medusa/issues/810#issuecomment-255877018
- [x] [Wrong title](https://github.com/pymedusa/Medusa/issues/1493#issuecomment-258853023) 
- [ ] [Wrong release group](https://github.com/pymedusa/Medusa/issues/810#issuecomment-259827803)
- [x] [Wrong release group](https://github.com/pymedusa/Medusa/issues/810#issuecomment-261100944)
- [ ] [Wrong release group](https://github.com/pymedusa/Medusa/issues/810#issuecomment-261377027)
- [ ] [Wrong numbering](https://github.com/pymedusa/Medusa/issues/810#issuecomment-261583833)
- [x] [Wrong numbering](https://github.com/pymedusa/Medusa/issues/810#issuecomment-261721029)
- [ ] [Wrong Title](https://github.com/pymedusa/Medusa/issues/810#issuecomment-261752356) 
- [ ] [Wrong Title](https://github.com/pymedusa/Medusa/issues/810#issuecomment-263158795)
- [ ] [Wrong Title](https://github.com/pymedusa/Medusa/issues/810#issuecomment-264337236)
- [ ] [Wrong release group](https://github.com/pymedusa/Medusa/issues/810#issuecomment-264665613)
- [ ] [Wrong release group](https://github.com/pymedusa/Medusa/issues/810#issuecomment-264722778)
- [ ] [Wrong episode_details](https://github.com/pymedusa/Medusa/issues/810#issuecomment-266252346)

medariox:
Release group is just BRISK, without [PRiME].

ratoaq2:
What's the full name? Or it's just that?

[...]

PR #886, opened 2016-08-13
Scene releases: release groups should drop all words after dash

ratoaq2:

This is a small enhancement for some cases in  #810 
- Removing words in a release group after a dash. e.g.: GROUP-BUYMORE will become GROUP

The scenario above works fine. We need to do regression test on other scenarios to make sure there's no side effect.


fernandog:
@duramato

fernandog:
Approved

Issue #872, opened 2016-08-11
[ FEATURE ] keep looking for better quality after manual snatch


NicoLeOca:
Branch/Commit: last develop
OS: synology last DSM
What you did: 
quality settings of a show set to HD1080p
I forced a manual search of the episode s06E04, snatched a release with 720p HDTV quality.
it seems that Medusa considers this as allowed (as it appears in yellow)
![suits main](https://cloud.githubusercontent.com/assets/9487685/17583288/789772c6-5fb0-11e6-8a65-4e32bb3dc4ce.JPG)
a few days later, it has NOT been replaced by another release matching the quality setup. (HD1080p)
another manual search shows that releases are present, matching the quality settings.
![suits manual search](https://cloud.githubusercontent.com/assets/9487685/17583342/cd9805c4-5fb0-11e6-83da-b6e8cbc704c7.JPG)
What you expected:
I thought that Medusa would have replaced my manually snatched episode

is it an expected behaviour?

NicoLeOca:
@fernandog 

![backlog suits](https://cloud.githubusercontent.com/assets/9487685/17635018/97996028-60d5-11e6-9646-fae6aead5e58.JPG)

fernandog:
force backlog and send full logs of backlog thread

NicoLeOca:
full logs, not filtered but last pieces are only about the backlog search
https://gist.github.com/NicoLeOca/332f27a3a7b5d66afc709840a4f7da56

One wierd message is:
Episode already exists with quality 720p HDTV and the found result has same/lower quality, ignoring found result for Suits S06E01 with quality 1080p WEB-DL

NicoLeOca:
@fernandog 
1080p was just snatched automatically for the last episode of Suits

The "undesired behaviour is that manually snatching a lower quality makes Medusa consider this snatch as of allowed quality and doesn't replace then

fernandog:
@NicoLeOca did you changed anything in quality settings? still on develop branch?

where did you got that sentence?

NicoLeOca:
@fernandog 
still on dev yes
no I didn't change anything

that sentence is mine, it's my personal opinion

fernandog:
@NicoLeOca by design any manual search episode turns off the auto search feature.

if you want to keep searching for better quality, just set it to wanted after you manuall searched it

NicoLeOca:
@fernandog 
ok thanks for the advice

PR #887, opened 2016-08-13
Fix not download new quality after quality change

fernandog:
as discussed in slack

Wanted quality is HD720P . and i have a SDTV when my show had SDTV wanted quality

From logs:
`Episode already exists with quality SDTV and the found result has same/lower quality, ignoring found result for Ripper Street S04E03 with quality 720p WEB-DL`

medusa ignored 720p result

image


tests/test_tv.py
```diff
@@ -0,0 +1,120 @@
+# coding=utf-8
+"""Tests for sickbeard.tv.py."""
+import pytest
+
+from sickbeard.common import DOWNLOADED, Quality, SNATCHED, SNATCHED_PROPER, WANTED
+from sickbeard.tv import TVEpisode
+
+
+@pytest.mark.parametrize('p', [
+    {  # p0: Downloaded 720p HDTV and found 720p BluRay: yes
```
fernandog: here it should be "no". only allowed = snatch any and stop

tests/test_tv.py
```diff
@@ -0,0 +1,120 @@
...
+    },
+    {  # p2: Downloaded 720p BluRay and found 720p HDTV, not a better quality: no
+        'ep_status': DOWNLOADED,
+        'cur_quality': Quality.HDBLURAY,
+        'new_quality': Quality.HDTV,
+        'allowed_qualities': [Quality.HDTV, Quality.HDWEBDL, Quality.HDBLURAY],
+        'preferred_qualities': [],
+        'expected': False
+    },
+    {  # p3: Snatched 720p HDTV and found 720p BluRay: yes
```
fernandog: also no. only allowed = any and stop

fernandog:
@ratoaq2 @markheloking 

If you have qualities in the Preferred list set, any episode matching a quality in this list will not be replaced with a better quality. AKA it is a "best" match and is final.

If you ONLY have qualities in the Allowed list (Or have a preset selected other than custom) it will also stop on the first match in this list and be considered a "best" match, and never be replaced with a better quality.

If you have qualities in BOTH lists, it will accept any quality in the Allowed list, and continue looking for a better quality in the Allowed list, until it finds one in the Preferred list.

fernandog:
Need to see if this needed to be changed also:
https://github.com/pymedusa/SickRage/blob/develop/sickbeard/search.py#L313

fernandog:
This one is reladed to backlog overview

All logic should be in one place only right?
https://github.com/pymedusa/SickRage/blob/develop/sickbeard/tv.py#L1574

fernandog:
@medariox milestone 0.1.5 right? we need make sure this works fine (keep some days in develop)

gui/slick/views/inc_qualityChooser.mako
```diff
@@ -38,5 +29,14 @@ selected = None
...
+        <div style="text-align: left; float: left;">
+            <h5>Allowed</h5>
+            <% anyQualityList = filter(lambda x: x > Quality.NONE, Quality.qualityStrings) %>
+            <select id="anyQualities" name="anyQualities" multiple="multiple" size="${len(anyQualityList)}" class="form-control form-control-inline input-sm">
+            % for curQuality in sorted(anyQualityList):
+                <option value="${curQuality}" ${'selected="selected"' if curQuality in anyQualities else ''}>${Quality.qualityStrings[curQuality]}</option>
+            % endfor
+            </select>
+        </div>
     </div>
```
fernandog: @medariox why not keep original "order" in UI to not confuse users?

fernandog:
My review about tests is missing? what happened?

need to add tests with status ARCHIVED  and status UNKNOWN

fernandog:
@ratoaq2 @medariox there is too place that still does some quality checks:

https://github.com/pymedusa/Medusa/blob/develop/medusa/search/core.py#L323
https://github.com/pymedusa/Medusa/blob/develop/medusa/tv.py#L1692-L1730


all logic should be in one place right?

tests/test_tv.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@pytest.mark.parametrize('p', [
+    {  # p0: Downloaded 720p HDTV and found 720p BluRay: yes
+        'ep_status': DOWNLOADED,
+        'cur_quality': Quality.HDTV,
+        'new_quality': Quality.HDBLURAY,
+        'allowed_qualities': [Quality.HDTV, Quality.HDWEBDL, Quality.HDBLURAY],
+        'preferred_qualities': [],
+        'expected': True
```
fernandog: this is 'no' right?

when user sets only allowed_qualities medusa downloads only the first available. it doesn't re-download.

NicoLeOca:
@fernandog 
not sure it is ok for me to remind it here but:
can you please have a look to the fact that a manual snatch is directly considered as being of the prefered quality no matter what the exact quality is?
https://github.com/pymedusa/Medusa/issues/872

the interest clearly depends on the ETA of this:
https://github.com/pymedusa/Medusa/issues/1266
and if you really review it completely as @OmgImAlexis and I suggested.

medariox:
_**Here is how it works**_

If you **only** have qualities in the Preferred list set (or have a preset selected other than custom), any episode matching a quality in this list will not be replaced with a better quality. AKA it is a "best" match and is final.

If you **only** have qualities in the Allowed list it will accept any quality in the Allowed list, and continue looking for a better quality in the Allowed list until the highest quality is reached.

If you have qualities in **both** lists, it will accept any quality in the Allowed or Preferred list, and continue looking for a better quality in the Allowed list, until it finds one in the Preferred list (consistent behavior).

fernandog:
@medariox today this "If you only have qualities in the Preferred list set" is in allowed. are we changing it? users will need to change their quality settings?

NicoLeOca:
@medariox 
I didn't know that
that's a very good news then

Cheers

medariox:
@fernandog 
Users that have **_custom preset_ with _only_ allowed qualities set** will eventually need to change that to preferred qualities, in order to keep the same behavior.

medariox:
As we can't agree on how to handle this, the current functionality remains unchanged until we decide what to do.

ratoaq2:
Just need to invert the logic in the method and its test

fernandog:
superseeded

NicoLeOca:
@medariox @ratoaq2 @fernandog 
Guys,
can you please add the explanations of how to handle quality settings somewhere easy to find for everyone?
like next to the settings
image

cause I use Medusa/SR for some time now and I just learned that if there are several prefered qualities setup, they will be downloaded successively up to the best.

PR #891, opened 2016-08-14
Added base_url to PageTemplate class.

p0psicles:

Okay, this looks like a much cleaner solution. Still need to do allot of tests, with different setups before we merge it.

I'd like to give a final Approval on this myself after some reviews/testing pls.

@OmgImAlexis you can ofcourse merge this into your topic branch if you like start playing with it.

Ok, now shoot.

### DIFF
--- a/sickbeard/server/web/core/base.py
+++ b/sickbeard/server/web/core/base.py
@@ -91,6 +91,10 @@ class PageTemplate(MakoTemplate):
             'newsBadge': '',
             'toolsBadge': '',
             'toolsBadgeClass': '',
+            'base_url': urljoin(rh.request.protocol + '://' + rh.request.host,
+                                sickbeard.WEB_ROOT) +
+                          ('', '/')
+                          [urljoin(rh.request.protocol + '://' + rh.request.host, sickbeard.WEB_ROOT)[-1] not in '/'],
         }
 
         if rh.request.headers['Host'][0] == '[':


sickbeard/server/web/core/base.py
```diff
@@ -91,6 +91,10 @@ def __init__(self, rh, filename):
             'newsBadge': '',
             'toolsBadge': '',
             'toolsBadgeClass': '',
+            'base_url': urljoin(rh.request.protocol + '://' + rh.request.host,
+                                sickbeard.WEB_ROOT) +
+                          ('', '/')
+                          [urljoin(rh.request.protocol + '://' + rh.request.host, sickbeard.WEB_ROOT)[-1] not in '/'],
```
medariox: I'm not sure why you do this. WEB_ROOT can't have an ending slash, so why don't we just always add it?
For example:
```urljoin(rh.request.protocol + '://' + rh.request.host, sickbeard.WEB_ROOT + '/')```

p0psicles: Yeah sure. If I change that, we still need to test this some more. But, then are we confident that we can use this?

medariox: Why would web root change? 
The issue is that when we don't have Web_root set, we add an ending /. 
So we either have / or /webroot
In case we don't have Web_root set, it would cause base_url to have a double /.

ratoaq2:
This is already in develop with xo changes

sickbeard/server/web/core/base.py
OmgImAlexis: @p0psicles @medariox this is already in develop with my changes and is working fine. If you want to change it I'd suggest opening a new PR.

PR #893, opened 2016-08-14
Only show location, filaname, size, download link and subtitles if

fernandog:
status is DOWNLOADED or ARCHIVED

If previous status is DOWNLOADED and we snatch another release, we shouldn't show those values from old status


medariox:
Approved

PR #908, opened 2016-08-17
Fix pymedusa/SickRage/issues/902

fernandog:
Fixes https://github.com/pymedusa/SickRage/issues/902

@p0psicles @duramato


p0psicles:
Nope wrong @fernandog. This isn't how it's the addShowById is working.

PR #914, opened 2016-08-18
Show watched (and removed from medusa) recommended show as....watched!

fernandog:
@p0psicles @medariox @ratoaq2 

image

### DIFF
--- a/gui/slick/js/core.js
+++ b/gui/slick/js/core.js
@@ -3707,7 +3707,7 @@ var SICKRAGE = {
 
             $('#traktlistselection').on('change', function(e) {
                 var traktList = e.target.value;
-                window.history.replaceState({}, document.title, '?traktList=' + traktList);
+                window.history.replaceState({}, document.title, 'addShows/trendingShows/?traktList=' + traktList);
                 $('#trendingShows').loadRemoteShows(
                     'addShows/getTrendingShows/?traktList=' + traktList,
                     'Loading trending shows...',
--- a/gui/slick/views/addShows_recommended.mako
+++ b/gui/slick/views/addShows_recommended.mako
@@ -133,7 +133,7 @@
                                 data-indexer-id="${cur_result.indexer_id}" data-show-name="${cur_result.title | u}"
                                 data-add-show>Add</a>
                             % endif
-                            % if trakt_b and (not cur_result.show_in_list or cur_result.indexer_id in removed_from_medusa):
+                            % if trakt_b and not (cur_result.show_in_list or cur_result.indexer_id in removed_from_medusa):
                                 <button data-indexer-id="${cur_result.indexer_id}" class="btn btn-xs" data-blacklist-show>Blacklist</a>
                             % endif
                         </div>
--- a/gui/slick/views/addShows_recommended.mako
+++ b/gui/slick/views/addShows_recommended.mako
@@ -81,6 +81,12 @@
             <% trakt_b = context.get('trakt_blacklist') %>
         % endif
 
+        % if not context.get('removed_from_medusa'):
+            <% removed_from_medusa = [] %>
+        % else:
+            <% removed_from_medusa = context.get('removed_from_medusa') %>
+        % endif
+ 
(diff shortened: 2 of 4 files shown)


p0psicles:
We already decided to display shows that are already in library.  But then grayed out. Why would we want to change that now?

OmgImAlexis:
@p0psicles so that means instead of just showing new shows we show stuff that the user already has added? That's kinda pointless.

p0psicles:
That's your opinion.  I've had that Pr open for 5 months.  Nobody thought it was pointless during that time.

fernandog:
@p0psicles please read the PR again! My english is that bad?

It to not show that user REMODED from medusa after watched show

I already watched Dexter and removed from Medusa. Why show it?!

fernandog:
@medariox is that clear for you the propose of the PR?

p0psicles:
I don't want to add specific logic for one recommended list provider.  That trakt has the blacklists already bugs me.  Cause now I had to implement trakt specific functionality. If you want to change functionality, sure.  But then it has to meet the following criteria: 1. Work the same for all recommended lists.  2. Everyone should agree on it. 3. Provide the user a way of displaying what's hidden from them. Those where some of the principles I've tried to follow while refactoring the lists.

fernandog:
Imdb doesnt have user custom lists. We dont do watch list tracking in imdb (like in trakt)

So only trakt would have this feature.

fernandog:
Also the blacklist was a existing feature you removed

p0psicles:
Yeah, and the reason why I forgot it, was because it did not fit in the unified UI for recommended lists. So I removed it back in jan, planning to deal with it later.

Also your criteria for not showing a show in the list are to vague. your using the /watched. When are shows added to this trakt list?

You are using a combination of criteria. It will not be clear to the user, why it's missing certain shows. 

A user can start on a show, watch a few eps, then delete is (for whatever reason). While going through the trakt list a few years he could stumble on the show and add it if he wants to. Just to name one user story.

If would be okay with it, if your using one trakt list, which is used to keep track of removed shows. Ow wait, we already have that, it's the blacklist. You could even improve on this functionality, by adding the blacklist as one of the recommended lists. That way, a user can also see, what he has blacklisted, from within medusa.

fernandog:
Im using watched (watched one and gave up or watched all) + collected (collected one or all)

That is easy to do. Just read blacklist name from config.

p0psicles:
Everything is easy to do. It's just hard to understand. Me as a developer, am having a hard time. What do you think of our average user?

fernandog:
Easy to understand what?

User see a show he already watched in the recommendation show and wonder why its showing a show he already watched and already removed from medusa.

@omgimalexis @medariox @ratoaq2 @supergonkas please give your thought as pops want we all to agree

@wimprybx you too.

duramato:
* @duramato 
I don't understand exacly what this does, does it remove the show completly from the recommend list? if so i would rather it be like it is on imdb , greyed out and with a checkmark

fernandog:
@supergonkas it hides (list comes from trakt). To appear as grey out show must be in medusa. And thats the point of  this PR. The show its not in medusa anymore

fernandog:
@p0psicles in develop it is removing from recommended list the collected shows. did you noticed that? If we shouldn't remove anything, we need to remove that code.

p0psicles:
Yeah good point.

OmgImAlexis:
Personally if we're still going for the polished look we should just grey it out instead of dropping that tick over it. It doesn't fit at all with the rest of the UI and greying it out should be enough indication to the user. Also maybe add a UI option like what trakt does where it either removes the show or just greys it out.

For me personally and I know this would be the case for other hoarders; I like to go through that page every so often to add new shows that're coming out. If I already have a show added I shouldn't see it there, neither should I see it if I've deleted it in the past as that means I didn't want it. If it's added back there but just greyed out that means instead of just seeing the new shows I want I end up seeing loads of others I've already seen and had no interest in.

PR #945, opened 2016-08-21
Refactoring issue submission and log files reader

ratoaq2:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read [contribution guide](https://github.com/PyMedusa/SickRage/blob/master/.github/CONTRIBUTING.md)
- [x] Refactored issue submission logic to a separate module
- [x] Log reading improvement to be efficient and to not be memory hungry.
- [x] Better gist logs
- [x] Flake8 fixes on the modified files
- [x] Log reader tests
- [x] issue submitter tests
- [x] Improved similar issue submission to be more generic: if title is 90% similar to an existing issue, it will comment the issue instead of creating a new one. Also restricting to the last 90 days to avoid full issue scan on github.
- [x] Added Period filter to log view

PR #962, opened 2016-08-23
Fix Third party apps only showing API and TORNADO logs after some calls

fernandog:
Example: 40 lines of FINDSUBTITLES and 10 of API/TORNADO
Next 'log' refresh it will show 30 and 20
Until reaches 50 of API/TORNADO
so user will only see API/TORNADO logs no matter what

@MGaetan89 (ShowsRage app)

@ratoaq2


fernandog:
@ratoaq2 if you want to add this fix to your PR feel free. Just close this

ratoaq2:
Thanks.
This is covered in #945

PR #972, opened 2016-08-24
Fixes pymedusa/SickRage/issues/874

fernandog:
@ratoaq2 
it will always download subs in PP folder when "postpone if no subs" enabled

User had setting "subtitles_download_in_pp" enabled and 'postpone' disabled.
No reason to have this setup.


fernandog:
he said this PR fixed his issue. No more redownloading

PR #974, opened 2016-08-24
Fix another tab issue

fernandog:
@OmgImAlexis agree?

Related to https://github.com/pymedusa/SickRage/commit/b64c81db48ba7f8f1e8a1382057a1763d3a0d501

I found a bug in backRestor.mako so I saw that all mako files were missing the "hack" we did here:
https://github.com/pymedusa/SickRage/pull/969/commits/db1ba6ff6f4a449d7da3bce3067fb4fa02bb6cba

### DIFF
--- a/gui/slick/views/config_anime.mako
+++ b/gui/slick/views/config_anime.mako
@@ -11,8 +11,9 @@
             <form id="configForm" action="config/anime/saveAnime" method="post">
                 <div id="config-components">
                     <ul>
-                        <li><a href="${base_url}config/anime/#animedb-settings">AnimeDB Settings</a></li>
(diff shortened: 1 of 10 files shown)


OmgImAlexis:
Yeah we're just going to have to add this back until I can work a way around the jQuery UI tabs. I was thinking we might just end up forking it and removing all the ajax crap since there isn't really anything new added between the old and new version.

Approved!

[Approved with PullApprove](https://pullapprove.com/pymedusa/SickRage/pull-request/974/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=pymedusa/SickRage)

PR #980, opened 2016-08-25
Also trim cache in manual search providers

fernandog:
Its was cleaning only daily/backlog search providers

This is when user has only manual search providers enable as fall back providers
Like provider A and B as daily and backlog search only
and
Provider C, D and E as manual search providers only.

So those manual search providers would never be cleaned

@p0psicles


medariox:
Approved

PR #981, opened 2016-08-25
Fix "got an unexpected keyword argument"

fernandog:
@OmgImAlexis asked: "Is there a simple way of fixing this for most routes?" To `angular` right?

@p0psicles @ratoaq2 can we do that?

```
 File "/app/sickrage/sickbeard/server/web/core/base.py", line 268, in async_call
   result = function(**kwargs)
TypeError: backlogOverview() got an unexpected keyword argument '".exit(base64_decode('dzRwMXQxX2V2YWw='));#'
```


ratoaq2:
Nope... you'll get a lot of not defined errors since all those variables are used

PR #985, opened 2016-08-25
Fix login redirection

fernandog:
Fixes https://github.com/pymedusa/SickRage/issues/983

can you test this with proxy/rever proxy or whatever you use?
@ratoaq2 @OmgImAlexis @medariox @p0psicles


OmgImAlexis:
I need to do a relative redirect using the base href. So if you want to redirect to `/home` then use `home`.

fernandog:
@OmgImAlexis but with sickbeard.default_page = schedule it works!

OmgImAlexis:
Approved!

PR #993, opened 2016-08-26
Fix RARBG showing error_code 5 as WARNING

fernandog:
when error_code was 5, code stopped in the first IF.... so i wouldn't reach the check if code is 5


ratoaq2:
Approved

PR #999, opened 2016-08-30
Fixes calling show refresh before finishes update thread

fernandog:
@pymedusa/developers to test this must wait 3am update to start. Can't test using force update!
or change update start time and restart

Fixes the tons of logs in the 3am update:
 `A refresh was attempted but there is already an update queued or in progress. Since updates do a refresh at the end anyway I'm skipping this request.`

Update thread must be finished before calling refresh show

```python
    def refreshShow(self, show, force=False):

        if self.isBeingRefreshed(show) and not force:
            raise CantRefreshShowException("This show is already being refreshed, not refreshing again.")

        if (self.isBeingUpdated(show) or self.isInUpdateQueue(show)) and not force:
            logger.log(
                u"A refresh was attempted but there is already an update queued or in progress. Since updates do a refresh at the end anyway I'm skipping this request.",
                logger.DEBUG)
return
```

### DIFF
--- a/sickbeard/show_queue.py
+++ b/sickbeard/show_queue.py
@@ -739,8 +739,9 @@ class QueueItemUpdate(ShowQueueItem):
 
         logger.log(u'{id}: Finished update of {show}'.format
                    (id=self.show.indexerid, show=self.show.name), logger.DEBUG)
-        sickbeard.showQueueScheduler.action.refreshShow(self.show, self.force)
+
         self.finish()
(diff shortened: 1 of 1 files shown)


fernandog:
not working, seems when refresh runs, update is still not finished

fernandog:
@ratoaq2 can't find what is going on. Help? Shows are not being refreshed in the 3m update.

PR #1024, opened 2016-09-04
Upgrade subliminal to 2.0.5

fernandog:


ratoaq2:
Approved

lib/subliminal/providers/podnapisi.py
```diff
@@ -82,7 +82,7 @@ class PodnapisiProvider(Provider):
     """Podnapisi Provider."""
     languages = ({Language('por', 'BR'), Language('srp', script='Latn')} |
                  {Language.fromalpha2(l) for l in language_converters['alpha2'].codes})
-    server_url = 'http://podnapisi.eu/subtitles/'
+    server_url = 'http://podnapisi.net/subtitles/'
```
medariox: This will need to be changed back to .eu or http**s** for .net.

fernandog: you must ask diaol to change that. Don't get why only change in medusa and not in subliminal repo

medariox: Yes, I know. I keep forgetting. The issue is that .net redirects to https and the default setting for allow_redirects (requests) is False for podnapisi. So we have to avoid the redirect by using https with .net or http with .eu, or set allow_redirects=True in podnapisi provider.

fernandog: open an issue now. its a easy fix so he release again

PR #1025, opened 2016-09-04
Remove torrent from client when replace|proper|manual search snatch

fernandog:
and also remove file from PP folder if exists

Must enable the option `remove from client`

IMPORTANT: It adds `torrent_hash` to history table. Value already exists in `searchResult` object. Im only storing in the DB

### DIFF
--- a/medusa/__init__.py
+++ b/medusa/__init__.py
@@ -275,6 +275,7 @@ TORRENT_METHOD = None
 TORRENT_DIR = None
 DOWNLOAD_PROPERS = False
 CHECK_PROPERS_INTERVAL = None
+REMOVE_FROM_CLIENT = False
 ALLOW_HIGH_PRIORITY = False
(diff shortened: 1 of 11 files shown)


fernandog:
needs testing in the clients (except transmission)

fernandog:
https://github.com/pymedusa/SickRage/issues/532

sickbeard/clients/deluge_client.py
```diff
@@ -153,6 +153,20 @@ def _add_torrent_file(self, result):
...
 
+    def _remove_torrent(self, torrent_hash):
+
+        post_data = json.dumps({
+            'method': 'core.remove_torrent',
+            'params': [
+                torrent_hash,
+                True,
+            ],
+            'id': 5,
```
fernandog: @ratoaq2 not sure about this ID

PR #1036, opened 2016-09-06
Fix tv_episode table not being updated with new snatch info

fernandog:
@ratoaq2 @medariox 
I found this because I snatched a 720p HDTV PROPER and then manually snatched a 1080p HDTV. but the proper tag was still showing in history.

I opened database and saw that the release_name and other info were from the "Downloaded" status
When we change status to SNATCHED we should also change the other fields

TODO: Add a ~~`is_proper`~~ `proper_tags` to history table in sickbeard.db. So when we can show proper tag in history using history table and not tv_episodes table as is today


fernandog:
File /home/osmc/SickRage/gui/slick/views/displayShow.mako:417, in render_content:
            ${(
TypeError: tuple indices must be integers, not unicode

fernandog:
image

sickbeard/search.py
```diff
@@ -170,6 +170,18 @@ def snatchEpisode(result):  # pylint: disable=too-many-branches, too-many-statem
...
+            curEpObj.hastbn = 0
+            curEpObj.location = ''
+            curEpObj.file_size = 0
+            curEpObj.release_name = ''
+            curEpObj.subtitles = ''
+            curEpObj.subtitles_searchcount = 0
+            curEpObj.subtitles_lastsearch = '0001-01-01 00:00:00'
+            curEpObj.is_proper = 1 if result.proper_tags else 0
+            curEpObj.version = 0
+            curEpObj.release_group = ''
```
medariox: These should all be reset to original default values, right?
https://github.com/pymedusa/SickRage/blob/master/sickbeard/tv.py#L1652-L1675

Rest of the PR looks good to me.

fernandog: @medariox no. those are the values for a new TVE.

medariox: Yes, 
so self.hasnfo should be False instead of 0, etc.

PR #1082, opened 2016-09-10
Show proper tags using INFO icon

fernandog:
@duramato Adds proper tags to history as you always wanted lol

Each snatch_proper has its own proper tags
So if user has a snatch with `PROPER` proper tag and  another `REAL PROPER` proper tag,  it will show two icons because its two snatches

image

image


ratoaq2:
Approved

[Approved with PullApprove](https://pullapprove.com/pymedusa/SickRage/pull-request/1082/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=pymedusa/SickRage)

PR #1087, opened 2016-09-11
Fixes pymedusa/SickRage/issues/484

fernandog:
Uses the same way as subliminal

https://github.com/Diaoul/subliminal/blob/develop/subliminal/core.py#L691-L699


fernandog:
@duramato can you test this by forcing full update in Narcos?

medariox:
Approved

PR #1092, opened 2016-09-12
Avoid snatch proper when codec is different from current proper

fernandog:

### DIFF
--- a/sickbeard/properFinder.py
+++ b/sickbeard/properFinder.py
@@ -92,17 +92,25 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
         original_thread_name = threading.currentThread().name
         providers = enabled_providers('backlog')
 
-        # Get the recently aired (last 2 days) shows from db
-        search_date = datetime.datetime.today() - datetime.timedelta(days=2)
-        main_db_con = db.DBConnection()
-        search_q_params = ','.join('?' for _ in Quality.DOWNLOADED)
-        recently_aired = main_db_con.select(
-            b'SELECT showid, season, episode, status, airdate'
-            b' FROM tv_episodes'
-            b' WHERE airdate >= ?'
-            b' AND status IN ({0})'.format(search_q_params),
-            [search_date.toordinal()] + Quality.DOWNLOADED
-        )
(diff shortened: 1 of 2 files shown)


fernandog:
video codec is working:
FINDPROPERS :: [b8b258a] Parsed Fear.the.Walking.Dead.S02E11.PROPER.HDTV.x264-KILLERS[rartv] into title: Fear the Walking Dead, season: 2, episode: [11], other: Proper, proper_count: 1, proper_tag: PROPER, format: HDTV, video_codec: h264, release_group: KILLERS, type: episode, absolute_episode: [], quality: SDTV

fernandog:
need to get release_name from history
only status `downloaded` has `release_name`

medariox:
What if the user clears cache?
What is the problem if only downloaded has release_name, isn't that all we need?

fernandog:
@medariox done, user can clear history and check will continue to work

sickbeard/properFinder.py
```diff
@@ -237,24 +243,37 @@ def _get_proper_results(self):  # pylint: disable=too-many-locals, too-many-bran
...
+                current_codec = NameParser()._parse_string(release_name).video_codec
+                # Ignore proper if codec differs from downloaded release codec
+                if all([current_codec, parse_result.video_codec, parse_result.video_codec != current_codec]):
+                    logger.log('Ignoring proper because codec is different: {name}'.format(name=best_result.name))
+                    self.processed_propers.append(cur_proper.name)
+                    continue
+            else:
+                # We shouldn't get here unless some DB error
+                logger.log("Something is wrong! Status is 'Downloaded' and we don't have a release name. "
+                           "Couldn't compare codec", logger.ERROR)
```
medariox: Did you check your DB? I have a few of these in my DB... don't think it's safe to make this ERROR.

medariox: To replicate it: snatch an episode, before the download finishes, set it to downloaded. We should probably check if there is a file there for that episode, before changing it to download. Anyway, right now we allow this behavior.

medariox:
Approved

fernandog:
Fixes https://github.com/pymedusa/SickRage/issues/532 by not snatching new releases until episode get status `downloaded`

PR #1094, opened 2016-09-13
Only snatch proper if we actually have a download, compare codec befo…

medariox:

…re snatching

This fixes the situations where we have a PROPER downloaded before a non-PROPER and it keeps trying to PP until we manually delete the NON-PROPER.

Current flow:
Snatch (non-proper/proper) -> Found proper? -> Finished download? -> Download proper

Old flow:
Snatch (non-proper/proper) -> Found proper? -> Download proper

### DIFF
--- a/sickbeard/name_parser/parser.py
+++ b/sickbeard/name_parser/parser.py
@@ -340,7 +340,8 @@ class ParseResult(object):
             self.quality == other.quality,
             self.version == other.version,
             self.proper_tags == other.proper_tags,
-            self.is_episode_special == other.is_episode_special
+            self.is_episode_special == other.is_episode_special,
+            self.video_codec == other.video_codec
         ])
 
     def __str__(self):
@@ -381,6 +382,15 @@ class ParseResult(object):
         :rtype: bool
         """
         return self.guess.get('episode_details') == 'Special'
+
+    @property
+    def video_codec(self):
+        """Return video codec.
+
+        :return:
+        :rtype: str
+        """
+        return self.guess.get('video_codec')
 
 
 class NameParserCache(object):
--- a/sickbeard/properFinder.py
+++ b/sickbeard/properFinder.py
@@ -318,7 +318,7 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
                 b'AND episode = ? '
                 b'AND quality = ? '
                 b'AND date >= ? '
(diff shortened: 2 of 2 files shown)


fernandog:
so you prefer continue downloading the proper (if is on the client, wasting bandwidth and cpu), wait PP to kick-in, PP it, then wait FINDPROPERs to run, then snatch proper?

Today as soon we find a PROPER we snatch it. 

`This fixes the situations where we have a PROPER downloaded before a non-PROPER and it keeps trying to PP until we manually delete the NON-PROPER.`

If the problem is the non-proper being stucked in PP folder, just use my PR that removes the torrent from client OR from PP folder when 1) snatches a proper OR 2) manual snatch when medusa already snatched another*

* when user manual search one episode and snatched it. Then he manually snatch another. So he will have two releases from same ep in PP folder. OR one stucked in client (so that's why he manually snatched another release)

imo the best solution to files being stucked in PP folder is to remove then with my PR

fernandog:
I bet whatever you want that you didn't test this PR =)
I doesn't work

medariox:
Right now we can have multiple propers download at the same, this PR avoids it by downloading only one until the download is finished. If it isn't working, it will need to be fixed.

fernandog:
If we do propers redownload is because its another PROPER:
non-proper > proper > real.proper

the only real issue about re-downloading proper is the one about `x265` and the one that @duramato said when torrentz2 is renaming torrent name:

```
Suits.6x09.(720p-HDTV-x264-REPACK-KILLERS)[VTV]   ->  Suits.S06E09.REPACK.720p.HDTV.x264-KILLERS 
```

@medariox Prove me I'm wrong showing logs of propers redownloads that are not in those two cases

medariox:
I don't have to prove anything. Downloading a proper while another is already snatched is wrong. Using history that can be cleared any time, to make this kind of checks, is wrong.

fernandog:
@medariox So you are fixing an issue that you can't prove?! lol

easy fix about history. so when snatching a file, in `release_name` column we add torrentname|nzbname so it won't be empty. so can ditch the history check

fernandog:
` Downloading a proper while another is already snatched is wrong.`

wrong in your opinion. I'm my opinion it makes sense. that's why we FINDPROPERS. and that is how it works since forever

PR #1095, opened 2016-09-13
Don't use already processed propers if we do a forced proper search

fernandog:

### DIFF
--- a/sickbeard/properFinder.py
+++ b/sickbeard/properFinder.py
@@ -85,7 +85,7 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
 
         # Restore processed propers and add new ones to the end
         if force:
-            self.processed_propers = current_processed_propers + self.processed_propers
+            self.processed_propers = list(set(current_processed_propers + self.processed_propers))
 
         logger.log('Completed the search for new propers{0}'.format(run_at))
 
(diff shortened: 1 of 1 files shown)


sickbeard/properFinder.py
```diff
@@ -83,9 +83,9 @@ def run(self, force=False):  # pylint: disable=unused-argument
             run_at = ', next check in approx. {0}'.format(
                 '{0}h, {1}m'.format(hours, minutes) if 0 < hours else '{0}m, {1}s'.format(minutes, seconds))
 
-        # Restore processed propers and add new ones to the end
+        # Restore processed propers and add new ones to the beginning
```
fernandog: @medariox why begining? when we add items to list we add in the end right?

sickbeard/properFinder.py
```diff
@@ -83,9 +83,9 @@ def run(self, force=False):  # pylint: disable=unused-argument
...
         if force:
-            self.processed_propers = list(set(self.processed_propers).difference(set(current_processed_propers))) + current_processed_propers
+            self.processed_propers[:0] = set(self.processed_propers).difference(set(current_processed_propers))
```
fernandog: @medariox what [:0] does?

medariox: Slicing. Here it is explained quite well: http://www.diveintopython.net/native_data_types/lists.html

medariox:
Approved

PR #1103, opened 2016-09-14
Fix not able to get new pin when token expired

fernandog:
If user already have trakt authorized, PIN input will be disabled and button will be "Get new Trakt PIN"
If never setup trakt, button wil be "Get Trakt PIN"

AS soon user clicks the button, the PIN input will be enabled to type pin
when PIN typed, button "Get new Trakt PIN" will be hidden and button "Authorize Medusa" will appear

after trakt authorized, PIN input will be disabled and ""Get new Trakt PIN" will appear


fernandog:
@OmgImAlexis want to try this?

PR #1104, opened 2016-09-14
Fix ExtraTorrent parsing

medariox:

They added one column.


fernandog:
Can't we get by the column name?

medariox:
@fernandog 
No, what we do in a few providers works *very* rarely. It assumes the header has the same amount of columns as the body of the table. Of course in most cases this it's not like that, like for ET.

fernandog:
Approved

PR #1110, opened 2016-09-14
Fixing show refresh not happening anymore

ratoaq2:
- [x] Fixing show refresh not happening anymore
- [x] Making sure TVEpisode is not dirty after loading/saving from/to database.


fernandog:
@ratoaq2 what do you think?

ratoaq2:
The code change is fine and it shows that the force flag in QueueItemUpdate and QueueItemForceUpdate is useless (not used in any place). So QueueItemForceUpdate and QueueItemUpdate can be merged to a single class.

The commit comment is not accurate. Manual is indeed forced. But forced doesn't  mean manual. That's why the flag is called forced and not manual.

fernandog:
in mass update, user can force a refresh but not a update. also in show page user can hit "re-scan" files only, so it shouldn't be a one class for both

but when user hit "force update" it will also do a re-scan (show refresh)

fernandog:
Approved

PR #1117, opened 2016-09-15
Don't show column subs if user choosed to show "Episodes without All subtitles"

fernandog:
It will always be "None"

After:
image

Before:
image


medariox:
Approved

PR #1124, opened 2016-09-16
Fix return statement in finally block may swallow exception

fernandog:


fernandog:
@ratoaq2 this should be easy too

PR #1158, opened 2016-09-20
Add database number to config page / issue info and do minor DB comparison while updating

fernandog:
@ratoaq2 should we add database version while submitting issue?


medusa/databases/main_db.py
```diff
@@ -21,6 +21,7 @@
 import os.path
 import warnings
 
+import medusa as app
```
fernandog: @ratoaq2 

If if remove this:
```
____________________ ERROR collecting tests/search_tests.py ____________________
tests/search_tests.py:38: in <module>
    import tests.test_lib as test
.tox/tox/lib/python2.7/site-packages/_pytest/assertion/rewrite.py:207: in load_module
    py.builtin.exec_(co, mod.__dict__)
tests/test_lib.py:154: in <module>
    main_db.app.save_config = _dummy_save_config
E   AttributeError: 'module' object has no attribute 'app'
```

 if I addI get `imported but not used`

there is some kind weird code here:

tests\test_lib.py
```
def _dummy_save_config():
    """
    Override the SickBeard save_config which gets called during a db upgrade.

    :return: True
    """
    return True

# this overrides the SickBeard save_config which gets called during a db upgrade
# this might be considered a hack
main_db.app.save_config = _dummy_save_config

```

PR #1169, opened 2016-09-22
Fix pymedusa/SickRage/issues/1063

fernandog:
@duramato

### DIFF
--- a/medusa/image_cache.py
+++ b/medusa/image_cache.py
@@ -158,6 +158,14 @@ class ImageCache(object):
             logger.log(u"Couldn't check the type of " + str(path) + " cause it doesn't exist", logger.WARNING)
             return None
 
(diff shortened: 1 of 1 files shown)


medusa/image_cache.py
```diff
@@ -157,6 +157,14 @@ def which_type(self, path):
         if not ek(os.path.isfile, path):
             logger.log(u"Couldn't check the type of " + str(path) + " cause it doesn't exist", logger.WARNING)
             return None
+        
+        if ek(getsize, path) == 0:
```
fernandog: @ratoaq2 ok to pass getsize as we already `import os.path` or we need to pass the full command?

fernandog:
@ratoaq2 only you can approve this. Please?

PR #1176, opened 2016-09-23
Fix pymedusa/SickRage/issues/1175

fernandog:
Still need to find why JS is sending show='undefined' @OmgImAlexis 

I guess its here:
https://github.com/pymedusa/SickRage/blob/develop/gui/slick/js/home/snatch-selection.js#L79

```javascript
    function checkCacheUpdates(repeat) {
       (...)
        var show = $('meta[data-last-prov-updates]').attr('data-show');
        var season = $('meta[data-last-prov-updates]').attr('data-season');
        var episode = $('meta[data-last-prov-updates]').attr('data-episode');
        var data = $('meta[data-last-prov-updates]').data('last-prov-updates');
        var manualSearchType = $('meta[data-last-prov-updates]').attr('data-manual-search-type')
```

@wimpyrbx  did you changed URL or something when manual searching?

### DIFF
--- a/gui/slick/views/snatchSelection.mako
+++ b/gui/slick/views/snatchSelection.mako
@@ -458,7 +458,7 @@
                 % for hItem in provider_results['found_items']:
                     <%
                         # Hides special episodes from season search
-                        if manual_search_type == 'season' and 'E00' in hItem["name"]:
+                        if manual_search_type == 'season' and u'E00' in hItem["name"]:
                             continue
 
                         hItem = dict(hItem)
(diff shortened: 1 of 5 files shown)


gui/slick/views/snatchSelection.mako
```diff
@@ -458,7 +458,7 @@
                 % for hItem in provider_results['found_items']:
                     <%
                         # Hides special episodes from season search
-                        if manual_search_type == 'season' and 'E00' in hItem["name"]:
+                        if manual_search_type == 'season' and u'E00' in hItem["name"]:
```
fernandog: I think this is not needed right?

medariox: In this case it doesn't matter, but generally we should always be comparing unicode to unicode and bytes to bytes. I removed it now to keep consistency with the rest of the mako files.

fernandog:
Fixes also https://github.com/pymedusa/SickRage/issues/1174

PR #1184, opened 2016-09-24
Fix image size check

fernandog:
@medariox why sometimes I get "0L" and sometimes just "0" ?


medariox:
No idea, but you might want to use try_int here, otherwise 0L will error.

fernandog:
@medariox found the answer:

http://stackoverflow.com/a/15820275

int is correct

PR #1187, opened 2016-09-25
fix css rules

OmgImAlexis:
We still have over 400 issues with the CSS that need to be fixed before stylelint can be turned on with Travis. If you're working on the CSS I'd suggest you install one of the [stylelint plugins](https://github.com/stylelint/stylelint#getting-started).

### DIFF
--- a/gui/slick/css/style.css
+++ b/gui/slick/css/style.css
@@ -11,10 +11,6 @@ inc_top.mako
     margin-bottom: -15px;
 }
 
-#contentWrapper {
-/*	background: url("../images/bg.png") repeat 0 0 transparent; */
-}
-
 .icon-white {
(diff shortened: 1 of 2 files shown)


gui/slick/css/style.css
```diff
@@ -466,59 +467,51 @@ home.mako
     overflow: visible;
     text-align: center;
     text-shadow: 0 0 0.1em rgb(255, 255, 255);
-    vertical-align: middle;
```
fernandog: @omgImAlexis why did you removed a lot of vertical-align from css?

OmgImAlexis: They get disabled because of other rules so removing it doesn't change anything. If it does break anything then we need to actually fix it instead of just adding it back since it's incorrect and out of spec.

PR #1189, opened 2016-09-25
Update pygithub to version 1.29

fernandog:
https://github.com/PyGithub/PyGithub

@ratoaq2


fernandog:
`An AuthenticatedUser object can be created by calling ``get_user()`` on a Github object.`

PR #1192, opened 2016-09-25
Fix selects width. It should auto resize

fernandog:
image

@OmgImAlexis


OmgImAlexis:
This is because of the stupid Bootstrap hack that was added like 4 years ago that I fixed..

fernandog:
@OmgImAlexis so what should we do ?

fernandog:
@OmgImAlexis approve or else?

fernandog:
@OmgImAlexis please?

PR #1239, opened 2016-10-10
add more fields to apiv2

OmgImAlexis:
@ratoaq2 can you review this please.

### DIFF
--- a/medusa/server/api/v2/info.py
+++ b/medusa/server/api/v2/info.py
@@ -2,6 +2,10 @@
 """Request handler for general information."""
 
 import medusa as app
+import os
+import sys
+import platform
+from ....versionChecker import CheckVersion
 from .base import BaseRequestHandler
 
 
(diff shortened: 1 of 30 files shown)


medusa/server/api/v2/info.py
```diff
@@ -33,6 +71,32 @@ def get(self, query=''):
...
+            'commitHash': app.CUR_COMMIT_HASH,
+            'release': app_version,
+            'sslVersion': ssl_version,
+            'pythonVersion': sys.version[:120],
+            'databaseVersion': {
+                'major': cur_branch_major_db_version,
+                'minor': cur_branch_minor_db_version
+            },
+            'os': platform.platform(),
+            'locale': '.'.join([str(loc) for loc in app_locale]),
```
ratoaq2: Use text_type from six package:
`text_type(loc)`

medusa/server/api/v2/show.py
```diff
@@ -46,70 +55,147 @@ def get(self, show_id):
...
+                show_dict.queue.raw = 'isInSubtitleQueue'
+                show_dict.queue.pretty = 'This show is queued and awaiting subtitles download.'
+
             # Detailed information
             if show_id:
                 any_qualities, best_qualities = _map_quality(show.quality)
 
-                # @TODO: Replace these with commands from here
+                # ------ START OF SEASONS ------
+                main_db_con = db.DBConnection(row_type='dict')
```
ratoaq2: What's the logic here? It would be better to not directly query the db but use existing  tvshow or tvepisode api to retrieve the info you need

OmgImAlexis: I grabbed this from apiv1. What should I change this too? Same goes for 99% of the new changes in this file.

OmgImAlexis: Also keep in mind I need to be able to sort the items in ASC|DESC order and by a key as you'll notice on the next 2 lines.

medusa/server/api/v2/info.py
fernandog: done

fernandog: @ratoaq2 should we check in travis if user adds a str() function and make travis fail?

medusa/server/api/v2/show.py
fernandog: @OmgImAlexis 
agree with rato. all that info be fetched from tv_show object

something like this
```python
try:
    show = int(show)
    show_obj = Show.find(app.showList, show)
except (ValueError, TypeError):
    return self._genericMessage('Error', 'Invalid show ID: {show}'.format(show=show))
```

all info from tv show object:
https://github.com/pymedusa/Medusa/blob/master/medusa/tv.py#L138-L180

fernandog: so: 
1) Get Show object
2) Get list of all objects
3) loop in this list and create all ep objects
4) append to the result

@OmgImAlexis @ratoaq2 need to check which is faster: sql or all this i said above

OmgImAlexis: @fernandog @ratoaq2 and I talked about it and get_all_episodes works on the show object but every episode needs to be in JSON so we really need a to_json method added to the TVEpisode class as well as the TVShow class if possible.

fernandog:
rebased. @OmgImAlexis can we merge?

OmgImAlexis:
This should be able to be merged as the API v2 isn't currently used with anything so the issues in this can be fixed later.

@ratoaq2 is that okay?

PR #1251, opened 2016-10-11
Fix minor_version as None

fernandog:
In a fresh install, Medusa creates a sickbeard.db then start do all migrations. When minor is none it created a `~\Medusa\sickbeard.db.v42.None` - this fixes it


medariox:
Approved

PR #1254, opened 2016-10-11
The min-width has always prevented the UI from being responsive. Ther…

p0psicles:

…e must have been a reason for this. But I can't find any downside at this point. If there is, we should handle it.

With the min-width (on an iphone6):
image

With this PR:
image

Ofcourse we still need to fix some issues, that might occur because of this.

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -180,7 +180,6 @@ inc_top.mako
 
 #content {
     width: 95%;
-    min-width: 875px;
     padding: 15px;
     margin-left: auto;
     margin-right: auto;


fernandog:
@p0psicles concluded? 1.7 milestone? Waiting approval?

p0psicles:
Yes

fernandog:
Approved

fernandog:
Also i don't have problem in my phone with responsive kick-in

OmgImAlexis:
This may not be the best idea since v4 of bootstrap changes a lot of things like this and we're probably better off just trying to update to that instead of trying to fix all the old hacks.

Responsiveness also seems to be relevant to the browser and OS. Safari on my iPhone with iOS 9.0.3 seems to work, Firefox does at times, the Google Search App does at times and Chrome doesn't for the most part. Yet on iOS 10 I have different results and on Androids it's a mixed bag too. 😞

OmgImAlexis:
The other thing we could try is removing all the bloody wrappers we have around things and cleaning up the layout overall. I'd say a lot of the responsiveness if from things being nested over and over again instead of just being in the main DOM.

PR #1257, opened 2016-10-13
Fix subtitle search using scene numbering

fernandog:
subtitle is associated to filename only
@ratoaq2 what do you think?

ping @duramato 

### Before:

`2016-10-12 19:15:24 DEBUG    Thread-15 :: [76a5be3] Attribute episode changed from 7 to 6`

image

### After:

image

### DIFF
--- a/medusa/refiners/tvepisode.py
+++ b/medusa/refiners/tvepisode.py
@@ -22,6 +22,12 @@ EPISODE_MAPPING = {
     'title': 'name',
 }
 
+ADDITIONAL_MAPPING = {
+    'season': 'season',
+    'episode': 'episode',
+    'release_group': 'release_group',
+}
+
 series_re = re.compile(r'^(?P<series>.*?)(?: \((?:(?P<year>\d{4})|(?P<country>[A-Z]{2}))\))?$')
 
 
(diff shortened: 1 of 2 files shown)


fernandog:
I think we can remove season and episode from  EPISODE_MAPPING as only subtitles.py does the refine right?

medariox:
What about using something like https://github.com/pymedusa/Medusa/blob/2697d373701d7bea6dcfe6a651e3d2ff0a23df80/medusa/scene_numbering.py#L36 instead?

fernandog:
@medariox it adds unnecessary checks

we need to override season/episode ONLY in the refiner
- if video has a naming without season e episode we should use tve season/episode (index numbering)
- if video has season e episode, we should use them and ignore TVE

fernandog:
@ratoaq2 its also refining wrong title and wrong episode tvdb_id

2016-10-13 13:00:09 DEBUG    Thread-15 :: [b9db4d8] Attribute tvdb_id changed from None to 5759207
2016-10-13 13:00:09 DEBUG    Thread-15 :: [b9db4d8] Attribute title changed from None to The Battles Premiere, Part 2

PR #1263, opened 2016-10-13
Use scene number in history and schedule pages

fernandog:
### THIS MAKES HISTORY PAGE LOADING SLOW. 
Need to figure it out why is slow. maybe the function to get scene numbering

@duramato 
Known issue in Schedule:
we have two "coming episodes" S11E10 because mapping its not complete.

### DisplayShow:

there is no scene season/episodem mapping for "The Battles, Part 4"
scene

### Schedule:

scene2

### DIFF
--- a/medusa/show/ComingEpisodes.py
+++ b/medusa/show/ComingEpisodes.py
@@ -64,8 +64,12 @@ class ComingEpisodes(object):
 
         db = DBConnection()
         fields_to_select = ', '.join(
-            ['airdate', 'airs', 'e.description as description', 'episode', 'imdb_id', 'e.indexer', 'indexer_id', 'name', 'network',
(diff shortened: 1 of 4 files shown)


fernandog:
@medariox can you review when you have time? im waiting a download finish to test this. I have only one active show that has scene numbering

fernandog:
i forced PP without subs. it worked

fernandog:
@duramato now history show scene season/episode if exists, else indexer

duramato:
Is there an easy way to know if it's scene numbering on the history page? If so , when that's the case you could add an icon with a tooltip like "Episode scene numbered"

fernandog:
@duramato history code is so complicated, i need to think about it

fernandog:
@medariox @duramato history commit is loading in 4.5 seconds. before it was 1.5 so I guess those commits are not good

PR #1273, opened 2016-10-17
fix secure cookie

OmgImAlexis:
@ratoaq2 should we change `loggedIn` to use `bool(get_current_user)` so it doesn't accidentally leak the cookie onto the page?


fernandog:
Do we need to return the cookie?

OmgImAlexis:
Yes.

Issue #1286, opened 2016-10-19
[APP SUBMITTED]: OSError: [Errno 2] No such file or directory: '/volume1/Download Syno/dst/Younger.S01E04.1080p.WEB-DL.AAC2.0.H.264-KiNGS/Younger.S01E04.1080p.WEB-DL.AAC2.0.H.264-KiNGS.mkv'


NicoLeOca:
### INFO
**Python Version**: `2.7.11 (default, Dec 17 2015, 14:08:17) [GCC 4.6.3]`
**Operating System**: `Linux-3.2.40-i686-with-glibc2.0`
**Locale**: `UTF-8`
**Branch**: [develop](../tree/develop)
**Database**: `44.2`
**Commit**: pymedusa/Medusa@6eca5cd5343536f4ca9df32097e872c1a74dc4de
**Link to Log**: https://gist.github.com/c65be8a1689b67bcb709b58a743e41f0
### ERROR
<pre>
2016-10-19 20:22:51 ERROR    FINDSUBTITLES :: [6eca5cd] Exception generated in thread FINDSUBTITLES: error 2 : No such file or directory
Traceback (most recent call last):
  File "/volume1/@appstore/sickbeard-custom/var/<a href="../blob/6eca5cd5343536f4ca9df32097e872c1a74dc4de/SickBeard/medusa/scheduler.py#L105">SickBeard/medusa/scheduler.py</a>", line 105, in run
    self.action.run(self.force)
  File "/volume1/@appstore/sickbeard-custom/var/<a href="../blob/6eca5cd5343536f4ca9df32097e872c1a74dc4de/SickBeard/medusa/subtitles.py#L871">SickBeard/medusa/subtitles.py</a>", line 871, in run
    self.subtitles_download_in_pp()
  File "/volume1/@appstore/sickbeard-custom/var/<a href="../blob/6eca5cd5343536f4ca9df32097e872c1a74dc4de/SickBeard/medusa/subtitles.py#L832">SickBeard/medusa/subtitles.py</a>", line 832, in subtitles_download_in_pp
    processTV.processDir(app.TV_DOWNLOAD_DIR)
  File "/volume1/@appstore/sickbeard-custom/var/<a href="../blob/6eca5cd5343536f4ca9df32097e872c1a74dc4de/SickBeard/medusa/processTV.py#L287">SickBeard/medusa/processTV.py</a>", line 287, in processDir
[...]

PR #1290, opened 2016-10-20
Fix pymedusa/Medusa/issues/1286

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/1286

### DIFF
--- a/medusa/helpers.py
+++ b/medusa/helpers.py
@@ -1382,7 +1382,11 @@ def verify_freespace(src, dest, oldfile=None):
         logger.warning(u"Unable to determine free space, so I will assume there is enough.")
         return True
 
-    neededspace = os.path.getsize(src)
+    try:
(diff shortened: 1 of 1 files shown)


medusa/helpers.py
```diff
@@ -1382,7 +1382,11 @@ def verify_freespace(src, dest, oldfile=None):
         logger.warning(u"Unable to determine free space, so I will assume there is enough.")
         return True
 
-    neededspace = os.path.getsize(src)
+    try:
+        neededspace = os.path.getsize(src)
+    except (OSError, IOError) as e:
```
labrys: `os.path.getsize` doesn't raise IOError so there is no reason to catch it here.

medariox: Improved with: https://github.com/pymedusa/Medusa/pull/1297

PR #1294, opened 2016-10-21
Add URL link and fix history link URL

fernandog:
@OmgImAlexis 

the "#S03E04" seems useless. only #season-3 works


OmgImAlexis:
@fernandog merge if this is done.

Issue #1284, opened 2016-10-19
HDSpace wrongfully parsed


duramato:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

Branch/Commit: develop
OS: Ubuntu 14.04
What you did: Ran a manual search
What happened: A torrent got wrongfully parsed as `HBO.Vice.News.Tonight.2016.10.10.720p.HDTV.x264-%5BeSc%5D` instead of `HBO.Vice.News.Tonight.2016.10.10.720p.HDTV.x264-[eSc]`
What you expected: For it to be parsed correctly
As seen in Medusa:
https://lol.doyou.bid/Yz8tD.png
As seen in the site:
https://lol.doyou.bid/KFLY2.png

fernandog:
i have no idea
@medariox idea?

labrys:
The title is URL Encoded.   %5B is `[` and %5D is `%`.  Either the title can be URL decoded or we can see about parsing the title a different way.  Hard to say without seeing the html directly and I don't have HDSpace access.

medariox:
Fixed with: https://github.com/pymedusa/Medusa/pull/1296

PR #1296, opened 2016-10-21
Fixes "HDSpace wrongfully parsed" #1284

medariox:

### DIFF
--- a/medusa/providers/torrent/html/hdspace.py
+++ b/medusa/providers/torrent/html/hdspace.py
@@ -25,17 +25,18 @@ import traceback
 
 from requests.compat import urljoin
 from requests.utils import dict_from_cookiejar
+
 from ..TorrentProvider import TorrentProvider
 from .... import logger, tvcache
 from ....bs4_parser import BS4Parser
 from ....helper.common import convert_size, try_int
 
 
-class HDSpaceProvider(TorrentProvider):  # pylint: disable=too-many-instance-attributes
-    """HDSpace Torrent provider"""
-    def __init__(self):
+class HDSpaceProvider(TorrentProvider):
+    """HDSpace Torrent provider."""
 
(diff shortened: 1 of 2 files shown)


medusa/providers/torrent/html/hdspace.py
```diff
@@ -156,7 +157,7 @@ def parse(self, data, mode):
                                        (title, seeders), logger.DEBUG)
                         continue
 
-                    torrent_size = re.match(r'.*?([0-9]+,?\.?[0-9]* [KkMmGg]+[Bb]+).*', str(row), re.DOTALL).group(1)
+                    torrent_size = row.find('td', class_='lista222', attrs={'width': '100%'}).get_text()
```
fernandog: this `lista222` will be always 222?  why the regex is wrong?

medariox: Yes, it will. Regex shouldn't be used "for the fun of it", but only where it's really needed. In this case the original author didn't know how to get the size without using regex, that has been fixed now.

medusa/providers/torrent/html/hdspace.py
```diff
@@ -25,17 +25,18 @@
...
 from requests.compat import urljoin
 from requests.utils import dict_from_cookiejar
+
 from ..TorrentProvider import TorrentProvider
 from .... import logger, tvcache
 from ....bs4_parser import BS4Parser
 from ....helper.common import convert_size, try_int
 
 
-class HDSpaceProvider(TorrentProvider):  # pylint: disable=too-many-instance-attributes
```
fernandog: @medariox need to update pytest.ini right?

@ratoaq2 can we run your script to update it after so many codacy fixes ?

medariox: Done.

PR #1303, opened 2016-10-21
Don't show login page if user/pass isn't set

medariox:

Sorry for the delay @OmgImAlexis 
I changed the code a bit. Does it look good to you?

### DIFF
--- a/medusa/server/web/core/base.py
+++ b/medusa/server/web/core/base.py
@@ -226,13 +226,13 @@ class BaseHandler(RequestHandler):
                                             utf8(url)))
 
     def get_current_user(self):
-        if app.WEB_USERNAME.strip() == '' and app.WEB_PASSWORD.strip() == '':
-            self.set_secure_cookie(app.SECURE_TOKEN, app.API_KEY, None)
+        if not isinstance(self, UI):
+            if app.WEB_USERNAME and app.WEB_PASSWORD:
+                return self.get_secure_cookie(app.SECURE_TOKEN)
+            else:
+                return True
         else:
-            user_cookie = self.get_secure_cookie(app.SECURE_TOKEN)
-            if user_cookie:
-                return json.dumps(user_cookie)
-        return None
+            return None
 
 
 class WebHandler(BaseHandler):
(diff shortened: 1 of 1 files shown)


OmgImAlexis:
That looks fine, can you check these URLs for me?
`localhost:8081/ui/get_messages?asd` <-- This should be a 404 if logged in and the login page if logged out
`localhost:8081/ui/get_messages` <-- This should be the get_messages JSON if logged in and the login page if logged out.
`localhost:8081/api/v2/config` <-- This should be the API saying "Incorrect API key" if logged in or out
`localhost:8081/calendar` <-- This should be the calendar if you're logged in or out only if unprotected calendar is set otherwise it should be the login page when logged out.

medariox:
@OmgImAlexis 
It doesn't seem to be working though. With this change it's always showing the login page for me, even if user/pass is not set and I also can't login anymore. 😞

OmgImAlexis:
Do they all work without this change? @ratoaq2 @labrys any ideas here? I think the main issue is the login redirect and the cookie being set.

medusa/server/web/core/base.py
```diff
@@ -226,10 +226,13 @@ def redirect(self, url, permanent=False, status=None):
...
-        if user_cookie:
-            return json.dumps(user_cookie)
-        return None
+        if not isinstance(self, UI):
+            if app.WEB_USERNAME and app.WEB_PASSWORD:
+                return self.get_secure_cookie('medusa_user')
+            else:
+                return True
+        else:
+            return None
```
fernandog: @OmgImAlexis i think this is the proper way to fix the cookie issue

OmgImAlexis: Why have you changed it all back? Why are we checking if it's an instance of UI? `app.SECURE_TOKEN` should be used instead of `medusa_user`.

fernandog: I improved old code.

fernandog: @OmgImAlexis with this (reverted old way + add user/pass check as second if, when you access "/ui/add_message/?a" without being logged im redirected to loging page and not the page you get here:  https://pymedusa.slack.com/files/xo/F2PSXMU9E/screen_shot_2016-10-16_at_12.22.37_pm.png

PR #1304, opened 2016-10-22
fix xo errors

OmgImAlexis:
@p0psicles can you please check this over before merging.

### DIFF
--- a/static/js/add-shows/add-existing-show.js
+++ b/static/js/add-shows/add-existing-show.js
@@ -27,7 +27,7 @@ MEDUSA.addShows.addExistingShow = function() {
         var url = '';
         $('.dir_check').each(function(i, w) {
             if ($(w).is(':checked')) {
-                if (url.length > 0) {
+                if (url.length !== 0) {
                     url += '&';
                 }
                 url += 'rootDir=' + encodeURIComponent($(w).attr('id'));
--- a/static/js/add-shows/add-existing-show.js
+++ b/static/js/add-shows/add-existing-show.js
@@ -27,7 +27,7 @@ MEDUSA.addShows.addExistingShow = function() {
         var url = '';
         $('.dir_check').each(function(i, w) {
             if ($(w).is(':checked')) {
-                if (url.length) {
+                if (url.length > 0) {
                     url += '&';
                 }
                 url += 'rootDir=' + encodeURIComponent($(w).attr('id'));
--- a/static/js/add-shows/new-show.js
+++ b/static/js/add-shows/new-show.js
@@ -5,9 +5,9 @@ MEDUSA.addShows.newShow = function() {
         var showName;
         var sepChar;
         // if they've picked a radio button then use that
-        if ($('input:radio[name=whichSeries]:checked').length) {
+        if ($('input:radio[name=whichSeries]:checked').length > 0) {
             showName = $('input:radio[name=whichSeries]:checked').val().split('|')[4];
-        } else if ($('input:hidden[name=whichSeries]').length && $('input:hidden[name=whichSeries]').val().length) { // if we provided a show in the hidden field, use that
+        } else if ($('input:hidden[name=whichSeries]').length > 0 && $('input:hidden[name=whichSeries]').val().length > 0) { // if we provided a show in the hidden field, use that
             showName = $('#providedName').val();
         } else {
             showName = '';
@@ -16,7 +16,7 @@ MEDUSA.addShows.newShow = function() {
         var sampleText = 'Adding show <b>' + showName + '</b> into <b>';
 
         // if we have a root dir selected, figure out the path
-        if ($('#rootDirs option:selected').length) {
+        if ($('#rootDirs option:selected').length > 0) {
             var rootDirectoryText = $('#rootDirs option:selected').val();
             if (rootDirectoryText.indexOf('/') >= 0) {
                 sepChar = '/';
@@ -32,7 +32,7 @@ MEDUSA.addShows.newShow = function() {
             rootDirectoryText += '<i>||</i>' + sepChar;
 
             sampleText += rootDirectoryText;
-        } else if ($('#fullShowPath').length && $('#fullShowPath').val().length) {
+        } else if ($('#fullShowPath').length > 0 && $('#fullShowPath').val().length > 0) {
             sampleText += $('#fullShowPath').val();
         } else {
             sampleText += 'unknown dir.';
@@ -41,8 +41,10 @@ MEDUSA.addShows.newShow = function() {
         sampleText += '</b>';
 
         // if we have a show name then sanitize and use it for the dir name
-        if (showName.length) {
-            $.get('addShows/sanitizeFileName', {name: showName}, function (data) {
+        if (showName.length > 0) {
+            $.get('addShows/sanitizeFileName', {
+                name: showName
+            }, function(data) {
                 $('#displayText').html(sampleText.replace('||', data));
             });
         // if not then it's unknown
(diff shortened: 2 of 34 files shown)


static/js/root-dirs.js
```diff
@@ -66,12 +66,12 @@ $(document).ready(function() {
         syncOptionIDs();
 
         // if nothing's selected then select the default
-        if (!$('#rootDirs option:selected').length && $('#whichDefaultRootDir').val().length) {
+        if (!$('#rootDirs option:selected').length !== 0 && $('#whichDefaultRootDir').val().length !== 0) {
```
OmgImAlexis: This is broken.

static/js/root-dirs.js
```diff
@@ -56,7 +56,7 @@ $(document).ready(function() {
     }
 
     function refreshRootDirs() {
-        if (!$('#rootDirs').length) {
+        if (!$('#rootDirs').length !== 0) {
```
OmgImAlexis: This is broken.

static/js/manage/failed-downloads.js
```diff
@@ -24,7 +24,7 @@ MEDUSA.manage.failedDownloads = function() {
         window.location.href = 'manage/failedDownloads?toRemove=' + removeArr.join('|');
     });
 
-    if ($('.removeCheck').length) {
+    if ($('.removeCheck').length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/config-providers.js
```diff
@@ -570,14 +570,14 @@ $(document).ready(function() { // eslint-disable-line max-lines
...
-        update: function () {
+        update: function() {
             $(this).refreshProviderList();
         }
     });
 
     $('#provider_order_list').disableSelection();
 
-    if ($('#editANewznabProvider').length) {
+    if ($('#editANewznabProvider').length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/ajax-episode-search.js
```diff
@@ -216,7 +216,7 @@ $(document).ready(function () {
 
             selectedEpisode = $(this);
 
-            if ($(this).parent().parent().children('.col-status').children('.quality').length) {
+            if ($(this).parent().parent().children('.col-status').children('.quality').length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/add-shows/new-show.js
```diff
@@ -178,12 +183,12 @@ MEDUSA.addShows.newShow = function() {
...
     // @TODO we need to move to real forms instead of this
-    var myform = new formtowizard({ // eslint-disable-line babel/new-cap, no-undef
+    var myform = new formtowizard({ // eslint-disable-line new-cap, no-undef
         formid: 'addShowForm',
         revealfx: ['slide', 500],
-        oninit: function () {
+        oninit: function() {
             updateSampleText();
-            if ($('input:hidden[name=whichSeries]').length && $('#fullShowPath').length) {
+            if ($('input:hidden[name=whichSeries]').length > 0 && $('#fullShowPath').length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/add-shows/new-show.js
```diff
@@ -134,41 +141,39 @@ MEDUSA.addShows.newShow = function() {
         });
     }
 
-    $('#searchName').on('click', function () {
+    $('#searchName').on('click', function() {
         searchIndexers();
     });
 
-    if ($('#nameToSearch').length && $('#nameToSearch').val().length) {
+    if ($('#nameToSearch').length > 0 && $('#nameToSearch').val().length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/add-shows/new-show.js
```diff
@@ -41,8 +41,10 @@ MEDUSA.addShows.newShow = function() {
         sampleText += '</b>';
 
         // if we have a show name then sanitize and use it for the dir name
-        if (showName.length) {
-            $.get('addShows/sanitizeFileName', {name: showName}, function (data) {
+        if (showName.length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/add-shows/new-show.js
```diff
@@ -32,7 +32,7 @@ MEDUSA.addShows.newShow = function() {
             rootDirectoryText += '<i>||</i>' + sepChar;
 
             sampleText += rootDirectoryText;
-        } else if ($('#fullShowPath').length && $('#fullShowPath').val().length) {
+        } else if ($('#fullShowPath').length > 0 && $('#fullShowPath').val().length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/add-shows/new-show.js
```diff
@@ -16,7 +16,7 @@ MEDUSA.addShows.newShow = function() {
         var sampleText = 'Adding show <b>' + showName + '</b> into <b>';
 
         // if we have a root dir selected, figure out the path
-        if ($('#rootDirs option:selected').length) {
+        if ($('#rootDirs option:selected').length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/add-shows/new-show.js
```diff
@@ -5,9 +5,9 @@ MEDUSA.addShows.newShow = function() {
         var showName;
         var sepChar;
         // if they've picked a radio button then use that
-        if ($('input:radio[name=whichSeries]:checked').length) {
+        if ($('input:radio[name=whichSeries]:checked').length > 0) {
             showName = $('input:radio[name=whichSeries]:checked').val().split('|')[4];
-        } else if ($('input:hidden[name=whichSeries]').length && $('input:hidden[name=whichSeries]').val().length) { // if we provided a show in the hidden field, use that
+        } else if ($('input:hidden[name=whichSeries]').length > 0 && $('input:hidden[name=whichSeries]').val().length > 0) { // if we provided a show in the hidden field, use that
```
OmgImAlexis: This should be `!== 0`.

static/js/add-shows/new-show.js
```diff
@@ -5,9 +5,9 @@ MEDUSA.addShows.newShow = function() {
         var showName;
         var sepChar;
         // if they've picked a radio button then use that
-        if ($('input:radio[name=whichSeries]:checked').length) {
+        if ($('input:radio[name=whichSeries]:checked').length > 0) {
```
OmgImAlexis: This should be `!== 0`.

static/js/add-shows/add-existing-show.js
```diff
@@ -27,7 +27,7 @@ MEDUSA.addShows.addExistingShow = function() {
         var url = '';
         $('.dir_check').each(function(i, w) {
             if ($(w).is(':checked')) {
-                if (url.length) {
+                if (url.length > 0) {
```
OmgImAlexis: This should be `!== 0`.

p0psicles:
What do you want me to check on this?

OmgImAlexis:
Just to make sure I didn't flip any of the if checks since I changed them from `!var.length` and I have a feeling a few of them are now double negative with things like `!var.length !== 0` instead of `!var.length !==0`.

fernandog:
@p0psicles @OmgImAlexis still need review?

OmgImAlexis:
Merge it please. I can't see anything wrong with it at all.

static/js/add-shows/new-show.js
```diff
@@ -5,9 +5,9 @@ MEDUSA.addShows.newShow = function() {
         var showName;
         var sepChar;
         // if they've picked a radio button then use that
-        if ($('input:radio[name=whichSeries]:checked').length) {
+        if ($('input:radio[name=whichSeries]:checked').length !== 0) {
```
fernandog: @OmgImAlexis why use strict comparison (not equal value or not equal type) when `.length` will always return INT ? can't we use just `!=` ?

is there any issue using existing code?

static/js/add-shows/new-show.js
fernandog: @OmgImAlexis why checking > 0 here and not "!=" too?   can't we always use ".length > 0" instead of != or !== ?

p0psicles: I think it's more what we agree on as being a standard. opposed to using 5 different ways of checking the length of an array.

static/js/parsers.js
```diff
@@ -18,18 +18,18 @@ $.tablesorter.addParser({
     },
     format: function(s) {
         var replacements = {
-            'custom': 11,
-            'bluray': 10, // Custom: Only bluray
-            'hd1080p': 9,
+            custom: 11,
+            bluray: 10, // Custom: Only bluray
+            hd1080p: 9,
```
fernandog: @OmgImAlexis shouldn't the key be a string and not a var?

OmgImAlexis: In JSON the left is a string even if it's not quoted so this is fine.

static/js/add-shows/new-show.js
OmgImAlexis: Exactly. I'd like to also move everything we have to a strict set of standards so it's always understandable and then write a doc staring out standards. This will also allow me to write a travis check with xo to only accept certain checks like this. 

If everything is always the same it's much easier to know why and where something breaks.

fernandog: Ok thanks!

PR #1311, opened 2016-10-24
Updated simpleanidb to version 0.1.0

p0psicles:

The publich method refresh was added, which can be used to periodically refresh the cached anime-list.xml.


p0psicles:
I'm going to try to implement the dogpile caching that's also implemented for our subliminal.
But I don't know how long that works, and in the mean time anidb recommended lists is not working. So I propose to let it update the cache, with every showUpdater run. And note that as technical debt. Then when the dogpile caching has been implemented, it can be removed from the showUpdater.

Issue #1309, opened 2016-10-23
Medusa doesn't delete files when PP better quality


NicoLeOca:
Branch/Commit: develop last commit
OS: windows 10 & Python 2.7
What you did: as I moved from having Medusa on Synology to Windows, I added Arrested development there. Medusa detects the 720p bluray files and starts to search for better quality, finds1080p bluray, downloads it and postprocesses it.
after postprocess finish, both 720p and 1080p remains present in the appropriate folder.

I expected Medusa to delete the 720p file

Logs:
https://gist.github.com/NicoLeOca/e344ca45ac59642132d8e59581e5d764

@medariox

medariox:
Thanks for opening the issue.
If I understand this correctly, Medusa is not replacing the from files of a lower quality with those of a higher quality in the Series folder. Is that correct?

In this case:
```arrested.development.s01e05.1080p.bluray.x264-bia.mkv```
should replace
```Arrested.Development.S01E05.720p.BluRay.X264-REWARD```
but it doesn't. So you end up having both in your series folder.

Please let me know if I understand this correctly before continuing.

NicoLeOca:
@medariox 
you got it right

medariox:
Fixed with: https://github.com/pymedusa/Medusa/pull/1312

Thanks again for the report.

NicoLeOca:
You got it

Le 23 oct. 2016 ? 22:32, Dario <notifications@github.com<mailto:notifications@github.com>> a ?crit :


Thanks for opening the issue.
If I understand this correctly, Medusa is not replacing the from files of a lower quality with those of a higher quality in the Series folder. Is that correct?

In this case:

should replace
```Arrested.Development.S01E05.720p.BluRay.X264-REWARD```
but it doesn't. So you end up having both in your series folder.

Please let me know if I understand this correctly before continuing.



-

NicoLeOca:
@medariox 
the fix isn't fully accurate, I forced the PP and watched it live.
![image](https://cloud.githubusercontent.com/assets/9487685/19700992/cfe895e4-9af9-11e6-8866-a553d2601a37.png)

https://gist.github.com/NicoLeOca/9a18d8df5bbf00116a0b2b2b9a63e03d

medariox:
You had the new versions snatched before the fix? The fix will only work for new snatches. Give it some days and I'm sure you won't have the issue anymore.

NicoLeOca:
@medariox 
[...]

PR #1312, opened 2016-10-24
Reset location is causing not replacing file issue

fernandog:
@medariox


fernandog:
https://github.com/pymedusa/Medusa/issues/1309

fernandog:
@medariox worked

PR #1317, opened 2016-10-24
Raise InvalidNameException for special episodes

fernandog:
@ratoaq2


duramato:
Breaks current functionality

PR #1322, opened 2016-10-25
Add DBDEBUG to UI - only if developer

fernandog:
Log is reconfigured on the fly with @ratoaq2 code.


fernandog:
@medariox approve?

Issue #1323, opened 2016-10-25
Proper not being downloaded over non-proper


NicoLeOca:
Branch/Commit: last develop
OS: windows
What you did: I added an existing folder of Black Mirror containing S03E04 released by SKGTV
![image](https://cloud.githubusercontent.com/assets/9487685/19699718/b92b7a42-9af4-11e6-8410-57939497ca01.png)
file is properly recognized and Medusa search as quality is only allowed

here is the search:
![image](https://cloud.githubusercontent.com/assets/9487685/19699816/14b05f2c-9af5-11e6-881e-e68b01ce673c.png)

there is a proper release which isn't downloaded, why?

FYI: a backlog search was forced on the date/time you see above.
and logs are from that time, as I don't have previous ones anymore.

Logs:
https://gist.github.com/NicoLeOca/65beb24b9efe05b2d7bcce1991132cc5

medariox:
That log doesn't help much. Force "Find Propers Search:" instead and post the logs right after it has finished.

NicoLeOca:
@medariox 
here you are:
https://gist.github.com/NicoLeOca/56dbd746c0d0d65e5eafa528e9f86740

medariox:
The proper search is not searching for Black Mirror at all. Propers are only being searched for 3(?) days after non-proper download or something similar. @fernandog can be more precise here. Your only way to get those propers would be manual download.

fernandog:
2 days after airdate

NicoLeOca:
why only 2 days after air-date? that seems very short to me.

HDTV are regularly available less than 24h after air date
webdl, about 3 days
Bluray a couple of months

fernandog:
@NicoLeOca proper search keep searching for propers for 2 days after airdate

so if ep airs today at 21:00, it will keep searching for propers until (today + 2 days) at 21:00

NicoLeOca:
@fernandog 
that's what I understood.
and I keep thinking it should be longer than 2 days, maybe a week
at least because sometimes even  initial HDTV release are not posted within 2 days following air.
and by definition a proper can't appear before an initial release.

as for myself, I don't download HDTV, only web-dl or bluray.
it is very rare that the initial web-dl release appear within 2 days after airdate; it's usually 3 to 4.
and web-dl can have propers as well.
Black Mirror above is a good example.
SKGTV released 4 days after the airdate and Deflate posted a proper 6-7 days after airdate.

duramato:
[...]

PR #1326, opened 2016-10-25
Use custom proper search days

fernandog:
@p0psicles this is my idea.

### DIFF
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -67,6 +67,8 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
             current_processed_propers = self.processed_propers
             self.processed_propers = []
 
+        logger.log('Using proper search days: {0}'.format(app.PROPERS_SEARCH_DAYS))
+
         propers = self._get_proper_results()
 
         if propers:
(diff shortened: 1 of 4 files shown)


medusa/search/proper.py
```diff
@@ -66,6 +67,8 @@ def run(self, force=False):  # pylint: disable=unused-argument
         if force:
```
p0psicles: So it will only apply the custom_search_days, when the 'force' applies? Meaning, you will need to forcibly search for a proper? That kind of defeats the purpose, as you never know when you need to search for a proper right? Also as i'm using nzb, propers can arrive later then at a good torrent provider. Meaning, I might want to extend the default search period for propers?

fernandog:
https://github.com/pymedusa/Medusa/issues/1323

medusa/search/proper.py
duramato: Using nzb doesn't influence that, but I agree with you propers download should and are something automated

medusa/search/proper.py
```diff
@@ -66,6 +66,8 @@ def run(self, force=False):  # pylint: disable=unused-argument
...
             current_processed_propers = self.processed_propers
             self.processed_propers = []
+            self.custom_search_days = True
```
p0psicles: Still don't understand the purpose of this?

PR #1331, opened 2016-10-26
Use torrasave.top instead of vr3mrzpy4ugpt2wi.onion.cab

medariox:


fernandog:
are you sure there is no captcha or JS blocking?
we disabled torcache in the past because of JS blocking

medariox:
@fernandog 
Tested and working just fine.

duramato:
Why the http link? No ssl version of the site? Same question goes for the existing ones too

medariox:
@duramato 
torrasave.top has an invalid certificate. Not sure about itorrents. Anyway, in this case the use of TLS is useless as we aren't transmitting any sensitive information.

duramato:
It's always better to use ssl :P

PR #1334, opened 2016-10-26
Check free space before running Medusa

fernandog:
@medariox as you suggested

### DIFF
--- a/start.py
+++ b/start.py
@@ -197,6 +197,17 @@ class Application(object):
             sys.exit('Sorry, you MUST add the Medusa folder to the PYTHONPATH environment variable\n'
                      'or find another way to force Python to use %s for string encoding.' % app.SYS_ENCODING)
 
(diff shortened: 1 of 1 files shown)


ratoaq2:
Don't think this is a fix. Every time I start the application it will create and delete a 10mb file. We should just let the ouf of disk space occur. If you didn't restart the application for some time and you have low disk space this "fix" will not prevent anything

fernandog:
No disk space will cause mako issues and all sort of errors. App cant continue to work

duramato:
This prevents it from running in an unusable environment, running out of disk mid operation is another story imo

medariox:
As discussed, we need a more reliable solution.

OmgImAlexis:
Can we not check if the free space is larger than 100mb otherwise don't start? I know *NIX OSs have a cli until to check that and so does Windows.

PR #1361, opened 2016-10-28
Refactor medusa/*.py files

labrys:
- [X] Read [contribution guide](https://github.com/PyMedusa/SickRage/blob/master/.github/CONTRIBUTING.md)


medusa/exception_handler.py
```diff
@@ -17,9 +17,9 @@ def handle(thread_name, err):
     """
     if isinstance(err, OSError):
         if err.errno == 28:
-            logger.warning('Out of disk space: {error_msg}', error_msg=err.message)
+            logger.warning('Out of disk space: {error_msg}', error_msg=err)
```
fernandog: @ratoaq2 fyi

medusa/history.py
```diff
@@ -22,7 +22,7 @@
 import db
 from .common import FAILED, Quality, SNATCHED, SUBTITLED
 from .helper.encoding import ss
-from .show.History import History
+from .show.history import History
```
fernandog: @labrys i would import as `show_history` to not confuse with history.py. maybe also rename show/show_history.py

labrys: `from .show.history` already differentiates between `show/history.py` and `history.py`.  If `history.py` also had a `class History:` then we could import it as ShowHistory to differentiate, but thats not the case here.

labrys:
Closing this as #1402 makes the conflict resolution more of a pain than just redoing the refactors.

PR #1391, opened 2016-10-30
Fix pymedusa/Medusa/issues/1387

duramato:

@medariox

### DIFF
--- a/medusa/providers/torrent/html/morethantv.py
+++ b/medusa/providers/torrent/html/morethantv.py
@@ -87,10 +87,10 @@ class MoreThanTVProvider(TorrentProvider):  # pylint: disable=too-many-instance-
 
         for mode in search_strings:
             logger.log('Search mode: {0}'.format(mode), logger.DEBUG)
-            
+
             if mode == 'Season':
-                search_strings[mode].append(re.sub(r'(.*)S0?', r'\1Season ', search_strings[mode]))
-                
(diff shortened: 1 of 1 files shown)


medariox:
Why append and Regex? I'd go for replace

duramato:
My tought here we need "Season 2" and "S02" becouse the 2 are avaiable on the site, so we need to search for the 2 formats becouse we don't know when there will be one or the other.

fernandog:
@medariox has to be re.sub : 

{u'Season': [u'Arrow S04']}

fernandog:
i have a better solution. testing now

medusa/providers/torrent/html/morethantv.py
```diff
@@ -88,6 +88,9 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
         for mode in search_strings:
             logger.log('Search mode: {0}'.format(mode), logger.DEBUG)
 
+            if mode == 'Season':
+                search_strings[mode].append(re.sub(r'(.*)S0?', r'\1Season ', search_strings[mode][0]))
```
fernandog: problem of this is that if there is scene exceptions it will use only the first one.

fernandog:
@medariox are you going to try parse season in this PR? 

@duramato can you test this?

fernandog:
rebased

PR #1412, opened 2016-10-31
Improved check for broken providers

fernandog:
Update only if last update wasn't done in last hour
@medariox @duramato


fernandog:
@duramato keep in mind that if config will still be saved multiple times as broken providers update was only a synthom

PR #1441, opened 2016-11-01
Upgrade certifi to 2016.9.26

fernandog:
@ratoaq2 @medariox


fernandog:
approve ?

ratoaq2:
Let's have ssl working before introducing a new change that might influence it

PR #1457, opened 2016-11-03
Fix season update

fernandog:
We are loading all eps from db:
```python
         # get episode list from DB
        DBEpList = self.show.load_episodes_from_db()
```
and for indexer we are loading only outdated season(s): 
```python
IndexerEpList = self.show.load_episodes_from_indexer(self.seasons)
```

@p0psicles @medariox 
My idea is to pass "season" to `self.show.load_episodes_from_db()` and only load episodes from that season(s)

Feel free to commit to this PR


fernandog:
will keep the branch if want to check code

PR #1464, opened 2016-11-04
Change displayShow invalid Timeformat message

fernandog:
Related:

pasted image at 2016_11_04 13_51

### DIFF
--- a/views/displayShow.mako
+++ b/views/displayShow.mako
@@ -162,11 +162,11 @@
                 % endif
                 % endif
                 % if show.network and show.airs:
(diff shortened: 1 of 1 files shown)


p0psicles:
is this still an issue? thought I resolved that?

fernandog:
If airs = '' the convertion will still fail in mako

ratoaq2:
So, it's only empty that needs to be handled, right? The new message is not the best one. Probably we should only leave the ui field empty

fernandog:
But if format is wrong users should know and edit info in tvdb

p0psicles:
no because in mako '' == false.
It will just not display anything.

So if airs is '' and network is '', it will show nothing.

fernandog:
Beetle Bailey
Chuck Norris: Karate Kommandos
Fanthomas <- this is a norwegian web series so not really a network noted on thetvdb

PR #1465, opened 2016-11-04
Fix missing network when value is empty

fernandog:
default value is '':
https://github.com/pymedusa/Medusa/blob/develop/medusa/tv.py#L1101

Should we change default to None?

@medariox @p0psicles

### DIFF
--- a/medusa/network_timezones.py
+++ b/medusa/network_timezones.py
@@ -116,7 +116,7 @@ def get_network_timezone(network, _network_dict):
     # Get the name of the networks timezone from _network_dict
     network_tz_name = _network_dict[network] if network in _network_dict else None
 
-    if network_tz_name is None and network not in missing_network_timezones:
+    if network_tz_name is None and network != '' and network not in missing_network_timezones:
         missing_network_timezones.add(network)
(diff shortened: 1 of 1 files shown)


medusa/network_timezones.py
```diff
@@ -116,7 +116,7 @@ def get_network_timezone(network, _network_dict):
     # Get the name of the networks timezone from _network_dict
     network_tz_name = _network_dict[network] if network in _network_dict else None
 
-    if network_tz_name is None and network not in missing_network_timezones:
+    if network_tz_name is None and network != '' and network not in missing_network_timezones:
```
ratoaq2: Why not? - >

```
    if not network_tz_name and network and network not in missing_network_timezones:
```

fernandog: @ratoaq2 default value for network is ''. imo it should be None so we don't need to change this code

medariox: @fernandog https://docs.python.org/2/library/stdtypes.html#truth-value-testing

fernandog:
@medariox approve?

PR #1466, opened 2016-11-04
Query series images with accept_languages to be consistent and get th…

medariox:

…e expected results

### DIFF
--- a/medusa/indexers/tvdbv2/tvdbv2_api.py
+++ b/medusa/indexers/tvdbv2/tvdbv2_api.py
@@ -312,16 +312,16 @@ class TVDBv2(BaseIndexer):
 
         This interface will be improved in future versions.
         """
-        key_mapping = {'file_name': 'bannerpath', 'language_id': 'language', 'key_type': 'bannertype', 'resolution': 'bannertype2',
-                       'ratings_info': {'count': 'ratingcount', 'average': 'rating'}, 'thumbnail': 'thumbnailpath', 'sub_key': 'sub_key', 'id': 'id'}
+        key_mapping = {'file_name': 'bannerpath', 'language_id': 'language', 'key_type': 'bannertype',
+                       'resolution': 'bannertype2', 'ratings_info': {'count': 'ratingcount', 'average': 'rating'},
+                       'thumbnail': 'thumbnailpath', 'sub_key': 'sub_key', 'id': 'id'}
 
         search_for_image_type = self.config['image_type']
 
         log().debug('Getting show banners for %s', sid)
         _images = {}
 
-        # Let's fget the different type of images available for this series
-
+        # Let's get the different types of images available for this series
         try:
             series_images_count = self.series_api.series_id_images_get(sid, accept_language=self.config['language'])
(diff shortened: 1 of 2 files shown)


medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -340,7 +341,8 @@ def _parse_images(self, sid):
                 if image_type not in _images:
                     _images[image_type] = {}
 
-                images = self.series_api.series_id_images_query_get(sid, key_type=image_type)
+                images = self.series_api.series_id_images_query_get(sid, key_type=image_type,
+                                                                    accept_language=self.config['language'])
```
medariox: This is the meat of the PR. I added an Italian show and got only some images, with the language added I get the expected images (all images available on thetvdb).

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -331,7 +331,8 @@ def _parse_images(self, sid):
         for image_type, image_count in self._object_to_dict(series_images_count).iteritems():
             try:
                 if search_for_image_type and search_for_image_type != image_type:
-                    if image_type == 'poster' and search_for_image_type != 'poster_thumb':
+                    # We want to use the 'poster' image also for the 'poster_thumb' type
+                    if image_type != 'poster' or image_type == 'poster' and search_for_image_type != 'poster_thumb':
```
medariox: This logic was erroneous, my fault, sorry.

medusa/indexers/tvdbv2/tvdbv2_api.py
fernandog: what if `images` returns Italian poster and an English poster? Which will be used? shouldn't we put meta data language first?

fernandog: @medariox ". I added an Italian show " - meta data language is Italian.

so are you 100% sure that if `images` has Italian poster and English poster, medusa will use the Italian (if exists) over English

medariox: It will only return images for the requested language.

p0psicles:
Just to share a little how the image parsing is working.
This is where the images are retrieved from tvdb:
https://github.com/pymedusa/Medusa/blob/d2f03ff66da41b37b80a827301f530adca71b383/medusa/indexers/tvdbv2/tvdbv2_api.py#L347

Bur we save a specific image from this large list, based on how it's rated.
I thought of it as a generic method. But maybe it's tvdb specific and needs to be moved. Anyway, maybe this is causing some issues?
https://github.com/pymedusa/Medusa/blob/d2f03ff66da41b37b80a827301f530adca71b383/medusa/indexers/indexer_base.py#L209

medariox:
@p0psicles 
Yes, but that happens after series_api.series_id_images_query_get() and I didn't notice any issue with it so far. Anyway, the issue here is that we didn't get any results from series_api.series_id_images_query_get() in the first place.

PR #1477, opened 2016-11-05
tvdb api v2 exception handling and cleanups

p0psicles:

I've used the PR to cleanup the exceptions, as allot of them still had the snake_case format. I've also added better exception handling for handling exceptions raised by the tvdb api.

The LanguageNotFound exception is not used here, but will in other PR's, for handling switching language for existing shows.


p0psicles:
@duramato can you share me your user experience?

duramato:
Will test tonight

duramato:
Sorry, but i don't find this much better that what it is atm
Take the german show `Knallerfrauen` for example, 
Searching with Portuguese for the meta:
`No results found in indexer TVDBv2 using language pt. Try changing language or indexer.`
image
Searching with English for the meta:
`No results found, try a different search.`
image
Searching with German for the meta:
It's the show's language it finds results.
image

English shows there are no results , while Portuguese recomends to try another language. How is one different than the other? The show's language is German

I think if this is going to be the de facto solution, i would just show the `No results found in indexer TVDBv2 using language **xyz**. Try changing language or indexer.` always...

p0psicles:
Then we close the PR, and leave it like it is?

p0psicles:
Can anyone do a quick review on this, and approve please?

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -139,8 +141,12 @@ def _show_search(self, show, request_language='en'):
         """
         try:
             results = self.search_api.search_series_get(name=show, accept_language=request_language)
+        except ApiException as e:
+            raise IndexerShowNotFound(
+                'Show search failed in getting a result with reason: %s (%s)' % (e.reason, e.status)
+            )
         except Exception as e:
-            raise IndexerException('Show search failed in getting a result with error: %r', e)
+            raise IndexerException('Show search failed in getting a result with error: %r' % e)
```
fernandog: @p0psicles can't you catch `IndexerException` instead of `exception` ?

p0psicles: I think it's still possible to get exceptions raised that are not catched with IndexerException. Like SSL or other connection errors. So at this stage, i'd like to normalize it, and use the IndexerExceptions.

labrys: Just to clarify you are raising an exception that serves to do nothing but say that there was an exception and reraising a different exception.  This swallows important information (such as the traceback) and provides no additional information to the exception chain.  If it provided additional details to make the exception clearer, or if it raised a saner exception type it might be useful.  But raising a known exception type because you don't know what type of exception you might get is generally not a good practice.

Some examples:
from requests
```python
            try:
                connect, read = timeout
                timeout = TimeoutSauce(connect=connect, read=read)
            except ValueError as e:
                # this may raise a string formatting error.
                err = ("Invalid timeout {0}. Pass a (connect, read) "
                       "timeout tuple, or a single float to set "
                       "both timeouts to the same value".format(timeout))
                raise ValueError(err)
```
Here they are catching a ValueError, and yet reraising another ValueError because it provides context to the ValueError about what it actually means, and a traceback is unnecessary because there is nothing higher up the chain to debug and modify.

from our numdict.py:
```python
        try:
            return int(key)
        except (TypeError, ValueError):
            if key is not None:
                raise KeyError(key)
```
Since numdict acts like a dict, an invalid number raising a TypeError goes against expectations, thus we raise a KeyError instead.

PR #1480, opened 2016-11-05
Improve existing about about replace qualities and fix not downloading after quality change. 

fernandog:
Improve existing code:

```
if ep_status in Quality.DOWNLOADED + Quality.SNATCHED + Quality.SNATCHED_PROPER and \
        new_quality in preferred_qualities and (new_quality > cur_quality or cur_quality not in preferred_qualities):
```
to a method like @ratoaq2 did

### DIFF
--- a/medusa/search/backlog.py
+++ b/medusa/search/backlog.py
@@ -158,9 +158,11 @@ class BacklogSearcher(object):
                 continue
 
             if cur_status != common.WANTED:
+                # If preferred and not best, keep redonwloading
                 if preferred_qualities:
-                    if cur_quality in preferred_qualities:
+                    if cur_quality == max(preferred_qualities):
                         continue
+                # If allowed only, any is good
                 elif cur_quality in allowed_qualities:
                     continue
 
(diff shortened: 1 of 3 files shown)


fernandog:
@pymedusa/developers 
found this code:
https://github.com/pymedusa/Medusa/blob/master/medusa/search/backlog.py#L161-L166

which goes against the code in tv.py
This code makes the preferred to stop at first download
```
            if cur_status != common.WANTED:
                if preferred_qualities:
                    if cur_quality in preferred_qualities:
                        continue
                elif cur_quality in allowed_qualities:
                    continue
```

The tv.py code is used only while daily searching and found another episode. So it would replace it.
In backlog it wouldn't replace because it wouldn't search for it.

So we have different behavior depending on the thread

fernandog:
@pymedusa/developers now all threads seems consistant:

## Allowed only (downloads any and stop):
allowed

## Allowed + Preferred (download any from allowed (which appears first) but stops when finds the best in preferred)
allowed preferred

## Preferred only (keep downloading until reaches best)
preferred

p0psicles:
If we're changing behavior. I expect also a proper update of the wiki, and a proper explanation in the quality selector. That counts as 50% for this pr. As is discussed multiple times, the problem is not solely in code.

PR #1509, opened 2016-11-07
Bring sanity to list_associated_files

medariox:

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -165,48 +165,37 @@ class PostProcessor(object):
         :param subfolders: check subfolders while listing files
         :return: A list containing all files which are associated to the given file
         """
-        def recursive_glob(treeroot, pattern):
-            results = []
-            for base, _, files in os.walk(treeroot):
-                goodfiles = fnmatch.filter(files, pattern)
-                results.extend(os.path.join(base, f) for f in goodfiles)
-            return results
+        def find_files(file_path, pattern='*', subfolders=None, base_name_only=None):
+            directory = os.path.dirname(file_path)
+            if base_name_only:
+                pattern = os.path.basename(file_path).rpartition('.')[0] + pattern
+
+            found_files = []
+            for root, _, filenames in os.walk(directory):
+                for filename in fnmatch.filter(filenames, pattern):
+                    found_files.append(os.path.join(root, filename))
+                if not subfolders:
+                    break
+            return found_files
 
         if not file_path:
             return []
 
-        # don't confuse glob with chars we didn't mean to use
-        globbable_file_path = helpers.fixGlob(file_path)
-
         file_path_list = []
-
         extensions_to_delete = []
+        base_name = os.path.basename(file_path)
 
         if subfolders:
-            base_name = os.path.basename(globbable_file_path).rpartition('.')[0]
-        else:
-            base_name = globbable_file_path.rpartition('.')[0]
-
-        if not base_name_only:
-            base_name += '.'
-
-        # don't strip it all and use cwd by accident
-        if not base_name:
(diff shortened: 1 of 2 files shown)


medusa/post_processor.py
```diff
@@ -209,14 +198,14 @@ def recursive_glob(treeroot, pattern):
                     new_file_name = file_name
                     sub_file_name = file_name.rpartition('.')[0]
 
-                if is_subtitle and sub_file_name.lower() == base_name.lower().replace('[[]', '[').replace('[]]', ']'):
+                if is_subtitle and sub_file_name.lower() == base_name.lower():
```
fernandog: @medariox  need to find why this was added (blame) and if we can safely remove now

medariox: It's only if glob is used and even then it's not needed in this case.

medusa/post_processor.py
```diff
@@ -17,20 +17,19 @@
...
 
 import adba
 
-from babelfish import language_converters
-
 import medusa as app
 
 from six import text_type
 
+from subtitles import from_code, from_ietf_code
```
ratoaq2: It should be
`from .subtitles`

medusa/post_processor.py
```diff
@@ -148,85 +147,54 @@ def list_associated_files(self, file_path, base_name_only=False,
...
 
-            # loop through all the files in the folder, and check if they are the same name
-            # even when the cases don't match
-            for filefound in checklist:
+            file_name = os.path.basename(found_file).lower()
 
-                file_name = filefound.rpartition('.')[0]
-                file_extension = filefound.rpartition('.')[2]
-                is_subtitle = None
+            if file_name[-3:] in subtitle_extensions:
```
ratoaq2: Shouldn't it be everything after the last dot?

medusa/post_processor.py
```diff
@@ -148,85 +147,54 @@ def list_associated_files(self, file_path, base_name_only=False,
...
         for associated_file_path in filelist:
             # Exclude the video file we are post-processing
             if associated_file_path == file_path:
                 continue
 
-            # Exlude non-subtitle files with the 'only subtitles' option (not implemented yet)
-            # if subtitles_only and not associated_file_path[-3:] in subtitle_extensions:
-            #     continue
+            # Exlude non-subtitle files with the 'only subtitles' option
+            if subtitles_only and not associated_file_path[-3:] in subtitle_extensions:
```
ratoaq2: Same as before. Probably we should have a method: is_subtitle, like the one: isMediaFile

medariox:
Thanks for the review @ratoaq2.
I agree that [-3:] is not a reliable way to get the subtitle extension, it should always be .rpartition('.')[2] instead. Was thinking about implementing is_subtitle, as we have also other atrocities further down the file, that could surely benefit from it.

There are a few questions that I'd like to discuss before adding the method:
1) What should be supported? E.g. should just passing the extension 'srt' or even '.srt' be able to determine if it's a subtitle as well or just file_name and full_path?
2) What about language subs? Should we differentiate about the two types (maybe optionally) or should we just check the extension?

I'd like some feedback about this before continuing.

ratoaq2:
Imo we should have like a fileutils that works on filepaths:
- get_extension(filepath): return 'abc' 
- is_subtitle(filepath) : return True
- get_subtitle_language(filepath): return Language('en')

medariox:
@ratoaq2 
Is this more or less what you were thinking of? https://github.com/pymedusa/Medusa/pull/1509/commits/9022783283278a11c5b663e0e2381cbb3ba8034b

ratoaq2:
Exactly

PR #1511, opened 2016-11-08
Feature/indexer/tmdb

p0psicles:
Added tmdb indexer.

Main reason for now, is to see what's still required to add additional indexers. Like for example, how usable are indexers that can't make use of the scene name exceptions. Or other services which require a thetvdb id.

A followup will be adding the tvmaze indexer. As is waiting in #145 

### Log:

This is a list to document what needs to be taken care of when implementing a new indexer. I'll update this as I move along.
- [x] Show search by name
- [x] Show search by id, indexing the show attributes, seasons, episodes, banner, etc. `Is working, except for the banner.`
- [x] Episode titles / summaries (episode titles not yet visible, small issue)
- [x] Poster and Banners are downloaded. But for now i'm using the 'original' quality. Meaning in some cases, very large images. Maybe we should decide on a different image size.
- [x] Add metadata creation
- [x] Add banner and poster support if available
- [x] Add actors information (Used for metadata creation)
- [x] Add multilingual support, (seach shows, get show data, get episode data and get images in the locale configured).
- [x] Add timezone and network information
- [ ] Add indexer mappings (when indexer providers indexer ids to other indexers)
- [x] Add the indexer icon
- [x] Implement a new method for show updating, as we cannot use the old tvdb xml's anymore. (well really need to test this.
- [ ] TMDB gives back a list with supported languages. Maybe we can utilize that.
- [ ] Cleanup

### Tvdb Apiv2 Lib related
- [x] Make sure the api returns unicode. Test by searching: 'guide to pretty'. Currently i've adjusted the mapping for the Series modal. Need to also check for others, like episode, actors, etc.
- [x] Pass the indexer requests object. This will improve caching.
- [x] Add searching with apostophes '.
### showUpdates related

More info: https://github.com/pymedusa/Medusa/wiki/Show-indexer-updating
For this the following needs to be implemented:
- [x] Support for force full update
- [x] ~~Update only seasons that reached their next_update.~~ Better yet, update seasons that have had an update since last show_update.
- [x] Add new shows, should set the next_updates
- [x] Episodes that start a new season, should update the cache indexer_update table with the next_update.

### Known isuses:
- [ ] For some reason it's not always downloading/updating the poster.
- [ ] TMDB doesn't provide banners. So we need to come up with a backup plan for this.
- [ ] Show updating hasn't been fully tested. I'm just gonna run it for a few days.

### Screenies


duramato:
```. Like for example, how usable are indexers that can't make use of the scene name exceptions``` you mean xem screen numbering? Becouse our scene exptions we have control over them

File: medusa/indexers/indexer_config.py [hunk 1]
```diff
@@ -59,5 +61,23 @@
...
+            'use_zip': True,
+            'session': helpers.make_session(cache_etags=False),
+        },
+        'trakt_id': 'tvdb_id',
+        'xem_origin': 'tvdb',
+        'icon': 'tmdb16.png',
+        'scene_loc': 'https://cdn.pymedusa.com/scene_exceptions/scene_exceptions.json',
+        'base_url': 'https://www.themoviedb.org',
+        'show_url': 'https://www.themoviedb.org/tv/',
+        'mapped_to': 'tvdbid'  # The attribute to which other indexers can map there thetvdb id to
```
duramato: *their

p0psicles:
Continuing in #1534

PR #1518, opened 2016-11-09
Fixed: For some shows there isn't an 'en' language available. But a result is provide, without the showname.

p0psicles:

Skipping the show when it doesn't have a result in the requested language.
your can reproduce the issue by searching for: '24 h berlin' in english.

### DIFF
--- a/medusa/indexers/tvdbv2/tvdbv2_api.py
+++ b/medusa/indexers/tvdbv2/tvdbv2_api.py
@@ -167,11 +167,7 @@ class TVDBv2(BaseIndexer):
 
         # Remove results with an empty series_name.
         # Skip shows when they do not have a series_name in the searched language. example: '24 h berlin' in 'en'
-        cleaned_results = []
-        for show in mapped_results:
-            if not show.get('seriesname'):
-                continue
-            cleaned_results.append(show)
+        cleaned_results = [show for show in mapped_results if show.get('seriesname')]
 
         return OrderedDict({'series': cleaned_results})['series']
 
--- a/medusa/indexers/tvdbv2/tvdbv2_api.py
+++ b/medusa/indexers/tvdbv2/tvdbv2_api.py
@@ -163,8 +163,20 @@ class TVDBv2(BaseIndexer):
             return
 
         mapped_results = self._object_to_dict(results, self.series_map, '|')
+        mapped_results = [mapped_results] if not isinstance(mapped_results, list) else mapped_results
 
-        return OrderedDict({'series': mapped_results})['series']
+        # Remove results with an empty series_name.
+        # Skip shows without configured series_name. example: '24 h berlin'
+        cleaned_results = []
+        for show in mapped_results:
+            if not show.get('seriesname', None):
+                if show.get('aliases', '') and len(show.get('aliases').split('|')):
+                    show['seriesname'] = show.get('aliases').split('|')[0]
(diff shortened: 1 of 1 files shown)


fernandog:
@p0psicles how that will impact updating exiting shows that doesn't have seriesname attribute (the issue we discussed some days ago) ?

p0psicles:
Hmm, your right. I think we should bann shows, without series name from medusa. Or at least when we get a result back, without series_name for a particular language. In that case, i only have to make a small adjustment.

fernandog:
@p0sicles i already said one solution. just catch this error in showupdate and show as WARNING to user asking to change language metadata languaga in editshow!

p0psicles:
This issue has nothing to do with language selection.

labrys:
It is a language selection issue, both for searching and for metadata.  

The series referenced in #1518 has a series title in the original language (German) as can be seen by visiting this url: http://www.thetvdb.com/?tab=series&id=114171&lid=14

This can also be tested with the API by passing "de" for "Accept-Language".  Then the result does have the proper `seriesname`.  Changing "Accept-Lanugage" to any other language then returns no `seriesname`.

labrys:
The TVDBv2 lib should support passing multiple languages to it.  Given current implementation of TVDBv2 api that would require one request per language, but that's a limitation of their API.

IMO the lib should support passing in a list of acceptable languages, and return results as a dict of results by language.  Possibly even supporting a cascading function that produces a single result where null values are replaced with updates from other languages.

Sample function:
```python
# coding = utf-8

def pop_nulls(data):
    return {k: v for k, v in data.items() if v}

def cascade_update(data, languages):
    result = {}
    language_priority = languages
    language_priority.reverse()

    for lang in language_priority:
        if lang in data:
            result.update(pop_nulls(data[lang]))
    return result

## CONTRIVED EXAMPLE
test_data = {
    'en': {'series_name': 'Attack on Titan', 'overview': 'An anime about sutff.', 'network': None},
    'ja': {'series_name': '進撃の巨人', 'overview': None, 'network': 'MBS'}
}

print(test_data)
# {'ja': {'series_name': '進撃の巨人', 'overview': None, 'network': 'MBS'}, 'en': {'series_name': 'Attack on Titan', 'overview': 'An anime about sutff.', 'network': None}}
print(cascade_update(test_data, ['fr', 'ja', 'en']))
# {'network': 'MBS', 'overview': 'An anime about sutff.', 'series_name': '進撃の巨人'}
print(cascade_update(test_data, ['en', 'ja', 'fr']))
# {'network': 'MBS', 'overview': 'An anime about sutff.', 'series_name': 'Attack on Titan'}
```

p0psicles:
Yes Labrys your right. Missed that it only had german language. I'm pretty sure i did one search for it.

Only proper solution is to search without language, and get back all shows, with supported languages per show.

Fortunately this is already in the works. (https://trello.com/b/7zl7PE1j/api-2-x-features-and-bug-fixes).
Tvdbv2 will support fallback languages. Meaning we can define a list of languages it will fallback to. Then it will at least always have a result. With that I also hope they will return the language it has fallen back into. As that's also one of the issues with the old api and our implementation.

The second FR, is about getting back a list of supported languages for a show. We could use that, to limit the list with languages you could change the language into.

For now i'll adjust this PR, to at least handle the error. I'll remove the usage of aliases.

PR #1519, opened 2016-11-09
Feature/fix switching tvdb show language

p0psicles:
- [x] Prevent the db from being updated to the new language, if that doesn't have any results.

### DIFF
--- a/medusa/server/web/home/handler.py
+++ b/medusa/server/web/home/handler.py
@@ -1222,6 +1222,27 @@ class Home(WebRoot):
             out.append('S{season}: {names}'.format(season=season, names=', '.join(names)))
         return '<br>'.join(out)
 
+    @staticmethod
+    def check_show_for_language(show_obj, language):
+        """Request the show in a specific language from the indexer.
+
+        If the indexer throws the ShowNotFoundInLanguage Exception, we catch it,
+        and report it's not available in this language.
+        :param show_obj: (TVShow) Show object.
+        :param language: Language passed as string as a two letter country ccde. For ex: 'en'.
+
+        :returns: True (show found in language) False (show not found in language)
+        """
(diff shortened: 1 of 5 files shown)


p0psicles:
@duramato Is this what you had in mind? I'm stopping the show refresh, and giving some feedback through the ui.notifications(). Don't really know, how to handle it any other better way.

medusa/server/web/home/handler.py
```diff
@@ -1222,6 +1222,27 @@ def sceneExceptions(show):
...
+        If the indexer throws the ShowNotFoundInLanguage Exception, we catch it,
+        and report it's not available in this language.
+        :param show_obj: (TVShow) Show object.
+        :param language: Language passed as string as a two letter country ccde. For ex: 'en'.
+
+        :returns: True (show found in language) False (show not found in language)
+        """
+        indexer_api_params = app.indexerApi(show_obj.indexer).api_params.copy()
+        indexer_api_params['language'] = language
+        indexer_api_params['episodes'] = False
```
p0psicles: To give you guys a small update. I've changed the way the episode download is triggered. Previously it was passed as a param, but always passed as True. Meaning for every show search, you had no choice other then to always also download the episodes. Now the param config['episodes_enabled'] can be set, to toggle it.

fernandog:
@p0psicles 

message is that long by design?  `2016-11-11 09:46:30 WARNING  SHOWQUEUE-UPDATE :: [2dd7161] 282670: Data retrieved from TVDBv2 was incomplete. The indexer does not provide show information in the searched language pt. Aborting: Missing attribute series_name, cant index in language: pt` (edited)

i miss the seriesname in the message. ofc i can use the `282670`

p0psicles:
There can be multiple reasons why it had to abort. So some parts are concatenated. Of course we add the show name or indexerid.

PR #1539, opened 2016-11-13
Fix genre pills in displayShow.mako when show can't be mapped to imdb.

p0psicles:

Now I think about it. I think it was like |drama|adventure| before. At least that's the only explanation I can come up with.


p0psicles:
@ratoaq2 I could but then i'm changing it functionally. Now i'm only fixing a possible bug.
In the current solution IMDB genre's have preference over the default indexers genres. That helper function, works exact the other way around.

PR #1554, opened 2016-11-16
Standard logging call should be used in rules.py 

ratoaq2:
- [x] Standard logging call should be used in rules.py since guessit rules are loaded before wrapped logger instance is initialized.

This fixes latest bug in logger call reported in #1093


fernandog:
Approved. 0.1.8 is released
so 0.1.8.1 or 0.1.9 ?

Issue #1555, opened 2016-11-16
Renamer issue?


wimpyrbx:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit: develop
**OS: windows
**What you did: tried to rename files
**What happened: it did not rename files
**What you expected: for it to rename files
**Logs:**
```
2016-11-16 09:38:36 DEBUG    Thread-33 :: [32ca6ae] 314334: Renaming/moving episode from the base path F:\TV\SHOWS\Chappelle's Show\Season 03\Chappelle's.Show.dvdrip.S03E03.mkv to F:\TV\SHOWS\Chappelle's Show\Season 03\Chappelle's Show - S03E03 - Los
t Episode 3 (SD.DVD)
2016-11-16 09:38:36 ERROR    Thread-33 :: [32ca6ae] Exception generated: path is on drive F:, start on drive D:
Traceback (most recent call last):
  File "C:\SickRageMedusa\**********\server\web\core\base.py", line 271, in async_call
    result = function(**kwargs)
  File "C:\SickRageMedusa\**********\server\web\home\handler.py", line 1891, in doRename
    root_ep_obj.rename()
  File "C:\SickRageMedusa\**********\tv.py", line 3191, in rename
    related_files = post_processor.PostProcessor(self.location).list_associated_files(
  File "C:\SickRageMedusa\**********\post_processor.py", line 69, in __init__
    self.rel_path = self._get_rel_path()
  File "C:\SickRageMedusa\**********\post_processor.py", line 109, in _get_rel_path
    rel_path = os.path.relpath(self.file_path, app.TV_DOWNLOAD_DIR)
  File "C:\SickRageMedusa\**********\init\filesystem.py", line 85, in <lambda>
    return lambda *args, **kwargs: handle_output(f(*[handle_arg(arg) for arg in args], **{k: handle_arg(arg) for k, arg in kwargs.items()}))
  File "C:\Python27\lib\ntpath.py", line 529, in relpath
    % (path_prefix, start_prefix))
ValueError: path is on drive F:, start on drive D:
```

wimpyrbx:
show is located at F:\TV\SHOWS\
not sure why it generates exception saying start on D:\

medariox:
Will be fixed with https://github.com/pymedusa/Medusa/pull/1558

PR #1558, opened 2016-11-16
Fix "ValueError: path is on drive F:, start on drive D:" for _get_rel…

medariox:

…_path() #1555

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -114,7 +114,6 @@ class PostProcessor(object):
             except ValueError:
                 pass
 
-        self._log(u"Couldn't get the relative path, using the parent folder instead", logger.DEBUG)
         parent_name = os.path.basename(os.path.dirname(self.file_path))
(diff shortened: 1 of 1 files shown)


medusa/post_processor.py
```diff
@@ -106,10 +106,13 @@ def _get_rel_path(self):
...
-            # check if we really found the relative path
-            if not rel_path.startswith('..'):
-                return rel_path
+            try:
+                rel_path = os.path.relpath(self.file_path, app.TV_DOWNLOAD_DIR)
+                # check if we really found the relative path
+                if not rel_path.startswith('..'):
+                    return rel_path
+            except ValueError:
+                pass
```
p0psicles: Does it make any sense to add an informative or warning log here?

medariox: You mean something like "Couldn't get the relative path, using the parent folder instead" I had in the first commit? We could add it with a bit of refactoring, but don't know how useful that would be - also the error, in this case, is not really something we need to know about in my opinion.

PR #1562, opened 2016-11-16
Fixing circular dependencies + refactoring __init__.py

ratoaq2:

The circular dependencies were not fixed during our rebranding release. This PR will solve that issue. The goal is to not have `import medusa as app` anymore and only use relative imports (that was not possible due to circular dependencies, mainly because __init__.py had all global variables together with several functions that imports several other modules)


fernandog:
approved. looks ok to me

PR #1589, opened 2016-11-18
Use standard logger in post_processor, return full path as fallback i…

medariox:

…n get_rel_path, lint

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -98,10 +98,13 @@ class PostProcessor(object):
         :rtype: text_type
         """
         if app.TV_DOWNLOAD_DIR:
-            rel_path = os.path.relpath(self.file_path, app.TV_DOWNLOAD_DIR)
-            # check if we really found the relative path
-            if not rel_path.startswith('..'):
-                return rel_path
+            try:
+                rel_path = os.path.relpath(self.file_path, app.TV_DOWNLOAD_DIR)
+                # check if we really found the relative path
+                if not rel_path.startswith('..'):
+                    return rel_path
+            except ValueError:
+                pass
 
         logger.debug(u"Couldn't get relative path, using full path instead")
         return self.file_path
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -18,6 +18,7 @@
 
 import fnmatch
 import glob
+import logging
 import os
 import re
 import stat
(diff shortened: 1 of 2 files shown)


medusa/post_processor.py
```diff
@@ -129,32 +122,28 @@ def _checkForExistingFile(self, existing_file):
...
+            logger.debug(u"There is no existing file so there's no worries about replacing it")
             return PostProcessor.DOESNT_EXIST
 
         # if the new file exists, return the appropriate code depending on the size
         if os.path.isfile(existing_file):
-
             # see if it's bigger than our old file
             if os.path.getsize(existing_file) > os.path.getsize(self.file_path):
-                self._log(u'File {0} is larger than {1}'.format(existing_file, self.file_path), logger.DEBUG)
+                logger.debug(u'File {0} is larger than {1}'.format(existing_file, self.file_path))
```
ratoaq2: You can use lazy logging: without formatting the msg:

`logger.debug('My msg {abc}',  abc='def')`

medusa/post_processor.py
```diff
@@ -83,22 +87,10 @@ def __init__(self, file_path, nzb_name=None, process_method=None, is_priority=No
...
-    def _log(self, message, level=logger.INFO):
-        """
-        A wrapper for the internal logger which also keeps track of messages and saves them to a string for later.
-
-        :param message: The string to log (unicode)
-        :param level: The log level to use (optional)
-        """
-        logger.log(message, level)
-        self.log += message + '\n'
-
```
fernandog: @medariox @ratoaq2 without this we can't show the result logs in manual PP page. 

you know it right?

medusa/post_processor.py
medariox: I know, my eyes were bleeding.

medusa/post_processor.py
medariox: It just shows the important stuff now and I would even reduce it further. The rest can be found in the logs.

fernandog: Did you tried with nzbtomedia ? cause it also shows medusa log there from self.log

medariox: I just removed the ones in post_processor, doesn't affect the other self._log. Thinking about it we might want to keep some more messages

medariox:
No time to finish this.

PR #1592, opened 2016-11-18
Skip tvchaosuk highlighted torrents

duramato:

### DIFF
--- a/medusa/providers/torrent/html/tvchaosuk.py
+++ b/medusa/providers/torrent/html/tvchaosuk.py
@@ -136,7 +136,7 @@ class TVChaosUKProvider(TorrentProvider):  # pylint: disable=too-many-instance-a
             for row in torrent_rows[1:]:
                 try:
                     # Skip highlighted torrents
-                    if row['class'] == 'highlight even' or row['class'] == 'highlight odd':
+                    if row.get('class'):
                         continue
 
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/html/tvchaosuk.py
```diff
@@ -135,9 +135,11 @@ def parse(self, data, mode, **kwargs):
             # Skip column headers
             for row in torrent_rows[1:]:
                 try:
+                    # Skip highlighted torrents
+                    if (row['class'] == 'highlight even') or (row['class'] == 'highlight odd'):
```
medariox: You don't need the brackets here.
Are you sure 'class' will always be there? If you are unsure and want to play it safe, use .get('class') instead.

duramato: That class will always be there

fernandog:
@duramato @medariox i think we should discard last commit. is safer to check the class name.

duramato:
Checking class name didn't worked at all. From this pr only that commit works

medariox:
I don't know if there is another way, as I don't have access. Checking for 'class' existence looks kind of excessive to me though.

medariox:
@duramato 
Please always add the milestone in the future.

PR #1595, opened 2016-11-18
add auth handler, fix CORS

OmgImAlexis:
@ratoaq2


fernandog:
did you tested? cause without my commit it wouldn't work

OmgImAlexis:
What wouldn't work?

I had notifications turned off so that's probably why. This also won't be merged until I've added the asset path so it's fine if there's bugs in it.

PR #1596, opened 2016-11-18
Don't run RSS search before backlog search

fernandog:
Backlog = specific search. Doesn't need to run daily. 
Daily runs a lot more frequent than backlog.

@pymedusa/developers 

image

Daily uses another function:
image


duramato:
RSS searches are slow with guessit, more than one gets even slower wich adds up in the end.
Since backlog searches are for searching for the "wanted" episodes , this doesn't break any behaviour and makes backlog search not waist time a get to the "point" of getting the episodes wanted.

This is a much needed change for good.

fernandog:
@ratoaq2 @p0psicles @labrys @OmgImAlexis missing your votes

PR #1613, opened 2016-11-20
Cleanup tv_cache.py + final flake

fernandog:


medusa/providers/torrent/rss/shazbat.py
```diff
@@ -81,9 +81,9 @@ def _getRSSData(self):
             'duration': '2 hours'
         }
 
-        return self.getRSSFeed(self.provider.urls['rss_recent'], params=params)
+        return self.get_rss_feed(self.provider.urls['rss_recent'], params=params)
 
-    def _checkAuth(self, data):
+    def _get_rss_data(self, data):
```
ratoaq2: Is this correct?

fernandog:
yes

PR #1614, opened 2016-11-20
Failed history

labrys:
Disable failed history trim on start (it's not like the file will suddenly be good in 30 days)
Improve failed history trim (allow for variable age trimming of failed history)


medusa/failed_history.py
```diff
@@ -17,62 +17,78 @@
 # You should have received a copy of the GNU General Public License
 # along with Medusa. If not, see <http://www.gnu.org/licenses/>.
 
-import datetime
+from datetime import datetime, timedelta
```
ratoaq2: Here

Issue #1633, opened 2016-11-21
Daily search scene exception lookup


duramato:
**Branch/Commit:** develop
**OS:** Ubuntu 14.04
**What you did:** Let daily run
**What happened:** It does an unecessary number of "Scene exception lookups" for show's i don't even have. For each show that it found and needed to do a "Scene exception lookups" it did it 6 times!
**What you expected:** For it to do only the number of necessary "Scene exception lookups"
**Logs:** https://gist.github.com/duramato/c0aa3d6d324e8e4f3d5b1a5134a045a0

fernandog:
when can't match to a show in the database it does:

```python
        # try scene exceptions
        if not show:
            show_id = scene_exceptions.get_scene_exception_by_name(name)[0]
            if show_id:
                show = Show.find(app.showList, int(show_id))
```

btw i don't have this logs:
```
DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [a737de1] Hit the old cached items, not parsing any more for: rarbg
DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [a737de1] Unable to match Keeping.Up.with.the.Kardashians.S12E21.No.Good.Deeds.1080p.ETV.WEBRip.AAC2.0.x264-BTW[rartv] to a show in your database. Parser result: title: Keeping Up with the Kardashians, season: 12, episode: [21], episode_title: No Good Deeds, screen_size: 1080p, format: WEBRip, audio_codec: AAC, audio_channels: 2.0, video_codec: h264, video_encoder: x264, release_group: BTW, type: episode, absolute_episode: [], quality: 1080p WEB-DL
DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [a737de1] Unable to match Cops.S29E21.1080p.WEB.x264-HEAT[rartv] to a show in your database. Parser result: title: Cops, season: 29, episode: [21], screen_size: 1080p, format: WEB-DL, video_codec: h264, video_encoder: x264, release_group: HEAT, type: episode, absolute_episode: [], quality: 1080p WEB-DL
INFO     SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [a737de1] Removing cache entries older than 3 days from rarbg
```

fernandog:
Looks like it try that when can't find show in the indexer

see if you can find `Unable to match Keijo.S01` this show in your indexers

duramato:
[...]

PR #1634, opened 2016-11-21
Check scene exceptions using season number

fernandog:
Possible fix to https://github.com/pymedusa/Medusa/issues/1633

@duramato

### DIFF
--- a/medusa/helpers.py
+++ b/medusa/helpers.py
@@ -926,7 +926,7 @@ def full_sanitizeSceneName(name):
     return re.sub('[. -]', ' ', sanitizeSceneName(name)).lower().lstrip()
 
 
-def get_show(name, tryIndexers=False):
+def get_show(name, tryIndexers=False, season_number=None):
     from . import classes, name_cache, scene_exceptions
     if not app.showList:
         return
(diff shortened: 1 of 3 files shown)


fernandog:
I also had those:

```
2016-11-21 11:21:41 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [a737de1] Scene exception lookup got indexer ID 95441, using that
2016-11-21 11:21:41 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [a737de1] Scene exception lookup got indexer ID 95441, using that
2016-11-21 11:21:37 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [a737de1] Scene exception lookup got indexer ID 95441, using that
```

fernandog:
@p0psicles can you review this please? i don't know what you did about scene exceptions with multiple indexers and i have some concerns. ping me when you have time

fernandog:
Related to last commit:
image

p0psicles:
I did allot of tests. And I came to this concolusion: passing the season number is wrong. As we shouldn't assume it's the indexer season number being passed. We are getting the scene name from a search result, and we want to map using the scene name to an indexer + (possible) indexer season number. So the functionality was okay like it was.

In fernandog's example using the America Horror Story.
If we would get a result for season 4, it could look like this: American Horror Story Freak Show - S1E13.
I know that for this show, they almost always also provide the correct indexer's season number (in this case 4). But bear with me just to make my point.

In the old sitution, you would get back: indexerid: 250487+ season: 4.
After this PR, you would not get back a result. As season 1, would have been passed in the sql, and it would only compare: 'American horror story murder house'.

PR #1639, opened 2016-11-22
Log wanted episodes in daily

fernandog:
@duramato ok to use `absolute` numbering when air_by_date or anime ?


duramato:
air by date doesn't use absolute at all. It's an anime thing

fernandog:
but if air_by_date we don't have season right?

duramato:
season is the year, "normal" is date

PR #1640, opened 2016-11-22
Upgrade pytimeparse to 1.1.5

fernandog:
@duramato

### DIFF
--- a/lib/pytimeparse/timeparse.py
+++ b/lib/pytimeparse/timeparse.py
@@ -34,8 +34,8 @@ kinds of time expressions.
 import re
 
 SIGN        = r'(?P<sign>[+|-])?'
-YEARS      = r'(?P<years>\d+)\s*(?:ys?|yrs?.?|years?)'
-MONTHS     = r'(?P<months>\d+)\s*(?:mos?.?|mths?.?|months?)'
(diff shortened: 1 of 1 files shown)


duramato:
Did it had local changes of ours? If so we need them. Also the things commented we need them uncommented for our use case.

ratoaq2:
What's the issue? We should try to avoid as much as possible patching our libs. It might be hard to track them when updating

duramato:
Iir I added support for years and months to the library, but as it is a relative time to the calendar, I left it commented to follow the author's "wish" so any one that wants to use year and month parsing the just needs to uncomment the relevant part

fernandog:
This code is in the library. Just upgrading

medariox:
I think we should stay on the latest [Pypi version](https://pypi.python.org/pypi/pytimeparse/1.1.5) to avoid future headaches.

fernandog:
This commit *IS* the 1.1.5

future headaches?
we don't use this lib right now in medusa.
I added this library for a future PR that @duramato will create

fernandog:
tired of wasting time explaining stuff when a simple check can be done

medariox:
Are we adding libs for future PRs now? This lib shouldn't have been added in the first place.

fernandog:
@medariox stop complaining! omg!

I did for @duramato to make his life easier

PR #1661, opened 2016-11-24
Add pubdate to providers

duramato:
- [x] Limetorrents and fix https://github.com/pymedusa/Medusa/issues/1675
- [x] Torrentz2
- [x] Bitsnoop 
- [x] HDSpace
- [x] TorrentProject
- [x] GFTracker
- [x] HoundDawgs
- [x] ExtraTorrent
- [x] Bit-HDTV
- [x] BlueTigers (RIP)
- [x] CPasBien


medusa/providers/torrent/html/limetorrents.py
```diff
@@ -112,14 +115,13 @@ def parse(self, data, mode):
 
         with BS4Parser(data, 'html5lib') as html:
             torrent_table = html('table', class_='table2')
+            # if mode != 'RSS' and torrent_table:
```
duramato: Note to self: Look in to this later

fernandog: @duramato why did you disable?

medusa/providers/torrent/html/gftracker.py
```diff
@@ -175,14 +174,15 @@ def process_column_header(td):
 
                     torrent_size = cells[labels.index('Size/Snatched')].get_text(strip=True).split('/', 1)[0]
                     size = convert_size(torrent_size, units=units) or -1
-
+                    pubdate = cells[labels.index('Added')].get_text(" ")
+
```
duramato: No need, already in a supported format

fernandog: the format needs to be `datetime` format. 2016-11-25 00:00:00-0000`

duramato: It's `2016-11-25 00:00:00`, absence of timezone means its utc 0 afaik? It's one less datetime import and usage..

fernandog: its not about that. need to be a datime object. medusa converts datetime object in generic_provider, if is not, it will fail

duramato: But it worked as it and got shown correctly in the manual page,  dates in python aren't  always objects? So that they can be manipulated

fernandog:
@duramato you need to start fixing codacy issue and avoid creating then

fernandog:
rebased and fixed conflict

p0psicles:
@duramato seems not to be that much work left. You think you can finish it?

duramato:
This is far from finished, the task list is what I was working on and not figurative of the whole apR progress, it still needs to be decided if we want to store dates as date objects or string representation.
I'm filled with study/exams IRL, I can only continue this after that.

labrys:
Dates should not be stored as date objects.  They should be stored as either timestamps or strings.  

I personally prefer timestamps if its only to be used for method usage and doesn't hold particular semantic value.  

For example a premier date has a specific semantic value and would be a good option for a string since you might want to know what year, month, or day something was released.  

PubDates is just for a date-ordering thus a timestamp should be more than sufficient.

labrys:
Also as much as possible I would store times in GMT or with explicit timezone references.

fernandog:
@labrys problem some providers uses universal time format, others like "x minutes ago" or "y weeks ago" (not precise), others have timezone, others not (causing conflicts storing naive datetime with aware datetime)

OmgImAlexis:
@duramato Can we merge this soon and open a new PR to handle the rest of them?

duramato:
If someone can fix the conflicts it can be merged already.

p0psicles:
@duramato please check that you didn't wrote new logging, using the old format!

duramato:
No new logging.

medusa/providers/torrent/html/gftracker.py
```diff
@@ -168,14 +168,16 @@ def process_column_header(td):
 
                     torrent_size = cells[labels.index('Size/Snatched')].get_text(strip=True).split('/', 1)[0]
                     size = convert_size(torrent_size, units=units) or -1
+                    pubdate_raw = cells[labels.index('Added')].get_text(" ")
```
p0psicles: double quotes

p0psicles: Think this is the last one

PR #1668, opened 2016-11-25
Fix Sequence index is not an int, slice, or instance with __index__

fernandog:


fernandog:
@medariox approve please

Issue #1676, opened 2016-11-26
post processing not working


poplolnman:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:**
**OS:** windows 10
**What you did:** turned on medusa
**What happened:**got an error and pp not working
**What you expected:**for pp to work
**Logs:**
```
PASTE LOGS HERE
```
2
3
4
5
6
7
8
9
10
11
12
13
2016-11-25 17:41:47 ERROR    POSTPROCESSOR :: [617ad17] Exception generated: download_subtitles() got an unexpected keyword argument 'force'
Traceback (most recent call last):
  File "C:\Medusa-develop\medusa\scheduler.py", line 103, in run
    self.action.run(self.force)
  File "C:\Medusa-develop\medusa\auto_post_processor.py", line 57, in run
    process_tv.processDir(app.TV_DOWNLOAD_DIR, force=force)
  File "C:\Medusa-develop\medusa\process_tv.py", line 288, in processDir
    process_media(processPath, videoFiles, nzbName, process_method, force, is_priority, ignore_subs, result)
  File "C:\Medusa-develop\medusa\process_tv.py", line 573, in process_media
    result.result = processor.process()
  File "C:\Medusa-develop\medusa\post_processor.py", line 1198, in process
    cur_ep.download_subtitles(force=True)
TypeError: download_subtitles() got an unexpected keyword argument 'force'

p0psicles:
Same @fernandog

PR #1678, opened 2016-11-26
Fix pymedusa/Medusa/issues/1677

fernandog:
@ratoaq2 why pycharm didn't showed this usage?

image


fernandog:
https://github.com/pymedusa/Medusa/issues/1676

p0psicles:
Because you shouldn't trust on pycharm. Better to just do search, and compare them one by one.

PR #1681, opened 2016-11-26
Not query xem for unsupported indexers

fernandog:
@p0psicles @labrys


labrys:
This ignores the fact that other providers (tvmaze for example) may provide a tvdb id

fernandog:
### Before:

image

### After:

image

fernandog:
@p0psicles why do not merge? is the best solution imo. did you see the SS from pycharm?

p0psicles:
Your solution was not wrong. It was just putting indexer configuration in code. Which i'm trying to prevent all over the medusa code. I did the same as you, but used the xem_origin, like it was supposed to.

If tvmaze and tmdb will be supported by xem in future, xem_origin, should be put back in the config.

p0psicles:
@labrys the confusion here came from us asuming this was doing the xem lookup for a show_id. But this is just a plain xem refresh, per indexer.

PR #1691, opened 2016-11-28
Create a method to check if we should replace existing quality

fernandog:
I wrote the method so it could be easy do understand and to "read"

@pymedusa/developers i will run this on my branch for a few days
would be nice if someone could review this.

### DIFF
--- a/medusa/common.py
+++ b/medusa/common.py
@@ -601,7 +601,8 @@ class Quality(object):
         return True
 
     @staticmethod
-    def should_replace(ep_status, old_quality, new_quality, allowed_qualities, preferred_qualities):
+    def should_replace(ep_status, old_quality, new_quality, allowed_qualities, preferred_qualities,
+                       download_current_quality=False, force=False):
         """Return true if the old quality should be replaced with new quality.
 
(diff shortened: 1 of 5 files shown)


fernandog:
@pymedusa/developers i will run this on my branch for a few days

would be nice if someone could review this.

fernandog:
Approve or close

fernandog:
@pymedusa/developers no one else has comments on this?

medusa/common.py
```diff
@@ -601,6 +601,46 @@ def should_search(status, show_obj):
         return True
 
     @staticmethod
+    def should_replace(ep_status, old_quality, new_quality, allowed_qualities, preferred_qualities):
```
p0psicles: Only remark I could have, is that "old_quality, new_quality, allowed_qualities, preferred_qualities" are all similar parameters. They could be easily passed as kwargs.

p0psicles:
Looks nice!

fernandog:
I'm coding a more complete `should_replace` function but i will leave to another PR
also using @p0psicles suggestion

PR #1695, opened 2016-11-28
Improved pushbullet.py to return the error message

p0psicles:

which is returned instead of a generic error message.

Replaced getURL with default session object. As there is no reason IMO to use getURL here. getURL isn't really usefull for json requests anyway, as it will return None when a 4xx response is returned.

### DIFF
--- a/medusa/notifiers/pushbullet.py
+++ b/medusa/notifiers/pushbullet.py
@@ -126,6 +126,7 @@ class Notifier(object):
         except ValueError:
             logger.log('Pushbullet notification failed. Could not parse pushbullet response.', logger.WARNING)
             push_result['error'] = 'Pushbullet notification failed. Could not parse pushbullet response.'
+            return push_result
 
         failed = response.pop('error', {})
         if failed:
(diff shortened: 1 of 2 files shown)


fernandog:
@p0psicles why milestone 0.2.0 ?

NicoLeOca:
much better
thank you @p0psicles 
It tells me that I need a pro account now.. :-/

p0psicles:
Jip. I got the same issue. At least now your seeing the reason it doesn't work. I would recommend to buy pushover account. Not that expensive and works really well with Medusa. If you wanna fix this. Stay under 500 messages per month. Or get in contact with Pb.

NicoLeOca:
@p0psicles 
you just reminded me that it was 500 messages per month cause my PB profile says 100.
Articles here and there say 500 as well.
This month for the first time, I sure hit 100; not sure about 500.

why do you tell me to contact PB?

p0psicles:
As when you google a little, you see more reports of users who are sure they didn't exceed 500, and still got this error. And didn't know anything about the 100. Thats new to me. In that case, i'm ditching it immediately.

PR #1704, opened 2016-11-29
Initial commit to get codec from guessit  in sceneQuality

fernandog:
@medariox @duramato 
as we talked. this is a initial idea.

ping @ratoaq2

### DIFF
--- a/medusa/common.py
+++ b/medusa/common.py
@@ -496,22 +496,6 @@ class Quality(object):
         """
         rel_type = ''
         name = name.lower()
-        codec = re.search(r'[xh].?26[45]', name) or ''
-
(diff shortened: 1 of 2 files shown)


medusa/common.py
```diff
@@ -545,7 +529,7 @@ def sceneQualityFromName(name, quality):
             elif re.search(r'web(-| |\.)?(rip|mux)', name):
                 rel_type = ' WEBRip'
 
-        return rel_type + codec
+        return rel_type
```
fernandog: @medariox @duramato do we need codec here too?

If so we add add in the call to sceneQuality, not here.

fernandog:
No one interested

PR #1705, opened 2016-11-29
Removed mapIndexersToShow, and replaced it with call to the show objects externals.

p0psicles:

This fixes the api v1 shows and show calls.

* Removed the module import for indexers from init. As that is used nowhere. Think it's legacy, and can cause circular imports.

I started this with the intention to fix the int casts on imdb ids. But ended with removing allot of crappy code.

### DIFF
--- a/medusa/indexers/__init__.py
+++ b/medusa/indexers/__init__.py
@@ -16,5 +16,3 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Medusa. If not, see <http://www.gnu.org/licenses/>.
-
-from . import indexer_api, indexer_exceptions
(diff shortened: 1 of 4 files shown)


medusa/indexers/__init__.py
```diff
@@ -16,5 +16,3 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Medusa. If not, see <http://www.gnu.org/licenses/>.
-
-from . import indexer_api, indexer_exceptions
```
p0psicles: @ratoaq2 @labrys any reason why I should not do this?

fernandog:
approved as fixed my issue

PR #1707, opened 2016-11-30
Do not return None, for non 2xx response, when the request type is 'response'.

p0psicles:

There is still the issue where some exceptions are swallowed and returned as None. But don't want to change too much in one go.


fernandog:
approved

medusa/helpers.py
```diff
@@ -1198,7 +1198,9 @@ def getURL(url, post_data=None, params=None, headers=None, timeout=30, session=N
 
             logger.debug(u'Requested url {url} returned status code {status}: {desc}'.format
                          (url=resp.url, status=resp.status_code, desc=http_code_description(resp.status_code)))
-            return None
+
+            if response_type and response_type != u'response':
+                return None
```
fernandog: @p0psicles what if we have `response_type` what we should return ?

p0psicles:
@duramato can you run some provider tests on develop?

PR #1721, opened 2016-12-02
Can't directly import showList as that will create symbol. Need to im…

p0psicles:

…port app, and reference showList from there.

* Added pagination to removal of indexer_updates rows. Although I doubt this will occur again. As that was probably caused by the missing showList.

### DIFF
--- a/medusa/show_updater.py
+++ b/medusa/show_updater.py
@@ -19,7 +19,7 @@
 import logging
 import threading
 import time
-from app import showList, showQueueScheduler
+import app
 
 from six import text_type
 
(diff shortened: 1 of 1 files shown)


p0psicles:
This should fix allot of indexer_update related issues. The indexer_updates table should become properly populated by this.

medusa/show_updater.py
```diff
@@ -224,16 +227,23 @@ def clean_expired_seasons(self, show_list):
...
-                b'DELETE FROM indexer_update '
-                b'WHERE indexer_update_id IN (%s)' % ','.join('?' * len(remove_row)),
-                remove_row
-            )
+            for paged_list in get_rows_by_slice(remove_row, 50):
+                self.action(
+                    b'DELETE FROM indexer_update '
+                    b'WHERE indexer_update_id IN (%s)' % ','.join('?' * len(paged_list)),
+                    paged_list
+                )
```
fernandog: nice!

medusa/show_updater.py
```diff
@@ -224,16 +227,23 @@ def clean_expired_seasons(self, show_list):
...
                     show.indexer == row['indexer'] and
                     show.indexerid == row['indexer_id']]:
+
                 remove_row.append(row['indexer_update_id'])
                 remove_show.append(row['indexer_id'])
 
+        def get_rows_by_slice(seq, row_len):
+            """Simple pagination/slice method. Will transform [1,2,3,4] into [[1,2],[3,4]] when provided rowlen=2."""
+            for start in xrange(0, len(seq), row_len):
+                yield seq[start:start + row_len]
```
labrys: Very nice! 

Just as an FYI, this could also be written as a gen exp. instead of a function:
```python
# Slice list into evenly sized chunks (e.g.  [1,2,3,4] --> [[1,2],[3,4]])
limit = 50  # max size for slice
rows_by_slice = (remove_row[i:i + n] 
                 for i in xrange(0, len(remove_list), limit))
```
Just keep in mind your function (and this version) don't work on iterables that don't have a length

PR #1722, opened 2016-12-02
More quality improvements v3

fernandog:
I'm testing this in my branch. But its safe.


fernandog:
@labrys just fyi, this PR will keep looking in the providers even we find everything in the first one (quality)

maybe a release_group from other providers is a preferred one and we could miss it.

Issue #1719, opened 2016-12-01
[APP SUBMITTED]: TypeError: cannot serialize 60 (type int)


duramato:
### INFO
**Python Version**: `2.7.9 (default, Sep 17 2016, 20:26:04) [GCC 4.9.2]`
**Operating System**: `Linux-4.4.34-v7+-armv7l-with-debian-8.0`
**Locale**: `UTF-8`
**Branch**: [develop](../tree/develop)
**Database**: `44.2`
**Commit**: pymedusa/Medusa@dd57c33e77ff6f7b3fac5db4b888336ddb3c3407
**Link to Log**: https://gist.github.com/e0e19276bb629584edc0a9ed0d65f163

Show: X-Ray Mega Airport
Indexer: Tvmaze

### ERROR
<pre>
2016-12-01 20:08:24 ERROR    SHOWQUEUE-REFRESH :: [dd57c33] 11478: Error while refreshing show X-Ray Mega Airport. Error: cannot serialize 60 (type int)
Traceback (most recent call last):
  File "/home/pi/Medusa/<a href="../blob/dd57c33e77ff6f7b3fac5db4b888336ddb3c3407/medusa/show_queue.py#L588">medusa/show_queue.py</a>", line 588, in run
    self.show.refresh_dir()
  File "/home/pi/Medusa/<a href="../blob/dd57c33e77ff6f7b3fac5db4b888336ddb3c3407/medusa/tv.py#L1481">medusa/tv.py</a>", line 1481, in refresh_dir
    self.load_episodes_from_dir()
  File "/home/pi/Medusa/<a href="../blob/dd57c33e77ff6f7b3fac5db4b888336ddb3c3407/medusa/tv.py#L727">medusa/tv.py</a>", line 727, in load_episodes_from_dir
    cur_episode = self.make_ep_from_file(os.path.join(self.location, media_file))
  File "/home/pi/Medusa/<a href="../blob/dd57c33e77ff6f7b3fac5db4b888336ddb3c3407/medusa/tv.py#L1142">medusa/tv.py</a>", line 1142, in make_ep_from_file
    **********_ep.create_meta_files()
  File "/home/pi/Medusa/<a href="../blob/dd57c33e77ff6f7b3fac5db4b888336ddb3c3407/medusa/tv.py#L2605">medusa/tv.py</a>", line 2605, in create_meta_files
    self.__create_nfo()
[...]

Issue #1686, opened 2016-11-27
Indexer "invalid timeformat"


duramato:
![image](https://lol.doyou.bid/j/R13Q5.jpeg)

- "invalid timeformat"
- Genre is cut

@p0psicles

Issue #1685, opened 2016-11-27
[APP SUBMITTED]: TypeError: object of type 'NoneType' has no len()


duramato:
### INFO
**Python Version**: `2.7.6 (default, Oct 26 2016, 20:32:47) [GCC 4.8.4]`
**Operating System**: `Linux-4.2.0-42-generic-i686-with-Ubuntu-14.04-trusty`
**Locale**: `None`
**Branch**: [develop](../tree/develop)
**Database**: `44.2`
**Commit**: pymedusa/Medusa@c86d3aa8f2d6c96e494625e42809b78c2a778715
**Link to Log**: https://gist.github.com/67661406c4b059af7f56712c8dcba775
### ERROR
<pre>
2016-11-27 01:15:53 ERROR    SHOWQUEUE-ADD :: [c86d3aa] 47670 Error while loading information from indexer TMDB. Error: "object of type 'NoneType' has no len()"
Traceback (most recent call last):
  File "/home/home/SickRage/<a href="../blob/c86d3aa8f2d6c96e494625e42809b78c2a778715/**********/show_queue.py#L348">**********/show_queue.py</a>", line 348, in run
    s = t[self.indexer_id]
  File "/home/home/SickRage/<a href="../blob/c86d3aa8f2d6c96e494625e42809b78c2a778715/**********/indexers/indexer_base.py#L254">**********/indexers/indexer_base.py</a>", line 254, in __getitem__
    self._get_show_data(key, self.config['language'])
  File "/home/home/SickRage/<a href="../blob/c86d3aa8f2d6c96e494625e42809b78c2a778715/**********/indexers/tmdb/tmdb.py#L454">**********/indexers/tmdb/tmdb.py</a>", line 454, in _get_show_data
    self._get_episodes(sid, specials=False, aired_season=None)
  File "/home/home/SickRage/<a href="../blob/c86d3aa8f2d6c96e494625e42809b78c2a778715/**********/indexers/tmdb/tmdb.py#L223">**********/indexers/tmdb/tmdb.py</a>", line 223, in _get_episodes
    if tmdb_id not in self.shows or not len(self.shows[tmdb_id].data.get('seasons')):
TypeError: object of type 'NoneType' has no len()
</pre>
---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

duramato:
[...]

PR #1724, opened 2016-12-02
Feature/indexer sanitizing

p0psicles:

Sanitized indexer inputs. Int and Floats are now cast to str by default. I'd like to use it by default like this, as the metadata doesn't handle integer fields very well.
* Sanitize tvmaze input.
* Use WebChannel as network, when available.
* Sanitize tmdb input.
* Use highest value runtime from tmdb.
* Removed some piece of old trakt code. It apparently removes a show from trakt watchlist, when that show can't be added because of "any" reason (is located in an catch Exception block. The main reason I removed it, is because it also still has references to tvrage. Leaving this can cause unexpected situations.

### DIFF
--- a/medusa/show_queue.py
+++ b/medusa/show_queue.py
@@ -21,7 +21,7 @@ import traceback
 
 from imdb import _exceptions as imdb_exceptions
 from six import binary_type, text_type
-from traktor import TraktApi, TraktException
+from traktor import TraktException
 from . import app, generic_queue, logger, name_cache, notifiers, scene_numbering, ui
 from .black_and_white_list import BlackAndWhiteList
 from .common import WANTED
--- a/medusa/show_queue.py
+++ b/medusa/show_queue.py
@@ -387,7 +387,7 @@ class QueueItemAdd(ShowQueueItem):
         # TODO: Add more specific indexer exceptions, that should provide the user with some accurate feedback.
         except IndexerShowIncomplete as e:
             logger.log(u"%s Error while loading information from indexer %s. "
-                       u"Error: %r" % (self.indexer_id, indexerApi(self.indexer).name, ex(e)), logger.ERROR)
+                       u"Error: %s" % (self.indexer_id, indexerApi(self.indexer).name, e.message), logger.WARNING)
             ui.notifications.error(
                 "Unable to add show",
                 "Unable to look up the show in %s on %s using ID %s "
--- a/medusa/show_queue.py
+++ b/medusa/show_queue.py
@@ -386,6 +386,8 @@ class QueueItemAdd(ShowQueueItem):
                 return
         # TODO: Add more specific indexer exceptions, that should provide the user with some accurate feedback.
         except IndexerShowIncomplete as e:
+            logger.log(u"%s Error while loading information from indexer %s. "
+                       u"Error: %r" % (self.indexer_id, indexerApi(self.indexer).name, ex(e)), logger.ERROR)
             ui.notifications.error(
                 "Unable to add show",
                 "Unable to look up the show in %s on %s using ID %s "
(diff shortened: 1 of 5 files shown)


p0psicles:
Should fix: #1719, #1685

medusa/show_queue.py
```diff
@@ -397,30 +408,32 @@ def run(self):
...
-                    data['shows'][0]['ids']['tvdb'] = self.indexer_id
-                else:
-                    data['shows'][0]['ids']['tvrage'] = self.indexer_id
-
-                try:
-                    trakt_api.traktRequest("sync/watchlist/remove", data, method='POST')
-                except TraktException as e:
-                    logger.log(u"Could not remove show '{0}' from watchlist. Error: {1}".format
-                               (title, e), logger.WARNING)
+            # FIXME: This shouldn't be here. I've commented it. If nobody can provide me with reasons to keep it.
```
p0psicles: @pymedusa/developers Please tell me, what to do with this?

fernandog: @p0psicles i used that so im going to explain.

I add a show 1) in trakt watchlist, medusa syncs and add show or 2) in medusa and show is added to trakt watchlist
I watch first episode but didn't like. then i remove show. then medusa removes this show from trakt

if you think it shouldn't be there, at least the call to the function that removes should be.

p0psicles: Okay that functionality is still there with this Pr. This does only try to remove it from trakt,  if an exception occurs. Which is poorly design and poorly implemented.

fernandog: but you removed entire code after `app.USE_TRAKT:` 

the call to remove is `trakt_api.traktRequest("sync/watchlist/remove", data, method='POST')`

p0psicles: Yes, and that's all part of the exception. If there is a specific reason to remove shows with an exception, it should be implemented a different way. Not through a 'catch all' exception.

medusa/show_queue.py
```diff
@@ -397,30 +408,32 @@ def run(self):
...
+            #             }
+            #         ]
+            #     }
+            #     if trakt_id == 'tvdb_id':
+            #         data['shows'][0]['ids']['tvdb'] = self.indexer_id
+            #     # else:
+            #     #     data['shows'][0]['ids']['tvrage'] = self.indexer_id
+            #
+            #     try:
+            #         trakt_api.traktRequest("sync/watchlist/remove", data, method='POST')
```
fernandog: you shouldn't remove this as this request that removes show from trakt

p0psicles:
@duramato can you run some diverse tests using this branch? If you say go, well merge.

duramato:
Yes i will, anything of more severity i should focus on?

p0psicles:
Focus on the issues you reported. Also includes the imdb link. Try to add some shows on different indexers. And maybe a show that's airing this week. To test the scedule page.

p0psicles:
ooh maybe also test this one again: https://github.com/pymedusa/Medusa/issues/1686.
As that's working fine for me.

duramato:
the invalid timeformat is still present:
image

Also the tt in the imdb link is fixed but for that show it mapped the totally wrong imdb http://www.imdb.com/title/tt0829754/

duramato:
https://github.com/pymedusa/Medusa/issues/1719 is fixed

p0psicles:
My guess is that the show does not have Imdb id mapped. Then it uses the name to attempt to get Imdb id. And there I t gets the wrong one. Can't really do anything about that. Except disabling the het Imdb I'd by name functionality.

duramato:
For https://github.com/pymedusa/Medusa/issues/1685, it shows me a notification i can't add becouse of no seasons but also an error for the exception
image

p0psicles:
Aah yes.  Will need to catch the exception also there.

p0psicles:
Should fix #1729

p0psicles:
Can you test the proper search?

fernandog:
Approved

PR #1730, opened 2016-12-02
Fix proper being redownloaded.

fernandog:
@ratoaq2 @duramato 

To check if is already in history we do:

if any(clean_proper_name == self._canonical_name(cur_result[b'resource'], clear_extension=True)

with `parsing_time` added to parse, it was comparing it with history, and history doesn't have this info, so proper was snatched again.

clean_proper_name = u'title:show|season:1|episode:1|other:proper|proper_count:1|proper_tag:repack|screen_size:1080p|format:hdtv|video_codec:h264|video_encoder:x264|release_group:dimension|type:episode|parsing_time:0.553999900818'

### DIFF
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -382,7 +382,7 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
 
     @staticmethod
     def _canonical_name(name, clear_extension=False):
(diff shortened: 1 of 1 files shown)


fernandog:
@pymedusa/developers approve please? this will keep downloading same proper over and over!

PR #1732, opened 2016-12-02
Snatch only one allowed

fernandog:
@labrys 

i need to fix any test that fails. i web commited


fernandog:
@labrys did you test? need to merge this soon otherwise users with preferred quality will have multiple snatches

PR #1733, opened 2016-12-03
Fix failed history not saving to DB when reverting episode

fernandog:
@labrys as i said in chat

i think @duramato had this issue before about composite FAILED status.

FAILED is temporary status.  After set to failed medusa set to WANTED so medusa starts a new search OR revert old episode status (as user already have the file in the library)

To reproduce:

1) Snatch & PP an episodes in allowed
2) Preferred quality is released
3) Snatch the preferred
4) Download of preferred failed

Without this PR:
5) Status is set to compound FAILED (preferred quality)
6) Revert of episode not saved to database
7) Medusa starts a search with episode status FAILED (which is wrong)

With this PR:
5) Episode is reverted to allowed quality and library file is untouched
6) Medusa starts a search with old episode status DOWNLOADED (allowed) , so it will still look for preferred.

### DIFF
--- a/medusa/failed_history.py
+++ b/medusa/failed_history.py
@@ -22,7 +22,7 @@ import re
 from datetime import datetime, timedelta
 
 from . import db, logger
-from .common import FAILED, Quality, WANTED
+from .common import FAILED, Quality, WANTED, statusStrings
 from .helper.exceptions import EpisodeNotFoundException
 from .show.history import History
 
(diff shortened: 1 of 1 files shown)


fernandog:
@labrys the failed compound is because of `mark_failed`

there is no sense: 

mark as compound failed and then revert episode!

```python
                logger.log(u"Marking episode as bad: [" + epObj.pretty_name() + "]")

                failed_history.mark_failed(epObj) # HERE IT MAKE STATUS: FAILED (SDTV)

                (release, provider) = failed_history.find_release(epObj)
                if release:
                    failed_history.log_failed(release)
                    history.log_failed(epObj, release, provider)

                failed_history.revert_episode(epObj) # HERE IT REVERT TO WANTED OR PREVIOUS STATUS
```

as `ep_obj.save_to_db()` was with wrong ident, the episode object status wasn't being reverted.

if you agree we should not call `mark_failed`

medusa/failed_history.py
```diff
@@ -156,43 +156,21 @@ def revert_episode(ep_obj):
...
-
-    :param ep_obj: Episode object to mark as failed
-    :return: empty string
-    """
-    log_str = u''
-
-    try:
-        with ep_obj.lock:
-            quality = Quality.splitCompositeStatus(ep_obj.status)[1]
-            ep_obj.status = Quality.compositeStatus(FAILED, quality)
```
fernandog: here it was creating the compound failed.

PR #1743, opened 2016-12-06
Remove torrent from client when provider ratio has been reach

fernandog:
This is a alpha feature! Use careful! =)
WARNING. It will upgrade your DB to 44.4

- [x] Support for Transmission for now (if others can code the method for your their own clients would be nice)
- [x] Torrent hash to history (DB version is 44.4 now)
- [x] Setting `remove_torrent_from_client`
- [x] TORRENTCHECKER scheduler
- [x] Setting to choose frequency of the thread
- [X] Add new thread to status page and log filter
- [x] Remove from Transmission when ratio rechead (remove data + torrent)

#### Clients:
Adds `remove_torrent` to:
- [x] Transmission
- [x] Deluged
- [x] Deluge
- [x] Qbittorrent
- [x] Utorrent

Add `remove_ratio_reached` method to:
- [x] Transmission

@pymedusa/developers need help from you for this method ^


duramato:
Well this has a big design flaw with transmission, that needs to be revised:
It is making the assumption that a stopped client (completed status) are ones that hit the seed ratio. Which aren't necessarely ones that hit the min ratio, they can be: user stopped, stopped by TorrentToMedia ( which stops them while doing the PP). After seeing a stopped torrent, and considering it completed, proceeds with it's removal regardless of it's ratio (ratio, which we know from each provider's settings).

fernandog:
@duramato please update. it will check 1) hash exists in history and (file already PPed or status downloaded)

fernandog:
@duramato are you still using this?

duramato:
Not currently, but i did for a long time

fernandog:
@p0psicles can you review this please when you have time? currently it only works for transmission as I can't code for all clients

p0psicles:
Please first look at these, then i'll have a look again.

medusa/clients/deluged_client.py
```diff
@@ -83,6 +83,9 @@ def _add_torrent_file(self, result):
 
         return remote_torrent or None
 
+    def _remove_torrent(self, info_hash):
```
fernandog: I got this method from CP. No use for now until all clients are implemented.

views/manage_subtitleMissedPP.mako
```diff
@@ -66,5 +70,36 @@
 </table>
     <input id="submit" class="btn" type="submit" value="Run Manual Post-Process" />
 </form>
+<br>
+    <h3 style="display: inline;">Releases waiting minimum ratio</h3>
```
fernandog: This only for users that postpone PP

medusa/helpers.py
```diff
@@ -1694,3 +1694,42 @@ def get_broken_providers():
 
     logger.info('Broken providers found: {0}'.format(response))
     return ','.join(response)
+
+
+def is_already_processed_media(full_filename):
```
fernandog: Can be. Will evaluate

medusa/clients/transmission_client.py
```diff
@@ -186,5 +192,104 @@ def _set_torrent_priority(self, result):
...
+            'arguments': arguments,
+            'method': 'torrent-remove',
+        })
+
+        self._request(method='post', data=post_data)
+
+        return self.response.json()['result'] == 'success'
+
+    def get_torrents_status(self):
+        """Get status of all snatches from Medusa.
```
fernandog: Suggestions?

medusa/clients/transmission_client.py
```diff
@@ -186,5 +192,104 @@ def _set_torrent_priority(self, result):
...
+
+        0 = Torrent is stopped
+        1 = Queued to check files
+        2 = Checking files
+        3 = Queued to download
+        4 = Downloading
+        5 = Queued to seed
+        6 = Seeding
+
+        :return: list of releases Medusa sent to Transmission
```
fernandog: Will fix

medusa/clients/transmission_client.py
```diff
@@ -186,5 +192,104 @@ def _set_torrent_priority(self, result):
...
+            if torrent.get('isStalled') and not torrent['percentDone'] == 1:
+                status = 'failed'
+            elif torrent['status'] == 0:
+                if torrent['percentDone'] == 1 and torrent.get('isFinished'):
+                    status = 'completed'
+                else:
+                    status = 'stopped'
+
+            if status == 'completed':
+                logger.warning("Torrent completed and reached minimum ratio: [{ratio}] or "
```
fernandog: this is only to better find the messages while dev testing

medusa/helpers.py
fernandog: i kept this here as it can be used for something else other than providers/clients. 
Others i moved

fernandog:
@p0psicles self in transmission is not the self from generic:

Failed to check torrent status. Error: 'TransmissionAPI' object has no attribute 'is_info_hash_in_history'

is there another way to use in helpers as before?

medusa/clients/transmission_client.py
labrys: Finding messages when dev testing can be easily done with grep :)

p0psicles: Yeah but I guess you want to move to a point where we can merge to develop right?

fernandog: @p0psicles. Yes. Then I will make it INFO when approved PR

labrys:
@fernandog generic doesn't have the method `is_info_hash_in_history`.  If it did, then `TransmissionClient` would have no trouble using it.  You defined it in `TorrentProvider`.  

You could extract the method to a helper file and redefine it as `def is_info_hash_in_history(obj, info_hash):`.  Then attach that method to the classes that need it.  When done like this, obj becomes the self of the class its attached to.  

Also if it doesn't need to be a method of the class but you would still like to make the class the "self" of the method at call-time you could call it like this  `is_info_hash_in_history(self, info_hash)`.

Additionally, the method is a staticmethod so needs no self anyways.  Remove the self reference and just make it an external helper.

Lastly, it's bad practice to `SELECT` by column number.  Use the column name explicitly.  So instead of `SELECT 1 FROM` you would `SELECT info_hash FROM` (assuming thats the correct column name).

Also your return statement can be simplified to either `return history_result` which is itself truthy or falsey, or if you want it to be explicitly True or False, `return bool(history_result)'.

Lastly, all this applies to `is_info_hash_processed` as well.

fernandog:
@labrys changed the return and column name

can't those functions be in helper? 
I don't agree with moving to generic as nzb providers can't use those

labrys:
As I said:

"Additionally, the method is a staticmethod so needs no self anyways. Remove the self reference and just make it an external helper."

fernandog:
@labrys it was like that before @p0psicles review.  He asked about moving to generic

labrys:
Any time you don't use `self` in the body of the method, chances are the method is a static method and should either be declared as such (`@staticmethod`) or should be moved outside of the class if it can be used by something other than the class itself.

fernandog:
all changes done. anything else?

labrys:
If it was only used by the GenericClient and it's subclasses I would agree with @p0psicles.  But since it's used in unrelated classes and the implementation does not differ, its a great candidate for an external function.

fernandog:
Last commit should be in develop for a while so devs don't have their torrent removed for some weird bug.
When PR approved i will change that log from warning to info

labrys:
I didn't review the whole PR, was just looking to answer your question about self usage.  But as for that part, that was all the changes.

labrys:
First thing I noticed on a quick review... `remove_torrents` is `_remove_torrents` in the other clients.

labrys:
I'm not convinced we need a thread to check torrents instead of just checking them as needed.

labrys:
Bumping the DB version (whether major or minor) should be one of the last things implemented as it messes up someones DB version when they switch to your branch to test your code and then switch back.  If your PR doesn't get implemented in Dev or a different PR that bumps DB version gets merged first, the DB version would then be messed up.

fernandog:
protected member in transmission.

fernandog:
@labrys but to test this, need a new field in the DB.

fernandog:
so when to check if we should remove torrents?

labrys:
You can add the field without bumping the DB version, just leave the version unchanged.
I withdrew my comment about scheduling, but I think the scheduler should be defaulted to off unless the user both has torrent clients enabled and has the option to remove torrents selected.

labrys:
As for protected member in transmission I assume you mean prefixing the method gives you that error.  If so then you misunderstood, I meant for you to remove the prefix from the other clients.

fernandog:
scheduler frequency setting added. setting will only show if you enabled the remove from client

medusa/helpers.py
p0psicles: Didn't know it was used by others. All good then!

p0psicles:
I'm good with this PR. If the warning is removed, i'll approve. Anyone else wanna take a look at this?

fernandog:
https://github.com/pymedusa/Medusa/pull/1743

PR #1746, opened 2016-12-07
Feature/add animetorrents

p0psicles:

- [x] Add season pack support. Thought of it later, probably will need to make an adjustment to the season search string.


p0psicles:
This provider works now. But because of the 'strange' release names from this provider, we still need to make some changes to our name/title parser, before it will become really useful.

Created an issue here: #1747

medusa/providers/torrent/html/animetorrents.py
```diff
@@ -245,9 +245,16 @@ def _get_episode_search_strings(self, episode, add_string=''):
             'Episode': []
         }
 
+        season_scene_names = scene_exceptions.get_scene_exceptions(episode.show.indexerid, season=episode.scene_season)
```
p0psicles: I'd like to implement something like this in genericprovider. But will possibly modify (or create new) allPossibleShowNames. As there get_scene_exceptions is already called. Just as well return it with the matching season if available.

medariox:
Approved 👍

medusa/providers/torrent/html/animetorrents.py
```diff
@@ -47,26 +47,27 @@ def __init__(self):
...
         self.anime_only = True
-        self.categories = {2: 'Anime Series',
-                           7: 'Anime Series HD'}
+        self.categories = {
+            2: 'Anime Series',
+            7: 'Anime Series HD'
+        }
 
         # Proper Strings
+        self.proper_strings = []
```
duramato: Are there no propers? or alias tags like v1/v2?

PR #1757, opened 2016-12-09
Adds manual search flag in history and tv_episodes. Show flag in hist…

fernandog:
WARNING: this will upgrade your DB to 43.3

image

### DIFF
--- a/medusa/common.py
+++ b/medusa/common.py
@@ -593,10 +593,13 @@ class Quality(object):
     }
 
     @staticmethod
-    def should_search(status, show_obj):
+    def should_search(status, show_obj, manual_searched):
         """Return true if that episodes should be search for a better quality."""
         cur_status, cur_quality = Quality.splitCompositeStatus(int(status) or UNKNOWN)
         allowed_qualities, preferred_qualities = show_obj.current_qualities
+
+        if manual_searched:
+            return False
 
         if cur_status not in (WANTED, DOWNLOADED, SNATCHED, SNATCHED_PROPER):
             return False
@@ -611,7 +614,7 @@ class Quality(object):
 
     @staticmethod
     def should_replace(ep_status, old_quality, new_quality, allowed_qualities, preferred_qualities,
-                       download_current_quality=False, force=False):
+                       download_current_quality=False, force=False, manual_searched=False):
         """Return true if the old quality should be replaced with new quality.
 
         If not preferred qualities, then any downloaded quality is final
@@ -626,6 +629,11 @@ class Quality(object):
         if old_quality == Quality.UNKNOWN:
             if not force:
                 return False, 'Existing quality is UNKNOWN. Ignoring new quality'
+
+        if manual_searched:
+            if not force:
+                # We only allow replace a manual searched episode if is a forced search
+                return False, 'Existing episode quality was manually snatched. Ignoring all new qualities'
 
         if not Quality.wanted_quality(new_quality, allowed_qualities, preferred_qualities):
             return False, 'New quality is not in any wanted quality lists. Ignoring new quality'
--- a/medusa/common.py
+++ b/medusa/common.py
@@ -593,12 +593,12 @@ class Quality(object):
     }
 
     @staticmethod
-    def should_search(status, show_obj, manual_searched):
+    def should_search(status, show_obj, manually_searched):
         """Return true if that episodes should be search for a better quality."""
         cur_status, cur_quality = Quality.splitCompositeStatus(int(status) or UNKNOWN)
         allowed_qualities, preferred_qualities = show_obj.current_qualities
 
-        if manual_searched:
+        if manually_searched:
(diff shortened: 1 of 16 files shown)


fernandog:
@p0psicles the `should_replace` method is used in daily and backlog searches. As I added the manual searched rule in the method, so both searches will check it

duramato:
*manually_searched

medusa/common.py
```diff
@@ -593,11 +593,14 @@ def statusFromName(name, anime=False):
     }
 
     @staticmethod
-    def should_search(status, show_obj):
+    def should_search(status, show_obj, manual_searched):
```
medariox: Why not making it optional like should_replace? E.g.:
```
def should_search(status, show_obj, manual_searched=False):
```

medusa/history.py
```diff
@@ -25,7 +25,8 @@
 from .show.history import History
 
 
-def _logHistoryItem(action, showid, season, episode, quality, resource, provider, version=-1, proper_tags=''):
+def _logHistoryItem(action, showid, season, episode, quality, resource,
+                    provider, version=-1, proper_tags='', manual_searched=False):
```
medariox: Should this here really default to ```False```?
What is it going to look like in DB?

medusa/post_processor.py
```diff
@@ -899,39 +903,27 @@ def _is_priority(self, ep_obj, new_ep_quality):
...
+        if self.in_history:
+
+            # If manual searched, then by pass any quality checks
+            if self.manual_searched:
+                self._log(u"This episode was manually snatched. Marking it as priority", logger.DEBUG)
                 return True
 
-            # if it's in history, we only want it if the new quality is higher or
-            # if it's a proper of equal or higher quality
-            if new_ep_quality > old_ep_quality and new_ep_quality != common.Quality.UNKNOWN:
```
medariox: If you remove this, UNKNOWN quality will replace all other qualities as it is the highest we have (32768).

medusa/post_processor.py
```diff
@@ -899,39 +903,27 @@ def _is_priority(self, ep_obj, new_ep_quality):
...
+            # We only want it if the new quality is higher
+            # Assuming the new quality is a wanted quality
+            if new_ep_quality > old_ep_quality:
                 self._log(u"Medusa snatched this episode and it is a higher quality. Marking it as priority",
                           logger.DEBUG)
                 return True
 
-            if self.is_proper and new_ep_quality >= old_ep_quality and new_ep_quality != common.Quality.UNKNOWN:
```
medariox: Same as above.

medusa/search/core.py
```diff
@@ -351,14 +353,16 @@ def wantedEpisodes(show, fromDate):
...
+        "FROM tv_episodes "
+        "WHERE showid = ? AND season > 0 and airdate > ?",
         [show.indexerid, fromDate.toordinal()]
     )
 
     # check through the list of statuses to see if we want any
     for result in sql_results:
         _, cur_quality = common.Quality.splitCompositeStatus(int(result["status"] or -1))
-        if not Quality.should_search(result['status'], show):
+        if not Quality.should_search(result['status'], show, result["manual_searched"]):
```
medariox: single quotes

medusa/server/web/home/handler.py
```diff
@@ -1749,6 +1749,10 @@ def setStatus(self, show=None, eps=None, status=None, direct=False):
                                    u'so obviously you want it replaced')
                         ep_obj.release_name = ''
 
+                    if ep_obj.manual_searched and int(status) == WANTED:
+                        logger.log(u"Reseting 'manual searched' flag as episode was change to WANTED", logger.DEBUG)
```
medariox: should be:
```
"Resetting 'manually searched' flag as episode was changed to WANTED", logger.DEBUG)```

File: medusa/history.py (hunk 2 above)
Comment (fernandog) [inline medusa/history.py]: @medariox other functions use this: log subtitle, log download, etc...
only log snatch will send True - if true

File: medusa/common.py (hunk 1 above)
Comment (fernandog) [inline medusa/common.py]: @medariox because is not optional in here. in should_replace is optional because is a multi use method

File: medusa/server/web/home/handler.py [hunk 7]
```diff
@@ -1750,7 +1750,7 @@ def setStatus(self, show=None, eps=None, status=None, direct=False):
                         ep_obj.release_name = ''
 
                     if ep_obj.manual_searched and int(status) == WANTED:
-                        logger.log(u"Reseting 'manual searched' flag as episode was change to WANTED", logger.DEBUG)
+                        logger.log(u"Reseting 'manually searched' flag as episode was changed to WANTED", logger.DEBUG)
```
medariox: Reseting -> Rese**tt**ing

fernandog: Grrrr

PR #1773, opened 2016-12-11
Fix creation of tmb and jpg thumbnail files for episodes.

p0psicles:

* Only supported by tvmaze & tmdb api. As the new v2 tvdb api, doesn't support getting the thumbnail url's in batch. And we are not going to do one api call per episode.


fernandog:
how to test this? only if @duramato can't test

p0psicles:
Create a fake episode for a show. And refresh from disk. Thumb and jpeg should be created for the episode

PR #1774, opened 2016-12-11
Fixed saving manual search results, when their downloaded through the…

p0psicles:

… season scene name.

This will make sure downloads that have used the season scene name, are still stored using the indexers season.

@medariox @ratoaq2 i'd like to make this part of the parser somehow. Do you have any ideas on this?

### DIFF
--- a/medusa/tv_cache.py
+++ b/medusa/tv_cache.py
@@ -24,7 +24,7 @@ import time
 import traceback
 
 from six import text_type
-from . import app, db, logger, show_name_helpers
+from . import app, db, logger, scene_exceptions, show_name_helpers
 from .helper.common import episode_num
 from .helper.exceptions import AuthException
 from .name_parser.parser import InvalidNameException, InvalidShowException, NameParser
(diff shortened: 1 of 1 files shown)


p0psicles:
I've changed this back to on hold, as i've noticed that the parsing towards indexer numbering, is already included in the parser. It was just not functioning for shows, which have a season scene name, but don't have a scene numbering updated. My solution does also not map to indexer season, but it mapped to scene season, which won't work for xem mapped shows. For. ex Fairy Tail.

So I need to do some more investigation. And need to get a better understanding of the current working of the parser logic (not guessit, but medusa's logic).

p0psicles:
Apparently season scene names, only have effect when used icw scene numbering. Which AFAIK is only possible with xem. What I wanted to accomplish was just mark a specific season (for ex. ajin season 2) with a season scene name. As currently in scene, there are releases names like ```Ajin 2nd Season - 05 (2016) [KamiFS] [10-bit] [HDTV] [HD 720p]```. As it's numbered 5, this should be searched with the show name 'Ajin 2nd Season'. And it should start searching from ep 1, for that season.

I was hoping that just adding 'Ajin 2nd season' was sufficient. But because there is not xem for this show, this is not possible.

PR #1783, opened 2016-12-13
Remove torrent hash from providers parsing

fernandog:
We get the torrent hash in `send_torrent` method right before update history table
No need to parse from providers

My history table has hashes even providers doesn't parse it:

image


fernandog:
any more comments on this? if not please approve

PR #1787, opened 2016-12-14
Feature/manage season exceptions

p0psicles:

Still WIP. But the CRUD routes can be reviewed. @ratoaq2 care to take a look?


medusa/server/api/v2/cache.py
```diff
@@ -0,0 +1,143 @@
...
+    def get(self, *args, **kwargs):
+        """Query scene_exception information.
+
+        :param show_indexer_id: The indexer id. Like 1 for tmdb and 3 for tvmaze.
+        :param show_id:
+        :type show_indexer_id: str
+        :param season:
+        """
+        try:
+            exception_id = int(kwargs.pop('row_id', 'throwme'))
```
ratoaq2: You could add `row_id=None` to `get(self, ...)` and use
```
exception_id = self._parse(row_id)
```

The super class has this _parse (if I'm not mistaken) that can parse arguments and by default it uses `int`

medusa/server/api/v2/cache.py
```diff
@@ -0,0 +1,143 @@
...
+            sql_base = b'SELECT * FROM scene_exceptions'
+            params = []
+            if exception_id:
+                sql_base += ' WHERE exception_id = ?'
+                params.append(exception_id)
+
+        if sql_base:
+            exceptions = cache_db_con.select(sql_base, params)
+        else:
+            self.api_finish(data={},error="Error, need to provide")
```
ratoaq2: What's the logic here? Shouldn't we return a `400 Bad Request` ? Or some other code?

medusa/server/api/v2/cache.py
```diff
@@ -0,0 +1,143 @@
...
+        last_changes = cache_db_con.connection.total_changes
+        cache_db_con.action(b'INSERT INTO scene_exceptions'
+                            b' (indexer, indexer_id, show_name, season, custom) '
+                            b' values (?,?,?,?,1)',
+                            [data.get('indexer'),
+                             data.get('indexer_id'),
+                             data.get('show_name'),
+                             data.get('season')])
+        if cache_db_con.connection.total_changes - last_changes > 0:
+            return self.api_finish(status=204)
```
ratoaq2: Since `POST` is creation, you can return `201 CREATED`

We're still deciding our conventions, but IMO we should use this [guide](http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#useful-post-responses) as a base for our work. 

>>> In case of a POST that resulted in a creation, use a HTTP 201 status code and include a Location header that points to the URL of the new resource.

medusa/server/api/v2/cache.py
```diff
@@ -0,0 +1,143 @@
...
+        try:
+            exception_id = int(kwargs.pop('row_id', 'throwme'))
+        except (ValueError, TypeError):
+            return self.api_finish(status=400, error="Require a valid exception row id.")
+
+        cache_db_con = db.DBConnection('cache.db')
+        last_changes = cache_db_con.connection.total_changes
+        cache_db_con.action(b'DELETE FROM scene_exceptions WHERE exception_id = ?', [exception_id])
+        if cache_db_con.connection.total_changes - last_changes > 0:
+            return self.api_finish(status=204, data={'result': 'Deleted {0} row.'
```
ratoaq2: If you're returning `204 No Content`, better to not have any data returned.

http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#http-status

medusa/server/api/v2/cache.py
```diff
@@ -0,0 +1,143 @@
...
+                                   .format(cache_db_con.connection.total_changes - last_changes)})
+        return self.api_finish(status=400, error="Could not update.")
+
+        return self.api_finish()
```
ratoaq2: Probably you missed.

medusa/server/api/v2/cache.py
OmgImAlexis: We should be using correct HTTP codes for all errors. 200 is only to be sent if it was actually successful.

medusa/server/api/v2/cache.py
```diff
@@ -0,0 +1,143 @@
...
+                 'indexer': row[1],
+                 'indexer_id': row[2],
+                 'show_name': row[3],
+                 'season': row[4],
+                 'custom': row[5]}
+                for row in exceptions]
+
+        self.api_finish(data=data)
+
+    def put(self, *args, **kwargs):
```
OmgImAlexis: This is to replace the WHOLE exception, PATCH also needs to be added to override a single field.

medusa/server/api/v2/cache.py
OmgImAlexis: @ratoaq2 we should look at writing up a doc for all of the conventions or providing a skeleton class that should be for new files.

OmgImAlexis:
9.6 PUT

If an existing resource is modified, either the 200 (OK) or 204 (No Content) response codes > SHOULD be sent to indicate successful completion of the request.

9.7 DELETE

A successful response SHOULD be 200 (OK) if the response includes an entity describing the status, 202 (Accepted) if the action has not yet been enacted, or 204 (No Content) if the action has been enacted but the response does not include an entity.

p0psicles:
Well the pr was actually started to provide users a UI control that makes them able to add custom season scene exceptions. The crud routes where just to enable that.

PR #1788, opened 2016-12-14
Fix pymedusa/Medusa/issues/1675

fernandog:
Fixes pymedusa/Medusa/issues/1675

  File "/home/pi/Medusa/medusa/providers/torrent/html/bithdtv.py", line 133, in parse
    title = cells[2].find('a')['title']
IndexError: list index out of range


medariox:
I would rather do something like
```
if len(cells) < 3:
    continue
```
right after ```cells = row('td')```

PR #1789, opened 2016-12-14
Don't fallback to other indexer id if if provider doesn't support it

fernandog:
@p0psicles @ripper1967

Fixes:
2016-12-14 08:27:16 WARNING  SEARCHQUEUE-FORCED-22414 :: [Rarbg] :: [aa2e8f5] Invalid search_tvdb Code: 7

this should fix BTN too. @duramato is there any error similar for btn?

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -477,9 +477,10 @@ class GenericProvider(object):
 
         return [search_string]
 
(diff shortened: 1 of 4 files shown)


fernandog:
Seems only providers uses this, so we can safely remove the fallback OR do as I proposed

PR #1794, opened 2016-12-14
Initial provider tests

medariox:

Todo:
- [ ] Test with non-ascii results
- [ ] Support magnet links
- [ ] Much more


ratoaq2:
Nice :-)

ratoaq2:
Take a look in `VCR` (the one that subliminal uses: https://github.com/Diaoul/subliminal/blob/master/tests/test_legendastv.py#L13-L16 )

You can just annotate the test method: https://github.com/Diaoul/subliminal/blob/master/tests/test_legendastv.py#L112

And it will do a real request to the site (only first time) and record it in a yaml file:
https://github.com/Diaoul/subliminal/blob/master/tests/cassettes/legendastv/test_search_titles_episode.yaml

It worth to take a look at it. It might help. If you do use it, don't forget to remove/change your user/password from the generate yaml files before pushing your code to github.

medariox:
Thanks for the feedback @ratoaq2. ```VRC``` looks promising, will see what I can do with it.

What do you think about the structure? I wanted to avoid having to deal with single tests for each provider (subliminal-like), as they can be hard to maintain, but I'm open to suggestions.

medariox:
@ratoaq2 
Sadly atm not even the basic tests are working. I will try to have something that at least works for the WE.

medariox:
This is now rebased and fixed. It should work with any provider type, although I just tried with public torrent providers (magnets) for now. Works offline as well after the first run. What should be added is some way to strip the sensitive data from private trackers. I won't have much time this week, but feedback is always appreciated.

p0psicles:
Let's try to get it to a mergeable state. The provider coverage is not important imo. So let's try to have the tests working for a subset. And expand in time.

medariox:
@p0psicles 
That's the plan. I will only add tests for public torrent providers for now. The problem is that with every new test, new issues come up (which is the whole point of tests) and those take time to be fixed.

medariox:
The PR is now concluded! 🎉

Issue #1785, opened 2016-12-14
name_cache beeing generated with wrong id


duramato:
@p0psicles, Follow up from irc 
logged info in the format `" Internal name cache for Quarry with id (show.indexerid) set is: [name_cache]" `
https://gist.github.com/duramato/4d5adc46ed173e79e29b9a2409edddfd

The issue is the show is saving to name cache with the imdb id `u'quarry': 269110`,and not the indexer id `Quarry with id (4158)`

Indexer used: tvmaze

p0psicles:
Do you have a full log?

duramato:
There are no other logs with info pertaining to the issue. The only other time I see it is an empty name cache is when performing a search, that it shows the name cache items,but it is the same as above.

p0psicles:
Or just remove one of the shows.

PR #1795, opened 2016-12-14
Prevent shows from being added multiple times by different indexers.

p0psicles:

- [x] Add all possible externals to shows/db when adding a show.
- [x] Check for already existing show, when adding a show, using all externals.
- [x] Added Indexers as external source.
- [x] Added Trakt as external source.
- [x] Implement some nice debug log messages.
- [x] Remove dir when a show add is started, but it was interupted because the show already existed in db. If dir empty, it should be deleted.

~~* Added StringMatcher helper. As i'm planning to use this to do some scored based show name comparing. Don't know if i'm going to keep this.~~
Am going to verify on exact name using a rest call in frontend. Then user will just get a warning. As some shows are named alike, I don't want to block anything. (not part of this pr)

* Added the externals.py, to use as a helper module for externals.

Goal of this PR is to find as many as possible dependable techniques for preventing shows being added multiple times through different indexers.

You can test this, by adding show Arrow with tvdb. Afterwards rename/remove the folder and add it through tvmaze. With this PR this should not be possible anymore. In this particular case it's blocked by matching IMDB ids.

@pymedusa/developers 
I'd also add to this pr, a way of removing the dir, after finishing early, when the dir is empty. Don't know what you guys think of that?


fernandog:
@p0psicles im ok with that. we have a helper function that cleans empty dirs

fernandog:
@p0psicles can we make it lowercase?

helper/externals.py:97:9: N806 variable in function should be lowercase

p0psicles:
Should be able to prevent this from happening: #1785

p0psicles:
Review pls.

fernandog:
I will soon. Wont be a PRO review but...

medusa/helper/externals.py
```diff
@@ -0,0 +1,161 @@
...
+                      'trakt_access_token': app.TRAKT_ACCESS_TOKEN,
+                      'trakt_refresh_token': app.TRAKT_REFRESH_TOKEN}
+    trakt_api = TraktApi(app.SSL_VERIFY, app.TRAKT_TIMEOUT, **trakt_settings)
+
+    id_lookup = '/search/{external_key}/{external_value}?type=show'
+    trakt_mapping = {'tvdb_id': 'tvdb', 'imdb_id': 'imdb', 'tmdb_id': 'tmdb', 'trakt_id': 'trakt'}
+    trakt_mapping_rev = {v: k for k, v in trakt_mapping.items()}
+
+    for external_key in externals:
+        if not trakt_mapping.get(external_key) or not externals[external_key]:
```
medariox: I think it's safer to use .get here as well.
```
externals.get(external_key)
```

medusa/helper/externals.py
```diff
@@ -0,0 +1,161 @@
...
+            continue
+
+        url = id_lookup.format(external_key=trakt_mapping[external_key], external_value=externals[external_key])
+        logger.debug(u"Looking for externals using Trakt and {indexer_name}'s {id}",
+                     indexer_name=trakt_mapping[external_key], id=externals[external_key])
+        result = trakt_request(trakt_api, url)
+        if result and len(result) and result[0].get('show') and result[0]['show'].get('ids'):
```
medariox: You could use all([]) instead of and .. and .. and .. and

medusa/indexers/tmdb/tmdb.py
```diff
@@ -583,3 +583,27 @@ def get_last_updated_seasons(self, show_list, from_time, weeks=1):
...
+        Accepts as kwargs, so you'l need to add the externals as key/values.
+        :param tvrage_id: The tvrage id.
+        :param tvdb_id: The tvdb id.
+        :param imdb_id: An imdb id (inc. tt).
+        :returns: A dict with externals, including the tvmaze id.
+        """
+        for external_id in ['tvdb_id', 'imdb_id', 'tvrage_id']:
+            if kwargs.get(external_id):
+                result = self.tmdb.Find(kwargs.get(external_id)).info(**{'external_source': external_id})
+                if result.get('tv_results', None) and result['tv_results'][0]:
```
medariox: .get already defaults to None

medusa/indexers/tvmaze/tvmaze_api.py
```diff
@@ -465,3 +465,30 @@ def get_last_updated_series(self, from_time, weeks=1, filter_show_list=None):
...
+                try:
+                    result = self.tvmaze_api.get_show(**{external_id: kwargs.get(external_id)})
+                    if result:
+                        externals = result.externals
+                        externals[external_id] = result.id
+                        return {mapping[external_id]: external_value
+                                for external_id, external_value
+                                in externals.items()
+                                if external_value and mapping.get(external_id)}
+                except ShowNotFound:
```
medariox: Maybe a log (debug) here?

medusa/show_queue.py
```diff
@@ -367,32 +369,51 @@ def run(self):
...
+                logger.log(u"Could not add the show %s, as it already is in your library."
+                           u" Error: %s" % (s['seriesname'], e.message), logger.WARNING)
+                ui.notifications.error(
+                    'Unable to add show',
+                    'reason: {0}' .format(e.message)
+                )
+                self._finishEarly()
+
+                # Clean up leftover if the newly created directory is empty.
+                delete_empty_folders(self.showDir)
```
medariox: This will delete the "Shows folder" if there aren't any other shows in that folder.

medusa/tv.py
```diff
@@ -1263,8 +1264,20 @@ def load_from_indexer(self, tvapi=None):
...
-        self.imdbid = getattr(indexed_show, 'imdb_id', '') or self.externals.get('imdb_id')
+
+        # Add myself (indexer) as an external
+        self.externals[mappings[self.indexer]] = self.indexerid
+
+        # Enrich the externals, using reverse lookup.
+        self.externals.update(get_externals(self))
+
+        if self.externals.get('imdb_id'):
+            self.imdbid = self.externals.get('imdb_id')
```
medariox: You could use ```or``` instead of the if .. else

medusa/show_queue.py
p0psicles: Seriously? I tested it by putting a text file in the Dir. With the file in it, it didn't delete. Without it did.

medusa/helper/externals.py
ratoaq2: It's correct... No change needed

medariox: You are right, didn't notice it, the second check can be removed completely then.

medusa/show_queue.py
```diff
@@ -367,32 +369,51 @@ def run(self):
...
             # has no proper english version of the show
             if getattr(s, 'seriesname', None) is None:
-                logger.log(u"Show in {} has no name on {}, probably searched with the wrong language.".format
+                logger.log(u"Show in {0} has no name on {1}, probably searched with the wrong language.".format
                            (self.showDir, indexerApi(self.indexer).name), logger.ERROR)
 
-                ui.notifications.error("Unable to add show", "Show in " + self.showDir + " has no name on " +
-                                       str(indexerApi(self.indexer).name) + ", probably the wrong language. "
-                                       "Delete .nfo and manually add the correct language.")
+                ui.notifications.error('Unable to add show Show',
```
ratoaq2: Show show

medusa/indexers/tvmaze/tvmaze_api.py
p0psicles: fya, I made a small change here. Of for one key are reverse external can't be found, it will try the next one now. Before that I exited. But a show can have for ex a imdb mapped but not a tvdb.

p0psicles:
@ratoaq2 @medariox can I get approval on this?

PR #1799, opened 2016-12-16
Fix RARBG regex when check if is media file

fernandog:
It was detecting RARBG.mp4 as valid media info file.
hey, @rarbg are there any other file formats other than `RARBG.mp4` ?

### DIFF
--- a/medusa/helpers.py
+++ b/medusa/helpers.py
@@ -108,7 +108,7 @@ def isMediaFile(filename):
             return False
 
         # ignore RARBG release intro
-        if re.search(r'^RARBG\.\w+\.(mp4|avi|txt)$', filename, re.I):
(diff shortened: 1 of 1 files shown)


fernandog:
@ratoaq2 can we merge this before you change the parser?

medariox:
You probably want to make the first point optional, not the second?

rarbg:
@fernandog 
RARBG(\.|\.(com|to))\.(mp4|avi|txt)

fernandog:
@rarbg thanks! but that won't match RARBG.mp4. I added the "?" to be optional

https://regex101.com/r/eBSpOp/1

ratoaq2:
Of course it matches.

rarbg:
@fernandog 
Sorry still drinking my coffee
RARBG(\.(com|to))\.(txt|avi|mp4)

Issue #1801, opened 2016-12-16
Missing network name


fernandog:
@p0psicles I have a tvmaze show that has no network:

![image](https://cloud.githubusercontent.com/assets/2620870/21275613/c9011e84-c3b4-11e6-8ff9-71c7d1f98c6a.png)

But in tvmaze info is available:  Streamed on: Netflix (2016 - now) 
http://www.tvmaze.com/shows/13162/3

Wondering if is the "Streamed on" that is not being parsed

duramato:
Afaik tvmaze has diff items for networks and streaming services, don't remember the exact item nane

fernandog:
@p0psicles 

![image](https://cloud.githubusercontent.com/assets/2620870/21280726/d73be2a8-c3ce-11e6-9484-2aee6b9571f1.png)

PR #1802, opened 2016-12-16
Fix pymedusa/Medusa/issues/1801

fernandog:
@p0psicles do agree with my fix?

image

Fixes https://github.com/pymedusa/Medusa/issues/1801


p0psicles:
I'll check tomorrow. But getting the name should be done through get()  as that's a dict.

fernandog:
Its a webChannel class. Not dict.

PR #1807, opened 2016-12-19
fix api auth when using no user/pass, add new field to api config

OmgImAlexis:
This allows the use of a form when posting to `/api/auth/login` instead of just basic auth. It also allows users to authenticate and get an API key when there's no username/password set. This is mainly used when Medusa is first installed.

### DIFF
--- a/medusa/server/api/v2/auth.py
+++ b/medusa/server/api/v2/auth.py
@@ -22,6 +22,7 @@ class LoginHandler(BaseRequestHandler):
 
     def post(self, *args, **kwargs):
         """Submit login."""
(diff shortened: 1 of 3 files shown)


p0psicles:
@OmgImAlexis we've had allot of issues with the login/auth related to vue and the new api. How confident are your that this will not cause regression, for when we would release to 1.9? Tested with/without reverse proxy (apache/nginx)?

OmgImAlexis:
This actually doesn't change anything for how the old system works and nothing currently uses the new way I've added.

I'd like in the future to add some tests that actually test reverse proxy, etc. using nginx in travis so we don't have to keep guessing if something breaks.

PR #1828, opened 2016-12-24
Fix ExtraTorrents provider (use RSS)

medariox:

### DIFF
--- a/medusa/providers/torrent/html/extratorrent.py
+++ b/medusa/providers/torrent/html/extratorrent.py
@@ -125,7 +125,7 @@ class ExtraTorrentProvider(TorrentProvider):
                     download_url = element.magneturi.get_text()
                     if not all([title, download_url]):
                         continue
-                    
+
                     download_url += self._custom_trackers
 
                     seeders = try_int(element.seeders.get_text())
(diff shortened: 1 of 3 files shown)


fernandog:
@medariox about pubdate for ET, its already in the other PR.  we need to discuss the better way to store it. so pubdate is on hold for now

medusa/providers/torrent/html/extratorrent.py
```diff
@@ -118,44 +112,22 @@ def parse(self, data, mode):
...
-            # Skip column headers
-            for result in torrent_rows[2:]:
+            for element in elements:
                 try:
-                    cells = result('td')
-
-                    torrent_info = cells[0].find('a')
-                    if not torrent_info:
+                    title = element.title.get_text()
+                    download_url = element.enclosure.get('url')
```
duramato: Here you should append our list of trackers like it's done on other trackers

medariox: It's not a magnet.

duramato: You should use magnets always they are more reliable ,only black hole should use torrent files

medariox:
All done, awaiting approval.

medariox:
This won't be needed for now, as ET removed the html protection they had in place.

p0psicles:
What are we going to do with this now?

medariox:
Let's close it and keep the branch?

PR #1845, opened 2016-12-26
Add XEM column/icon in home

fernandog:
### Simple | Banner | Small Poster
image

### Poster
image


fernandog:
@duramato if is all good, approve?

PR #1846, opened 2016-12-27
API v2 updates

OmgImAlexis:
This adds scopes to the API so we can eventually support more than a single user along with granular controls on said users.

~~I've also added some new fields to the config route which I still need to test with the current mako files before this is merged.~~

This removes some methods we had on the `/home` route like the ones for home and schedule layout changing as I've moved them to patch methods with the api. The main reason for this is they shouldn't be in the home method for starters as schedule != home and the api has validation built in.

Assets can now be loaded by using `asset=""` on an element, for example to get a show's banner you would use this. This works with `a` and `img` elements for now.

`asset="show/${show.id}?type="banner"` I also went with camelcase for type instead of underscores like `posterThumb` vs `poster_thumb`.

@p0psicles I've fixed an issue where plot tooltips only worked for tvdb shows. They now work for all indexers and use the API.


medusa/server/api/v2/base.py
```diff
@@ -47,6 +47,9 @@ def api_finish(self, status=None, error=None, data=None, headers=None, **kwargs)
             if data is not None:
                 self.set_header('Content-Type', 'application/json; charset=UTF-8')
                 self.finish(json.JSONEncoder(default=json_string_encoder).encode(data))
+            elif stream:
```
OmgImAlexis: @p0psicles this finally fixes the redirect from showPoster issue so it should serve the image instantly. This will help with page loads for larger libraries.

OmgImAlexis:
@p0psicles can you please fix the conflicts when you get a chance. I have a feeling this conflicts with some changes you were making.

OmgImAlexis:
I'm not adding anymore until this is tested and merged.

medusa/server/api/v2/asset.py
```diff
@@ -1,46 +1,38 @@
...
-            if asset_group == 'show':
-                asset_type = self.get_argument('type', default='banner')
-                # @TODO: Replace this with real posterThumb as we only have poster ATM
-                if asset_type == 'posterThumb':
-                    asset_type = 'poster'
-                return self._serve_asset(path=os.path.join(app.CACHE_DIR, 'images/'), filename=query + '.' + asset_type)
-            if asset_group == 'network':
-                return self._serve_asset(path=os.path.join(app.PROG_DIR, 'static/images/network/'), filename=query)
+        # http://localhost:8081/api/v2/asset/show/295519?api_key=xxx&type=banner
+        asset_type = self.get_argument('type', default='banner')
```
medariox: Why not after the if?

medusa/server/api/v2/asset.py
```diff
@@ -1,46 +1,38 @@
...
-                            data = f.read(16384)
-                            if not data:
-                                break
-                            self.write(data)
-                    self.finish()
-                except IOError:
-                    self.api_finish(status=404, error='Asset or Asset Type Does Not Exist')
-            else:
-                self.api_finish(status=404, error='Asset or Asset Type Does Not Exist')
+            if asset_type[0:6] == 'banner':
```
medariox: Instead of all the if .. else, you could do:

```
if asset_type.startswith(('banner', 'fanart', 'poster', 'network')):
    media = ShowNetworkLogo(show_id, media_format)
```

medusa/server/api/v2/asset.py
```diff
@@ -1,46 +1,38 @@
...
+            elif asset_type[0:6] == 'poster':
+                media = ShowPoster(show_id, media_format)
+            elif asset_type[0:7] == 'network':
+                media = ShowNetworkLogo(show_id, media_format)
+
+            if media is not None:
+                self.set_header('Content-Type', media.get_media_type())
+                self.api_finish(stream=media.get_media())
+
+            return None
```
medariox: No return needed here

medusa/server/api/v2/config.py
```diff
@@ -3,15 +3,21 @@
 
 import platform
 import sys
-
 from six import text_type
+import tornado
```
medariox: Think about only importing what you need from tornado (e.g. escape)

medusa/server/api/v2/asset.py
OmgImAlexis: How would that work since each if contains a call to a different method?

The ShowNetworkLogo isn't for banners, fanart, poster, etc.

medusa/server/api/v2/asset.py
OmgImAlexis: This was just copied from the old method.

medusa/server/api/v2/asset.py
labrys: well if you had a dict of methods then you could call the method from the dict by asset name... e.g. (assuming that pattern is <type>-<remaining info>)
```python
methods = {
    'banner': banner_method,
    'poster': poster_method,
    ....
}
if asset_type.startswith(('banner', ...)):
    asset = asset_type.split('-')
    media = method[asset](show_id, media_format)
```
also using slicing to see if a string startswith something is poor practice... startswith is much clearer and generally a better option.....

medariox: @xo You are right ofc,  sorry about that, I thought they all used the same method. 
To add to what labrys already said, you could also use something like this (with a dict of methods):
```
for key in methods:
    if asset_type.startswith(key):
        methods[key](show_id, media_format)
        break
```

medusa/server/api/v2/config.py
```diff
@@ -3,15 +3,21 @@
 
 import platform
 import sys
-
 from six import text_type
+import json_decode from tornado.escape
```
medariox:
```
from tornado.escape import json_decode
```

OmgImAlexis:
@p0psicles can you please review, fix the conflicts and then merge once you're free.

medusa/server/api/v2/asset.py
labrys: At least change the slice notation to startswith since it's much less error-prone and clearer as to the intention.

labrys: Also you probably want to add a `.lower()` e.g.:
```python
asset_type.lower().startswith('banner')
```

OmgImAlexis:
@ratoaq2 thoughts on removing our tox tests for this and moving to [dredd](https://github.com/apiaryio/dredd) for the whole of API v2?

medusa/server/api/v2/config.py
```diff
@@ -71,23 +74,195 @@ def get(self, query=''):
...
+            'nzb': {
+                'enabled': app.USE_NZBS,
+                'username': app.NZBGET_USERNAME,
+                'password': app.NZBGET_PASSWORD,
+                # app.NZBGET_CATEGORY
+                # app.NZBGET_CATEGORY_BACKLOG
+                # app.NZBGET_CATEGORY_ANIME
+                # app.NZBGET_CATEGORY_ANIME_BACKLOG
+                'host': app.NZBGET_HOST,
+                'pririty': app.NZBGET_PRIORITY
```
p0psicles: pririty?

lib/jwt/__main__.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python
```
p0psicles: Better to include the libs, using a separate PR per lib. Then after, rebase this PR on that.

medusa/server/api/v2/config.py
```diff
@@ -71,23 +72,206 @@ def get(self, query=''):
...
+                    if 'allSeasons' in data['layout']['show'] and str(data['layout']['show']['allSeasons']).lower() in ['true', 'false']:
+                        app.DISPLAY_ALL_SEASONS = int(data['layout']['show']['allSeasons'])
+                        done_data['layout']['show'].setdefault('allSeasons', bool(app.DISPLAY_ALL_SEASONS))
+                    if 'specials' in data['layout']['show'] and str(data['layout']['show']['specials']).lower() in ['true', 'false']:
+                        app.DISPLAY_SHOW_SPECIALS = int(data['layout']['show']['specials'])
+                        done_data['layout']['show'].setdefault('specials', bool(app.DISPLAY_SHOW_SPECIALS))
+        # Make sure to update the config file after everything is updated
+        app.instance.save_config()
+        if len(done_errors):
+            print('Can\'t PATCH [' + ', '.join(done_errors) + '] since ' + ["it's a static field.", "they're static fields."][len(done_errors) > 1])
```
p0psicles: Maybe there's a reason for it. But do we really need to use print here?

medusa/server/api/v2/config.py
p0psicles: Sorry, was looking at old commit.

medusa/server/api/v2/config.py
OmgImAlexis: For now I've added this to debug why certain routes can't be patched.

lib/jwt/__main__.py
OmgImAlexis: Ref: https://github.com/pymedusa/Medusa/pull/1882

medusa/server/api/v2/config.py
ratoaq2: You should use logging api instead of print

labrys: Never use print to log.

p0psicles: I'll fix it.

PR #1853, opened 2016-12-28
Feature/metadata reduce indexer calls

p0psicles:

### DIFF
--- a/medusa/metadata/__init__.py
+++ b/medusa/metadata/__init__.py
@@ -46,8 +46,33 @@ def _getMetadataClass(name):
     return module.metadata_class()
 
 
+class MetaDataProvidersDict(dict):
+    """A custom dict class, used to reset the indexer_api attribute of an individual indexer.
+
+    therefor the app.metadata_provider_dict may only be itterated through the .values() method.
+    """
+    def __init__(self, *args, **kwargs):
+        """MetaDataDict init."""
+        super(MetaDataDict, self).__init__(*args, **kwargs)
+
+    def __getitem__(self, key):
+        """Get the dict value, and reset the indexer_api to None. This was needed because we initialize the metadata
+        provider object once, and then store it in this dict. We've implemented the indexer_api, to cache the indexer
+        api information, when using it for images, season banners, actor data, show info etc."""
+        val = dict.__getitem__(self, key)
+        val.indexer_api = None
+        return val
+
+    def values(self):
+        """Override values method."""
+        providers = []
+        for metadata_provider in self:
+            providers.append(self[metadata_provider])
+        return providers
+
+
 def get_metadata_generator_dict():
-    result = {}
+    result = MetaDataProvidersDict()
     for cur_generator_id in available_generators():
         cur_generator = _getMetadataClass(cur_generator_id)
         if not cur_generator:
--- a/medusa/metadata/__init__.py
+++ b/medusa/metadata/__init__.py
@@ -53,7 +53,7 @@ class MetaDataProvidersDict(dict):
     """
 
     def __init__(self, *args, **kwargs):
-        """MetaDataDict init."""
+        """Initialize the metadatadict object instance."""
         super(MetaDataProvidersDict, self).__init__(*args, **kwargs)
 
     def __getitem__(self, key):
@@ -61,7 +61,8 @@ class MetaDataProvidersDict(dict):
 
         This was needed because we initialize the metadata provider object once, and then store it in this dict.
         We've implemented the indexer_api, to cache the indexer api information, when using it for images,
-        season banners, actor data, show info etc."""
+        season banners, actor data, show info etc.
+        """
         val = dict.__getitem__(self, key)
         val.indexer_api = None
         return val
--- a/medusa/metadata/__init__.py
+++ b/medusa/metadata/__init__.py
@@ -53,7 +53,7 @@ class MetaDataProvidersDict(dict):
     """
     def __init__(self, *args, **kwargs):
         """MetaDataDict init."""
-        super(MetaDataDict, self).__init__(*args, **kwargs)
+        super(MetaDataProvidersDict, self).__init__(*args, **kwargs)
 
     def __getitem__(self, key):
         """Get the dict value, and reset the indexer_api to None. This was needed because we initialize the metadata
--- a/medusa/metadata/__init__.py
+++ b/medusa/metadata/__init__.py
@@ -49,16 +49,19 @@ def _getMetadataClass(name):
 class MetaDataProvidersDict(dict):
     """A custom dict class, used to reset the indexer_api attribute of an individual indexer.
 
-    therefor the app.metadata_provider_dict may only be itterated through the .values() method.
+    Therefor the app.metadata_provider_dict may only be itterated through the .values() method.
     """
+
     def __init__(self, *args, **kwargs):
         """MetaDataDict init."""
         super(MetaDataProvidersDict, self).__init__(*args, **kwargs)
 
     def __getitem__(self, key):
-        """Get the dict value, and reset the indexer_api to None. This was needed because we initialize the metadata
-        provider object once, and then store it in this dict. We've implemented the indexer_api, to cache the indexer
-        api information, when using it for images, season banners, actor data, show info etc."""
+        """Get the dict value, and reset the indexer_api to None.
+
+        This was needed because we initialize the metadata provider object once, and then store it in this dict.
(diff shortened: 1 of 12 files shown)


medusa/metadata/mede8er.py
```diff
@@ -99,48 +100,14 @@ def _show_data(self, show_obj):
...
-        l_indexer_api_params['actors'] = True
-
-        if indexer_lang and not indexer_lang == app.INDEXER_DEFAULT_LANGUAGE:
-            l_indexer_api_params['language'] = indexer_lang
-
-        if show_obj.dvdorder != 0:
-            l_indexer_api_params['dvdorder'] = True
-
-        t = indexerApi(show_obj.indexer).indexer(**l_indexer_api_params)
+        my_show = self._get_show_data(show_obj)
```
medariox: Also here.

medusa/metadata/media_browser.py
```diff
@@ -234,46 +234,10 @@ def _show_data(self, show_obj):
...
-        l_indexer_api_params['actors'] = True
-
-        if indexer_lang and not indexer_lang == app.INDEXER_DEFAULT_LANGUAGE:
-            l_indexer_api_params['language'] = indexer_lang
-
-        if show_obj.dvdorder != 0:
-            l_indexer_api_params['dvdorder'] = True
-
-        t = indexerApi(show_obj.indexer).indexer(**l_indexer_api_params)
+        my_show = self._get_show_data(show_obj)
```
medariox: Also here.

medusa/metadata/kodi_12plus.py
```diff
@@ -107,43 +108,10 @@ def _show_data(self, show_obj):
...
-        l_indexer_api_params['actors'] = True
-
-        if indexer_lang and not indexer_lang == app.INDEXER_DEFAULT_LANGUAGE:
-            l_indexer_api_params['language'] = indexer_lang
-
-        if show_obj.dvdorder != 0:
-            l_indexer_api_params['dvdorder'] = True
-
-        t = indexerApi(show_obj.indexer).indexer(**l_indexer_api_params)
+        my_show = self._get_show_data(show_obj)
```
medariox: I think we need a try .. except here.

medusa/metadata/generic.py
```diff
@@ -722,23 +722,22 @@ def _retrieve_show_image(self, image_type, show_obj, which=None):
         indexer_lang = show_obj.lang
 
         try:
-            # There's gotta be a better way of doing this but we don't wanna
-            # change the language value elsewhere
-            lINDEXER_API_PARMS = indexerApi(show_obj.indexer).api_params.copy()
+            if not self.indexer_api:
```
medariox: We have this same code in 3 different places:
```
if not self.indexer_api:
                # There's gotta be a better way of doing this but we don't wanna
                # change the language value elsewhere
                l_indexer_api_params = indexerApi(show_obj.indexer).api_params.copy()

                l_indexer_api_params['banners'] = True

                if indexer_lang and not indexer_lang == app.INDEXER_DEFAULT_LANGUAGE:
                    l_indexer_api_params['language'] = indexer_lang

                if show_obj.dvdorder != 0:
                    l_indexer_api_params['dvdorder'] = True

                self.indexer_api = indexerApi(show_obj.indexer).indexer(**l_indexer_api_params)

            indexer_show_obj = self.indexer_api[show_obj.indexerid]
```

This could be optimized.

medusa/metadata/generic.py
```diff
@@ -897,6 +894,54 @@ def _season_banners_dict(self, show_obj, season):
...
+                if show_obj.dvdorder != 0:
+                    l_indexer_api_params['dvdorder'] = True
+
+                self.indexer_api = indexerApi(show_obj.indexer).indexer(**l_indexer_api_params)
+
+                my_show = self.indexer_api[int(show_id)]
+        except IndexerShowNotFound:
+            logger.log(u'Unable to find {indexer} show {id}, skipping it'.format
+                       (indexer=indexerApi(show_obj.indexer).name,
+                        id=show_id), logger.ERROR)
```
medariox: Should this really be ERROR? Maybe WARNING instead.

medusa/metadata/generic.py
```diff
@@ -897,6 +894,54 @@ def _season_banners_dict(self, show_obj, season):
...
+        except IndexerError:
+            logger.log(u'{indexer} is down, can\'t use its data to add this show'.format
+                       (indexer=indexerApi(show_obj.indexer).name), logger.ERROR)
+            raise
+
+        # check for title and id
+        if not (getattr(my_show, 'seriesname', None) and getattr(my_show, 'id', None)):
+            logger.log(u'Incomplete info for {indexer} show {id}, skipping it'.format
+                       (indexer=indexerApi(show_obj.indexer).name,
+                        id=show_id), logger.ERROR)
```
medariox: Same.

p0psicles:
@medariox i've tried to implement your review comment as good as possible. I did solve it a little different. As now i'm catching the exception in an earlier stage.

I then noticed a bug I implemented, which was caused by the caching in the indexer_api attribute. It can happen that you index show x from indexer y and then show b from indexer z. Because the metadata  objects (kodi, kodi12, etc) are initialized once at the start of medusa, it could then reuse the indexer_api which was saved for indexer y, to search an indexer x show id.

To prevent this, i've created a custom dict class, that will empty indexer_api, when values() is called for the iteration. Let me know if I need to walk you through it.

medusa/metadata/generic.py
p0psicles: This has been reduced.

medusa/tv.py
```diff
@@ -635,6 +635,8 @@ def __write_show_nfo(self):
         result = False
 
         logger.log(u'{id}: Writing NFOs for show'.format(id=self.indexerid), logger.DEBUG)
+        # You may only call .values() on metadata_provider_dict! As on values() call the indexer_api attribute
+        # is reset. This will prevent errors, when using multiple indexers and caching.
         for cur_provider in app.metadata_provider_dict.values():
```
p0psicles: All these metadata_provider_dict iterations should be reduced to one iteration.

medusa/metadata/generic.py
```diff
@@ -897,6 +824,55 @@ def _season_banners_dict(self, show_obj, season):
...
+                self.indexer_api = indexerApi(show_obj.indexer).indexer(**l_indexer_api_params)
+
+            my_show = self.indexer_api[int(show_id)]
+        except IndexerShowNotFound:
+            logger.log(u'Unable to find {indexer} show {id}, skipping it'.format
+                       (indexer=indexerApi(show_obj.indexer).name,
+                        id=show_id), logger.WARNING)
+            return False
+
+        except (IndexerError, RequestException):
```
medariox: Shouldn't this also catch ```IndexerShowIncomplete``` and ```IndexerUnavailable```?

medusa/metadata/generic.py
```diff
@@ -970,7 +947,12 @@ def _retrieve_show_images_from_tmdb(self, show, img_type):
         # get TMDB configuration info
         tmdb.API_KEY = app.TMDB_API_KEY
         config = tmdb.Configuration()
-        response = config.info()
+        try:
+            response = config.info()
+        except RequestException as e:
```
medariox: Why ```RequestException``` and not ```IndexerUnavailable```?

medusa/metadata/generic.py
p0psicles: IndexerUnavailable is handled by IndexerError. IndexerShowIncomplete is too specific. We could choose to replace all with IndexerException?

medusa/metadata/generic.py
p0psicles: Because this is a direct call to the tmdb_api. And the indexerUnavailable is a Medusa indexer exception.
It could be a good change, to replace this with a call to our own tmdb indexer though.

medusa/metadata/__init__.py
```diff
@@ -46,8 +46,37 @@ def _getMetadataClass(name):
...
+        val = dict.__getitem__(self, key)
+        val.indexer_api = None
+        return val
+
+    def values(self):
+        """Override values method."""
+        providers = []
+        for metadata_provider in self:
+            providers.append(self[metadata_provider])
+        return providers
```
medariox: You could write this as a list comprehension:
```
return [self[metadata_provider] for metadata_provider in self]
```

medusa/metadata/generic.py
medariox: I think it would be better.

medariox:
Approved

Issue #1410, opened 2016-10-31
[APP SUBMITTED]: TypeEError: 'int' object has no attribute '__getitem__'


fernandog:
### INFO
**Python Version**: `2.7.9 (default, Aug 13 2016, 17:56:53) [GCC 4.9.2]`
**Operating System**: `Linux-4.4.27-1-osmc-armv7l-with-debian-8.6`
**Locale**: `UTF-8`
**Branch**: [optimized2](../tree/optimized2)
**Database**: `44.3`
**Commit**: pymedusa/Medusa@ff8dcfa6034366fa6c9d0b526b67e3cdb02fafe6
**Link to Log**: https://gist.github.com/2f6af155ba86624566e4d31c88f94e40
### ERROR
<pre>
2016-10-31 12:46:51 ERROR    SEARCHQUEUE-MANUAL-248835 :: [T411] :: [ff8dcfa] Unknown exception while searching T411. Error: "'int' object has no attribute '__getitem__'"
Traceback (most recent call last):
  File "/home/osmc/Medusa/<a href="../blob/ff8dcfa6034366fa6c9d0b526b67e3cdb02fafe6/medusa/search/core.py#L621">medusa/search/core.py</a>", line 621, in searchProviders
    downCurQuality, manual_search, manual_search_type)
  File "/home/osmc/Medusa/<a href="../blob/ff8dcfa6034366fa6c9d0b526b67e3cdb02fafe6/medusa/providers/generic_provider.py#L182">medusa/providers/generic_provider.py</a>", line 182, in find_search_results
    items_list += self.search(search_string, ep_obj=episode)
  File "/home/osmc/Medusa/<a href="../blob/ff8dcfa6034366fa6c9d0b526b67e3cdb02fafe6/medusa/providers/torrent/json/t411.py#L106">medusa/providers/torrent/json/t411.py</a>", line 106, in search
    results += self.parse(jdata, mode)
  File "/home/osmc/Medusa/<a href="../blob/ff8dcfa6034366fa6c9d0b526b67e3cdb02fafe6/medusa/providers/torrent/json/t411.py#L121">medusa/providers/torrent/json/t411.py</a>", line 121, in parse
    torrent_rows = sorted(unsorted_torrent_rows, key=itemgetter('added'), reverse=True)
TypeError: 'int' object has no attribute '__getitem__'
</pre>

---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

[...]

PR #1857, opened 2016-12-28
Fix BitSnoop

duramato:

Fix https://github.com/pymedusa/Medusa/issues/1410#issuecomment-269413532

 @OmgImAlexis

### DIFF
--- a/medusa/providers/torrent/xml/bitsnoop.py
+++ b/medusa/providers/torrent/xml/bitsnoop.py
@@ -114,7 +114,7 @@ class BitSnoopProvider(TorrentProvider):
                     # unless it is not torcache or we are not using blackhole
                     # because we want to use magnets if connecting direct to client
                     # so that proxies work.
-                    download_url = row.enclosure['url']
+                    download_url = row.enclosure.get('url')
                     if app.TORRENT_METHOD != 'blackhole' or 'torcache' not in download_url:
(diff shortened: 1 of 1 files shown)


medariox:
The issue is that row.enclosure is None, this won't fix it.

duramato:
won't the .get return none? there's a check for download_url right ahead of this

medariox:
If you try .get on a None, you will still get an exception. get is useful if you have e.g. dict, but you don't know if that dict has a specific key. I removed the enclosure now, as it had a toracache link anyway. Let me know what you think.

medusa/providers/torrent/xml/bitsnoop.py
```diff
@@ -101,29 +101,28 @@ def parse(self, data, mode):
...
+                return items
+
+            for element in elements:
                 try:
-                    if not row.category.text.endswith(('TV', 'Anime')):
+                    if not element.category.text.endswith(('TV', 'Anime')):
                         continue
 
-                    title = row.title.text
-                    # Use the torcache link bitsnoop provides,
```
duramato: This should stay

medariox: What should stay and why?

duramato: Ignore me lol, all good here

duramato:
LGTM

PR #1862, opened 2016-12-29
Relative time in manual search updated date

fernandog:
won't touch pubdate now as is undefined how we will handle the naive/aware pubdates from providers

- [x] Add relative time using JS @OmgImAlexis

### DIFF
--- a/medusa/search/manual.py
+++ b/medusa/search/manual.py
@@ -27,7 +27,7 @@ from dateutil import parser
 from .queue import FORCED_SEARCH_HISTORY, ForcedSearchQueueItem
 from .. import app, db, logger
 from ..common import Overview, Quality, cpu_presets, statusStrings
-from ..helper.common import enabled_providers, pretty_date, pretty_file_size
+from ..helper.common import enabled_providers, pretty_file_size
 from ..sbdatetime import sbdatetime
 from ..show.show import Show
 from ..show_name_helpers import containsAtLeastOneWord, filterBadReleases
@@ -276,8 +276,7 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
         cached_results = [dict(row) for row in sql_total]
         for i in cached_results:
             i['quality_name'] = Quality.split_quality(int(i['quality']))
-            i['time_pretty'] = pretty_date(datetime.fromtimestamp(i['time']))
-            i['time'] = datetime.fromtimestamp(i['time']).strftime(app.DATE_PRESET + ' ' + app.TIME_PRESET)
+            i['time'] = datetime.fromtimestamp(i['time'])
             i['release_group'] = i['release_group'] or 'None'
             i['provider_img_link'] = 'images/providers/' + i['provider_image'] or 'missing.png'
             i['provider'] = i['provider'] if i['provider_image'] else 'missing provider'
--- a/medusa/search/manual.py
+++ b/medusa/search/manual.py
@@ -27,7 +27,7 @@ from dateutil import parser
 from .queue import FORCED_SEARCH_HISTORY, ForcedSearchQueueItem
 from .. import app, db, logger
 from ..common import Overview, Quality, cpu_presets, statusStrings
-from ..helper.common import enabled_providers, pretty_file_size
+from ..helper.common import enabled_providers, pretty_date, pretty_file_size
 from ..sbdatetime import sbdatetime
 from ..show.show import Show
 from ..show_name_helpers import containsAtLeastOneWord, filterBadReleases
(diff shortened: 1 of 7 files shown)


fernandog:
no one interested. closing

fernandog:
still waiting some review here. PR is pretty simple

medusa/search/manual.py
```diff
@@ -276,7 +276,8 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
         cached_results = [dict(row) for row in sql_total]
         for i in cached_results:
             i['quality_name'] = Quality.split_quality(int(i['quality']))
-            i['time'] = datetime.fromtimestamp(i['time']).strftime(app.DATE_PRESET + ' ' + app.TIME_PRESET)
+            i['time'] = datetime.fromtimestamp(i['time'])
+            i['time_pretty'] = pretty_date(i['time'])
```
labrys: Don't create a new variable.  pretty_date should be applied in the UI (and imo only if a pretty date is prefered over absolute)

views/snatchSelection.mako
```diff
@@ -148,7 +148,7 @@
                             </td>
                             <td class="col-size">${hItem["pretty_size"]}</td>
                             <td align="center">${hItem["provider_type"]}</td>
-                            <td class="col-date">${hItem["time"]}</td>
+                            <td class="col-date">${hItem["time_pretty"]}</td>
```
labrys: This is where the pretty_date should be applied.

medusa/search/manual.py
fernandog: @labrys my goal was to keep the raw time so in case we need it for some reason.

views/snatchSelection.mako
fernandog: replied in the other message

medusa/search/manual.py
duramato: Yes no point in losing accuracy, the universal time format can still be show on a title tooltip or even a qTip tooltip.

labrys: @fernandog I know you were also keeping the original date.  My point was that creating a variable for the pretty date is unnecessary, the function should just be applied to the UI.  Also I think it should be treated like the option for series times to be in local time or network timezone.

views/partials/home/banner.mako
```diff
@@ -139,7 +139,8 @@
                 <span style="display: none;">${cur_show.name}</span>
                 <div class="imgbanner banner">
                     <a href="home/displayShow?show=${cur_show.indexerid}">
-                    <img asset="show/${cur_show.indexerid}?type=banner" class="banner" alt="${cur_show.indexerid}" title="${cur_show.name}"/>
+                    <img src="show/${cur_show.indexerid}?type=banner" class="banner" alt="${cur_show.indexerid}" title="${cur_show.name}"/>
```
fernandog: @p0psicles pycharm set asset here is not allowed and missing `src`.  were you that added `asset` here?

OmgImAlexis: I added asset and it's documented. /show isn't a valid path it's for the API. Leave this.

views/snatchSelection.mako
```diff
@@ -141,7 +141,7 @@
                             <td class="col-size">${hItem["pretty_size"]}</td>
                             <td align="center">${hItem["provider_type"]}</td>
                             <td class="col-date">
-                                <span data-qtip-my="top left" data-qtip-at="bottom left" title='${hItem["time"]}'>${hItem["time_pretty"]}</span>
+                                <span data-qtip-my="top left" data-qtip-at="bottom left" title='${hItem["time"]}' class="addQTip">${hItem["time_pretty"]}</span>
```
labrys:
```
<span data-qtip-my="top left" data-qtip-at="bottom left" title='${hItem["time"]}' class="addQTip">${pretty_date(hItem["time"])}</span>
```

fernandog:
@OmgImAlexis added. I didn't touched pubdate as its a hole new problem (pubdate PR)

the only thing i didn't liked is the "about"  in the "about 5 hours ago"

OmgImAlexis:
If need be you can just change that, really I've been wanting to change that for a while.

In `static/js/common/init.js` just remove the about on LN 99.

fernandog:
problem of using that in JS, while the table is being updated, it shows the raw date

OmgImAlexis:
That's to be to expected since we currently just pull in the whole div with the table in it. This will be fixed in vue and it's not that big of a deal if it's like this for now since we already have weird flashes on this page from just reloading the results.

static/js/common/init.js
```diff
@@ -95,8 +95,8 @@ MEDUSA.common.init = function() {
             prefixFromNow: 'In ',
             suffixAgo: 'ago',
             suffixFromNow: '',
-            seconds: 'less than a minute',
-            minute: 'about a minute',
+            seconds: 'few seconds',
```
OmgImAlexis: this should be "a few seconds" otherwise you'll get "few seconds ago" or "in few seconds".

PR #1864, opened 2016-12-29
Remove proper strings from providers which return "propers" without them

duramato:

Some providers return the same results with `Show S01E01` or `Show S01E01 PROPER`, so this makes them just use `Show S01E01` propers, as they will still return "proper" results

### DIFF
--- a/medusa/providers/torrent/html/freshontv.py
+++ b/medusa/providers/torrent/html/freshontv.py
@@ -58,7 +58,7 @@ class FreshOnTVProvider(TorrentProvider):
         }
 
         # Proper Strings
-        self.proper_strings = [' ']
+        self.proper_strings = ['']
 
         # Miscellaneous Options
         self.freeleech = False
--- a/medusa/providers/torrent/html/freshontv.py
+++ b/medusa/providers/torrent/html/freshontv.py
@@ -58,7 +58,7 @@ class FreshOnTVProvider(TorrentProvider):
         }
 
         # Proper Strings
-        self.proper_strings = ['PROPER', 'REPACK', 'REAL']
+        self.proper_strings = [' ']
 
         # Miscellaneous Options
         self.freeleech = False
--- a/medusa/providers/torrent/html/freshontv.py
+++ b/medusa/providers/torrent/html/freshontv.py
@@ -58,6 +58,7 @@ class FreshOnTVProvider(TorrentProvider):
         }
 
         # Proper Strings
+        # Provider always returns propers and non-propers in a show search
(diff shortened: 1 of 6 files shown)


duramato:
If any one knows of any providers that do this please add them here or tell me. You will know if they behave like this if they return normal releases on a proper search.

medusa/providers/torrent/html/freshontv.py
```diff
@@ -58,7 +58,7 @@ def __init__(self):
         }
 
         # Proper Strings
-        self.proper_strings = ['PROPER', 'REPACK', 'REAL']
+        self.proper_strings = [' ']
```
medariox: The string needs to be empty, otherwise a space will be added.

medusa/providers/torrent/html/tvchaosuk.py
```diff
@@ -50,6 +50,7 @@ def __init__(self):
         }
 
         # Proper Strings
+        self.proper_strings = [' ']
```
medariox: Same.

fernandog:
@duramato i didn't understood this.

duramato:
There are sites which return the same results when searching with proper tag as when without, so there's no point in using them for that site.

fernandog:
@duramato ok, got it. i would add a comment for future reference:

```python  
# Proper Strings
# Provider always returns propers and non-propers in a show search
self.proper_strings = ['']
```

p0psicles:
Can you add a list of TODO"s, or use labels, so we can track the progress a little.

duramato:
If any one doesn't know of any other providers like this it's concluded.

medusa/providers/torrent/html/freshontv.py
```diff
@@ -58,7 +58,8 @@ def __init__(self):
         }
 
         # Proper Strings
-        self.proper_strings = ['PROPER', 'REPACK', 'REAL']
+        # Provider always returns propers and non-propers in a show search
+        self.proper_strings = ['']
```
p0psicles: any reason we need a list with empty string here? and not just [] ?

duramato: There is a loop on all the items in the list when doing a proper search, no items to loop for there won't be a search.

p0psicles: Yes, I would expect that?! If there are no propers configured, it shouldnt search for it?

duramato: But we want to do a search, if we didn't want it would be []

PR #1867, opened 2016-12-30
Fix NameError: 'allowed_qualities' is not defined

fernandog:
@p0psicles you were right. I forgot to double check

the code inside the script block can't be used in the content block.

@OmgImAlexis fixed


fernandog:
approve please?

PR #1851, opened 2016-12-27
Fix https://github.com/pymedusa/Medusa/issues/1515

duramato:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

Specials have `0` as the season

image
image

PR #1879, opened 2017-01-02
Fix post processor not processing episodes in rare cases, removed som…

medariox:
…e terrible code

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -516,7 +516,7 @@ class PostProcessor(object):
         """
         show = season = version = airdate = quality = None
         episodes = []
-        name_list = [self.nzb_name, self.file_name, self.rel_path]
+        name_list = [self.nzb_name, self.file_name, self.rel_path, self.folder_path]
 
         for counter, name in enumerate(name_list):
 
@@ -627,7 +627,7 @@ class PostProcessor(object):
         try:
             parse_result = NameParser().parse(name)
         except (InvalidNameException, InvalidShowException) as error:
-            logger.log(u"Couldn't parse file name: {0}. Error: {1}".format(name, error), logger.DEBUG)
+            self._log(u'{0}'.format(error), logger.DEBUG)
             return to_return
 
         if parse_result.is_air_by_date:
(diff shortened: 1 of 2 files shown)


medusa/post_processor.py
```diff
@@ -510,51 +510,36 @@ def _add_to_anidb_mylist(self, file_path):
...
             show = cur_show
 
-            if att_num < (len(attempt_list) - 1):
-                if common.Quality.qualityStrings[cur_quality] == 'Unknown':
-                    continue
-            quality = cur_quality
+            if cur_season is not None:
+                season = cur_season
+
+            if cur_episodes:
```
duramato: doesn't it need to be `if cur_episodes is not None:` too? otherwise specials won't get PPed ie:
`The.Bachelor.S21E00.Countdown.to.Nick.720p.HDTV.x264-ALTEREGO` or examples from here https://github.com/pymedusa/Medusa/pull/1851

p0psicles: I suspect it's a list.

medariox: @p0psicles Is right, all specials are correctly processed.

medusa/post_processor.py
```diff
@@ -510,51 +510,36 @@ def _add_to_anidb_mylist(self, file_path):
...
     def _find_info(self):
         """
-        For a given file try to find the showid, season, and episode.
+        For a given file try to find the show, season, epsiodes, version and quality.
 
-        :return: A (show, season, episodes, quality, version) tuple
+        :return: A (show, season, episodes, version, quality) tuple
         """
-        show = season = quality = version = None
+        show = season = version = quality = None
```
labrys: This is a really bad practice.  Should always prefer:
```python
show = None
season = None
....
```

p0psicles: As long as you don't assign a mutable type, it's okay.
So x = y = z = None, is okay.
x = y = z = [], is dangerous.

labrys: Which is why it's a bad practice.

PR #1888, opened 2017-01-03
fix history search

OmgImAlexis:
@duramato apparently this was way more broken than you said


duramato:
It's still not working?

This is:
image
This isn't:
image
Neither is this way:
image

PR #1899, opened 2017-01-04
fix nzbget test

OmgImAlexis:
This fixes nzbget using https at all times in the test.

We need the True/False section as current values are stored as on/off so next time the user runs this it'll fix it all up.


medusa/server/web/home/handler.py
```diff
@@ -201,6 +201,10 @@ def testSABnzbd(host=None, username=None, password=None, apikey=None):
 
     @staticmethod
     def testNZBget(host=None, username=None, password=None, use_https=False):
+        if use_https == 'false':
+            use_https = False
+        else:
+            use_https = True
```
p0psicles: use_https = bool(use_https) didn't work?

medusa/server/web/home/handler.py
```diff
@@ -201,6 +201,10 @@ def testSABnzbd(host=None, username=None, password=None, apikey=None):
...
         connected_status = nzbget.testNZB(host, username, password, use_https)
```
p0psicles: or put it here?

OmgImAlexis: The current values are 'on' 'off', they're now 'true' 'false'. Python sees all 4 values as truthy values and sets it as true. 😞

p0psicles: check: https://github.com/pymedusa/Medusa/blob/develop/medusa/config.py#L397

PR #1944, opened 2017-01-07
Feature/poc request police

p0psicles:
### PolicedSession features 
- [x] Newznab indexers (more notably, api limits)
- [ ] API torrent sites (same as previous one, api limits)
~~- [ ] Torrent sites (Time Outs, 5xx codes, etc)~~
~~- [ ] XEM (if requests fail, it keeps hammering in which is no bueno)~~
~~- [ ] Show Indexers?~~
- [x] Reserve x amount of api hits for daily search

### Known Errors
~~* Enable api hit cooldown settings is saved, but then not checked when opening the page.~~


medusa/providers/nzb/newznab.py
```diff
@@ -80,6 +81,10 @@ def __init__(self, name, url, key='0', cat_ids='5030,5040', search_mode='eponly'
 
         self.cache = tv_cache.TVCache(self, min_time=30)  # only poll newznab providers every 30 minutes max
 
+        self.request_police = RequestPolice()
+        self.request_police.enabled_police_request_hooks = [self.request_police.request_check_nzb_api_limit]
+        self.request_police.enabled_police_response_hooks = [self.request_police.response_check_nzb_api_limit]
```
p0psicles: The idea is to configure a RequestPolice object, for  which you can enable request and response methods.

**Response method**
The response method is used to check if a certain threshold has been breached, or maybe a specific response code has been returned. You can pretty much, create the checks how you like them.

**Request method**
The request method should match the response method. And it checks for certain criteria. Like for the request_check_nzb_api_limit, it checks if the cooldown has been passed. If not it will throw an exception, which is catched. Then get_url will return from the function.

OmgImAlexis:
From Slack.

> Alright. Basically I have a limit of 2.5k I want at least 100 api calls saved for daily so new episodes get snatched and Medusa can use the other 2.4k for whatever it feels like. I'd like this to reset at midnight or something. I don't know when dog resets exactly.

medusa/helpers.py
```diff
@@ -1189,7 +1189,19 @@ def get_url(url, post_data=None, params=None, headers=None, timeout=30, session=
...
+    # Get RequestPolice instance object.
+    rpolice = kwargs.pop(u'rpolice', None)
+
+    from helper.request_police import RequestPoliceException
+
     try:
+        if rpolice:
+            try:
+                [rpolice_check() for rpolice_check in rpolice.enabled_police_request_hooks]
+            except RequestPoliceException as e:
```
p0psicles: This might look a little useless. But i've put it here as a proof of concept.
Also if not catched here, it will be swallowed by the Exception down below.

labrys: These types of checks for API limit, etc should be part of the service's session.  You should only use a response hook as a last resort.  A response hook can only be checked after the request has already been sent and only applies the condition if the hook has already been attached.  If you want to check before sending a request, a better approach would be to subclass requests.Request and implement the functionality there.

p0psicles: That's why i've split them up in response and request hooks.

labrys: What I mean is instead of creating a bunch of hooks, create a custom session class from requests.Session like:
```
class PolicedSession(requests.Session):
```
And subclass the methods there.  That way they are all a part of the session directly.  Then you can use the PolicedSession in place of the requests.Session.  You can even hook the session.auth to get the pre-requisite API limits etc from the service.

labrys: In the init you can then do a `self.hook = ...` etc.   The required hooks for a service can then be passed as params to the PolicedSession's init.  As well as any user preferences such as API limit to reserve, etc.

medusa/helper/request_police.py
```diff
@@ -0,0 +1,356 @@
...
+        self.mount('https://', adapter)
+        self.cache_controller = adapter.controller
+        self.headers.update({'User-Agent': USER_AGENT, 'Accept-Encoding': 'gzip,deflate'})
+
+        # This can be used to attach a police object
+        self.police = police
+
+    @classmethod
+    def cloudflare_hook(cls, r, **kwargs):
+
```
p0psicles: This was it @labrys. Hope it's worth my time! Man. Pain here, was re-setting the new cookies on a PreparedRequest object!

fernandog:
@p0psicles not sure if is an issue or not:

image


image


Do we need a protected member?
image

p0psicles:
@fernandog good eye. Yes that's not gonna work. I'll remove it all together, as i'm not gonna use it anytime soon.

labrys:
@fernandog that was just an example of how it would be implemented, and you are correct it would not work as is, `self.reset_time()` would be a function to get the reset time, for example from the `X-RateLimit-Reset` header or a user entered value (for example the daily reset time for a providers API limit) and would be application specific to the subclass (thus why I didn't explicitly define it).  A default method could be something like X time since last reset or something.

labrys:
Session should be modularized:  Make a medusa.session package for session-related code.
Don't try to do too much with the base session

p0psicles:
Those are two things. Making a package.. sure np.
"don't try to do too much"? Can you give me an example where i'm trying todo too much? And maybe provide an alternative?

labrys:
They are sort of related.  After modularizing the base session would be simpler and it would provide more flexibility because you could choose the components you need to a session instead of trying to include them all from the start.  I'll push a commit with the start of a modularization, feel free to revert it if you wish, but I think its a better way to go.  Keep in mind this is only a start to the modularization but  it should be enough for you to get the idea of where I'm going with this.

fernandog:
@OmgImAlexis are you still testing this?

p0psicles:
@labrys, what's next on this one, to get it to a merge-able state?
Shall I just go through my list? Or do you think we should refactor some more?

labrys:
First of all the code I wrote was in parallel to yours (mostly as a concrete example of how the whole approach should be modularized).  

If you agree with my approach we can continue to add additional functionality to it (for example i have a working thread-safe throttling decorator that I plan on adding).  

If you don't agree with this approach feel free to cut it from the PR and continue on with your approach, but I feel this would be a more flexible, simpler, and more powerful alternative.  I plan on pushing the throttling decorator this weekend.

p0psicles:
Your approach is fine. It's the same as what I have only you gave it a better place.
I do want to put a hold on adding features. As we've refactored this much. I think we first should implement the sessions. And then start thinking about adding new functionality. 

You okay with that?

labrys:
I would say its fine to start adding the basic session usage to everything, we can work on adding in the throttling etc later as the methods get completed.  That's the nice thing about modular session design you can plug in the additional functionality on an as-needed basis.

p0psicles:
aah yes!! Only 87 occurrences of get_url in the providers!!

medusa/session/factory.py
```diff
@@ -0,0 +1,26 @@
...
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+def add_cache_control(session, cache_control_config):
+    if not all(['cache_etags' in cache_control_config,
+                'serializer' in cache_control_config,
+                'heuristic' in cache_control_config]):
+        logging.warning('Insufficient paramaters provided for the add_cache_control factory.')
+        return
```
labrys: Instead of a warning and a return None this should be an assert since we want to assert that the code is implemented properly.  Also this could be simplified using set syntax.
```python
def add_cache_control(session, cache_control_config):
    assert {'cache_etags', 'serializer', 'heuristic'}.issubset(cache_control_config.keys), cache_control_config
    ...
```

p0psicles:
@labrys by creating a BaseSession -> MedusaSession -> xSession structure, i'm also complying to your requirement, in making it flexible. By default i'm adding the exception handling to MedusaSession, as that's pretty much the way we've already had it. If you really want to utilize a bare session, use BaseSession.

medusa/scene_exceptions.py
```diff
@@ -28,13 +28,14 @@
...
 from .indexers.indexer_config import INDEXER_TVDBV2
+from .session.core import MedusaSession
 
 exceptionsCache = {}
 exceptionsSeasonCache = {}
 
 exceptionLock = threading.Lock()
 
-xem_session = helpers.make_session()
+xem_session = MedusaSession()
```
p0psicles: Can we disable this check in codacy? I tried disabling it. But I only have the option to disable all naming convention checks.

PR #1963, opened 2017-01-11
Log reason of should_process method and add tests

fernandog:
@OmgImAlexis this should give you the exact reason

TODO: 
- [x] Add more tests. Coded this in 15min so didn't had time to think much about other tests.
- [x] Review method. Some rules are wrong imo

### DIFF
--- a/medusa/server/web/config/subtitles.py
+++ b/medusa/server/web/config/subtitles.py
@@ -28,7 +28,7 @@ class ConfigSubtitles(Config):
 
         return t.render(submenu=self.ConfigMenu(), title='Config - Subtitles',
                         header='Subtitles', topmenu='config',
-                        controller='config', action='subtitles')
+                        controller='config', action='subtitlesPage')
 
     def saveSubtitles(self, use_subtitles=None, subtitles_plugins=None, subtitles_languages=None, subtitles_dir=None, subtitles_perfect_match=None,
                       service_order=None, subtitles_history=None, subtitles_finder_frequency=None,
--- a/static/js/config/subtitles.js
+++ b/static/js/config/subtitles.js
@@ -1,4 +1,4 @@
-MEDUSA.config.subtitles = function() {
+MEDUSA.config.subtitlesPage = function() {
     $.fn.showHideServices = function() {
         $('.serviceDiv').each(function() {
             var serviceName = $(this).attr('id');
--- a/views/config_subtitles.mako
+++ b/views/config_subtitles.mako
@@ -196,13 +196,12 @@ $('#subtitles_dir').fileBrowser({ title: 'Select Subtitles Download Directory' }
                         <h3>Subtitle Providers</h3>
                         <p>Check off and drag the plugins into the order you want them to be used.</p>
                         <p class="note">At least one plugin is required.</p>
-                        <p class="note"><span style="font-size: 16px;">*</span> Web-scraping plugin</p>
                     </div>
                     <fieldset class="component-group-list" style="margin-left: 50px; margin-top:36px;">
                         <ul id="service_order_list">
                         % for curService in subtitles.sorted_service_list():
                             <li class="ui-state-default" id="${curService['name']}">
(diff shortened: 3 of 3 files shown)


medusa/post_processor.py
```diff
@@ -844,18 +844,26 @@ def _should_process(current_quality, new_quality, allowed, preferred):
...
-            return True
+                if new_quality > current_quality:
+                    return True, 'New quality is higher than current Preferred. Accepting quality'
+                else:
+                    return False, 'New quality is same|lower than current Preferred. Ignoring quality'
+            return True, 'New quality is Preferred'
         elif new_quality in allowed:
             if current_quality in preferred:
-                return False
+                return False, 'Current quality is Allowed but we already have a current Preferred. Ignoring quality'
```
p0psicles: If your going to work on this. Might be a good moment to use Exceptions instead of tuples. This line could be rewritten to: 
```python
raise PostProcesQualityException('Current quality is Allowed but we already have a current Preferred. Ignoring quality')
```

tests/test_should_process.py
```diff
@@ -0,0 +1,67 @@
...
+def test_should_process(p):
+    # Given
+    current_quality = p['cur_quality']
+    new_quality = p['new_quality']
+    allowed_qualities = p['allowed_qualities']
+    preferred_qualities = p['preferred_qualities']
+    expected = p['expected']
+
+    # When
+    replace, msg = PostProcessor._should_process(current_quality, new_quality, allowed_qualities, preferred_qualities)
```
p0psicles: And you would rewrite this to:
```python
msg = ''
try:
    PostProcessor._should_process(current_quality, new_quality, allowed_qualities, preferred_qualities)
finally:
    except PostProcesQualityNotReplace as e:
    print 'do stuff when you do not want to replace'
    msg = e.message
except PostProcesQualityReplace as e:
    print 'do stuff when you want to replace'
    msg = e.message

....

Comment (p0psicles) [inline tests/test_should_process.py]: There are also other solutions. You can by default return a message when it should replace. And throw exception when it shouldn't.

Or you can create one Exception with a custom atribute PPQualityAssertion().should_replace.
And always thow this exception. And catch the e.message and e.should_replace.

Comment (fernandog) [inline tests/test_should_process.py]: @p0psicles imo the old code (before my test changes) is still good and we don't need this method but I'm not the author of it. Also I couldn't complete my quality PRs before that method got merged.

I'm just adding the tests because some tests are wrong imo so with them I can show is wrong.

That method is only for **replace existing files** so if current quality is allowed we shouldn't replace it - except user manually force it

Also if current quality is preferred we also shouldn't replace because we don't keep redownloading preferred qualities

By design quality system download:
2 qualities if preferred set ( allowed + preferred or preferred first)
or
1 quality if allowed only

**Comment by fernandog** [comment]
This second commit changes the method in the way i think it should be to be consistent with quality system.

**Comment by fernandog** [comment]
As you can see, the only scenario that method tests returns True is when replacing existing file (allowed quality) with a preferred quality

So old code is still valid and we don't need the method `should_process`

Unless someone tell me another scenario that is should return True
```python
_, preferred_qualities = common.Quality.splitQuality(int(show.quality))
if new_ep_quality not in preferred_qualities:
    raise EpisodePostProcessingFailedException(
        u'File exists and new file quality is not in a preferred '
        u'quality list, marking it unsafe to replace')
else:
    self._log(u'File exists and new file quality is in a preferred quality list, '
              u'marking it safe to replace')
```

p0psicles:
"if we don't search for better preferred we shouldn't PP better preferred"
But we we don't search for better preferred, and still it's asked to postproces it, that means the file was forced right? Then we should still postproces it?

fernandog:
@p0psicles in that scenario, user must force it because it wasn't snatched my Medusa.

If Medusa snatched or manually snatched, it will be PPed no matter what

fernandog:
imo the "not priority" code (not snatched by medusa) should be removed (line 997 to 1032)
not priority files shouldn't be PPed (except if forced)

PR #1977, opened 2017-01-15
Feature/fix show updates2

p0psicles:

The season update tables have been ditched. As all indexer support some sort of season level or episode level update update information. Tvdb has added an attribute 'lastUpdated' to the episode route.
https://trello.com/c/agtdOmNc/24-add-lastupdated-to-api-series-id-episodes-results

This means that for tvdb we can now also get the seasons updated in two stages.
1. Get list of updated shows.
2. For that list of shows, get a list of updated episodes.

The episodes are used to get a list of updated seasons, which we feed the per season show_updater.

TODO:
- [x] Remove leftovers of the indexer_update table (cache.db).


p0psicles:
@labrys pls switch over to this branch. It should fix most of your issues.

labrys:
Switched today.  Also if you could please add a log message of each url requested (like that used in getURL) it would be appreciated.

p0psicles:
@labrys you want me to put a log message for every url call, that's made during a show update and refresh? You do know, that's pretty much 99% of all calls don by medusa when you don't cound the providers right?

labrys:
Yes I realize, but its also a critical aspect to Medusa's function so it would make sense to have it logged.  It should be fairly simple to add and also makes it readily apparent when a search is unnecessarily repeated or fails altogether.

p0psicles:
Well it would be some work, as I need to add the hooks to each get_url call. Instead of going through that now, i'd rater first finish the MedusaSession, and then go through them all. Feel a little like double work.

p0psicles:
"fixed" #1994

medusa/helpers.py
```diff
@@ -1021,7 +1021,7 @@ def validate_show(show, season=None, episode=None):
         if show.dvdorder != 0:
             indexer_api_params['dvdorder'] = True
 
-        t = indexerApi(show.indexer).indexer(**indexer_api_params)
+        t = show.get_indexer_api()
```
labrys: One letter variables should be avoided except in cases where a one letter argument is the most clear (for example using x and y as properties of a cartesian coordinate class)

medusa/tv.py
```diff
@@ -195,13 +195,33 @@ def __init__(self, indexer, indexerid, lang='', quality=None, flatten_folders=No
...
         self.externals = {}
+        self._cached_indexer_api = None
 
         other_show = Show.find(app.showList, self.indexerid)
         if other_show is not None:
             raise MultipleShowObjectsException("Can't create a show if it already exists")
 
         self._load_from_db()
 
+    def get_indexer_api(self, indexer_api=None, refresh=False):
```
labrys: This method is unpythonic.  Having a getter that serves dual purpose as a setter is poor practice.

medusa/show_updater.py
```diff
@@ -151,7 +151,7 @@ def __del__(self):
 
 class ShowUpdate(db.DBConnection):
```
labrys: This entire class seems pointless.  

First of all if you're only getting the last update time of an indexer, making it a class called ShowUpdate seems counter-intuitive as it appears like it's the update of that particular show.  

Secondly a class where just a simple function would do seems unnecessary. 

Since this is an attribute related to an indexer, it seems to me it would make more sense as part of the Indexer code, and as such it would make more sense to make it a property.  
Then you could get the last update time from the indexer_api directly, and if it's None then get it from the database.  That way the only time you need to hit the DB for getting the last update time, is when you first get the property from the indexer.  As currently written that would mean that every time the indexer is recreated (which happens a lot) you would have to hit the database, but as its a property, its lazy and would only do so if you needed that data.  Also if we went to a singleton indexer object for each indexer (which we should) then it should cut that down to only a single get per indexer until the next restart.

p0psicles:
Rebased so I now also have the fixes for the tbn and nfo info.

PR #1982, opened 2017-01-16
Log should_process reason and add test

fernandog:
Re-sending
@p0psicles i did some readings and its ok to return status like this.

Anyway this may be removed soon with new quality system.
At least now we have tests and better logging.

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -867,6 +867,8 @@ class PostProcessor(object):
         :param preferred: Qualities that are preferred
         :return: Tuple with Boolean if the quality should be processed and String with reason if should process or not
         """
+        if current_quality is common.Quality.NONE:
+            return False, 'There is no current quality. Skipping as we can only replace existing qualities'
         if new_quality in preferred:
             if current_quality in preferred:
                 if new_quality > current_quality:
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -844,8 +844,8 @@ class PostProcessor(object):
             level = logger.DEBUG
             logger.log(u'In history: {0}'.format(self.in_history), level)
             logger.log(u'Manual snatch: {0}'.format(self.manually_searched), level)
-            logger.log(u'Existing quality: {0}'.format(old_ep_quality), level)
-            logger.log(u'New quality: {0}'.format(new_ep_quality), level)
+            logger.log(u'Existing quality: {0}'.format(common.Quality.qualityStrings[old_ep_quality]), level)
+            logger.log(u'New quality: {0}'.format(common.Quality.qualityStrings[new_ep_quality]), level)
             logger.log(u'Proper: {0}'.format(self.is_proper), level)
 
             # If in_history is True it must be a priority download
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -857,6 +857,9 @@ class PostProcessor(object):
         Determine if a quality should be processed according to the quality system.
 
         This method is used only for replace existing files
+        Despite quality system rules (should_search method), in should_process method:
+         - New higher Allowed replaces current Allowed (overrrides rule where Allowed is final quality)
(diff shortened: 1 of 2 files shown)


tests/test_should_process.py
```diff
@@ -0,0 +1,67 @@
...
+        'preferred_qualities': [Quality.HDWEBDL, Quality.HDBLURAY],
+        'expected': True
+    },
+    {  # p5: New quality not in quality system: no
+        'cur_quality': Quality.HDTV,
+        'new_quality': Quality.SDTV,
+        'allowed_qualities': [Quality.HDTV],
+        'preferred_qualities': [Quality.HDWEBDL, Quality.HDBLURAY],
+        'expected': False
+    },
```
fernandog: @OmgImAlexis @duramato any other scenario that I'm missing?

tests/test_should_process.py
```diff
@@ -0,0 +1,74 @@
...
+    {  # p6: Preferred lower quality replacing current higher allowed: yes
+        'cur_quality': Quality.HDWEBDL,
+        'new_quality': Quality.HDTV,
+        'allowed_qualities': [Quality.HDWEBDL],
+        'preferred_qualities': [Quality.HDTV],
+        'expected': True
+    },
+])
```
fernandog: @OmgImAlexis @duramato any other scenario that I'm missing?

medusa/post_processor.py
```diff
@@ -856,22 +855,37 @@ def _should_process(current_quality, new_quality, allowed, preferred):
...
+
         :param current_quality: The current quality of the episode that is being processed
         :param new_quality: The new quality of the episode that is being processed
         :param allowed: Qualities that are allowed
         :param preferred: Qualities that are preferred
-        :return: True if the quality should be processed, False or None otherwise.
+        :return: Tuple with Boolean if the quality should be processed and String with reason if should process or not
         """
+        if current_quality is common.Quality.NONE:
+            return False, 'There is no current quality. Skipping as we can only replace existing qualities'
```
medariox: This is useless and should be removed.

medusa/post_processor.py
```diff
@@ -856,22 +855,37 @@ def _should_process(current_quality, new_quality, allowed, preferred):
...
+                    return False, 'New quality is same|lower than current Preferred. Ignoring quality'
+            return True, 'New quality is Preferred'
         elif new_quality in allowed:
             if current_quality in preferred:
-                return False
+                return False, 'Current quality is Allowed but we already have a current Preferred. Ignoring quality'
             elif current_quality not in allowed:
-                return True
-            return new_quality > current_quality
+                return True, "New quality is Allowed and we don\'t have a current Preferred. Accepting quality"
```
medariox: No ```\``` needed here.

medusa/post_processor.py
```diff
@@ -1013,12 +1027,17 @@ def process(self):
                     self._log(u'New file is a PROPER, marking it safe to replace')
                     self.flag_kodi_clean_library()
                 else:
-                    allowed, preferred = show.current_qualities
-                    if not self._should_process(old_ep_quality, new_ep_quality, allowed, preferred):
+                    allowed_qualities, preferred_qualities = show.current_qualities
+                    self._log(u'Checking if new quality {0} should reaplace current quality: {1}'.format
```
medariox: ```replace```

p0psicles:
@fernandog i think all medariox comments are valid, can you create a new PR, to resolve these?

medusa/post_processor.py
fernandog: Its not useless. It make sure it will run only when there is valid quality. Fail safe.

medariox: How can we have an "invalid" quality? If the quality is "invalid" it will be Unknown.

fernandog: did you check the code before you say that?

medariox: Yes. Like I said, it's useless and needs to be removed.

PR #1988, opened 2017-01-17
Search term should be URL-encoded params for search

labrys:
Search term should be URL-encoded
Use params for search
Fix no results returned (thanks to @fernandog #1992)

- [ ] PR is based on the DEVELOP branch
- [ ] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [ ] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

Fixes: #1985

PR #1992, opened 2017-01-17
Fix T411 check for no results in daily and episode search

fernandog:
@duramato already tested but please double check:

```
2017-01-17 14:32:51 DEBUG    SEARCHQUEUE-MANUAL-260449 :: [T411] :: [a3dcf6e] Added RSS item: [Vikings.S04E12.720p.HDTV.x264.AC3-GOBO2S] to cache: [t411]
2017-01-17 14:31:54 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [T411] :: [a3dcf6e] Added RSS item: [the.middle.s07e06.french.720p.hdtv.x264-sh0w] to cache: [t411]
```

### DIFF
--- a/medusa/providers/torrent/json/t411.py
+++ b/medusa/providers/torrent/json/t411.py
@@ -119,12 +119,18 @@ class T411Provider(TorrentProvider):
         """
         items = []
 
-        unsorted_torrent_rows = data.get('torrents') if mode != 'RSS' else data
+        if not data:
+            logger.log('No data returned from provider', logger.DEBUG)
+            return items
 
-        if not unsorted_torrent_rows or not isinstance(unsorted_torrent_rows, dict):
+        # Data is `dict` when episode search and `list` when RSS search
+        if not ((isinstance(data, dict) and data.get('torrents')) or
+                isinstance(data, list) and len(data)):
             logger.log('Data returned from provider does not contain any {0}torrents'.format(
                 'confirmed ' if self.confirmed else ''), logger.DEBUG)
             return items
+
+        unsorted_torrent_rows = data.get('torrents') if mode != 'RSS' else data
 
         torrent_rows = sorted(unsorted_torrent_rows, key=itemgetter('added'), reverse=True)
 


medusa/providers/torrent/json/t411.py
```diff
@@ -119,13 +119,15 @@ def parse(self, data, mode):
         """
         items = []
 
-        unsorted_torrent_rows = data.get('torrents') if mode != 'RSS' else data
```
labrys: This will make either extract the torrents from the dict if its a standard search, or it will return the list if its a daily search.  So this should remain where it is.

medusa/providers/torrent/json/t411.py
```diff
@@ -119,13 +119,15 @@ def parse(self, data, mode):
...
-
-        if not unsorted_torrent_rows or not isinstance(unsorted_torrent_rows, dict):
```
labrys: The second part of this if statement is the real issue.  The if statement should just be `if not unsorted_torrent_rows:`.  The `unsorted_torrent_rows` would never be a dict so that's why this is failing.

medusa/providers/torrent/json/t411.py
```diff
@@ -119,13 +119,15 @@ def parse(self, data, mode):
...
+        # Data is `dict` when episode search and `list` when RSS search
+        if not data or not ((isinstance(data, dict) and data.get('torrents')) or
+                            isinstance(data, list) and len(data) > 0):
```
labrys: See my previous comments as to the correct fix.  That aside, here you are checking `len(data) > 0` which first of all could be simplified to `len(data)` since that evaluates to `False` if `data == 0`.  Secondly, it would never be false at this point anyway since you already checked `if not data`.

labrys: Thanks for identifying the source of the error (the if statement logic).  I'll add the changes to my t411 url-encoding fix. #1988

medusa/providers/torrent/json/t411.py
fernandog: Yes it will be. As I commented the code.

medusa/providers/torrent/json/t411.py
fernandog: This code is to check if we have results. Not if we have data. You missed the point.

fernandog: You fixed the URL encode I fixed another thing. Mine is already tested. Dont see a valid reason to pick my fix

fernandog: @labrys

PR #1995, opened 2017-01-18
Made cdn.pymedusa.com configurable in app.

p0psicles:

@OmgImAlexis, here you go.,
Can you double check I got everything?

### DIFF
--- a/medusa/app.py
+++ b/medusa/app.py
@@ -24,7 +24,8 @@ LOG_FILENAME = 'application.log'
 CONFIG_INI = 'config.ini'
 GIT_ORG = 'pymedusa'
 GIT_REPO = 'Medusa'
-CHANGES_URL = 'https://cdn.pymedusa.com/news/CHANGES.md'
+BASE_PYMEDUSA_URL = 'https://cdn.pymedusa.com'
+CHANGES_URL = '{base_url}/news/CHANGES.md'.format(base_url=BASE_PYMEDUSA_URL)
 APPLICATION_URL = 'https://github.com/{org}/{repo}'.format(org=GIT_ORG, repo=GIT_REPO)
 DONATIONS_URL = '{0}/wiki/Donations'.format(APPLICATION_URL)
 WIKI_URL = '{0}/wiki'.format(APPLICATION_URL)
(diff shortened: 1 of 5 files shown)


duramato:
What is the benefit of this? I see this as a superfluous config that can allows users to break Medusa's behaviour (scene exceptions etc) by changing it to something not our "CDN".

p0psicles:
Because without this change, users can't change source code? I did it because it was a request from OMG. As she'd like to have the domain in one location. As she was experimenting with other hostings.

duramato:
But users **shouldn't** change it, we can't foresee possible issues if people try to use something not our "CDN". If it's for developing it can be changed in the source code.

labrys:
Personally I think it *should * be a config var (not something set in the UI ofc) as then it separates logic from data.... the URL is just data, data should be easy to change, the code should contain logic only as much as possible.

p0psicles:
well for most data yes, only this is so tightly bound to the application, that I don't really see a use for making this available as configuration option.

labrys:
Well for example, if the CDN is down for any reason a user could fix it by pointing to the github.io repo which is the source.

p0psicles:
I feel we shouldn't suggest this as an option to users, as 1. it will only confuse them, 2. users trying to change it but make a mistake will only increase the amount of issue reports, and 3. one of the points of using a CDN is it's higher availability.

p0psicles:
Well I have more reasons, but i'd hope 3 is enough ;-)

labrys:
Its not an option for the users.  Its a configuration file.  I also advocate splitting the configurations as well.  For example user, provider, and application configs.  As this would be an app config it wouldn't normally be touched by the user.

labrys:
As a further example X11 config on linux.  The conf has settings that generally should not be changed by a user and its recommended that you use a utility to change them to avoid problems (the medusa UI in our case).  If you go against advice, it's on you.

fernandog:
Its hidden setting. Only editing config.ini users will be able to change it.
Lets move on as there is other stuff that needed to be tested/coded.

p0psicles:
Yeah your right fernandog, dropping the discussion.

Issue #1978, opened 2017-01-15
NyaaTorrent does not give back results when using non-zero padded episode numbers


p0psicles:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:** Develop
**OS:** windows/linux
**What you did:** Searched for show 'Youjo Senki' episode 2.
**What happened:** NyaaTorrent gave back no results.
**What you expected:** Around 5 results.
**Logs:**
```
PASTE LOGS HERE
```
It seems that when searching nyaa, the show + non padded absolute ep number is used. Meaning for Youjo Senki, is searches using the string: 'Youjo+Senki+2'. All releases for Youjo Senki are released with a padded episode like: ```Youjo Senki Saga of Tanya the Evil - 01 [720p][AAC][07B357FF].mkv```.
There are no wildcards available that we can use for this.

IDerr:
Hmm, it's searching with 02 for ao no exorcist and it's getting snatched, it's weird...
And all releases for Ao no exorcist are padded...

p0psicles:
Yes that's by design. As I made an exception for season exceptions. So if it's a seasion exception it now uses 02. If it's not, it will use absolute number, and now that is '2'. But nyaa doesn't like that.

IDerr:
Yeah, not really Nyaa, but it's like a convention to name episodes like that.

p0psicles:
Sickrage has recently also made the change to double digit. Well they moved to three digit before that, don't why youd want that. but sure. I think we should do the same.

IDerr:
In fact, the real way to name it, it's with 3 digits, but no one respect that.

IDerr:
It's more like : 
Showname 01-09
Showname 10-999

p0psicles:
They did some more canges like first try double digit, then fallback to three digit padding.
https://github.com/SickRage/SickRage/commit/9b6c552e9fb496d6df5ad741a4a6e1001d164739
Apparently some providers require that, as some shows release in 001, 002, 003.

IDerr:
I think that try both (2 and 3 digits) like them is a good idea.

h3llrais3r:
> They did some more canges like first try double digit, then fallback to three digit padding.
SickRage/SickRage@9b6c552
[...]

PR #1999, opened 2017-01-19
Fix for first scene search

h3llrais3r:
@p0psicles  Possible fix for https://github.com/pymedusa/Medusa/issues/1978#issuecomment-273605769

### DIFF
--- a/medusa/name_parser/parser.py
+++ b/medusa/name_parser/parser.py
@@ -139,14 +139,14 @@ class NameParser(object):
 
         elif result.show.is_anime and result.is_anime:
             logger.debug("Scene numbering enabled show '{name}' is anime", name=result.show.name)
-            scene_season = scene_exceptions.get_scene_exception_by_name(result.series_name)[1]
+            result.scene_season = scene_exceptions.get_scene_exception_by_name(result.series_name)[1]
             for absolute_episode in result.ab_episode_numbers:
                 a = absolute_episode
 
                 if result.show.is_scene:
                     a = scene_numbering.get_indexer_absolute_numbering(result.show.indexerid,
                                                                        result.show.indexer, absolute_episode,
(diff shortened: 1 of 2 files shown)


p0psicles:
Hi @h3llrais3r , has your issue been resolved now? And was that using your fix or mine?

h3llrais3r:
@p0psicles, the problem was solved with my fix. 
However, the work you did seems very promising, but I didn't have the time yet to test it.

p0psicles:
I think that's more of a feature. Like it will also parse and accept batches. Like ```[BKT] Naruto Shippuden 321-348 (Grande guerra mondiale dei ninja, Sasuke ed Itachi)[Regolare]```. Searching for 325, would also accept this as a result.

p0psicles:
I just tested your fix. That does seem to fix it. I'm going to keep this branch, for when the refactor has been finished.

h3llrais3r:
@p0psicles: updated previous comment with reference to the code.
In fact I don't think my fix caused the problem. It was probably the case before because we are checking if the episode number appears in the list of parsed episode numbers. Because the batch is also found for a normal episode search, and the episode number is in it, it snatches it.

p0psicles:
That's one the issues i'm trying to fix. The logic is currently for sick* way all over the place. So they all have search logic in determining if something should be a result, or if not then should be cached.

I'm removing the logic for "if cached". So then we can simplify to only evaluate if should be a valid result.

p0psicles:
I've fixed it in a similar way. I'm making sure i'm translating scene_season back to indexers season, and am comparing those. Tested that with all of the scene seasons of jojo's 2012. Still doing some more tests with other shows.

h3llrais3r:
Nice work. Will check your PR with your changes.

PR #2001, opened 2017-01-20
Fix manageTorrents not showing up if client url is not https and glob…

fernandog:
…al setting enable_https is set

enable_https has nothing to do with client url.
for example transmission doesn't have https webserver


fernandog:
@p0psicles can you quick review this please?

p0psicles:
Why should we fix this in code? Shouldn't the user just disable https, if he's not using https?

fernandog:
@p0psicles that config is related to medusa webserver (if you have local cert or not). There is nothing to do with clients. got it ?

PR #2011, opened 2017-01-21
Don't refresh xem data when loading home page even data is outdated

fernandog:
@p0psicles @duramato 

sorry about that @ripper1967

### DIFF
--- a/medusa/scene_numbering.py
+++ b/medusa/scene_numbering.py
@@ -367,7 +367,7 @@ def get_scene_numbering_for_show(indexer_id, indexer):
     return result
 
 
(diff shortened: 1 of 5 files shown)


p0psicles:
I think this is okay for now, until someone starts on scheduling the updates, and removing the persistent next_refresh data in db. Just perform a refresh on show_update. And add a button for manually refreshing it, would be enough, and remove allot of complexity.

fernandog:
Imo thos should be done in show refresh only (3am)

labrys:
@fernandog I agree.  With possibly an option to force it later.

ripper1967:
@fernandog i agree an option to force it later would be ok.

PR #2012, opened 2017-01-22
Fix BTN search

duramato:

- Fixes BTN search by name
- Fixes Scene names with colons, ie: `MythBusters: The Search` released as `MythBusters-The Search` by the Scene

### DIFF
--- a/medusa/helpers.py
+++ b/medusa/helpers.py
@@ -4,7 +4,7 @@
 # This file is part of Medusa.
 #
 # Medusa is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published byFRWL
+# it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
 #
--- a/medusa/helpers.py
+++ b/medusa/helpers.py
@@ -4,7 +4,7 @@
 # This file is part of Medusa.
 #
 # Medusa is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
+# it under the terms of the GNU General Public License as published byFRWL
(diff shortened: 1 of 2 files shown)


medusa/helpers.py
```diff
@@ -716,7 +716,7 @@ def sanitize_scene_name(name, anime=False):
         name = name.replace(x, '')
 
     # tidy up stuff that doesn't belong in scene names
-    name = name.replace('- ', '.').replace(' ', '.').replace('&', 'and').replace('/', '.')
+    name = name.replace('- ', '.').replace(' ', '.').replace('&', 'and').replace('/', '.').replace(': ', '-')
```
labrys: Is there any benefit of replacing with `-` over just a space?  I would think with a space it would return results for any of:
- `Mythbusters: The Search` 
- `Mythbusters-The Search`
- `Mythbusters, The Search`
- `Mythbusters The Search`

duramato: I don't know if all sites' search engines return results with a space in a character's place, on the other hand with the dash i do know they will, becouse thats how the releases are named.

labrys: Maybe for SCENE releases, but that doesn't mean all releases are named that way.  Also most sites are keyword searches so it *should* be safer to use ' ' over '-'.

medusa/helpers.py
```diff
@@ -4,7 +4,7 @@
 # This file is part of Medusa.
 #
 # Medusa is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
+# it under the terms of the GNU General Public License as published byFRWL
```
fernandog: @duramato typo?

fernandog:
Anything else to commit? All done?

duramato:
Concluded label is set

PR #2015, opened 2017-01-23
Fix main sanity check not running update_old_propers DUH!

fernandog:

### DIFF
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -47,6 +47,7 @@ class MainSanityCheck(db.DBSanityCheck):
         self.fix_show_nfo_lang()
         self.convert_tvrage_to_tvdb()
         self.convert_archived_to_compound()
+        self.update_old_propers()
         self.fix_subtitle_reference()
         self.clean_null_indexer_mappings()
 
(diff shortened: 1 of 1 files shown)


medusa/databases/main_db.py
```diff
@@ -260,6 +262,7 @@ def fix_unaired_episodes(self):
                                    [common.UNAIRED, cur_unaired["episode_id"]])
 
     def fix_tvrage_show_statues(self):
+        # Remove?
         status_map = {
             'returning series': 'Continuing',
             'canceled/ended': 'Ended',
```
fernandog: @p0psicles do we have something similar to the others indexers?

"tvmaze" has the 'to be determined' status. Its not 'ended' but its not 'continuing'
Keep as it is or make a list of default medusa show status so we map all?

medusa/databases/main_db.py
```diff
@@ -260,6 +262,7 @@ def fix_unaired_episodes(self):
                                    [common.UNAIRED, cur_unaired["episode_id"]])
 
     def fix_tvrage_show_statues(self):
+        #  Remove?
```
p0psicles: No this one can stay. Still need to check all the show status we currently have. As I didn't sanitize these. Then we just rename this one.

p0psicles:
But how is this really a bug related to propers?

fernandog:
Its not updating old propers with proper tags (before we added the proper tags code)

medusa/databases/main_db.py
fernandog: Rename then? Remove tvrage from name

fernandog:
waiting your OK @p0psicles

Issue #2022, opened 2017-01-24
[APP SUBMITTED]: TypeError: string indices must be integers, not str


ripper1967:
### INFO
**Python Version**: `2.7.12 (v2.7.12:d33e0cf91556, Jun 27 2016, 15:24:40) [MSC v.1500 64 bit (AMD64)]`
**Operating System**: `Windows-10-10.0.14393`
**Locale**: `cp1252`
**Branch**: [develop](../tree/develop)
**Database**: `44.4`
**Commit**: pymedusa/Medusa@872d7bf3428ca7ec4314c583518d4bfca0bcf34e
**Link to Log**: https://gist.github.com/678ea5b7f5b82a5a76ad3769d240778a
### ERROR
<pre>
2017-01-24 04:03:14 ERROR    SHOWUPDATER :: [872d7bf] Exception generated: string indices must be integers, not str
Traceback (most recent call last):
  File "C:\pymedusaSickRage\medusa\scheduler.py", line 103, in run
    self.action.run(self.force)
  File "C:\pymedusaSickRage\medusa\show_updater.py", line 72, in run
    updated_seasons = t.get_last_updated_seasons([show.indexerid], last_update, update_max_weeks)
  File "C:\pymedusaSickRage\medusa\indexers\tvdbv2\tvdbv2_api.py", line 595, in get_last_updated_seasons
    if int(episode['lastupdated']) > from_time:
TypeError: string indices must be integers, not str
</pre>
---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

wimpyrbx:
can confirm this happens here as well for me on develop

p0psicles:
I looked at it yesterday. But couldn't get it to reproduce.
Looks like somehow episode is a string and not a dict, like expected. Maybe something goes wrong when it tries to get the episodes for the show.

I have a PR ready, that will fix tvmaze updating. Maybe that also helps here. but honestly i doubt that.

ripper1967:
I get this on a 3am  auto update  not  mass update  and update all shows. if that helps.

fernandog:
@ripper1967 can you access your cache.db and send a screenshot of tje lastupdated table?

p0psicles:
I don't think it has anything to do with that. The 'lastupdated' in the exception is an episode attribute returned by the indexer, which we use to check against our own lastUpdated in db. As the exception is referring the lastupdated and not the from_time var.

p0psicles:
[...]

PR #2028, opened 2017-01-24
Improved show_update to be smarter in when to request the indexer for…

p0psicles:
… updated series and updated seasons (if the indexer supports this).

* Fixed tvdb bug, where it crashed on parsing an actor, without image.
* Changed some debug.warning to debug.warn.
* tmdb get_last_updated_series should return a list. Not a set.

### DIFF
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -560,7 +560,7 @@ class Tmdb(BaseIndexer):
                     new_list.append(show.indexerid)
             total_updates = new_list
 
-        return total_updates
+        return list(total_updates)
 
     # Public methods, usable separate from the default api's interface api['show_id']
     def get_last_updated_seasons(self, show_list, from_time, weeks=1):
--- a/medusa/indexers/tvdbv2/tvdbv2_api.py
+++ b/medusa/indexers/tvdbv2/tvdbv2_api.py
@@ -276,7 +276,7 @@ class TVDBv2(BaseIndexer):
             )
 
         mapped_episodes = self._object_to_dict(results, self.series_map, '|')
-        return OrderedDict({'episode': mapped_episodes})
+        return OrderedDict({'episode': mapped_episodes if isinstance(mapped_episodes, list) else [mapped_episodes]})
 
     def _parse_episodes(self, tvdb_id, episode_data):
         """Parse retreived episodes."""
--- a/medusa/indexers/tvdbv2/tvdbv2_api.py
+++ b/medusa/indexers/tvdbv2/tvdbv2_api.py
@@ -71,10 +71,10 @@ class TVDBv2(BaseIndexer):
             access_token = auth_api.login_post(authentication_string)
             auth_client = ApiClient(api_base_url, 'Authorization', 'Bearer ' + access_token.token)
         except ApiException as e:
-            log().warning("could not authenticate to the indexer TheTvdb.com, with reason '%s',%s)", e.reason, e.status)
+            log().warn("could not authenticate to the indexer TheTvdb.com, with reason '%s',%s)", e.reason, e.status)
             raise IndexerUnavailable("Indexer unavailable with reason '%s' (%s)" % (e.reason, e.status))
         except (MaxRetryError, RequestError) as e:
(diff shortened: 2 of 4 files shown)


fernandog:
approved

medusa/indexers/tmdb/tmdb.py
```diff
@@ -560,7 +560,7 @@ def get_last_updated_series(self, from_time, weeks=1, filter_show_list=None):
                     new_list.append(show.indexerid)
             total_updates = new_list
 
-        return total_updates
+        return list(total_updates)
```
labrys: Why even bother converting the set to a list.  The set is iterable just like a list and the conversion isn't really necessary here.

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -71,10 +71,10 @@ def __init__(self, *args, **kwargs):  # pylint: disable=too-many-locals,too-many
             access_token = auth_api.login_post(authentication_string)
             auth_client = ApiClient(api_base_url, 'Authorization', 'Bearer ' + access_token.token)
         except ApiException as e:
-            log().warning("could not authenticate to the indexer TheTvdb.com, with reason '%s',%s)", e.reason, e.status)
+            log().warn("could not authenticate to the indexer TheTvdb.com, with reason '%s',%s)", e.reason, e.status)
```
labrys: Why `log()`?  Create the log a single time not every time you use it.

medusa/indexers/tvmaze/tvmaze_api.py
```diff
@@ -371,7 +371,7 @@ def _parse_actors(self, tvmaze_id):
         for order, cur_actor in enumerate(actors.people):
             save_actor = Actor()
             save_actor['id'] = cur_actor.id
-            save_actor['image'] = cur_actor.image.get('original')
+            save_actor['image'] = cur_actor.image.get('original') if cur_actor.image else ''
```
labrys: If `cur_actor.image` is a dict with data, instead of being `None` when there's no data it should be an empty dict instead.  Then `cur_actor.image.get()` would work whether you have data or not.

labrys: Also if you wanted to make sure save_actor['image'] was an empty string when non-existent and not `None`, you can do the following:  `cur_actor.image.get('original', '')`  or `cur_actor.image.get('original') or ''`.   Lastly you can also chain conditions such as `cur_actor.image.get('medium') or cur_actor.image.get('original') or ''`

medusa/indexers/tmdb/tmdb.py
p0psicles: Reason is because this method should return a list. And a set() although it is also an iterable is not a list. You should always return what you advertise.

medusa/indexers/tvmaze/tvmaze_api.py
p0psicles: I know, but this is out of my control, as it's being serialized/deserialized by the lib. This is what it's giving me back.

p0psicles: None of your alternatives work. Only alternative could be: ```getattr(cur_actor, 'image', {}).get('originall', '')```

Ben then I couldn't say if that would be any better then what I currently have.

medusa/indexers/tmdb/tmdb.py
labrys: Then advertise that it returns an iterable.  Python is a duck-typing language not statically typed. This isn't C++ or Java :)

labrys: I'm not saying that there isn't valid reasons to return a list instead, for example iterating a list is faster than iterating a set, but checking contents or avoiding duplicates is faster than a list.  But unnecessarily converting types isn't an optimization.  I am just challenging to see if there is a valid reason.

p0psicles:
could fix #2022

medusa/indexers/tvmaze/tvmaze_api.py
labrys: Well if modifying the original lib is out of the question then instead you could do:

Issue #2034, opened 2017-01-26
Errors in show refresh crash the UI


labrys:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:** Dev / ee778e94195098ec9d602703ba0dcbb0b3db8192
**OS:** Win 10; Python 2.7.13
**What you did:**  Editing a show (changing the directory) triggered a show-refresh
**What happened:**  Show-refresh failed with SSL error (likely cause TVDB api is down).  Error caused the edit to fail and caused the page to crash (yielding a white screen with no input, response or other interface options)
**What you expected:**  The exception to be properly caught and to either finish the edit (and queueing the refresh for a later time) or a clear error page in the UI
**Logs:**
```
WARNING:tvdbv2_api:could not authenticate to the indexer TheTvdb.com, with reason 'SSLError
EOF occurred in violation of protocol (_ssl.c:661)',0)
WARNING:tvdbv2_api:could not authenticate to the indexer TheTvdb.com, with reason 'SSLError
EOF occurred in violation of protocol (_ssl.c:661)',0)
2017-01-25 22:57:37 ERROR    Thread-35 :: [ee778e9] Exception generated: (0)
Reason: SSLError
EOF occurred in violation of protocol (_ssl.c:661)
Traceback (most recent call last):
  File "H:\usr\local\medusa\medusa\server\web\core\base.py", line 259, in async_call
    result = function(**kwargs)
  File "H:\usr\local\medusa\medusa\server\web\home\handler.py", line 1377, in editShow
    if self.check_show_for_language(show_obj, indexerLang):
  File "H:\usr\local\medusa\medusa\server\web\home\handler.py", line 1298, in check_show_for_language
    t[show_obj.indexerid]
  File "H:\usr\local\medusa\medusa\indexers\indexer_base.py", line 254, in __getitem__
    self._get_show_data(key, self.config['language'])
  File "H:\usr\local\medusa\medusa\indexers\tvdbv2\tvdbv2_api.py", line 515, in _get_show_data
    series_info = self._get_show_by_id(sid, request_language=get_show_in_language)
  File "H:\usr\local\medusa\medusa\indexers\tvdbv2\tvdbv2_api.py", line 201, in _get_show_by_id
    results = self.series_api.series_id_get(tvdbv2_id, accept_language=request_language)
[...]

PR #2042, opened 2017-01-26
Catch and handle IndexerException, when switching languages.

p0psicles:

### DIFF
--- a/medusa/server/web/home/handler.py
+++ b/medusa/server/web/home/handler.py
@@ -1290,15 +1290,19 @@ class Home(WebRoot):
 
         :returns: True (show found in language) False (show not found in language)
         """
+
         indexer_api_params = indexerApi(show_obj.indexer).api_params.copy()
         indexer_api_params['language'] = language
         indexer_api_params['episodes'] = False
-        try:
-            t = indexerApi(show_obj.indexer).indexer(**indexer_api_params)
-            t[show_obj.indexerid]
-        except IndexerShowNotFoundInLanguage:
-            return False
-        return True
+        indexer_api = indexerApi(show_obj.indexer).indexer(**indexer_api_params)
+
+        if language and language in indexer_api.config['valid_languages']:
+            try:
+                indexer_api[show_obj.indexerid]
+            except IndexerShowNotFoundInLanguage:
+                return False
+            return True
+        return False
 
     def editShow(self, show=None, location=None, allowed_qualities=None, preferred_qualities=None,
                  exceptions_list=None, flatten_folders=None, paused=None, directCall=False,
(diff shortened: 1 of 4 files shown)


p0psicles:
Fixes #2034

medusa/server/web/home/handler.py
```diff
@@ -1293,8 +1293,8 @@ def check_show_for_language(show_obj, language):
         indexer_api_params = indexerApi(show_obj.indexer).api_params.copy()
         indexer_api_params['language'] = language
         indexer_api_params['episodes'] = False
-        t = indexerApi(show_obj.indexer).indexer(**indexer_api_params)
         try:
+            t = indexerApi(show_obj.indexer).indexer(**indexer_api_params)
```
labrys: Once again a single-letter variable.

labrys: Also this likely shouldn't be inside the `try: except` block.

labrys: Also you're recreating the indexer repeatedly here again.

labrys:
I'll review some more tomorrow.

medusa/server/web/home/handler.py
p0psicles: I'll change the letter. Np.
It's inside the try block. Because the init of tvdb and tmdb do some magic in refreshing the bearer. If an indexer is unavailable, it will throw IndexerUnavailable exception.
I'll try to create one indexer.

labrys: Except you're not catching `IndexerUnavailable` in that `try: except`.  You're catching `IndexerShowNotFoundInLanguage`.  Instead you should create the Indexer once (inside a `try: except IndexerUnavailable` and then use that instance throughout.  If `IndexerUnavailable` can occur at any time during the usage of the indexer, then the usage of the function would be the appropriate place to catch `IndexerUnavailable`. 
E.g.
```python
try:
   check_show_for_language(...)
except IndexerUnavailable:
   ...
```

p0psicles: Personally this  is the best solution as I have it now. As the creating the indexer_api can raise IndexerUnavailable. But chose to use IndexerException, as that's a somewhat broader catch. I don't want to give tvdb the room to prevent me from editing a show.

Also the IndexerException is around the check_show_for_language method, because the creation of the api aws well as the api call itself could in theory raise an exception other then IndexerShowNotFoundInLanguage.

labrys: You aren't catching `IndexerException`, you are catching `IndexerShowNotFoundInLanguage`.  So if the creation of the indexer raises `IndexerException` then it won't be caught.  If you don't intend for it to be caught then there's no reason for the indexer creation to be inside the `try: except`.  If you do want it to be caught you need to change the `except` clause to `except IndexerException`.

medusa/server/web/home/handler.py
```diff
@@ -1290,15 +1290,19 @@ def check_show_for_language(show_obj, language):
...
         indexer_api_params['episodes'] = False
-        try:
-            t = indexerApi(show_obj.indexer).indexer(**indexer_api_params)
-            t[show_obj.indexerid]
-        except IndexerShowNotFoundInLanguage:
-            return False
-        return True
+        indexer_api = indexerApi(show_obj.indexer).indexer(**indexer_api_params)
+
+        if language and language in indexer_api.config['valid_languages']:
```
labrys:
```python
if language in indexer_api.config['valid_languages']
```

p0psicles:
I'm good with the PR. Let's not spend any more time on this. Enough to do.

PR #2051, opened 2017-01-27
Add option to erase subtitles cache file on next boot

fernandog:
@ratoaq2 as you requested some time ago

```
2017-01-27 13:31:03 DEBUG    MAIN :: [de0be9e] Removing subtitles cache file: /home/osmc/Medusa/cache/subliminal.dbm
2017-01-27 13:31:03 DEBUG    MAIN :: [de0be9e] Removing subtitles cache file: /home/osmc/Medusa/cache/application.dbm
```

### DIFF
--- a/medusa/__main__.py
+++ b/medusa/__main__.py
@@ -1000,10 +1000,10 @@ class Application(object):
                     for cache_file in ['application.dbm', 'subliminal.dbm']:
                         file_path = os.path.join(app.CACHE_DIR, cache_file)
                         if os.path.isfile(file_path):
-                            logger.debug(u"Removing subtitles cache file: {cache_file}", cache_file=file_path)
+                            logger.info(u"Removing subtitles cache file: {cache_file}", cache_file=file_path)
                             os.remove(file_path)
                 except OSError as e:
-                    logger.debug(u"Unable to remove subtitles cache files. Error: {error}", error=e)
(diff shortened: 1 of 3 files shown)


fernandog:
can somebody else review/approve this please?

medusa/__main__.py
```diff
@@ -994,6 +995,18 @@ def initialize(self, console_logging=True):
...
 
+            if app.SUBTITLES_ERASE_CACHE:
+                try:
+                    for cache_file in ['application.dbm', 'subliminal.dbm']:
+                        file_path = os.path.join(app.CACHE_DIR, cache_file)
+                        if os.path.isfile(file_path):
+                            logger.debug(u"Removing subtitles cache file: {cache_file}", cache_file=file_path)
+                            os.remove(file_path)
+                except OSError as e:
+                    logger.debug(u"Unable to remove subtitles cache files. Error: {error}", error=e)
```
p0psicles: What's your reasoning here to use debug and not info?

ratoaq2: I would say it should be a warning

p0psicles: Yeah that was also my first impression. But with a warning, a user should know from the warning which action it should take.

p0psicles: Then the debug before that better be an info.

PR #2054, opened 2017-01-28
Invert isinstance logic

labrys:
`isinstance` should be checking to see if the instance is a string type not a list type.  Checking for list type is a bug.  For example:

```python
# Current behaviour

>>> arg = 'Some Text'
>>> [arg] if not isinstance(arg, list) else arg
['Some Text']  # Expected result so we don't iterate through the characters of the string

>>> arg = (a, b, c)  # but with this...
>>> [arg] if not isinstance(arg, list) else arg
[(a, b, c)]  # Tuple inside a list is a bug

# New behaviour

>>> arg = 'Some Text'
>>> [arg] if isinstance(arg, string_types) else arg
['Some Text']  # Expected result

>>> arg = (a, b, c)  # now with this...
>>> [arg] if isinstance(arg, string_types) else mapped_results
(a, b, c)  # Still an iterable, bug is fixed
```

### DIFF
--- a/medusa/indexers/tvdbv2/tvdbv2_api.py
+++ b/medusa/indexers/tvdbv2/tvdbv2_api.py
@@ -21,6 +21,7 @@ from collections import OrderedDict
 
 from requests.compat import urljoin
 from requests.packages.urllib3.exceptions import MaxRetryError, RequestError
+from six import string_types
 
 from tvdbapiv2 import (ApiClient, AuthenticationApi, SearchApi, SeriesApi, UpdatesApi)
 from tvdbapiv2.rest import ApiException
@@ -104,7 +105,7 @@ class TVDBv2(BaseIndexer):
 
         tvdb_response = getattr(tvdb_response, 'data', tvdb_response)
 
-        if not isinstance(tvdb_response, list):
+        if isinstance(tvdb_response, string_types):
             tvdb_response = [tvdb_response]
 
         for parse_object in tvdb_response:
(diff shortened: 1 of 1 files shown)


fernandog:
@labrys what about all this others?

image

labrys:
@fernandog They still need to be done.  These were meant to be a part of p0psicles previous PR as they were the ones directly related to it but it got merged before I got to add these edits.

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -104,7 +105,7 @@ def _object_to_dict(self, tvdb_response, key_mapping=None, list_separator='|'):
 
         tvdb_response = getattr(tvdb_response, 'data', tvdb_response)
 
-        if not isinstance(tvdb_response, list):
+        if isinstance(tvdb_response, string_types):
```
p0psicles: tvdb_response is not a string, it's an object or list of objects.

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -180,7 +181,7 @@ def search(self, series):
             return
 
         mapped_results = self._object_to_dict(results, self.series_map, '|')
-        mapped_results = [mapped_results] if not isinstance(mapped_results, list) else mapped_results
+        mapped_results = [mapped_results] if isinstance(mapped_results, string_types) else mapped_results
```
p0psicles: not a string, but OrderedDict or list of OrderedDict

p0psicles:
there are more changes done here, that will simply break when adding a show.
Please look carefully at each evaluated isinstance, as most of them are not strings.

p0psicles:
@labrys are you planning on finishing this?

labrys:
Yes I am but this is a low priority for me.  If you or anyone else wants to pick up on it I'm not opposed.

labrys:
Closing as this is a low priority for me right now and don't have time to commit to it.

PR #2059, opened 2017-01-29
Fix for issue https://github.com/guessit-io/guessit/issues/406

ratoaq2:
Thanks @IDerr for reporting it -> https://github.com/guessit-io/guessit/issues/406

Btw, this is a medusa issue, not a guessit issue. These kind of issues you should initially report it here. If we realize that's indeed a guessit problem, then we can instruct to open an issue there.


IDerr:
I am sorry to report at the wrong place :(
But thanks the fix.

Issue #2046, opened 2017-01-27
Excessive hits to the cache.db during backlog search


labrys:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:** Dev / ee778e9
**OS:** Win 10
**What you did:**  Observed backlog search logs with db logging enabled
**What happened:**  5 hits to the cache.db per episode searched.  2 of which are exact duplicates.
**What you expected:**  Get scene exceptions from the database with a single query.
**Logs:**
```
2017-01-26 20:55:19 INFO     SEARCHQUEUE-BACKLOG-**ID**:: [**PROVIDER**] :: [ee778e9] Performing season pack search for **SERIES**
2017-01-26 20:55:19 DB       SEARCHQUEUE-BACKLOG-**ID** :: [**PROVIDER**] :: [ee778e9] cache.db: SELECT * FROM [**PROVIDER**] WHERE indexerid = ? AND season = ? AND episodes LIKE ? with args [**ID**, 2002, '%|31|%']
2017-01-26 20:55:19 DB       SEARCHQUEUE-BACKLOG-**ID** :: [**PROVIDER**] :: [ee778e9] cache.db: UPDATE [lastSearch] SET time = ? WHERE provider = ? with args [1485482119, '**PROVIDER**']
2017-01-26 20:55:19 DB       SEARCHQUEUE-BACKLOG-**ID** :: [**PROVIDER**] :: [ee778e9] cache.db: SELECT show_name FROM scene_exceptions WHERE indexer = ? AND indexer_id = ? AND season = ? with args [1, **ID**, 2002]
2017-01-26 20:55:19 DB       SEARCHQUEUE-BACKLOG-**ID** :: [**PROVIDER**] :: [ee778e9] cache.db: SELECT show_name FROM scene_exceptions WHERE indexer = ? AND indexer_id = ? AND season = ? with args [1, **ID**, -1]
2017-01-26 20:55:19 DB       SEARCHQUEUE-BACKLOG-**ID** :: [**PROVIDER**] :: [ee778e9] cache.db: SELECT show_name FROM scene_exceptions WHERE indexer = ? AND indexer_id = ? AND season = ? with args [1, **ID**, -1]
```

p0psicles:
Maybe I have an hour spare time tonight. But depending the amount of work involved to fix this, I might or might not finish it before releasing.

fernandog:
[...]

PR #2074, opened 2017-02-01
Split up get_scene_exceptions in get and update.

p0psicles:
get_scene_exceptions did two things. 1. Check if a indexerid, indexer, (optional) season combination is available in exceptionsCache. If not search in db, and update the exceptionCache in the process.

Now those are two methods. We now "trust" the cache, as in, when it can't find in cache, it isn't there.
And the second methods constructs the cache.

This PR will also improve startup time, as previously it was creating the cache, trough looping by show and creating the cache per show.

Now i'm doing one db query, and creating two caches in one go.

* ~~Update is only run on startup and on editShow.~~
* Update is only run on editShow.
* get_scene_exceptions only get's exceptions from cache
* update_scene_exceptions_cache, updates the cache (db) for a specific show.
* refresh_exceptions_cache, re-caches the exceptions_cache and exceptions_season_cache cache.
* refresh_exceptions_cache is run on startup and show_updater.

## TODO:
- [ ] Refactor the same for get_scene_seasons.

~~## Force retrieve exception cache on show add.
With this PR we are forcing the retrieval of the exception in "retrieve_exceptions". I added this because, i'd like to know for sure I have all required scene exceptions for a show I just added. But the fact is, that the retrieve_exceptions is not really an efficient written method. So i'm having doubt, we should do this right now~~

### DIFF
--- a/medusa/scene_exceptions.py
+++ b/medusa/scene_exceptions.py
@@ -113,25 +113,6 @@ def set_last_refresh(ex_list):
     )
 
 
-def update_scene_exceptions_cache(indexer_id, indexer, season=-1):
-    """Update the scene_exceptions cache from db. And return the scene_exceptions available."""
-    cache_db_con = db.DBConnection('cache.db')
-    exceptions = cache_db_con.select(b'SELECT show_name FROM scene_exceptions WHERE '
-                                     b'indexer = ? AND indexer_id = ? AND season = ?',
-                                     [indexer, indexer_id, season])
-    if exceptions:
-        # TODO: I'd like to know in which use-cases we can't find a result in cache, but can from db?
-        exceptions_list = list({cur_exception[b'show_name'] for cur_exception in exceptions})
-
-        # TODO: Refactor exceptionCache to be usable by multiple indexers.
-        if indexer_id not in exceptions_cache:
-            exceptions_cache[indexer_id] = {}
-        exceptions_cache[indexer_id][season] = exceptions_list
-        # I'm only calling get_scene_exceptions here because that adds the -1 by default when there are
-        # no season exceptions defined. Apparently this is needed.
-    return get_scene_exceptions(indexer_id, indexer, season)
-
-
 def get_scene_exceptions(indexer_id, indexer, season=-1):
     """Given a indexer_id, return a list of all the scene exceptions from the scene_exception cache."""
     exceptions_list = []
@@ -150,24 +131,18 @@ def get_all_scene_exceptions(indexer_id):
     Get all scene exceptions for a show ID.
 
     :param indexer_id: ID to check
-    :return: dict of exceptions
+    :return: dict of exceptions. For example: exceptions_cache[season][exception_name]
     """
-    exceptions_dict = {}
-
-    cache_db_con = db.DBConnection('cache.db')
-    exceptions = cache_db_con.select(b'SELECT show_name, season FROM scene_exceptions WHERE indexer_id = ?',
-                                     [indexer_id])
-    if exceptions:
-        for cur_exception in exceptions:
-            if not cur_exception[b'season'] in exceptions_dict:
-                exceptions_dict[cur_exception[b'season']] = []
-            exceptions_dict[cur_exception[b'season']].append(cur_exception[b'show_name'])
-
-    return exceptions_dict
+    return exceptions_cache.get(indexer_id, {})
 
 
 def get_scene_seasons(indexer_id):
-    """Return a list of season numbers that have scene exceptions."""
+    """
+    Get a list of season numbers form the scene_exceptions, for which exceptions have been added.
+
+    :param indexer_id: ID to check
+    :return: list of seasons.
+    """
     return exceptions_season_cache.get(indexer_id, [])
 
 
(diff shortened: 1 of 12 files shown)


duramato:
Explain me, if a new exception is added to the repo it will only get updated startup and a show edit? Or this only affects scene exceptions added via edit show?
If so, it can't be like so. A normal user's Medusa only gets restarted on updated which is at min each week at each release.

p0psicles:
Correct me if i'm wrong. But the new cache is retrieved through the method: retrieve_exceptions. https://github.com/pymedusa/Medusa/blob/develop/medusa/scene_exceptions.py#L195.

That is called only through: https://github.com/pymedusa/Medusa/blob/develop/medusa/name_cache.py#L101

That in turn is called through:
show_queue.py QueueItemAdd() (https://github.com/pymedusa/Medusa/blob/develop/medusa/show_queue.py#L521).
and
search.core.py: search_providers()
(https://github.com/pymedusa/Medusa/blob/develop/medusa/search/core.py#L478)

And it's updating the db on those occasions, if it has passed the "cooldown".

So we know, that this change won't change anything about the way we get new season scene exceptions or show name exceptions in the db.

Then old way of updating the cache, was also not a failsave system. Because it would only get info from db, when it couldn't find it in cache. Meaning you could not sync cache with db, if for ex a season scene name was added for a season, where we already had a scene exception for. That would only be updated on restart.

What I think we should do is, is manually refresh the entire cache for: ```exceptionsCache``` and ```exceptionsSeasonCache``` once every 24 hours. As the db won't have any new data before that.

p0psicles:
@medariox the unittest related to the scene exception is failing. But i'm pretty confident I didn't caused regression. I still can't debug the unittest locally. Could you look into this?

medusa/scene_exceptions.py
```diff
@@ -20,22 +20,71 @@
...
-exceptionsCache = {}
-exceptionsSeasonCache = {}
+exceptions_cache = {}
+exceptions_season_cache = {}
 
 exceptionLock = threading.Lock()
 
 xem_session = helpers.make_session()
 
+log = logging.getLogger(__name__)
```
medariox: I think we usually call this ```logger```.
For consistency reasons I would call it like that instead:
```
```

medusa/scene_exceptions.py
```diff
@@ -20,22 +20,71 @@
...
+            if season not in exceptions_cache[indexer_id]:
+                exceptions_cache[indexer_id][season] = []
+
+            # exceptions_cache[indexerid][season] = ['showname 1', 'showname 2']
+            if exception[b'show_name'] not in exceptions_cache[indexer_id][season]:
+                exceptions_cache[indexer_id][season].append(exception[b'show_name'])
+
+            # Do the same for the exceptions_season_cache
+            if int(indexer_id) not in exceptions_season_cache:
+                exceptions_season_cache[int(indexer_id)] = []
```
medariox: No need for the int() here, already done on top (both times).

medusa/scene_exceptions.py
```diff
@@ -20,22 +20,71 @@
...
+
+            # exception_season_cache[indexerid] = [-1, 2, 3]
+            if season not in exceptions_season_cache[indexer_id]:
+                exceptions_season_cache[indexer_id].append(season)
+
+    log.info('Finished processing {nr} scene exceptions.', nr=len(exceptions))
```
medariox: Wrong indent? Otherwise this can be an issue if ```exceptions``` is None.

medusa/scene_exceptions.py
```diff
@@ -20,22 +20,71 @@
...
+
+
+def refresh_exceptions_season_cache():
+    """Query the db for show exceptions and update the exceptions_season_cache."""
+    return True
```
medariox: Not sure what this is for. Not used yet? Maybe add a TODO.

p0psicles: Should be gone. tnx

p0psicles:
Implements #2046

p0psicles:
@medariox could you check again?

medusa/scene_exceptions.py
```diff
@@ -218,11 +193,23 @@ def update_scene_exceptions(indexer_id, indexer, scene_exceptions, season=-1):
     # TODO: make sure we add indexer, when the global exceptionsCache var has been changed.
     if indexer_id in exceptions_cache:
         exceptions_cache[indexer_id] = {}
```
medariox: Here we clear ```exceptions_cache[indexer_id] = {}```
if it exists...

medusa/scene_exceptions.py
```diff
@@ -218,11 +193,23 @@ def update_scene_exceptions(indexer_id, indexer, scene_exceptions, season=-1):
...
-        exceptions_cache[indexer_id][season] = [se.decode('utf-8') for se in scene_exceptions]
 
-    for cur_exception in [se.decode('utf-8') for se in scene_exceptions]:
-        cache_db_con.action(b'INSERT INTO scene_exceptions (indexer_id, show_name, season, indexer) VALUES (?,?,?,?)',
-                            [indexer_id, cur_exception, season, indexer])
+    decoded_scene_exceptions = [se.decode('utf-8') for se in scene_exceptions]
+    for cur_exception in decoded_scene_exceptions:
+        if indexer_id not in exceptions_cache:
+            exceptions_cache[indexer_id] = {}
```
medariox: Here we create it if it doesn't exists...
I suggest we just create it once without any checks before the loop.
If it exists, it will be cleared, if it doesn't, it will be created.

p0psicles:
I've reviewed and all seems okay. Therefor i'm merging this, as it's pretty save now develop is open again.

PR #2078, opened 2017-02-01
Fix episode title notifications when there are unicode characters in it.

p0psicles:

This is reproducable by (manual) searching (and getting a result) for an episode which has unicode chars in the title.

Like for ex. Dragon ball super S04E13.
Protect Kai```ō```-shin Gowasu - Destroy Zamasu!

I tested this on Windows 7 x64.
Maybe it's not even an issue on linux like OS'.

image

### DIFF
--- a/medusa/search/queue.py
+++ b/medusa/search/queue.py
@@ -337,7 +337,8 @@ class ForcedSearchQueueItem(generic_queue.QueueItem):
                     ui.notifications.message("We have found season packs for {0}".format(self.show.name),
                                              "These should become visible in the manual select page.")
                 else:
-                    ui.notifications.message("We have found results for {0}".format(self.segment[0].pretty_name()),
+                    ui.notifications.message("We have found results for {0}".format(self.segment[0].pretty_name()
+                                                                                    .encode('utf-8')),
                                              "These should become visible in the manual select page.")
             else:
                 ui.notifications.message('No results were found')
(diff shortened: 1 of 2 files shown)


p0psicles:
Need to perform some tests on different os. To see if this is safe.

fernandog:
Confirm it fixed:

My concerns:
1) if we should do in the above log to with `"We have found season packs for {0}".format(self.show.name)`
2) If we should add this in the ui.py to apply to all messages: `self._messages.append(Notification(title, message, MESSAGE))`

Error without the PR:
```
2017-02-06 13:18:48 DEBUG    SEARCHQUEUE-MANUAL-295068 :: [51ff22a] Traceback (most recent call last):
  File "C:/Users/Fernando/Documents/Medusa\medusa\search\queue.py", line 345, in run
    ui.notifications.message("We have found results for {0}".format(self.segment[0].pretty_name()),
  File "C:\Python27\lib\encodings\cp1252.py", line 12, in encode
    return codecs.charmap_encode(input,errors,encoding_table)
UnicodeEncodeError: 'charmap' codec can't encode character u'\u014d' in position 40: character maps to <undefined>
```

After:
after

ratoaq2:
The same as in #2096 
I dont think that's the right place to encode and decode. I haven't check the source for those strings, but the encode/decode should be done quite earlier in the flow.

p0psicles:
This is different then the read from file imo. We should come with a better solution for this.

p0psicles:
I was thinking about utilizing the __str__ magic method on the series and episode objects. But then we should remove the current __str__ mess. As I don't think it's really used anywhere.
Then we can by default encode the name using encode('utf-8').

I still have to work out some details though.

Issue #2087, opened 2017-02-03
Post processing reports wrong download type


labrys:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:**  Dev
**OS:** Win 10
**What you did:**  Manually post processed an episode
**What happened:**  PP logs report `Processing <video_file> (Torrent)`.  Source was an NZB, no torrent providers are enabled
**What you expected:**  Correct download type to be reported
**Logs:**
```
Postprocessing results
Processing folder G:\tmp\download
PostProcessing Path: G:\tmp\download
PostProcessing Dirs: []
PostProcessing Files: [<files>]
PostProcessing VideoFiles: [<video_files>]
PostProcessing RarContent: 
PostProcessing VideoInRar: 
Processing <video_file> (Torrent)
Retrieving episode object for <series> <episode>
Looking up quality for <video_file>, got Unknown
Looking up quality for <video_file>, got Unknown
Trying other methods to get quality for <video_file>, got 1080p HDTV
Looks like <video_file> has quality 1080p HDTV, using that
Snatch in history: False
Manually snatched: False
Current quality: 1080p WEB-DL
New quality: 1080p HDTV
Proper: False
This episode is a priority download: False
File <existing_file> is larger than <video_file>
Checking if new quality 1080p HDTV should replace current quality: 1080p WEB-DL
Processing failed for <video_file>: File exists. Marking it unsafe to replace. Reason: New quality is lower than current Preferred. Ignoring quality
Problem(s) during processing, failed the following files/folders: 
[<video_file> : Processing failed: File exists. Marking it unsafe to replace. Reason: New quality is lower than current Preferred. Ignoring quality]```

fernandog:
a simple search in code you would find this:

`self._log(u'Processing {0} ({1})'.format(self.file_path, self.nzb_name or 'Torrent'))`

just search your nzbtomedia logs to check the nzb name param it sent to medusa

fernandog:
Don't assign me unless im ok with it

labrys:
@fernandog I did find that in code, and its obvious that the nzb name was not passed or it would have logged the nzb name, thus that line of code has a bug.

fernandog:
well, you can debug it as i don't download nzb

PR #2090, opened 2017-02-03
Remove nzb_name log as manual PP doesn't have nzbname input. Only nzbtomedia/etc

fernandog:
@medariox as we discussed

https://github.com/pymedusa/Medusa/issues/2087


fernandog:
someone approve please?

PR #2092, opened 2017-02-03
While logging the nzb download, add size to it so we can know is a di…

fernandog:
…fferent release@size nzb

~~needs testing as i don't use NZB~~. All nzb has sizes

UPDATE:
Tested without nzbget running:
`2017-02-03 15:32:08 INFO     SNATCHQUEUE-MANUALSNATCH-274431 :: [51ff22a] Downloading Gotham.S03E13.1080p.WEB-DL.DD5.1.H.264-R2D2-Obfuscated with size: 1.73 GB from NZBGeek`

@duramato what about torznab?

### DIFF
--- a/medusa/failed_processor.py
+++ b/medusa/failed_processor.py
@@ -1,6 +1,5 @@
 # coding=utf-8
 # Author: Tyler Fenby <tylerfenby@gmail.com>
-
 #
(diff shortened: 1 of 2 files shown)


fernandog:
@labrys done

@duramato still waiting your feedback

Issue #2091, opened 2017-02-03
Backlog Concerns and potential options


jxer:
Guys after having medusa add 300 torrents to my client and no way to clear it after simply changing quality i think some changes need to be made if possible.

a warning and an option needs to be added to indicate a quality change will re download all existing files . also if possible an option to only download future episodes in the new quality.

limit global backlog. not everyone holds on to everything, if i have 5 seasons of a show, and suddenly i want to change from sd to hd for season 6, most people are not going to want to redownload all 5 seasons now in hd, since we have prolly already watched them. I am asking for a "how far back do you want to go option" for those who want it all, they can leave it on unlimited, for me 2 weeks is prolly all i would use. "You have this option for forced backlog" which in my mind is pointless since it goes farther back in 12 hours anyway

finally archive on first download was removed. this needs to be put back, this needs to be a per show option.

people have to realize if private trackers are in play, then medusa can really really use alot of bandwidth and ratio constantly redownloading files that generally will prolly not be watched again.

these options need to be added for users who have small storage system limited bandwidth or simply don't keep multiple seasons of shows ( i delete everything after the season is over)  

also after the 300 hundred downloads started i tried to revert back to the original quality, that did not stop backlog from trying to keep downloading the new quality i tried to pick. i actually had to restore the entire server from a backup to get the original quality to stick without backlog killing me. something needs to be done here. either a clear backlog db or something

jxer:
also where is my delete options in medusa already without removing the show, a per show or per season option needs to be added!

fernandog:
My proposed solutions:
1) Add back the `Archive first match`
2) Add a WARNING while editing show quality that it could let to a lot of re-downloads if existing episodes doesn't have this new quality
3) Ask users with a modal while editing show quality if he wants to change status of all existing episodes to ARCHIVED

others?

p0psicles:
Personally Im for option 3. As archiving is actually for these kind of use cases. Only I understand if users are not aware of this. Providing a user with a warning, if the user chooses for it, we could automatically archive existing DOWNLOADED episodes?

fernandog:
yes, option 3 is to only archive episodes with `downloaded` status. don't know if too hard to do it
or just warn use and he picks the ones he wants to archive

p0psicles:
I think its the easiest to implement and most user friendly. Will not be hard to code.

jxer:
how about other issues p0psicles is it possible to code those as well?

fernandog:
[...]

PR #2093, opened 2017-02-03
Convert final quality episodes with DOWNLOADED status to ARCHIVED

fernandog:
As discussed in here: https://github.com/pymedusa/Medusa/issues/2091

@OmgImAlexis 

about layout, i think we can improve but im not good with it =)

### Before archive and changing quality:

image

------------------------------------------------

### After archive:

image

### DIFF
--- a/medusa/server/api/v2/show.py
+++ b/medusa/server/api/v2/show.py
@@ -87,7 +87,7 @@ class ShowHandler(BaseRequestHandler):
                                                                 preferred_qualities=preferred_qualities)
                 data = {'new': new, 'existing': existing}
             elif query == 'archiveEpisodes':
-                data = {'archived': 1 if tv_show.set_all_episodes_archived() else 0}
+                data = {'archived': 1 if tv_show.set_all_episodes_archived(final_status_only=True) else 0}
             elif query == 'queue':
                 action, message = app.show_queue_scheduler.action.get_queue_action(tv_show)
                 data = {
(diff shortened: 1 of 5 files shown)


fernandog:
@OmgImAlexis need to pick your brain here

In editShow we have a form. 
And we need to detect used changed quality before/when he saves quality (when he saves a backlog is started and first we need to call this function that convers to ARCHIVED only if he agrees it (modal?) 
I looked the code and I have no idea what to do

fernandog:
@p0psicles @OmgImAlexis now it will only convert those that have final quality

medusa/server/api/v2/show.py
```diff
@@ -86,6 +86,8 @@ def _handle_detailed_show(self, tv_show, query):
                 new, existing = tv_show.get_backlogged_episodes(allowed_qualities=allowed_qualities,
                                                                 preferred_qualities=preferred_qualities)
                 data = {'new': new, 'existing': existing}
+            elif query == 'archiveEpisodes':
```
OmgImAlexis: No no no. This is for GET routes. GET means retrieve data. You need to add this to the POST, PUT or PATCH as they're for adding and changing data.

fernandog: Ops!

fernandog: @OmgImAlexis i need to receive a yes|no if we archived episodes or not. so it can't be a POST

OmgImAlexis: Oh this looked like you were using this TO archive the files.
I need to stop reviewing shit when I've just woken up.

Use true/false not 1/0 or True/False since true/false are booleans in js. This should work like  how passing an int turns into a js number(remember the parseInt thing).

PR #2096, opened 2017-02-04
Shows that have a unicode encoded show name with special chars, store…

p0psicles:
…d in the local nfo file. Will cause an endless reload of the addShows/existingShows/ page, because the mako is crashing on the unicode special char.

Content of nfo.
image

You can reproduce by following these steps:
1. Make sure you have enabled metadata generator (kodi12).
2. Add Naruto Shippuuden using ~~tvmaze~~ tmdb.
3. Remove the show, but leave files on disk.
4. Try to add the same show from ~~tvmaze~~ tmdb, it will redirect you to the addExisting page, and that page will keep reloading.

### DIFF
--- a/medusa/server/web/home/add_shows.py
+++ b/medusa/server/web/home/add_shows.py
@@ -167,6 +167,11 @@ class HomeAddShows(Home):
                     if not (indexer_id and show_name):
                         (indexer_id, show_name, indexer) = cur_provider.retrieveShowMetadata(cur_path)
 
+                        # unicode encoded special character show names, should be encoded to byte, as the nfo can have a
+                        # show title with unicode encoded special chars. Like naruto shippuuden on tvmaze.
+                        if show_name:
+                            show_name = show_name.encode('utf-8')
+
                         # default to TVDB if indexer was not detected
                         if show_name and not (indexer or indexer_id):
                             (_, idxr, i) = helpers.search_indexer_for_show_id(show_name, indexer, indexer_id)


fernandog:
isn't safe enough for 0.1.10 ?

fernandog:
which OS caused this? which locale?

p0psicles:
I reproduced on win7

medusa/server/web/home/add_shows.py
```diff
@@ -167,6 +167,11 @@ def massAddTable(self, rootDir=None):
                     if not (indexer_id and show_name):
                         (indexer_id, show_name, indexer) = cur_provider.retrieveShowMetadata(cur_path)
 
+                        # unicode encoded special character show names, should be encoded to byte, as the nfo can have a
+                        # show title with unicode encoded special chars. Like naruto shippuuden on tvmaze.
+                        if show_name:
+                            show_name = show_name.encode('utf-8')
+
```
fernandog: @p0psicles @ratoaq2 shoudln't this be done `name = showXML.findtext('title')` in metada\generic.py

ratoaq2: I have the feeling that this is in the wrong place. That shouldn't be a concern of this logic, but the source from where show_name came from. If we spread encode() all over the place it will be hard to maintain

fernandog: @ratoaq2 but `etree.ElementTree(file=xmlFileObj)` reads a xml file. so i don't get why this is the best place.

ratoaq2: My comment was about the code in this line, not about your comment

p0psicles: What do you suggest as an alternative?

p0psicles: After looking into it another time. I don't think it's in the wrong place. As i'm reading unicode here from file. You can't fully control what your are parsing. I think the application should be ready to handle it.

On a sidenote, I do agree that we should look into other issues, where we often log or parse unicode in mako. As for ex mako only supports bytes.

PR #2098, opened 2017-02-04
Fix PP season pack with nzbtomedia

fernandog:
As @ratoaq2 reported:
 https://gist.github.com/ratoaq2/94042449045a9fd6f410fb92fc3581d2

As expected first commit fails as GoT episode expected is [10] but got []
Second commit passed in all tests

image

### DIFF
--- a/tests/test_postprocessor_find_info.py
+++ b/tests/test_postprocessor_find_info.py
@@ -54,7 +54,7 @@ class TestPP(PostProcessor):
                                    'game.of.thrones.s02e10.1080p.bluray.x264-rovers.mkv',
                          nzb_name='Game.of.Thrones.S02.Season.2.1080p.BluRay.x264-PublicHD',
                          ),
-        'series_name': 'Game of Thrones',
+        'series_name': 'game of thrones',
         'season': 2,
         'episodes': [10],
         'quality': Quality.FULLHDBLURAY,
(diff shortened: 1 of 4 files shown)


fernandog:
@pymedusa/developers anyone has another scenario that i didn't tested to make the nzb_name be used last?

tests/test_postprocessor_find_info.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _analyze_name(self, name):
+        """Override original name to only parse release name and not match show database."""
+        to_return = (None, None, [], None, None)
+
+        if not name:
+            return to_return
+
+        #  This was changed from original method and removed Try/Expect
+        parse_result = NameParser()._parse_string(name)
```
fernandog: FYI i changed this to not check if show exists in DB. So it only parses the release name

tests/test_postprocessor_find_info.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if parse_result.show and all([parse_result.show.air_by_date, parse_result.is_air_by_date]):
+            season = -1
+            episodes = [parse_result.air_date]
+        else:
+            season = parse_result.season_number
+            episodes = parse_result.episode_numbers
+
+        #  Change show to series_name
+        to_return = (parse_result.series_name, season, episodes, parse_result.quality, parse_result.version)
```
fernandog: And here i changed to series_name instead of parse_result.show

fernandog:
@ratoaq2 refactor done?

PR #2111, opened 2017-02-06
PP: Define name_list once

fernandog:
@medariox ideas to the comment above the name_list ?

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -22,6 +22,8 @@ import re
 import stat
 import subprocess
 
+from collections import OrderedDict
+
 import adba
 
 from six import text_type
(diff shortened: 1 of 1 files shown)


fernandog:
oh sh*.... dict loses the order when converting to list @#$%¨&

fernandog:
Its failing because of the wrong order in items_resource

@medariox do you think is worth to use OrderedDict to keep this new logs?

medusa/post_processor.py
```diff
@@ -87,7 +89,9 @@ def __init__(self, file_path, nzb_name=None, process_method=None, is_priority=No
 
         self.manually_searched = False
 
-        self.item_resources = {'file name': self.file_name, 'relative path': self.rel_path, 'nzb name': self.nzb_name}
+        self.item_resources = OrderedDict({'file name': self.file_name,
```
ratoaq2: This way you have the same issue since you're using a regular dict to Instantiate an ordered one

fernandog:
@medariox all good here: 

2017-02-07 09:46:39 DEBUG    FINDSUBTITLES :: [edfb56d] Show information parsed from file name

approve?

PR #2120, opened 2017-02-07
Fix SNATCHED (BEST) render pill not working

fernandog:
@OmgImAlexis @duramato  (finally fixed this one)

UPDATE: this issue happens only when you forced a search and gets a SNATCHED BEST and don't restart medusa, so JS keeps updating status. See this comment: https://github.com/pymedusa/Medusa/pull/2120#issuecomment-278142762

Existing regex creates this issue:
image

image

----------------------------
With this PR:
image

https://regex101.com/r/2yhi7u/3

### DIFF
--- a/static/js/ajax-episode-search.js
+++ b/static/js/ajax-episode-search.js
@@ -37,8 +37,6 @@ function updateImages(data) {
                 img.prop('alt', 'Searching');
                 img.prop('src', 'images/' + loadingImage);
                 disableLink(el);
-                // Update Status and Quality
-                rSearchTerm = /(\w+)\s\((.+?)\)/;
                 htmlContent = ep.searchstatus;
             } else if (ep.searchstatus.toLowerCase() === 'queued') {
                 // el=$('td#' + ep.season + 'x' + ep.episode + '.search img');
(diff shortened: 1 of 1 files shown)


fernandog:
Also i noted that `ajax-episode-search.js` keeps updating status every 2 seconds when don't need
the status is already SNATCHED BEST. No search was runned.

This JS is called when user forced a searched and it founds results, so as soon the ep is snatched, JS will update status without need to update the page

So i don't know why it keep updating without a search running

fernandog:
Here are my findings. If you do a search in Medusa, the EP is not removed from queue

JS calls `collect_episodes_from_search_thread` every 5 seconds:

https://github.com/pymedusa/Medusa/blob/develop/medusa/search/manual.py#L152-L153

That code check `get_all_ep_from_queue`. Even aftter search is finished it still have those ep in queue.
https://github.com/pymedusa/Medusa/blob/develop/medusa/search/queue.py#L150-L153

Somehow the `self.finish()` is not removing those from queue
https://github.com/pymedusa/Medusa/blob/develop/medusa/search/queue.py#L365

Which result in tons o JS process without need. The only way to fix is to restart

PR #2122, opened 2017-02-07
POC - Add a method to remove item from queue

fernandog:
@labrys @ratoaq2 no idea how to proceed but this is related to the other PR about JS i tagged you

https://github.com/pymedusa/Medusa/pull/2120#issuecomment-278142762


fernandog:
not related to the self.queue

PR #2126, opened 2017-02-08
Show number of backlogged episodes with new and existing qualities

fernandog:
Before user save new quality, a WARNING label in html will be shown with number of current backlogged episodes and the number of new backlogged episodes after quality change

ping @jxer

**User current settting:**

load

-----------------------------------
**After changes quality:**
load2

### DIFF
--- a/medusa/server/api/v2/show.py
+++ b/medusa/server/api/v2/show.py
@@ -83,8 +83,8 @@ class ShowHandler(BaseRequestHandler):
                 allowed_qualities = map(int, allowed_qualities.split(',')) if allowed_qualities else []
                 preferred_qualities = self._parse(self.get_argument('preferred', default=None), str)
                 preferred_qualities = map(int, preferred_qualities.split(',')) if preferred_qualities else []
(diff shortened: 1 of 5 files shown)


fernandog:
@OmgImAlexis 

this should work right? I get nothing

http://localhost:8081/api/v2/backlogged_episodes/?api_key=***&show_id=256655&allowed_qualities=[]&preferred_qualities=[256]

fernandog:
I added to v1 to do some tests with the new TV method and it works:

http://localhost:8081/api/v1/api_key/?cmd=show.backloggedepisodes&indexerid=248682&allowed=16|64&preferred=256

{"data": [3, 1], "message": "", "result": "success"}

OmgImAlexis:
Either add the method to the home route or to the api v2 not to api v1 please that's not meant to be changed at all.

If you need help adding a new api v2 method jump on slack and I'll talk you through it.

fernandog:
I added to v2 too. Check it.

What I dont know is how to call it

jxer:
looking good fernandog

PR #2132, opened 2017-02-09
Fix forced backlogged not being limited by BACKLOG_DAYS

fernandog:
as @ratoaq2 pointed

dividing integers will result in integer so
`self.cycleTime = app.BACKLOG_FREQUENCY / 60 / 24` will result 0.5 but value will be 0

so a backlog of 12 hours, will always have a cycleTime of  0 in case of the default of 720

So when checking if we should run a forced search, it will never reach that limit code:
`if not which_shows and not ((curDate - self._last_backlog) >= self.cycleTime):`
as `(curDate - self._last_backlog)` will always be higher than cycleTime

Imo this code was done to check if user runned a backlog before waiting for next backlog (so, a forced backlog)

image

### DIFF
--- a/medusa/search/backlog.py
+++ b/medusa/search/backlog.py
@@ -44,7 +44,7 @@ class BacklogSearcher(object):
     def __init__(self):
 
         self._last_backlog = self._get_last_backlog()
-        self.cycleTime = app.BACKLOG_FREQUENCY / 60 / 24
+        self.cycleTime = app.BACKLOG_FREQUENCY / 60.0 / 24
         self.lock = threading.Lock()
         self.amActive = False
         self.amPaused = False
--- a/medusa/search/backlog.py
+++ b/medusa/search/backlog.py
@@ -49,6 +49,7 @@ class BacklogSearcher(object):
         self.amActive = False
         self.amPaused = False
         self.amWaiting = False
+        self.force = False
         self.currentSearchInfo = {}
 
         self._resetPI()
@@ -90,9 +91,11 @@ class BacklogSearcher(object):
         curDate = datetime.date.today().toordinal()
         from_date = datetime.date.fromordinal(1)
 
-        if not which_shows and not ((curDate - self._last_backlog) >= self.cycleTime):
-            logger.log(u"Running limited backlog on missed episodes " + str(app.BACKLOG_DAYS) + " day(s) and older only")
+        if not which_shows and self.force:
+            logger.log(u"Running limited backlog search on missed episodes from last {0} days".format(app.BACKLOG_DAYS))
             from_date = datetime.date.today() - datetime.timedelta(days=app.BACKLOG_DAYS)
+        else:
+            logger.log(u"Running full backlog search on missed episodes for selected shows")
 
         # go through non air-by-date shows and see if they need any episodes
         for cur_show in show_list:
@@ -186,6 +189,8 @@ class BacklogSearcher(object):
 
     def run(self, force=False):
         try:
+            if force:
+                self.force = True
             self.search_backlog()
         except:
             self.amActive = False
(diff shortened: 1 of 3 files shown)


duramato:
What is the current behaviour?
If i'm reading correctly, now it will be: episodes with WANTED status older than "Backlog Days" won't ever get searched without user interaction?

fernandog:
How ep become wanted? How long is wanted?

duramato:
Became wanted becouse it aired, let's say it has been WANTED for longer than the "backlog days setting" value, what is the current behaviour and what will be the behaviour with this PR?

fernandog:
I changed this PR and now it only fixed issue described in PR body:

Also added a note in the UI:
image

medusa/search/backlog.py
```diff
@@ -49,6 +49,7 @@ def __init__(self):
         self.amActive = False
         self.amPaused = False
         self.amWaiting = False
+        self.force = False
```
medariox: Why do you make a new attribute for this? Just pass it to self.search_backlog(force)

medusa/search/backlog.py
```diff
@@ -90,9 +91,11 @@ def search_backlog(self, which_shows=None):
         curDate = datetime.date.today().toordinal()
         from_date = datetime.date.fromordinal(1)
 
-        if not which_shows and not ((curDate - self._last_backlog) >= self.cycleTime):
-            logger.log(u"Running limited backlog on missed episodes " + str(app.BACKLOG_DAYS) + " day(s) and older only")
+        if not which_shows and self.force:
+            logger.log(u"Running limited backlog search on missed episodes from last {0} days".format(app.BACKLOG_DAYS))
```
medariox: Use single quotes

medusa/search/backlog.py
fernandog: because force is also a attribute of the backlog.

medariox: What do you mean?

p0psicles:
If I run a python interpreter and run 4 / 0.5 / 3 it results in 2.666666666...?

labrys:
@p0psicles you have a float in the division thus the result is a float.  Same if you did any of the following:
```python
>>> float(3)/4
0.75

>>> 3.0/4
0.75

>>> 3/4.0
0.75

>>> from  __future__ import division
>>> 3/4  # after future import
0.75
>>> 3 // 4  # for integer division after future import
0
```

p0psicles:
aah yeah, now I see what he meant. well then if you expect a float, just divide by 24.0 right?

labrys:
You should actually promote first element to avoid this kind of issue:
```python
>>> 3 / 4 / 0.5
0

>>> 3.0 / 4 / 0.5
1.5
```

labrys:
Personally I prefer to use `__future__ division` everywhere and use `//` if I need explicit integer division

labrys:
Regardless integer division isnt actually a bug here.
It's being used at L93 in backlog.py as follows:
```
        if not which_shows and not ((curDate - self._last_backlog) >= self.cycleTime):
```
And curDate is a date (min resolution is a day) so 0 is the correct answer for a backlog frequency.

medusa/search/backlog.py
labrys: If you mean `force` as an attribute (meaning this backlog was triggered as the result of a "forced backlog" action) then the attribute name should be called `forced`.  But as @medariox asked: What do you mean?

fernandog: @labrys Yes. Forced because user clicked the force button on manage searches. I also added a note there

fernandog:
@labrys the forced is better than do all that math right? Its what the feature is called "forced search" will use backlog days

fernandog:
@labrys about frequency, in server status it show cycle time of 12 hours

fernandog:
@labrys changed to `forced` and promoted to float. btw he backlog_days limit doesn't use that cycleTime anymore

fernandog:
Another PR ready waiting for approval

medusa/search/backlog.py
fernandog: @medariox can you accept changes please?

PR #2133, opened 2017-02-09
 Improve extension check and log unwanted files if we are not going to remove then

fernandog:
@medariox

### DIFF
--- a/medusa/process_tv.py
+++ b/medusa/process_tv.py
@@ -270,7 +270,6 @@ def processDir(dirName, nzbName=None, process_method=None, force=False, is_prior
                     videoFiles += [x for x in rarContent if helpers.is_media_file(x)]
 
                 videoInRar = [x for x in rarContent if helpers.is_media_file(x)] if rarContent else ''
-                allowed_extensions = app.ALLOWED_EXTENSIONS.split(',')
 
                 # Don't Link media when the media is extracted from a rar in the same path
                 if process_method in (u'hardlink', u'symlink') and videoInRar:
@@ -294,8 +293,9 @@ def processDir(dirName, nzbName=None, process_method=None, force=False, is_prior
                         continue
 
                     delete_folder(os.path.join(processPath, u'@eaDir'))
-                    notwantedFiles = [x for x in fileList
-                                      if x not in videoFiles and helpers.get_extension(x) not in allowed_extensions]
+                    allowed_extensions = app.ALLOWED_EXTENSIONS.split(',')
+                    notwantedFiles = list(x for x in fileList
+                                          if x not in videoFiles and helpers.get_extension(x) not in allowed_extensions)
                     if notwantedFiles:
                         result.output += logHelper(u"Found unwanted files: %s" % notwantedFiles, logger.DEBUG)
                         delete_files(processPath, notwantedFiles, result)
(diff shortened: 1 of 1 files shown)


fernandog:
someone approve this please?

labrys:
iirc there's a reason rarfiles were unwanted files.. maybe for issues like the recursive rars?

fernandog:
@labrys didn't get it

we don't move/copy/hardlink .rar files.
the unwanted files in the end will be deleted and we can't delete the .rar files

labrys:
@fernandog When you are done processing a directory, you **want** to remove any remaining rarfiles (like for example when you have recursive rar files).. if you exclude them from the notwantedFiles you then don't remove them at line 303 and end up with leftover garbage in your PP folder.

fernandog:
Ok. To "move" its valid point.
For others they cant be touched.
Add another check for "move" ?

medusa/process_tv.py
```diff
@@ -569,7 +572,7 @@ def process_media(processPath, videoFiles, nzbName, process_method, force, is_pr
                                                        u"Continuing the post-process of this file: %s" % cur_video_file)
                         else:
                             associated_files = processor.list_associated_files(cur_video_file_path, subtitles_only=True)
-                            if not [f for f in associated_files if f[-3:] in subtitle_extensions]:
+                            if not [f for f in associated_files if helpers.get_extension(f) in subtitle_extensions]:
```
labrys: While i like improving the extension checking by using `os.path.splitext`, is it really necessary to use a helper for such a simple function?

Using `os.path.splitext` directly is more explicit and doesn't require an imported helper just to get the index of the result.  

Also to that end, in the `get_extension` helper instead of `[1:]` to strip the `.` from the extension, `.lstrip('.')` is more explicit and easier to understand.

medusa/process_tv.py
```diff
@@ -271,7 +271,10 @@ def processDir(dirName, nzbName=None, process_method=None, force=False, is_prior
 
                 videoInRar = [x for x in rarContent if helpers.is_media_file(x)] if rarContent else ''
                 allowed_extensions = app.ALLOWED_EXTENSIONS.split(',')
-                notwantedFiles = [x for x in fileList if x not in videoFiles and x[-3:] not in allowed_extensions]
+                notwantedFiles = [x for x in fileList
+                                  if x not in videoFiles and
+                                  helpers.get_extension(x) not in allowed_extensions and
+                                  x not in rarFiles]
```
labrys: As mentioned in a previous comment, I don't think we should be excluding rarfiles from the unwanted files.

labrys:
@fernandog if you're using a method that requires it to remain, then why are you removing any files?  For example if you extract an original rar you should process the extracted files and not any of the original folder contents.

fernandog:
Hard linking allowed file extensions.
For example I dont like nfo (not in my allowed extensions). So it log as unwanted, dont hardlink but also dont erase it.

Its used for two things:
Process it or not
Erase it or not

medusa/process_tv.py
fernandog: @labrys rarfiles are unwanted only in MOVE. For hardlink/copy/symlink they are unwanted. We only delete the folder in move. For the others, rarfiles should be kept untouched

fernandog:
@labrys i changed it all, check it out

medusa/process_tv.py
```diff
@@ -271,9 +271,6 @@ def processDir(dirName, nzbName=None, process_method=None, force=False, is_prior
 
                 videoInRar = [x for x in rarContent if helpers.is_media_file(x)] if rarContent else ''
                 allowed_extensions = app.ALLOWED_EXTENSIONS.split(',')
```
labrys: You're separating the allowed_extensions from where they are used now

fernandog:
@labrys done

PR #2134, opened 2017-02-10
add vue route

OmgImAlexis:
@labrys this adds the route we talked about on Slack.


OmgImAlexis:
I'll be moving the vue repo into this route so it can be tested along side the main UI.

OmgImAlexis:
@p0psicles it's just the name for now @labrys said to just keep it with vue until we think of something else to use. I was going to use `/new` but didn't want the weird tornado regex issue of it overriding newShow, etc.

fernandog:
are we going to merge this?

labrys:
It's not labeled as having been concluded yet.  When its concluded and reviewed we can merge it.

OmgImAlexis:
Closing this as I think it may be easier to add a file to handle this like @labrys and I talked about.

PR #2139, opened 2017-02-10
Adds actions (search, manual search and archive) to backlogOverview

fernandog:
@OmgImAlexis ideas. I just started to code

TODO:
- [x] Manual search doesn't get redirected
- [x] Forced search doesn't run and don't add spin wheel to icon
- [x] If ep gets archived need to remove that row
- [x] If ep gets snatched in forced search need to remove that row
- [x] Resize archive icon to 32x32

pasted image at 2017_02_11 11_22


static/js/manage/backlog-overview.js
```diff
@@ -6,6 +6,10 @@ MEDUSA.manage.backlogOverview = function() {
         }
     });
 
+    $.ajaxEpSearch({
```
OmgImAlexis: @fernandog we can add another option to the ajaxEpSearch to remove the row if found or even replace the whole row with a new one if we want.

We can also base the archive button's code form what this uses.

fernandog:
Its painting as snatched but its wrong as its searching

image

OmgImAlexis:
Painting? You mean the colour is changing to snatched? I can add some extra code to the ajaxEpisodeSearch function to fix that.

static/js/manage/backlog-overview.js
```diff
@@ -7,7 +7,7 @@ MEDUSA.manage.backlogOverview = function() {
     });
 
     $.ajaxEpSearch({
-        colorRow: true
+        colorRow: False
```
OmgImAlexis: `false` is lowercase in js

fernandog:
Issues:
Somehow the forced search is redirecting to a json result
spinning wheel takes too long to appear after the click

views/manage_backlogOverview.mako
```diff
@@ -8,6 +8,7 @@
 %>
 <%block name="scripts">
 <script type="text/javascript" src="js/ajax-episode-search.js?${sbPID}"></script>
+<script type="text/javascript" src="js/manage/backlog-overview.js?${sbPID}"></script>
```
OmgImAlexis: This isn't needed as it's loaded in the main mako file and then the contents are run when the js loads.

static/js/manage/backlog-overview.js
```diff
@@ -1,4 +1,60 @@
 MEDUSA.manage.backlogOverview = function() {
+
+    $(document).ready(function() {
```
OmgImAlexis: Remove the document ready as this whole function is run when the page starts so it should work just by using the `checkManualSearches();`.

static/js/manage/backlog-overview.js
```diff
@@ -1,4 +1,60 @@
...
+            },
+            timeout: 15000 // timeout every 15 secs
+        }).done(function(data) {
+            if (data.episodes) {
+                pollInterval = 5000;
+            } else {
+                pollInterval = 15000;
+            }
+            updateForcedSearch(data);
+            // cleanupManualSearches(data);
```
OmgImAlexis: Guessing this commented out bit can be removed too?

views/manage_backlogOverview.mako
```diff
@@ -2,15 +2,15 @@
...
     import datetime
-    from medusa.common import Overview, Quality
+    from medusa.common import ARCHIVED, DOWNLOADED,Overview, Quality, qualityPresets, statusStrings
     from medusa.helper.common import episode_num
     from medusa import sbdatetime, network_timezones
 %>
 <%block name="scripts">
-<script type="text/javascript">
-</script>
+<script type="text/javascript" src="js/manage/backlog-overview.js?${sbPID}"></script>
```
OmgImAlexis: Remove this please as it's added in the main mako file.

static/js/manage/backlog-overview.js
fernandog: i tested and it doesn't work without it

OmgImAlexis: If it's not working without then it needs to be fixed before merging as the whole `MEDUSA` object only runs when the page is loaded.

static/js/manage/backlog-overview.js
```diff
@@ -1,4 +1,59 @@
...
+        });
+    }
+
+    function updateForcedSearch(data) {
+        $.each(data.episodes, function(name, ep) {
+            var el = $('a[id=' + ep.show + 'x' + ep.season + 'x' + ep.episode + ']');
+            var img = el.children('img[data-ep-search]');
+            var parent = el.parent();
+            if (el) {
+                if (ep.searchstatus.toLowerCase() === 'searching' || ep.searchstatus.toLowerCase() === 'queued') {
```
OmgImAlexis: Why not just save `ep.searchStatus.toLowerCase()` to a var?
Also you can use data attributes so you don't have to parse the text. This will future proof it for when we add in translations.

PR #2147, opened 2017-02-13
Add show plot to imdb_info table

fernandog:
Now info can be displayed in displayShow

TODO:
- [x] Remove comment the code that increase minor version
- [x] Improve layout to show plot

ping @wimpyrbx and @Thraxis

### DIFF
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -528,4 +528,8 @@ class AddPlot(AddInfoHash):
         logger.log(u"Adding column plot in imdb_info")
         if not self.hasColumn("imdb_info", "plot"):
             self.addColumn("imdb_info", "plot", 'TEXT', None)
+
+        logger.log(u"Adding column plot in tv_show")
+        if not self.hasColumn("tv_shows", "plot"):
+            self.addColumn("tv_shows", "plot", 'TEXT', None)
         # self.inc_minor_version()
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -30,7 +30,7 @@ MIN_DB_VERSION = 40  # oldest db version we support migrating from
 MAX_DB_VERSION = 44
 
 # Used to check when checking for updates
-CURRENT_MINOR_DB_VERSION = 4
+CURRENT_MINOR_DB_VERSION = 5
 
 
 class MainSanityCheck(db.DBSanityCheck):
@@ -532,4 +532,4 @@ class AddPlot(AddInfoHash):
         logger.log(u"Adding column plot in tv_show")
         if not self.hasColumn("tv_shows", "plot"):
             self.addColumn("tv_shows", "plot", 'TEXT', None)
-        # self.inc_minor_version()
+        self.inc_minor_version()
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -301,7 +301,7 @@ class InitialSchema(db.SchemaUpgrade):
             queries = [
                 "CREATE TABLE db_version(db_version INTEGER);",
                 "CREATE TABLE history(action NUMERIC, date NUMERIC, showid NUMERIC, season NUMERIC, episode NUMERIC, quality NUMERIC, resource TEXT, provider TEXT, version NUMERIC DEFAULT -1);",
-                "CREATE TABLE imdb_info(indexer_id INTEGER PRIMARY KEY, imdb_id TEXT, title TEXT, year NUMERIC, akas TEXT, runtimes NUMERIC, genres TEXT, countries TEXT, country_codes TEXT, certificates TEXT, rating TEXT, votes INTEGER, last_update NUMERIC);",
+                "CREATE TABLE imdb_info(indexer_id INTEGER PRIMARY KEY, imdb_id TEXT, title TEXT, year NUMERIC, akas TEXT, runtimes NUMERIC, genres TEXT, countries TEXT, country_codes TEXT, certificates TEXT, rating TEXT, votes INTEGER, last_update NUMERIC, plot TEXT);",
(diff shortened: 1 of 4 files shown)


fernandog:
Proof is working:

image

OmgImAlexis:
Not all shows have an imdb ID and not all shows that have an imdb linked to them have the best plot on there. Some don't even have a plot at all as I found out.

I'd rather see this try and use the plot from the indexer and then fall back to using the imdb's plot if possible. This would keep with the idea that you user is selecting a certain indexer for a reason.

fernandog:
@p0psicles can we somehow use get_externals and get show plot from indexer as a fall back do imbd plot (or maybe imdb as fallback?)

wimpyrbx:
just a thought. depending on how big these plots are. would it be possible to truncate them at a certain length?

it would help towards keeping the plot-area somewhat consistant in the layout.

example; if plot length above x chars use a "... (click for more)" to extend the info ?

labrys:
@fernandog I think imdb should be fallback to indexer.  Also can we get rid of the prefix "Plot: " move it to the first element, expand it full width of the section, and add a line break between the plot paragraph and the rest of the info?  I think it would look cleaner that way.

p0psicles:
I agree with labrys, let's use indexer plot when possible. Maybe the user choose that indexer for a reason.
@labrys i think you meant, use a paragraph and style it, instead of adding a line break right?

fernandog:
@p0psicles I added one show, but plot didn't show up. only after a did a full refresh

fernandog:
displayShow layout is not in the scope of my commits. If someone wants to commit any .mako changes feel free

labrys:
@p0psicles I meant linebreak typographically not as in the <br> tag, so yes a paragraph would be appropriate.

medusa/databases/main_db.py
```diff
@@ -511,3 +511,21 @@ def execute(self):
...
+        """
+        return self.connection.version >= (44, 5)
+
+    def execute(self):
+        backupDatabase(self.connection.version)
+
+        logger.log(u"Adding column plot in imdb_info")
+        if not self.hasColumn("imdb_info", "plot"):
+            self.addColumn("imdb_info", "plot", 'TEXT', None)
+        # self.inc_minor_version()
```
labrys: Don't forget to uncomment this before merging.

labrys:
@fernandog mako edits are most definitely within the scope of the PR since it affects the view of the shows display.  As is it makes the show display look worse since it's more cluttered.

fernandog:
@labrys I said of MY commits. Not of the PR. Im nlt good with design. I already said if someone wants to change mako feel free

medusa/databases/main_db.py
fernandog: already in TODO in PR body

fernandog:
@pymedusa/developers 
ideas to show the plot? some shows have 15 lines of plot. others 1 or 2

p0psicles:
This could be an option http://dotdotdot.frebsite.nl.

fernandog:
but how user will read the full plot?  hint?

wimpyrbx:
could either have hint or make a modal popup when clicking "...(more)" link?
for example
https://jsfiddle.net/jasny/ZcLSE/
obviously without a "save" button hehe.. could perhaps also display some other artwork etc as well in the modal to fanzy it up a bit.

OmgImAlexis:
Guys instead of overcomplicating this why not just use the same thing we use for episode plots on the schedule page?

p0psicles:
you mean a tooltip?

wimpyrbx:
image

p0psicles:
I was thinking more in the line of making the div expandable, by clicking on it.
Like: https://www.w3schools.com/jquerymobile/tryit.asp?filename=tryjqmob_collapsible
It's not very mobile first, if you need to press on tiny information icons.

p0psicles:
At least the solution should be done in native js. Jquery plugins are okay if needed. But I think you have the right idea, with your commits @wimpyrbx

fernandog:
why instead of click a icon, just click the plot text to expand?

p0psicles:
Yeah @fernandog i was also thinking of that. And if you expand it, give it a small shadow, so it pops out a little. We could also make it absolute, meaning it would not mess with the rest of the layout.

wimpyrbx:
image

but when clicking this happens:

image

fernandog:
nice. imo needs some JS to increase that. but have no idea how

static/js/home/display-show.js
```diff
@@ -1,4 +1,13 @@
 MEDUSA.home.displayShow = function() { // eslint-disable-line max-lines
+	$('.imdbPlot').click(function() {
+		$(this).prev('span').toggle();
+		if ($(this).html() === "..show less") {
+			$(this).html("..show more");
+		} else {
+			$(this).html("..show less");
+		}
+	});
```
Thraxis: Add a call to moveSummaryBackground(); after the if.. else... check here to resize the background bar

static/js/home/display-show.js
```diff
@@ -1,4 +1,14 @@
 MEDUSA.home.displayShow = function() { // eslint-disable-line max-lines
+	$('.imdbPlot').click(function() {
```
OmgImAlexis: Please use `$('.imdbPlot').on('click', function() {` as click is deprecated.

p0psicles:
@OmgImAlexis why did you approve this? I though we didn't want to add this through mako code? I was working on replacing it with js?

OmgImAlexis:
@p0psicles @fernandog asked me to approve the main changes to it. I didn't notice the mako stuff until you just tagged me. I had assumed that was all fixed before it was going to be merged.

PR #2152, opened 2017-02-13
Use RSS instead of html for ExtraTorrent provider

medariox:
ExtraTorrent started to encrypt their website source code again, so we
are forced to use the RSS feed. It is usually a few hours behind the
website, but better than nothing.

### DIFF
--- a/medusa/providers/torrent/rss/extratorrent.py
+++ b/medusa/providers/torrent/rss/extratorrent.py
@@ -60,7 +60,7 @@ class ExtraTorrentProvider(TorrentProvider):
         self.minleech = None
 
         # Cache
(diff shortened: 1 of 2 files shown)


medusa/providers/torrent/rss/extratorrent.py
```diff
@@ -58,7 +60,7 @@ def __init__(self):
         self.minleech = None
 
         # Cache
-        self.cache = tv.Cache(self, min_time=30)  # Only poll ExtraTorrent every 30 minutes max
+        self.cache = tv.Cache(self, min_time=1)
```
fernandog: @medariox forgot this

PR #2154, opened 2017-02-13
Improve backlogged episodes count

fernandog:
@OmgImAlexis @duramato

Same:
image

Increase:
image

Decrease:
image

Related to: https://github.com/pymedusa/Medusa/issues/2091

### DIFF
--- a/static/js/quality-chooser.js
+++ b/static/js/quality-chooser.js
@@ -37,27 +37,25 @@ $(document).ready(function() {
         $('#allowed_qualities :selected').each(function(i, selected){
             selectedAllowed[i] = $(selected).val();
         });
-        var url = '/api/v2/' +
-                  'show/' +  $('#showIndexerName').attr('value') + $('#showID').attr('value') +
+        var url = 'show/' +  $('#showIndexerName').attr('value') + $('#showID').attr('value') +
                   '/backlogged' +
-                  '?api_key=' + apiKey +
-                  '&allowed=' + selectedAllowed +
+                  '?allowed=' + selectedAllowed +
                   '&preferred=' + selectedPreffered
-        $.getJSON(url, function(data) {
-            var new_backlogged = data[0]
-            var current_backlogged = data[1]
-            var variation = Math.abs(new_backlogged - current_backlogged)
-            var html =  'Currently you have <b>' + current_backlogged + '</b> backlogged episodes.<br>'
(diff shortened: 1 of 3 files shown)


OmgImAlexis:
This just makes it look tacky. @p0psicles @Thraxis thoughts?

fernandog:
@OmgImAlexis despite UI, code is good?

fernandog:
@p0psicles I already changed what @OmgImAlexis asked. He is ok now

static/js/quality-chooser.js
```diff
@@ -44,7 +44,22 @@ $(document).ready(function() {
                   '&allowed=' + selectedAllowed +
                   '&preferred=' + selectedPreffered
         $.getJSON(url, function(data) {
-            $('#backlogged_episodes').text('Currently you have ' + data[1] + ' backlogged episodes. With this change you would have ' + data[0] + ' backlogged episodes');
+            var new_backlogged = parseInt(data[0])
+            var current_backlogged = parseInt(data[1])
+            var variation = Math.abs(new_backlogged - current_backlogged)
```
OmgImAlexis: Also not sure why you're using Math.abs when it'll strip negative numbers, etc. just use `-`.

```
Math.abs('-1');     // 1
Math.abs(-2);       // 2
Math.abs(null);     // 0
Math.abs("string"); // NaN
Math.abs();         // NaN
```

static/js/quality-chooser.js
```diff
@@ -44,7 +44,22 @@ $(document).ready(function() {
                   '&allowed=' + selectedAllowed +
                   '&preferred=' + selectedPreffered
         $.getJSON(url, function(data) {
-            $('#backlogged_episodes').text('Currently you have ' + data[1] + ' backlogged episodes. With this change you would have ' + data[0] + ' backlogged episodes');
+            var new_backlogged = parseInt(data[0])
```
OmgImAlexis: Change this, the data being passed back should already be an int if it's not then you need to fix the python.

After that it's fine

static/js/quality-chooser.js
fernandog: @OmgImAlexis im using it to just calculate the variation. So i just display the variation number

If is greater or not i check in the IF.
Anyway i would need to use ABS to display a positive number
"increased 5"
"decreased 5"

OmgImAlexis: All good, missed seeing the decrease listed below.

Issue #1716, opened 2016-12-01
[APP SUBMITTED]: OperationalError: too many SQL variables


ripper1967:
### INFO
**Python Version**: `2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:40:30) [MSC v.1500 64 bit (AMD64)]`
**Operating System**: `Windows-10-10.0.14393`
**Locale**: `cp1252`
**Branch**: [develop](../tree/develop)
**Database**: `44.2`
**Commit**: pymedusa/Medusa@dd57c33e77ff6f7b3fac5db4b888336ddb3c3407
**Link to Log**: https://gist.github.com/d1f3e3609f5b733cddca07c0461ff131
### ERROR
<pre>
2016-12-01 08:00:00 ERROR    SHOWUPDATER :: [dd57c33] DB error: too many SQL variables
Traceback (most recent call last):
  File "C:\pymedusaSickRage\medusa\db.py", line 266, in action
    sql_results = self._execute(query, args, fetchall=fetchall, fetchone=fetchone)
  File "C:\pymedusaSickRage\medusa\db.py", line 106, in _execute
    sql_results = self.connection.cursor().execute(query, args)
OperationalError: too many SQL variables
</pre>
---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

fernandog:
@p0psicles show updater

fernandog:
@ripper1967 

can you easily reproduce that error?
stop medusa, edit config.ini, set dbdebug = 1, save, run medusa. reproduce error
report issue again

p0psicles:
Or maybe post the full log? That would also really help. Then I can see what happened just before the exception.

fernandog:
@p0psicles there is not more log than that. logs already have debug

maybe add more logs to that part of code.

ripper1967:
yes i can kinda produce this   it only does it on the set time of Choose hour to update .  i can do a manual mass update and dont get it will reproduce  it  at 1500

fernandog:
Just update develop and let auto show update starts

ripper1967:
ok

ripper1967:
got it it produced now what

ripper1967:
here is the error now  from 1500  is the time i have it set to auto update https://gist.github.com/ripper1967/9bb1acf0d0e11ceab835a59896a70a2a

fernandog:
@ripper1967 please update and report

[...]

PR #2158, opened 2017-02-13
Fix OperationalError: too many SQL variables

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/1716

### DIFF
--- a/medusa/network_timezones.py
+++ b/medusa/network_timezones.py
@@ -57,8 +57,10 @@ def update_network_dict():
             if key is None or val is None:
                 continue
             d[key] = val
-    except (IOError, OSError):
-        pass
+    except (IOError, OSError) as error:
+        logger.log('Unable to build the network dictionary. Aborting update. Error: {error}'.format
+                   (error=error), logger.WARNING)
+        return
 
     cache_db_con = db.DBConnection('cache.db')
 


fernandog:
@labrys you reviewed an old merged PR about popsm is that right?

fernandog:
I looked all queries that somehow uses a join of "?" and I don't see any other that should use this helper

p0psicles:
What's bugging me in this PR, is that we are adding a work-around, for a situation that should never occur right? Why would we want to delete a very large list of networks?

fernandog:
@p0psicles i see no errors while getting the networks in his log

what code basic does is load all cached networks to network_list var,
if the remote network is found in cached, we remove from the network_list var
so in the end, we delete all networks from network_list

what we can do is: if we read 0 networks from remote, don't do anything to the cached one

fernandog:
what is wrong is this:
```
    except (IOError, OSError):
        pass
```
maybe some ERROR and the variable that keeps the remove network is empty, so it delete it all

fernandog:
now i abort if exception. I see no other reason that could cause a massive delete.
thanks for the feedback

medusa/network_timezones.py
```diff
@@ -70,14 +73,17 @@ def update_network_dict():
...
+            queries.append(['UPDATE OR IGNORE network_timezones '
+                            'SET timezone = ? WHERE network_name = ?;', [timezone, network]])
 
         if existing:
             del network_list[network]
 
     if network_list:
         purged = [x for x in network_list]
-        queries.append(['DELETE FROM network_timezones WHERE network_name IN (%s);' % ','.join(['?'] * len(purged)), purged])
+        for paged_list in get_rows_by_slice(purged, 50):
```
p0psicles: This can be reverted.

medusa/helpers.py
```diff
@@ -1729,3 +1729,9 @@ def is_info_hash_processed(info_hash):
                                         'WHERE d.action LIKE "%04"',
                                         [info_hash])
     return bool(history_result)
+
+
+def get_rows_by_slice(seq, row_len):
```
p0psicles: As this.

medusa/network_timezones.py
```diff
@@ -25,6 +25,7 @@
 from six import iteritems
 from . import db, helpers, logger
 from .helper.common import try_int
+from .helpers import get_rows_by_slice
```
p0psicles: As this.

medusa/network_timezones.py
fernandog: @p0psicles Why not keep for safety?

fernandog:
@p0psicles waiting approval as its ready

Issue #2073, opened 2017-02-01
TVShow.is_recently_deleted does not support multi-indexers.


labrys:
```python
    @property
    def is_recently_deleted(self):
        return self.indexerid in app.RECENTLY_DELETED
```
This doesn't check which indexer the indexerid comes from so if you have the same ID number from two different and one of the shows is deleted, this would incorrectly flag the remaining show as recently deleted.

p0psicles:
There are currently probably a hundred bugs, where we assume indexerid is always unique over all indexers. Until we seriously going to handle them all, we'll have to park this one.

Best you can do, is PR and add a ```# TODO: multiindexer fix needed```
Or something like that.

labrys:
Which is why I opened the issue.  TODO's alone tend to just get forgotten.

p0psicles:
Well just as easily as GitHub issues. But then it's in code, and flagged. Adding todos is how everyone can contribute. We should do that more often to mark tech debt imo.

labrys:
Key phrase "TODO's alone..".  I documented it in my edits to tv.py with a todo already.

PR #2162, opened 2017-02-14
 Add support to recently deleted global var for multi-indexer

fernandog:
IMO show object must have this unique identifier as an attribute, example 'tvmaze1234'

With new api V2, we will need it in mako and in JS, so it would be good to have in the show object
ping @OmgImAlexis 

Fixes https://github.com/pymedusa/Medusa/issues/2073

### DIFF
--- a/medusa/show_queue.py
+++ b/medusa/show_queue.py
@@ -33,6 +33,7 @@ from .helper.exceptions import (
 from .helper.externals import check_existing_shows
 from .helpers import chmod_as_parent, delete_empty_folders, get_showname_from_indexer, make_dir
 from .indexers.indexer_api import indexerApi
+from .indexers.indexer_config import indexerConfig
 from .indexers.indexer_exceptions import (IndexerAttributeNotFound, IndexerError, IndexerException,
                                           IndexerShowAllreadyInLibrary, IndexerShowIncomplete,
                                           IndexerShowNotFoundInLanguage)
@@ -242,7 +243,7 @@ class ShowQueue(generic_queue.GenericQueue):
         self.add_item(queue_item_obj)
 
         # Show removal has been queued, let's updaste the app.RECENTLY_DELETED global, to keep track of it
-        app.RECENTLY_DELETED.update([show.indexerid])
+        app.RECENTLY_DELETED.update(['{0}{1}'.format(indexerConfig[show.indexer].get('identifier'), show.indexerid)])
 
         return queue_item_obj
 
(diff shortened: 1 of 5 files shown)


medusa/show_queue.py
```diff
@@ -242,7 +243,7 @@ def removeShow(self, show, full=False):
         self.add_item(queue_item_obj)
 
         # Show removal has been queued, let's updaste the app.RECENTLY_DELETED global, to keep track of it
```
duramato: `updaste` -> `update`

fernandog: ops

p0psicles:
Yeah that does make sense. Although I don't think tvmaze1234 is meant as a unique identifier. Or at least it is now. But it's more a kind of slug. And at some time, we should start using the internal tv_show row_id or something similar. But until then, I think it makes sense to use the slug for this.

p0psicles:
Let's have a discussion on this. We should do that short time.
If we approve this one, can you pick it up again, if we make the decision to add the "slug" to the series object.

fernandog:
i tried to add the slug but i got confused when/where it should be added. in the __init__ or a static method?

p0psicles:
Let's discuss it first, then see how to implement it.

ratoaq2:
Create a read only property in Series object that returns the indexer name. And another one that returns this indexer slug id

fernandog:
If later we decide to change slug pattern, just need to change the property

fernandog:
Tested this with backlogged episodes count in editShow and it's working ok. also in displayShow

ping @OmgImAlexis

fernandog:
anyone have an alternative idea to @ratoaq2 idea of a property in Series object?

fernandog:
Renamed to indexer_slug as @ratoaq2 requested

labrys:
We should eventually remove all the various types of indexer_id, indexer combinations and even indexer_slug and replace with a simple ID class.  e.g.:
```python
class SeriesID:
    def __init__(self, source, ident):
        self.source = source
        self.id = ident

    def __repr__(self):
        return '{cls}({source}, {id})'.format(cls=self.__class__.__name__, source=self.source, id=self.id)

    def __str__(self):
        return '{source}_id {id}'.format(source=self.source, id=self.id)

    @property
    def slug(self):
        return '{source}{id}'.format(source=self.source, id=self.id)

>>> series = SeriesID('tvdb', 1234)
>>> series
SeriesID(tvdb, 1234)

>>> 'The series is {series}'.format(series=series)
'The series is tvdb_id 1234'

>>> '/series/{series}'.format(series=series.slug)
'/series/tvdb1234'
```

Issue #1989, opened 2017-01-17
Opening a page triggers a massive hit on the database, UI non-responsive until finished


labrys:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:**  Dev
**OS:** N/A
**What you did:** Opened the UI in browser
**What happened:** Slow response, viewing log saw tons of DB hits
**What you expected:**  Responsive UI
**Logs:**

Hundreds of the following lines appear: 
```
 main.db: SELECT last_refreshed FROM xem_refresh WHERE indexer = ? and indexer_id = ? with args [1, XXXXX]
 main.db: SELECT season, episode, scene_season, scene_episode FROM tv_episodes WHERE indexer = ? and showid = ? and (scene_season or scene_episode) != 0 ORDER BY season, episode with args [1, XXXXX]
 main.db: UPDATE [xem_refresh] SET last_refreshed = ?, indexer = ? WHERE indexer_id = ? with args [1484620490, 1, XXXXX]
```

fernandog:
already fixed in develop

labrys:
This was never fixed.

fernandog:
Yes it is. Look at that PR

labrys:
That was only half of the issue.. The other half was display show getting the entire db by requesting one show at a time instead of the entire results in a single query.

fernandog:
Would you point me the query that displayShow "getting the entire db"
What I see is querying by indexer

fernandog:
i can't find this MASSIVE hit on database.

my displayShow loads in Mako: 0.1796s  in a Rpi1. 

![image](https://cloud.githubusercontent.com/assets/2620870/23088464/fbd47df2-f563-11e6-983c-41e2547c5767.png)

A show with 12 seaons loads in Mako: 0.9527s.
So the issue we in fact XEM_REFRESH

labrys:
From initial issue:
```
 main.db: SELECT season, episode, scene_season, scene_episode FROM tv_episodes WHERE indexer = ? and showid = ? and (scene_season or scene_episode) != 0 ORDER BY season, episode with args [1, XXXXX]
```
That triggers one hit to db for every show.  If you have 10 shows, you wouldn't notice much, but this is O(n) complexity and since db access is slow, this is a significant performance issue.  Instead this should query for all shows in the db, making it O(1) complexity.

fernandog:
@labrys 
[...]

PR #2172, opened 2017-02-15
Don't refresh XEM while loading mako. Only in show_refresh

fernandog:
Only Show update will update xem data by calling xem_refresh,so it will improve mako loading

ping @p0psicles @duramato


fernandog:
https://github.com/pymedusa/Medusa/issues/1989

PR #2174, opened 2017-02-16
SnatchSelection improve:  better header

wimpyrbx:

image

### DIFF
--- a/views/partials/seasonEpisode.mako
+++ b/views/partials/seasonEpisode.mako
@@ -1,8 +1,6 @@
-<h1 class="title" id="scene_exception_${show.indexerid}">
-    <a href="home/displayShow?show=${show.indexerid}" class="snatchTitle">
-        ${show.name} / Season ${season}
+Manual search for:<br>
+    <a href="home/displayShow?show=${show.indexerid}" class="snatchTitle">${show.name}</a> / Season ${season}
         % if manual_search_type != 'season':
             Episode ${episode}
         % endif
     </a>
-</h1>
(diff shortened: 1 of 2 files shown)


wimpyrbx:
image

fernandog:
The episode status change cant be in snatchSelection too.

wimpyrbx:
can you elaborate?

fernandog:
That listbox "changed selected to" + button "go"

fernandog:
In displayShow I would show that listbox on the left and the colored keys om the right.

As the checkboxes to select episodes are on the left. And the column status is on the right

wimpyrbx:
yeah i understand. for snatchselection we'll need to hide those.

fernandog:
@thraxis can you review/approve this please?

fernandog:
@wimpyrbx look:

@Thraxis moved away from the top menu because of that

image

fernandog:
With my commit:
### DisplayShow
image


-----------------

### Snatch selection:

image

wimpyrbx:
image

this happens on both displayShow and snatchSelection pages. it might need a empty row to seperate showheader from what's below.

also @Thraxis could you look at the <div> and </div> situtation? i'm not 100% sure it's correct now, but unable to see it clearly.

Thraxis:
Why did you move the checkboxes and selection back the way they were to begin with?   that change wipes out the move that was done to tighten up the slack space under the summary table?

You now have change first , auto-selects on the right, then manual selects on the left.

fernandog:
@Thraxis the checkboxes to change status are on the left. That's why i moved the `select` to the left to be together with the checkboxes

The colored keys i moved to the right, as the "status" column is in the right part of the screen.

space is still optimized, right?

Thraxis:
Look at the space difference below the summary table of these two images.  Of anything I'd think we should change "Select Filtered Episodes" to "Select Episodes" to shorten that button up a bit

image

image

wimpyrbx:
"Clear all" can also be changed to "Reset" or just "Clear"

Issue #2084, opened 2017-02-02
Fanart doesn't work properly on Mobile


duramato:
The image is self explanatory, the fanart doesn't cover the full screen. Same goes for the show details.
While the screenshoots are from the UI PR, this affects all versions of Medusa.

![captura de ecra de 2017-02-02 20 45 11](https://cloud.githubusercontent.com/assets/10121067/22567652/c2739b10-e988-11e6-963c-1bf27461ff8a.png)

![captura de ecra de 2017-02-02 20 46 09](https://cloud.githubusercontent.com/assets/10121067/22567654/c2764752-e988-11e6-9dce-90636ca199e0.png)

![captura de ecra de 2017-02-02 20 46 22](https://cloud.githubusercontent.com/assets/10121067/22567653/c274d2aa-e988-11e6-84b0-71fc272aa225.png)

fernandog:
at least to me is fixed

PR #2177, opened 2017-02-16
Fix History and Schedule Mobile View

### LINKED ISSUE #2159: Phone UI scaling wrong in Schedule and History tab
When in phone web browset UI gets messed up scaling when going to Schedule and History tab(like selecting "Request desktop site" in browser).I didnt have this problem in Sickrage/sickrage branch.Please fix if you can

Thraxis:
+ Fixes #2159 and Partial Fix for #2084

### DIFF
--- a/views/schedule.mako
+++ b/views/schedule.mako
@@ -77,6 +77,7 @@
     </div>
 </div>
 
+<div class="horizontal-scroll">
 % if 'list' == layout:
 <!-- start list view //-->
 <% show_div = 'listing-default' %>
@@ -379,6 +380,7 @@
 <!-- end calender view //-->
 </div>
 % endif
+</div>
 <div class="clearfix"></div>
 </div>
 </div>
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -4206,6 +4206,16 @@ css helper classes
     margin-bottom: 25px;
 }
 
+.horizontal-scroll {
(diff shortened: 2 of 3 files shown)


views/schedule.mako
```diff
@@ -77,6 +77,7 @@
     </div>
 </div>
 
+<div class="horizontal-scroll">
```
OmgImAlexis: Should this not be inside the if since nothing will be inside the div if layout isn't list?

Thraxis: It's outside the if so it will encompass any of the layouts.  They all have issues that having a horizontal scroll on mobile view will fix.

OmgImAlexis: Oh I didn't see the other section of it. All good. 👍

duramato:
With this change I can't use the history page to it's full, as I can't see the full page I have to scroll right or left..

image

duramato:
vs. before (where i can see the rows in their entirety)

captura de ecra de 2017-02-02 20 45 11

fernandog:
@duramato enable "desktop version"

duramato:
If I have to enable the desktop view, there is no need for a mobile version then..

The porpuse of a Mobile view is to be able to view it on mobile , and be able to use it. It may or may not have all of the info as the desktop, but should be usable.

PR #2179, opened 2017-02-16
add yarn.lock, add underscore.string, update vender file

OmgImAlexis:
@p0psicles


OmgImAlexis:
To add a new dev dep to the vender file follow this. `package.json` lives in the root but we need to move it into the .build for all build steps. @labrys @p0psicles maybe we should look at adding package.json just to the .build dir?

```
cd .build
cp ../package.json .
yarn add --dev package_name_here
cp package.json ../
rm package.json
```

To build the vender files use this.

```
cd .build
cp ../package.json .
yarn install && yarn build
rm package.json
```

OmgImAlexis:
@p0psicles are we merging this in?

p0psicles:
I've added the mixin, so it's ready to be used out of the box. I think it will definitely be useful in future. So you have my approval.

labrys:
@OmgImAlexis I'm uncertain, i find its convenient having all build related stuff in a single location for clarity, however package.json is only used by the js and doesn't have any project specific variables (like medusa version number) although it does have the hard-coded urls.

OmgImAlexis:
package.json is used for all static content. It's used for js and css lint tests as well as soon it'll be used for unit tests for the js.

If we leave it in the root then I'll change the config over to enter `.build` before it runs it's code otherwise I'll move it to `.build`.

If we leave it in the root then things like `xo` and `stylelint` can be run from the root which will check all the code for js and css errors. If we move it to `.build` it won't work properly.

labrys:
@OmgImAlexis no urls that are likely to change (except the homepage which currently points to the readme)

OmgImAlexis:
The links in there are auto generated. It doesn't matter what's in the package.json as this isn't a node project. The package.json is just used here for testing and linting.

Issue #1921, opened 2017-01-04
[APP SUBMITTED]: OSError: [Errno 2] No such file or directory: '/media/ssd/Shows/The 100/Season 01/The.100.S01E11.The.Calm.720p.HDTV-SickRage.en.srt'


duramato:
### INFO
**Python Version**: `2.7.9 (default, Sep 17 2016, 20:26:04) [GCC 4.9.2]`
**Operating System**: `Linux-4.4.38-v7+-armv7l-with-debian-8.0`
**Locale**: `UTF-8`
**Branch**: [feature/fix-add-show-from-api](../tree/feature/fix-add-show-from-api)
**Database**: `44.4`
**Commit**: pymedusa/Medusa@ea5dc4dbde61e68b4dd7176f7ca29e1b1f4d867d
**Link to Log**: https://gist.github.com/3b39a46fbeaa1bd126c461f7d33196af
### ERROR
<pre>
2017-01-04 19:54:30 ERROR    Thread-19 :: [ea5dc4d] Unable to get size for file /media/ssd/Shows/The 100/Season 01/The.100.S01E11.The.Calm.720p.HDTV-SickRage.en.srt Error: u'error 2 : No such file or directory'
Traceback (most recent call last):
  File "/home/pi/Medusa/<a href="../blob/ea5dc4dbde61e68b4dd7176f7ca29e1b1f4d867d/medusa/helpers.py#L1328">medusa/helpers.py</a>", line 1328, in get_size
    total_size += os.path.getsize(fp)
  File "/home/pi/Medusa/<a href="../blob/ea5dc4dbde61e68b4dd7176f7ca29e1b1f4d867d/medusa/init/filesystem.py#L85">medusa/init/filesystem.py</a>", line 85, in <lambda>
    return lambda *args, **kwargs: handle_output(f(*[handle_arg(arg) for arg in args], **{k: handle_arg(arg) for k, arg in kwargs.items()}))
  File "/usr/lib/python2.7/genericpath.py", line 49, in getsize
    return os.stat(filename).st_size
  File "/home/pi/Medusa/<a href="../blob/ea5dc4dbde61e68b4dd7176f7ca29e1b1f4d867d/medusa/init/filesystem.py#L85">medusa/init/filesystem.py</a>", line 85, in <lambda>
    return lambda *args, **kwargs: handle_output(f(*[handle_arg(arg) for arg in args], **{k: handle_arg(arg) for k, arg in kwargs.items()}))
[...]

PR #2194, opened 2017-02-17
Don't get size of file if file doesn't exit

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/1921


fernandog:
@duramato

Issue #2043, opened 2017-01-27
BIT-HDTV "list index out of range"


duramato:
### INFO
**Python Version**: `2.7.9 (default, Sep 17 2016, 20:26:04) [GCC 4.9.2]`
**Operating System**: `Linux-4.4.38-v7+-armv7l-with-debian-8.0`
**Locale**: `UTF-8`
**Branch**: [develop](../tree/develop)
**Database**: `44.4`
**Commit**: pymedusa/Medusa@6efed56a59a9bfe5abcf5afb030a6ff378623ff0
**Link to Log**: https://gist.github.com/8e8ed03a7bc5280efdbb38eef764b9cc
### ERROR
<pre>
2017-01-03 23:12:13 ERROR    SEARCHQUEUE-DAILY-SEARCH :: [BITHDTV] :: [8628e37] Failed parsing provider. Traceback: 'Traceback (most recent call last):\n  File "/home/pi/Medusa/medusa/providers/torrent/html/bithdtv.py", line 143, in parse\n    seeders = try_int(cells[8].get_text(strip=True))\nIndexError: list index out of range\n'
Traceback (most recent call last):
  File "/home/pi/Medusa/<a href="../blob/6efed56a59a9bfe5abcf5afb030a6ff378623ff0/medusa/providers/torrent/html/bithdtv.py#L143">medusa/providers/torrent/html/bithdtv.py</a>", line 143, in parse
    seeders = try_int(cells[8].get_text(strip=True))
IndexError: list index out of range
</pre>
---

fernandog:
@duramato can you fix this?

PR #2196, opened 2017-02-17
Fix list index out of range

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2043

@duramato

### DIFF
--- a/medusa/providers/torrent/html/bithdtv.py
+++ b/medusa/providers/torrent/html/bithdtv.py
@@ -147,8 +147,8 @@ class BithdtvProvider(TorrentProvider):
                     if not all([title, download_url]):
                         continue
 
-                    seeders = try_int(cells[8].get_text(strip=True))
-                    leechers = try_int(cells[9].get_text(strip=True))
+                    seeders = try_int(cells[8].get_text(strip=True)) if len(cells) > 7 else 1
+                    leechers = try_int(cells[9].get_text(strip=True)) if len(cells) > 8 else 0
 
                     # Filter unseeded torrent
                     if seeders < min(self.minseed, 1):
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/html/bithdtv.py
```diff
@@ -147,8 +147,8 @@ def parse(self, data, mode):
                     if not all([title, download_url]):
                         continue
 
-                    seeders = try_int(cells[8].get_text(strip=True))
-                    leechers = try_int(cells[9].get_text(strip=True))
+                    seeders = try_int(cells[8].get_text(strip=True)) if len(cells) > 7 else 1
```
medariox: What you probably want is ```>= 8```

fernandog: in math is the same result =)

medariox: Hmm, not really. ```> 9``` would be...

fernandog: so 8 > 7 is false? lol

medariox: Just think about what you are doing can be right....

medariox: It has nothing to do with Math. You have to think about what you are doing and what you want to do.

medusa/providers/torrent/html/bithdtv.py
```diff
@@ -147,8 +147,8 @@ def parse(self, data, mode):
                     if not all([title, download_url]):
                         continue
 
-                    seeders = try_int(cells[8].get_text(strip=True))
-                    leechers = try_int(cells[9].get_text(strip=True))
+                    seeders = try_int(cells[8].get_text(strip=True)) if len(cells) > 9 else 1
```
medariox: It's still wrong 😢

PR #2197, opened 2017-02-17
Feature/unicode fixes

p0psicles:
'
The PR is scoped to:
- [x] Change queue.py to use unicode_literals. As the combination of b''.format('unicode string with special cars') was causing errors.
- [x] Change old style to new style logger for Series.py
- [x] Implement future unicode_literals for Series.py
- [x] Change old style to new style logger for Episode.py
- [x] Implement future unicode_literals for Episode.py
- [x] Change old style to new style logger for queue.py


medusa/tv/episode.py
```diff
@@ -751,8 +751,7 @@ def to_json(self, detailed=True):
     def create_meta_files(self):
         """Create episode metadata files."""
         if not self.show.is_location_valid():
-            logger.warning('{id}: The show dir is missing, unable to create metadata', id=self.show.indexerid),
-
+            logger.warning('{id}: The show dir is missing, unable to create metadata', id=self.show.indexerid)
```
labrys: Use `.format` to apply arguments to the message until a proper brace-message adapter is written.

fernandog:
No need to change that logger because its used everywhere

labrys:
Arguments for a brace-message adapter should be in the form of a single dictionary, not kwargs, to avoid name collisions if multiple logging adapters are applied.  Until our existing brace-message adapter is fixed to allow for a single dict of kwargs, we need to stick `.format`.  Also converting from `.format` to a single dict of kwargs is a trivial regex while the reverse (switching from kwargs to a single dict) is less so.

fernandog:
No point change it for just one log. when/if that change happens, will need to change all.

labrys:
The point is that changing them from kwargs to a single dict is much more difficult to do later than from `.format` to a single dict, so adding more log messages that use kwargs will make the job much more difficult later on than it needs to be.  Also `.format` works with the current logging and would work with the fixed brace-message adapter so its a safer choice.  The logs as written would break with a fixed brace-message adapter.

p0psicles:
The current logging adapter meets all our needs.  It was designed to be used like this. And widely communicated at the time. Added to that, .format doesn't provide lazy logging. @ratoaq2's logger does. I'm not rewriting these, unless we all decide the current logging implementation is faulty.

PR #2198, opened 2017-02-17
Change label "Find a show on theTVDB"

fernandog:


medariox:
There is also this:
image

Imo it would be nicer to have something dynamic.

fernandog:
Its the same label:

image

Issue #1652, opened 2016-11-23
Edge browser issues


medariox:
All issues are confirmed on current develop branch.
@OmgImAlexis It would be great if you could take a look.

1) Trying to add an existing show:
![image](https://cloud.githubusercontent.com/assets/2893509/20581457/46c8c2b0-b1d9-11e6-8e6d-172c1b4d1a04.png)

2) Using hide/show episodes button:
![image](https://cloud.githubusercontent.com/assets/2893509/20581495/85bc095a-b1d9-11e6-80b5-83ead4403706.png)

3) [Can't change episode status from ignored to wanted](https://github.com/pymedusa/Medusa/issues/1574)

OmgImAlexis:
I'll have to fire up a Windows VM to test these. @p0psicles any chance you have access to a Windows machine to look into this?

p0psicles:
Don't have w10

ripper1967:
I have windows machine  if u need to look into this,  but it does the same thing  on my machine

OmgImAlexis:
1. It seems to ignore the base href in Edge for some stupid reason and the js needs to be changed to add that before it.
We should probably do that for all `window.location.href` changes.
For example `window.location.href = $('base').attr('href') + 'addShows/existingShows'`

2. I can't confirm this with my setup.

ripper1967:
want to try on my system? i will give u the info to log in if u want

ripper1967:
or create a commit with it fixed and i will test it

OmgImAlexis:
I've got a windows 10 vm booted so that's fine. I'm looking into anything else that may have broken in IE and Edge because of Microsoft being stupid... 😞

ripper1967:
ok

OmgImAlexis:
Issue 2 which hide/show episodes I can't check at the moment as I'm waiting for a few PRs to be merged which may affect it.

labrys:
Good to know that Edge is so much better than IE :/

fernandog:
Changing status in EDGE now works. Just tested

fernandog:
Adding show also works

Except i saw this: `C:\Users\Fernando\Documents\SÃ©ries\Foreign Legion`

notice the "Séries" is showing as SÃ©ries. 

But it added ok

In display show is SÃ©ries too

PR #2200, opened 2017-02-17
Load config file as UTF-8

fernandog:
Fix root dir not decoded

Related: https://github.com/pymedusa/Medusa/issues/1652#issuecomment-280756677

### Before:
image

### After:

image

### DIFF
--- a/medusa/__main__.py
+++ b/medusa/__main__.py
@@ -290,7 +290,7 @@ class Application(object):
         if self.console_logging and not os.path.isfile(app.CONFIG_FILE):
             sys.stdout.write('Unable to find %s, all settings will be default!\n' % app.CONFIG_FILE)
 
-        app.CFG = ConfigObj(app.CONFIG_FILE)
+        app.CFG = ConfigObj(app.CONFIG_FILE, encoding='UTF-8', default_encoding='UTF-8')
 
         # Initialize the config and our threads
         self.initialize(console_logging=self.console_logging)


fernandog:
@ratoaq2 in app.py.

So decode there then?
What is odd that without decode in python I can add shows ok. Seems only mako have issues.

fernandog:
@ratoaq2 is it safe to decode here?

image

fernandog:
@medariox can you also test this in windows please?

fernandog:
@duramato still has linux with no locale right? would you help me test it ? thanks

duramato:
nope

fernandog:
@medariox still have linux without locale? would you help me test it?

medariox:
Not anymore, but I can remove the locale and test tomorrow, no problem.

fernandog:
worked with linux locale utf-8

need a test in linux without locale

fernandog:
@medariox still need help here. would you?

medariox:
I tested this with Windows and Debian without locale. Without locale posters are missing and probably even more stuff is broken. This happens with and without this PR. Anyway, even without this fix, all paths showed correctly for me, so I'm not sure what this is going to solve?

fernandog:
@medariox see the screenshots in the PR. it was supposed to fix mako

medariox:
Which system? I tried Windows, Debian with and without locale. All handled folder with non-ascii chars correctly on latest develop.

fernandog:
@medariox Win 7

labrys:
Looks like it may be an issue with getting the info from the config.  Upon adding on Win 10:
image
After saving and restarting:
image

medariox:
Okay, I will have to test everything again then. Would have been nice to have this info beforehand.

fernandog:
@wimpyrbx all good? any issues?

wimpyrbx:
tested this PR last night and only thing i noted was:
* adding show called "Mot i Brøstet" was not found when searching (norwegian). however searching for "Mot i" found it.
- also related. i generate tvshow.nfo (metadata) from medusa and to me it seems contents does not seem to handle special chars like æøå in title or plot etc.

other than that i noticed nothing else out of order.

fernandog:
@wimpyrbx not related to this PR. Confirm same issue happens in develop?

wimpyrbx:
confirmed. but.. that was the only special-char issue i noted on this PR so thought i'd mention it.

fernandog:
@medariox did you checked this? @labrys told you did the ConfigObj change while ago

PR #2211, opened 2017-02-18
Make get_xem_numbering_for_show lazy

labrys:

This is a partial fix for #1989 as discussed in the issue it does not schedule any refresh of the numbering.

### DIFF
--- a/medusa/scene_numbering.py
+++ b/medusa/scene_numbering.py
@@ -21,15 +21,20 @@
 # @copyright: Dermot Buckley
 #
 
+from collections import defaultdict
 import datetime
 import time
 import traceback
+import warnings
 
 from . import app, db, helpers, logger
 from .helper.exceptions import ex
 from .indexers.indexer_api import indexerApi
 from .scene_exceptions import xem_session
 from .show.show import Show
+
+
+XEM_NUMBERING = None
 
 
 def get_scene_numbering(indexer_id, indexer, season, episode, fallback_to_xem=True):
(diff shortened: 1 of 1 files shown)


medusa/scene_numbering.py
```diff
@@ -367,36 +379,58 @@ def get_scene_numbering_for_show(indexer_id, indexer):
...
+        series = int(indexer), int(indexer_id)
+    except TypeError:
+        logger.log(
+            'Unable to determine series. Indexer: {indexer}, ID: {id}'.format(
+                indexer=indexer, id=indexer_id,
+            ),
+            logger.WARNING
+        )
+    finally:
+        return get_xem_numbering()[series]
```
labrys: This is intentional.  The primary purpose of this statement is to convert the indexer and id to integers for use as a key for the xem numbering dictionary.  If a TypeError is raised the indexer and id are not valid and thus we return the results for `series = None, None`.   Since `None, None` should not exist in the xem numbering it will return the default, an empty dict.

medusa/scene_numbering.py
```diff
@@ -367,36 +379,58 @@ def get_scene_numbering_for_show(indexer_id, indexer):
...
+    """
+    Returns a dict of (season, episode) : (scene season, scene episode)
+    mappings for an entire show.  Both the keys and values of the dict
+    are tuples.
 
-    return result
+    Will be empty if there are no scene numbers set in xem
+    """
+    if 'refresh_data' in kwargs:
+        warnings.warn('refresh_data is deprecated!', DeprecationWarning)
```
fernandog: Don't do this. Do a proper refactor. Just take 5min to make sure everything is replaced

ping @p0psicles

labrys: This is proper refactoring.  It deprecates an option so missed refactors can easily be found and removed.  Also if you would note the PR is still in progress so I have not yet removed the obvious refactors.

p0psicles:
@labrys what should we do with this?

labrys:
@p0psicles I think this should probably still be implemented, but I think there have been enough changes to xem numbering in the code since that this PR would likely need a (nearly) full rewrite.  Close if you want.

PR #2217, opened 2017-02-18
will fix the mobile address bar color being blue

Thraxis:
+ changes mobile address bar color for dark theme
+ left the if statement even though both are same color now in case this
changes


OmgImAlexis:
@Thraxis you might as well just hardcode it then if it'll be the same for both themes.

labrys:
Except then if it changed in the future, you'd have to reverse the hard-coding... imo better to keep it as two separate values.

Thraxis:
Technically it is hard-coded... it's just hard-coded to change based on the theme name instead of being able to be set with CSS

PR #2231, opened 2017-02-20
Choose show's root at /home

fernandog:
Choose show's root at /home

image

@wimpyrbx @Thraxis ~~need help to make the select more aligned with other objects~~


medusa/server/web/home/handler.py
```diff
@@ -56,19 +56,33 @@ def _genericMessage(self, subject, message):
         t = PageTemplate(rh=self, filename='genericMessage.mako')
         return t.render(message=message, subject=subject, topmenu='home', title='')
 
-    def index(self):
+    def index(self, *args, **kwargs):
         t = PageTemplate(rh=self, filename='home.mako')
+        # @TODO: Uncomment this
+        # shows_root = int(kwargs.get('root'))
+        shows_root = 0
```
fernandog: Until we can do this using api v2, it defaults use the index 0

wimpyrbx:
image

fernandog:
@OmgImAlexis I can't figure it out. im getting "Method Not Allowed"

medusa/__main__.py
```diff
@@ -1466,6 +1467,7 @@ def save_config():
         new_config['General']['display_all_seasons'] = int(app.DISPLAY_ALL_SEASONS)
         new_config['General']['news_last_read'] = app.NEWS_LAST_READ
         new_config['General']['broken_providers'] = helpers.get_broken_providers() or app.BROKEN_PROVIDERS
+        new_config['General']['shows_root'] = int(app.SHOWS_ROOT)
```
OmgImAlexis: Is this meant to be an int?

static/js/home/index.js
```diff
@@ -320,4 +320,18 @@ MEDUSA.home.index = function() {
...
     });
+
+    $('#showRootDir').on('change', function(){
+        api.patch('config', {
+            showsRoot: {
+                id: $(this).val()
+            }
+        }).then(function(response) {
+            log.info(response);
+            window.location.reload();
```
OmgImAlexis: You may not want this here as it'll reload the page on change.

OmgImAlexis: @fernandog this may have been what was causing the other page to reload, no?

medusa/__main__.py
fernandog: Yes. Its a index of which root folder was used

static/js/home/index.js
fernandog: By design to only get shows with that root

PR #2232, opened 2017-02-20
Auto refresh page after change theme

fernandog:
I liked this stuff lol

@OmgImAlexis

### DIFF
--- a/static/js/config/index.js
+++ b/static/js/config/index.js
@@ -4,6 +4,20 @@ MEDUSA.config.index = function() {
         $('label[for="proxy_indexers"]').hide();
     }
 
+
+    $('#theme_name').on('change', function(){
+        api.patch('config', {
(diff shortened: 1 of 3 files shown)


static/js/config/index.js
```diff
@@ -4,6 +4,20 @@ MEDUSA.config.index = function() {
...
 
+
+    $('#theme_name').on('change', function(){
+        api.patch('config', {
+            theme: {
+                name: $(this).val()
+            }
+        }).then(function(response) {
+            log.info(response);
+            window.location.reload();
```
OmgImAlexis: Same here.

fernandog: Need to reload the page so theme changes

OmgImAlexis: Can you not swap out the CSS?

PR #2260, opened 2017-02-21
Filter backlog by airdate

fernandog:
Filter backlog by airdate

As requested by @wimpyrbx 

image


medusa/server/web/manage/handler.py
```diff
@@ -341,6 +342,15 @@ def backlogOverview(self):
...
 
+        backlog_periods = {
+            'all': None,
+            'one_day': datetime.timedelta(days=1),
+            'three_days': datetime.timedelta(days=3),
+            'one_week': datetime.timedelta(days=7),
+            'one_month': datetime.timedelta(days=30),
+        }
+        backlog_period = backlog_periods.get(app.BACKLOG_PERIOD)
+
```
fernandog: I copied this from @ratoaq2 error_logs =)

fernandog:
backlogOverview is taking 8seconds to load instead of 2 seconds (without this PR)

I have no idea why. I moved the airdate code from mako to python
So it runs the same code
How to find out what is taking too long?

fernandog:
backlogOverview is taking 8seconds to load instead of 2 seconds (without this PR)

I have no idea why. I moved the airdate code from mako to python
So it runs the same code
How to find out what is taking too long?

fernandog:
All good now: Load time: 0.6726s / Mako: 0.0451s

medusa/server/api/v2/config.py
```diff
@@ -279,6 +279,10 @@ def patch(self, *args, **kwargs):
                 theme_name = data['theme']['name']
                 app.THEME_NAME = theme_name
                 done_data['themeName'] = theme_name
+            if key == 'backlogPeriod':
```
OmgImAlexis: If you add support for GET, DELETE, PATCH, etc. then you need to add the others. So this needs a `GET` method added.

medusa/server/api/v2/config.py
```diff
@@ -279,6 +279,10 @@ def patch(self, *args, **kwargs):
...
+                backlog_period = data['backlogPeriod']['period']
```
OmgImAlexis: So you're asking for something like this?
```
{
    backlogPeriod: {
        period: 7
    }
}
```

medusa/server/api/v2/config.py
fernandog: Didnt understood

medusa/server/api/v2/config.py
fernandog: not asking, im setting the config to 7

OmgImAlexis: What I'm saying is PATCH is expecting something like the code above, right?

Why not send a PATCH of just this?
```
{
    backlogPeriod: 7
}
```

PR #2268, opened 2017-02-22
will fix several javascript lint errors

Thraxis:
Fixes problems in the javascript files

### DIFF
--- a/static/js/common/init.js
+++ b/static/js/common/init.js
@@ -224,14 +224,12 @@ function rgb2hex(rgb) {
 
 var revertBackgroundColor; // used to revert back to original background-color after highlight
 var allCells = $('.triggerhighlight');
-allCells
-.on('mouseover', function() {
+allCells.on('mouseover', function() {
     var el = $(this);
     var pos = el.index();
     var revertBackgroundColor = rgb2hex($(this).parent().css('background-color')); // fetch the original background-color to revert back to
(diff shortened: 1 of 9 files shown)


static/js/common/init.js
```diff
@@ -195,42 +195,43 @@ MEDUSA.common.init = function() {
...
+var allCells = $('.triggerhighlight');
+allCells
+.on('mouseover', function() {
+    var el = $(this);
+    var pos = el.index();
+    var revertBackgroundColor = rgb2hex($(this).parent().css('background-color')); // fetch the original background-color to revert back to
+    var highlightBackgroundColor = colorLuminance(revertBackgroundColor, -0.15); // change highlight color based on original color
+    el.parent().find('.triggerhighlight').css('background-color', highlightBackgroundColor); // setting highlight background-color
+})
+.on('mouseout', function() {
```
OmgImAlexis: This should really be inline. Like so.

```
}).on('mouseout', function() {
```

static/js/home/snatch-selection.js
```diff
@@ -1,20 +1,20 @@
 MEDUSA.home.snatchSelection = function() {
     $('.imdbPlot').on('click', function() {
```
OmgImAlexis: This should be moved into an init function or something similar as it's used on multiple pages. For another PR though.

PR #2271, opened 2017-02-22
Fix mako no detecting add|remove of root folders. Populate with JS

fernandog:
@OmgImAlexis

Mako loads once the code in the import, so it won't see that app.ROOT_DIRS changed


static/js/home/index.js
```diff
@@ -349,4 +349,27 @@ MEDUSA.home.index = function() {
...
+    if (backend_dirs.length > 1) {
+        $('#showRoot').show()
+        var item = ['All Folders'];
+        rootDirOptions = item.concat(backend_dirs);
+        $.each(rootDirOptions, function (i, item) {
+            $('#showRootDir').append($('<option>', {
+                value: i - 1,
+                text : item,
+            }));
+        $('select#showRootDir').prop('selectedIndex', rootDirIndex + 1);
```
OmgImAlexis: Please indent this.

static/js/home/index.js
```diff
@@ -349,4 +349,27 @@ MEDUSA.home.index = function() {
             log.info(err);
         });
     });
+
+    var rootDir = MEDUSA.config.rootDirs
+    var rootDirIndex = MEDUSA.config.showsRoot.id
+    if (rootDir) {
+        backend_pieces = rootDir.split('|')
+        backend_default = 'rd-' + backend_pieces[0]
+        backend_dirs = backend_pieces.slice(1)
```
OmgImAlexis: Vars need to be camelCase, please stop using snake_case.

medusa/server/api/v2/config.py
```diff
@@ -129,6 +129,9 @@ def get(self, query=''):
                     'allSeasons': bool(app.DISPLAY_ALL_SEASONS),
                     'specials': bool(app.DISPLAY_SHOW_SPECIALS)
                 }
+            },
+            'showsRoot': {
+                'id': int(app.SHOWS_ROOT) if app.SHOWS_ROOT else None
```
OmgImAlexis: Is there a reason not to just return all the show roots dirs with their id?

fernandog: I only need id

OmgImAlexis: I know YOU only need the id but is there a reason not to just return the whole show root object/string? It'd make more sense for the API to expose it all. Think about API(third party) first before how Medusa uses it.

static/js/home/index.js
OmgImAlexis: @fernandog this should be inline with the line above..

static/js/home/index.js
```diff
@@ -348,17 +346,16 @@ MEDUSA.home.index = function() {
...
     if (rootDir) {
-        backend_pieces = rootDir.split('|')
-        backend_default = 'rd-' + backend_pieces[0]
-        backend_dirs = backend_pieces.slice(1)
+        backendPieces = rootDir.split('|');;
+        backendDirs = backendPieces.slice(1)
     }
-    if (backend_dirs.length > 1) {
-        $('#showRoot').show()
+    if (backendDirs.length > 1) {
```
OmgImAlexis: This should be `>= 1` right?

fernandog: no. we only show if users has TWO root dirs

PR #2272, opened 2017-02-22
Create a new method rename_associated_files

fernandog:
TODO:
- [ ] ~~Add pytest to this new method~~
- [ ] ~~Add pytest to list_associated_files~~

UPDATE: will add pytest in another PR. too much complicated

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -333,11 +333,11 @@ class PostProcessor(object):
                 notifiers.synoindex_notifier.deleteFile(cur_file)
 
     @staticmethod
-    def rename_associated_files(new_path, new_base_name, cur_file_path):
+    def rename_associated_file(new_path, new_base_name, cur_file_path):
         """Rename associated file using media basename.
 
-        :param new_path: destination path where we want the file to
-        :param new_base_name: the media base filename (no extension) to use during the action
+        :param new_path: full show folder path were the file will be moved|copied|linked
+        :param new_base_name: the media base filename (no extension) to use during the rename
         :param cur_file_path: full path of the associated file
         :return: renamed full file path
         """
@@ -391,8 +391,8 @@ class PostProcessor(object):
 
         Can rename the file as well as change its location, and optionally move associated files too.
 
-        :param file_path: The full path of the media file to act on
-        :param new_path: Destination path where we want to the file to
+        :param file_path: The full path of the file to act on
+        :param new_path: full show folder path were the file will be moved|copied|linked
         :param new_base_name: The base filename (no extension) to use during the action. Use None to keep the same name.
         :param associated_files: Boolean, whether we should copy similarly-named files too
         :param action: function that takes an old path and new path and does an operation with them (move/copy/)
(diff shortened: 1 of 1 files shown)


fernandog:
@medariox if you want to change docstring feel free to commit.

medusa/post_processor.py
```diff
@@ -332,6 +332,57 @@ def _delete(self, file_path, associated_files=False):
...
+        changed_extension = None
+        # file extension without leading dot
+        extension = helpers.get_extension(cur_file_path)
+        # initally set current extension as new extension
+        new_extension = extension
+
+        if is_subtitle(cur_file_path):
+            code = cur_file_path.rsplit('.', 2)[1].lower().replace('_', '-')
+            if from_code(code, unknown='') or from_ietf_code(code, unknown=''):
+                if code == 'pt-br':
```
ratoaq2: I know the code was just moved, but it would be better to put a TODO here to remove this hardcoded language and handle like in subtitles.py

medariox: In fact it could already be removed, it's just there to ensure support with some software (Kodi? Plex?) that has/had problems with pt-br (not recognized) vs. pr-BR (yes, case sensitive...).

medusa/post_processor.py
```diff
@@ -332,6 +332,58 @@ def _delete(self, file_path, associated_files=False):
                 # do the library update for synoindex
                 notifiers.synoindex_notifier.deleteFile(cur_file)
 
+    @staticmethod
+    def rename_associated_files(new_path, new_base_name, cur_file_path):
```
medariox: We are renaming only one file, so this should be singular.

medusa/post_processor.py
```diff
@@ -332,6 +332,58 @@ def _delete(self, file_path, associated_files=False):
...
+        """Rename associated file using media basename.
+
+        :param new_path: destination path where we want the file to
```
medariox: Sentence makes no sense.

medusa/post_processor.py
```diff
@@ -332,6 +332,58 @@ def _delete(self, file_path, associated_files=False):
...
+        :param new_base_name: the media base filename (no extension) to use during the action
```
medariox: during the action? maybe during rename

medusa/post_processor.py
```diff
@@ -340,10 +392,10 @@ def _combined_file_operation(self, file_path, new_path, new_base_name, associate
         Can rename the file as well as change its location, and optionally move associated files too.
 
         :param file_path: The full path of the media file to act on
-        :param new_path: Destination path where we want to move/copy the file to
-        :param new_base_name: The base filename (no extension) to use during the copy. Use None to keep the same name.
+        :param new_path: Destination path where we want to the file to
```
medariox: Same as above.

medusa/post_processor.py
fernandog: its the show folder where the file will be moved|copied|hardlinked to

fernandog:
@medariox anything else?

PR #2277, opened 2017-02-23
fixes space between dropdown and text in header elements.

wimpyrbx:

image

becomes:

image


fernandog:
shouldn't CSS handle this?
how in the other objects there is space without that?

wimpyrbx:
not sure we have other pages where there is text right after a selectbox. i thought adding a &nbsp; would solve it without having to change CSS that might affect other places.

fernandog:
Fixed in develop

PR #2280, opened 2017-02-23
Method _search_files should only return files

fernandog:
@medariox 

it should improve PP performance in list_associated_files as we won't do all the checks for folders

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -249,6 +249,11 @@ class PostProcessor(object):
         :return: list with found files or empty list
         :rtype: list
         """
+        def filter_files(it):
+            for item in it:
+                if os.path.isfile(item):
+                    yield item
+
         directory = os.path.dirname(path)
 
         if base_name_only:
@@ -264,7 +269,9 @@ class PostProcessor(object):
         path = Path(directory)
         glob = path.rglob(pattern) if subfolders else path.glob(pattern)
 
-        files = [text_type(match) for match in glob]
+        found_items = (text_type(match) for match in glob)
+        # We should only returns files and not folders.
+        files = filter_files(found_items)
 
         if sort:
             files = sorted(files, key=os.path.getsize, reverse=True)


medariox:
I disagree, _search_files() is intentionally kept generic and creating another list actually makes the performance worse.

fernandog:
not creating a list. im "filtering" the list
so which way i can filter without losing performance?

fernandog:
@medariox anything else?

fernandog:
@medariox is there a way to change glob pattern to only return files?
```python
        pattern = "*"
        path = Path(directory)
        glob = path.rglob(pattern) if subfolders else path.glob(pattern)
```

labrys:
@fernandog use a generator expression or generator function... use expression form if its only intended for use in a single location, function form if intended for reuse in multiple locations... (most likely you want a function)
```python
>>> files = (f for f in found_items if os.path.isfile(f))  # generator expression
>>> files
<generator object <genexpr> at ...>
```

medusa/post_processor.py
```diff
@@ -264,7 +264,9 @@ def _search_files(path, pattern='*', subfolders=None, base_name_only=None, sort=
         path = Path(directory)
         glob = path.rglob(pattern) if subfolders else path.glob(pattern)
 
-        files = [text_type(match) for match in glob]
```
labrys: This should be a generator
`files = (text_type(match) for match in glob)`

fernandog:
What about performance as medariox said?

medusa/post_processor.py
```diff
@@ -264,7 +264,9 @@ def _search_files(path, pattern='*', subfolders=None, base_name_only=None, sort=
...
+        found_items = [text_type(match) for match in glob]
+        # We should only returns files and not folders.
+        files = [f for f in found_items if os.path.isfile(f)]
 
         if sort:
             files = sorted(files, key=os.path.getsize, reverse=True)
```
labrys: Then this could be a function to be applied to _search_files later:
```python
def sorted_by_size(files, direction='asc'):
     reversed = dict(asc=False, desc=True)
     return sorted(files, key=os.path.getsize, reverse=reversed.get(direction))
```
This would keep _search_files lightweight as a generator, and still give an easy way to get a sorted list later

fernandog:
@labrys like this?

fernandog:
@medariox all good now ?

medariox:
I've already said my opinion, I won't be reviewing this anymore.

fernandog:
but i still don't get your reason

`_search_files` is called only in `list_associated_files` - there is no reason to return folders

If we use it elsewhere ok. But we only need files

medariox:
```list_associated_files``` already returns only files.

**Comment by fernandog** [comment]
yes, but it it run all code `list_associated_files` with folders
don't you think its wasted resources run all that with folders?!

**Comment by fernandog** [comment]
@labrys thtat function doesn't work with generator

```python
        def filter_files(it):
            for item in it:
                if os.path.isfile(item):
                    yield item
```

labrys:
@fernandog closing this PR as this is based on pathlib which as was mentioned in IRC will not meet our needs.

PR #2283, opened 2017-02-23
Improve archive text

fernandog:
@p0psicles 

it will open backlog Overview in a new tab:

it also fixes showing the archive button in decrease

image

### DIFF
--- a/views/inc_qualityChooser.mako
+++ b/views/inc_qualityChooser.mako
@@ -43,14 +43,24 @@ selected = None
         <h5><b>Quality setting explanation:</b></h5>
         <h5 id="allowed_text">This will download <b>any</b> of these qualities and then stops searching: <label id="allowed_explanation">${', '.join([Quality.qualityStrings[i] for i in allowed_qualities])}</label></h5>
         <h5 id="preferred_text1">Downloads <b>any</b> of these qualities: <label id="allowed_preferred_explanation">${', '.join([Quality.qualityStrings[i] for i in allowed_qualities + preferred_qualities])}</label></h5>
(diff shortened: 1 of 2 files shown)


views/inc_qualityChooser.mako
```diff
@@ -43,13 +43,19 @@ selected = None
         <h5><b>Quality setting explanation:</b></h5>
         <h5 id="allowed_text">This will download <b>any</b> of these qualities and then stops searching: <label id="allowed_explanation">${', '.join([Quality.qualityStrings[i] for i in allowed_qualities])}</label></h5>
         <h5 id="preferred_text1">Downloads <b>any</b> of these qualities: <label id="allowed_preferred_explanation">${', '.join([Quality.qualityStrings[i] for i in allowed_qualities + preferred_qualities])}</label></h5>
-        <h5 id="preferred_text2">But it will stop searching when one of these is downloaded:  <label id="preferred_explanation">${', '.join([Quality.qualityStrings[i] for i in preferred_qualities])}</label></h5>   
+        <h5 id="preferred_text2">But it will stop searching when one of these is downloaded:  <label id="preferred_explanation">${', '.join([Quality.qualityStrings[i] for i in preferred_qualities])}</label></h5>
```
p0psicles: xo: fya. I think we should really write down some standards. I personally prefer hyphanated, and then camelCase. And today there are different  methodologies out there, that mix them, with different purposes. But maybe we should now prefer one. And I don't think snakecase is a candidate. Non-the-less, i'm approving this. As we havent' clearly set out a standard yet.

PR #2287, opened 2017-02-24
will make table views on main page reflow for mobile

Thraxis:
This will cause the tables on the main page to reflow for mobile view.

image

image

image

image

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1070,7 +1070,7 @@ displayShow.mako
 ========================================================================== */
 
 .modal-wide .modal-dialog {
-  width: 90%;
+    width: 90%;
 }
 
 .modal-dialog {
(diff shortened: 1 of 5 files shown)


fernandog:
is it possible to be an option? personally i like to swipe left-right in the table as it is in develop

p0psicles:
@xo shouldn't we also make these packages part of the build script? As in adding them to bower?

Thraxis:
@fernandog No this is not something that can easily be made into an option. short of using a variable load of CSS

OmgImAlexis:
@p0psicles ideally we should actually be adding libs that don't require jQuery instead of just adding jQuery plugins. @Thraxis I've mentioned that before.

Keep in mind all of this will need to be added to Vue at some point or another and we shouldn't have to rely on jQuery for DOM manipulation so we shouldn't be using jQuery plugins when there's alternatives.

OmgImAlexis:
@Thraxis can you show what res this was tested on as honestly this is just going to take a LOT of extra space on mobile now. Imagine someone with 100+ shows instead of 100 rows they'll now have 100*11(1,100) rows to scroll through on a small *MOBILE* screen.

Issue #410, opened 2016-04-18
[ Feature ] Force update scene exceptions button


OmgImAlexis:
It'd be nice to have a button added to the config page where you could force the scene exceptions file to be updated.

fernandog:
scene exceptions is updated before every dailysearch run

OmgImAlexis:
If I add a show and then manually search it won't necessarily be updated or if I update the exceptions on Github it won't be updated right away so adding a manual update button would help.

PR #2288, opened 2017-02-24
Feature/force exception refresh

p0psicles:

image

@OmgImAlexis @ratoaq2 can you guys please check the route ivé added, to get scene exception last refresh, and force an update? I expect that we need to make changes to that.

Currently i'm using: ```api/v2/manage/searches/sceneexception/retrieve```
For the get and post (starting the syncing from source).

### DIFF
--- a/medusa/server/api/v2/scene_exception.py
+++ b/medusa/server/api/v2/scene_exception.py
@@ -25,6 +25,8 @@ class SceneExceptionTypeHandler(BaseRequestHandler):
         self.set_header('Access-Control-Allow-Methods', 'GET')
 
     def get(self, exception_type):
+        """Return a list or instance of exception types."""
+        exception_types = []
         if exception_type is not None and exception_type not in ['medusa', 'xem', 'anidb']:
             return self.api_finish(status=400)
 
@@ -32,12 +34,12 @@ class SceneExceptionTypeHandler(BaseRequestHandler):
             mapped_exception_type = {'medusa': 'custom_exceptions'}.get(exception_type, exception_type)
             for k, v in get_last_updates().items():
                 if k == mapped_exception_type:
-                    last_updated = {"id": {'custom_exceptions': 'medusa'}.get(k, k), "lastUpdate": v}
+                    exception_types = {"id": {'custom_exceptions': 'medusa'}.get(k, k), "lastUpdate": v}
         else:
-            last_updated = [{"id": {'custom_exceptions': 'medusa'}.get(k, k), "lastUpdate": v}
-                            for k, v in get_last_updates().items()]
+            exception_types = [{"id": {'custom_exceptions': 'medusa'}.get(k, k), "lastUpdate": v}
+                               for k, v in get_last_updates().items()]
 
-        self.api_finish(data=last_updated)
+        self.api_finish(data=exception_types)
 
 
 class SceneExceptionAllTypeOperationHandler(BaseRequestHandler):
--- a/medusa/server/api/v2/scene_exception.py
+++ b/medusa/server/api/v2/scene_exception.py
@@ -89,7 +89,7 @@ class SceneExceptionHandler(BaseRequestHandler):
         super(SceneExceptionHandler, self).set_default_headers()
         self.set_header('Access-Control-Allow-Methods', 'GET, OPTIONS, PUT, POST')
 
-    def get(self, *args, **kwargs):
+    def get(self, exception_id, **kwargs):
         """Query scene_exception information.
 
         :param show_indexer_id: The indexer id. Like 1 for tmdb and 3 for tvmaze.
@@ -97,15 +97,10 @@ class SceneExceptionHandler(BaseRequestHandler):
         :type show_indexer_id: str
         :param season:
         """
-        exception_id = self._parse(kwargs.pop('id', None))
         slug = self.get_query_argument('indexer', None)
         indexer, indexer_id = slug_to_indexer_id(slug)
         season = self.get_query_argument('season', None)
-        detailed = self._parse_boolean('detailed')
         exception_type = bool(self.get_query_argument('type', None) == 'custom')
-
-        if not detailed:
-            return self.api_finish(data={"lastUpdates": get_last_updates()})
 
         cache_db_con = db.DBConnection('cache.db')
         sql_base = b'SELECT * FROM scene_exceptions'
@@ -136,8 +131,6 @@ class SceneExceptionHandler(BaseRequestHandler):
             sql_base += b' WHERE ' + b' AND '.join([where + b' = ? ' for where in sql_where])
 
         exceptions = cache_db_con.select(sql_base, params)
-        if not exceptions:
-            self.api_finish(status=404, error='SceneException(s) not found.')
 
         exceptions = [{'id': row[0],
                        'indexer': indexer_id_to_slug(row[1], row[2]),
@@ -146,6 +139,8 @@ class SceneExceptionHandler(BaseRequestHandler):
                        'type': 'custom' if row[5] else None}
                       for row in exceptions]
 
+        if exception_id:
+            return self.api_finish(data=exceptions[0]) if exceptions else self.api_finish(status=404)
         return self._paginate(exceptions, 'id')
 
     def put(self, *args, **kwargs):
--- a/medusa/server/api/v2/scene_exception.py
+++ b/medusa/server/api/v2/scene_exception.py
@@ -13,15 +13,19 @@ class SceneExceptionRetrieveHandler(BaseRequestHandler):
         super(SceneExceptionRetrieveHandler, self).set_default_headers()
         self.set_header('Access-Control-Allow-Methods', 'GET POST')
 
-    def get(self):
-        """Return statistical information on the scene_exceptions table."""
+    @staticmethod
+    def get_last_updates():
         last_updates = {}
         for scene_exception_source in ['custom_exceptions', 'xem', 'anidb']:
             last_updates[scene_exception_source] = get_last_refresh(scene_exception_source)[0]['last_refreshed']
+        return last_updates
 
-        return self.api_finish(data={'last_update': last_updates})
+    def get(self):
+        """Return statistical information on the scene_exceptions table."""
+        return self.api_finish(data={'last_update': self.get_last_updates()})
 
     def post(self):
         """Start fetch retrieving scene name exceptions from ."""
         retrieve_exceptions(force=True)
-        return self.api_finish(data={'result': 'Started retrieval of scene name exceptions.'})
+        return self.api_finish(data={'result': 'Updated scene exceptions',
+                                     'last_update': self.get_last_updates()})
--- a/medusa/server/api/v2/scene_exception.py
+++ b/medusa/server/api/v2/scene_exception.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 """Request handler for scene_exceptions."""
 
+from medusa.indexers.indexer_config import indexer_id_to_slug, slug_to_indexer_id
 from medusa.scene_exceptions import get_last_refresh, retrieve_exceptions
 from tornado.escape import json_decode
 from .base import BaseRequestHandler
@@ -49,48 +50,56 @@ class SceneExceptionHandler(BaseRequestHandler):
         :type show_indexer_id: str
         :param season:
         """
-        exception_id = self._parse(kwargs.pop('row_id'))
-        indexer = self.get_query_argument('indexer', None)
-        indexer_id = self.get_query_argument('indexerId', None)
+        exception_id = self._parse(kwargs.pop('id', None))
+        slug = self.get_query_argument('indexer', None)
+        indexer, indexer_id = slug_to_indexer_id(slug)
         season = self.get_query_argument('season', None)
         detailed = self._parse_boolean('detailed')
+        exception_type = bool(self.get_query_argument('type', None) == 'custom')
 
         if not detailed:
             return self.api_finish(data={"lastUpdates": get_last_updates()})
 
         cache_db_con = db.DBConnection('cache.db')
-        if indexer and indexer_id and season:
-            sql_base = b'SELECT * FROM scene_exceptions WHERE indexer = ? AND indexer_id = ?'
-            params = [indexer, indexer_id]
-            if season:
-                sql_base += b' AND season = ?'
-                params.append(season)
-        else:
-            sql_base = b'SELECT * FROM scene_exceptions'
-            params = []
-            if exception_id:
-                sql_base += ' WHERE exception_id = ?'
-                params.append(exception_id)
+        sql_base = b'SELECT * FROM scene_exceptions'
+        sql_where = []
+        params = []
+
+        if exception_id:
+            sql_where.append(b'exception_id')
+            params += [exception_id]
+
+        if indexer and indexer_id:
+            sql_where.append(b'indexer')
+            params += [indexer]
+
+        if indexer_id:
+            sql_where.append(b'indexer_id')
+            params += [indexer_id]
+
+        if season:
+            sql_where.append(b'season')
+            params += [season]
+
+        if exception_type:
+            sql_where.append(b'custom')
+            params += [exception_type]
+
+        if sql_where:
+            sql_base += b' WHERE ' + b' AND '.join([where + b' = ? ' for where in sql_where])
 
         exceptions = cache_db_con.select(sql_base, params)
         if not exceptions:
             self.api_finish(status=404, error='SceneException(s) not found.')
 
         exceptions = [{'id': row[0],
-                       # TODO: move to 'indexer': 'tvdb1234'
-                       # instead of indexer and indexer id
-                       'indexer': row[1],
-                       'indexerId': row[2],
-                       'indexerId': row[2],
+                       'indexer': indexer_id_to_slug(row[1], row[2]),
                        'showName': row[3],
                        'season': row[4] if row[4] >= 0 else None,
                        'type': 'custom' if row[5] else None}
                       for row in exceptions]
 
-        # TODO: return a list of scene exceptions:
-        # return self._paginate(exceptions, 'id')
-        self.api_finish(data={"exceptions": exceptions, "lastUpdates": get_last_updates()})
-
+        return self._paginate(exceptions, 'id')
 
     def put(self, *args, **kwargs):
         """Update show information.
--- a/medusa/server/api/v2/scene_exception.py
+++ b/medusa/server/api/v2/scene_exception.py
@@ -1,0 +1,27 @@
+# coding=utf-8
+"""Request handler for scene_exceptions."""
+
+from .base import BaseRequestHandler
+from medusa.scene_exceptions import get_last_refresh, retrieve_exceptions
+
+
+class SceneExceptionRetrieveHandler(BaseRequestHandler):
+    """Scene Exception request handler."""
+
+    def set_default_headers(self):
+        """Set default CORS headers."""
+        super(SceneExceptionRetrieveHandler, self).set_default_headers()
+        self.set_header('Access-Control-Allow-Methods', 'GET POST')
+
+    def get(self):
+        """Return statistical information on the scene_exceptions table."""
+        last_updates = {}
+        for scene_exception_source in ['custom_exceptions', 'xem', 'anidb']:
+            last_updates[scene_exception_source] = get_last_refresh(scene_exception_source)[0]['last_refreshed']
+
+        return self.api_finish(data={'last_update': last_updates})
+
+    def post(self):
+        """Start fetch retrieving scene name exceptions from ."""
+        retrieve_exceptions(force=True)
+        return self.api_finish(data={'result': 'Started retrieval of scene name exceptions.'})
--- a/medusa/server/api/v2/scene_exception.py
+++ b/medusa/server/api/v2/scene_exception.py
@@ -158,5 +158,3 @@ class SceneExceptionHandler(BaseRequestHandler):
         if cache_db_con.connection.total_changes - last_changes > 0:
             return self.api_finish(status=204)
         return self.api_finish(status=400, error="Failed to delete.")
-
-        return self.api_finish()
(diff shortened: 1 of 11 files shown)


medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,32 @@
...
+        last_updates = {}
+        for scene_exception_source in ['custom_exceptions', 'xem', 'anidb']:
+            last_updates[scene_exception_source] = get_last_refresh(scene_exception_source)[0]['last_refreshed']
+        return last_updates
+
+    def get(self):
+        """Return statistical information on the scene_exceptions table."""
+        return self.api_finish(data={'last_update': self.get_last_updates()})
+
+    def post(self):
```
OmgImAlexis: `POST` is for sending data to be saved. @ratoaq2 do we have anything for RPC calls?

static/js/manage/manage-searches.js
```diff
@@ -0,0 +1,48 @@
+MEDUSA.manage.manageSearches = function() {
+    // Get total number current scene exceptions per source. Will request medusa, xem and anidb name exceptions.
+    var url = 'manage/searches/sceneexception/retrieve';
+    var updateExceptionTable = function(exceptions) {
+        var status = $('#sceneExceptionStatus');
+
+        var table = $('<ul></ul>')
```
OmgImAlexis: `.append`, `.on`, etc. should be on the same line as the ending of the last.

static/js/manage/manage-searches.js
```diff
@@ -0,0 +1,48 @@
...
+                new Date(exceptions.data.last_update.xem * 1000).toLocaleDateString())
+            .append('<li>Last updated anidb exceptions ' +
+                new Date(exceptions.data.last_update.anidb * 1000).toLocaleDateString());
+        status.append(table);
+
+        $('.forceSceneExceptionRefresh').removeClass('disabled');
+    };
+
+    api.get(url)
+        .then(function(response) {
```
OmgImAlexis: Same here.

medusa/server/api/v2/scene_exception.py
p0psicles: Yes I was also wondering. Post looked to be the best option at the time.

p0psicles: I do hope this doest need to become a really long discussion, and that we can be practical about it. Yes we've choose for rest. But these  rpc calls don't perform a clear action on a single resource. We could add sceneexceptions as a resource. But then there should be some room to run these kind of rpc.

duramato:
What are the "costum exceptions"? Ours? I'd name them like Global exceptions or "Community Exceptions"
I personally prefer my 2nd suggestion.

fernandog:
"Community Exceptions" - never heard of it =)

duramato:
- Community made exceptions
- Community contributed exceptions
- Blabla

All better and more descriptive than "Costum exceptions"

static/js/manage/manage-searches.js
p0psicles: I tried to improve it. But some lines need to be broken, to prevent them from getting too long.

static/js/manage/manage-searches.js
p0psicles: This one I could continue

p0psicles:
```Medusa's Scene Exceptions```

File: static/js/manage/manage-searches.js [hunk 4]
```diff
@@ -0,0 +1,48 @@
...
+            "Last updated medusa's exceptions</a> " +
+            new Date(exceptions.data.last_update.custom_exceptions * 1000)
+                .toLocaleDateString()).append('<li>' +
+            '<a href="' + MEDUSA.config.anonRedirect +
+            'http://thexem.de">' +
+            'Last updated xem exceptions</a> ' +
+            new Date(exceptions.data.last_update.xem * 1000)
+                .toLocaleDateString()).append('<li>Last updated anidb exceptions ' +
+            new Date(exceptions.data.last_update.anidb * 1000)
+                .toLocaleDateString());
```
p0psicles: @OmgImAlexis i'm not sure how to format this. I'd personally start every .append on a new line.

static/js/manage/manage-searches.js
```diff
@@ -0,0 +1,48 @@
...
+        status.append(table);
+
+        $('.forceSceneExceptionRefresh').removeClass('disabled');
+    };
+
+    api.get(url).then(function(response) {
+        updateExceptionTable(response);
+    }).catch(function(err) {
+        console.log('Trying to get scene exceptions failed with error: ' + err);
```
OmgImAlexis: This should be using `log.error()` instead of `console.log()`.

static/js/manage/manage-searches.js
OmgImAlexis: Appends are annoying like this. Do it whichever way looks best to you and is readable.

Personally I'd set `new Date(exceptions.data.last_update.xem * 1000).toLocaleDateString()` and others to a var so it's easier to see what the append is doing.

ratoaq2:
IMO, we should have a generic resource called operations, initially as follow:
```
{ type: REFRESH, created: 1970-01-01, user: 'super_user'}
```

The resource `sceneexception` should have operations as a sub-resource:
`sceneexception/operation`

Initially, we could support only creation:
```
PUT api/v2/sceneexception/operation
{ type: REFRESH}

HTTP RESPONSE
HTTP 201 CREATED
```

Later, if we decide so, we could persist these operations and search and list them using `GET /api/v2/sceneexception/operation`


This very same sub-resource could be used in other entities/resources as `application/operation` (to restart, shutdown, upgrade, etc), or a TVShow to update with indexer data:
```
updating the full series
PUT series/tvdb1234/operation
{ type: INDEXER_REFRESH}
```

```
updating only season 9
PUT series/tvdb1234/S09/operation
{ type: INDEXER_REFRESH}
```

In summary:

For this endpoint I'll go with:
```
PUT api/v2/sceneexception/operation
{ type: REFRESH}

HTTP RESPONSE
HTTP 201 CREATED
```

p0psicles:
@ratoaq2 i'd like to get the last_update for every scene exception refresh. Shoul I requests that like 
```GET api/v2/sceneexception/lastupdate``` ?
As it's a different table the the scene_exceptions them self.

**Comment by ratoaq2** [comment]
We never did like that. For a specific scene exception you could:
`GET api/v2/sceneexception/{scene_exception_id_here}/lastupdate`

But for all scene exceptions, we usually use detailed=True/False
`GET api/v2/sceneexception/?detailed=False`
where it returns a detailed or limited information for each item.

I would recommend you to use `detailed=False`, so it would return all sceneexceptions but with fewer information, and of course the lastupdate field should be included

**Comment by p0psicles** [comment]
I've used the routes:
```GET api/v2/sceneexception/?detailed=False``` To get the last updates for all.
And
```PUT api/v2/sceneexception/operation
{ type: REFRESH}

HTTP RESPONSE
HTTP 201 CREATED
```
To force the operation.

p0psicles:
Used some code from #1787 for the sceneexception route.

fernandog:
tested and no issues. 
wondering if we can add time in the UI for "last updated"

p0psicles:
Yeah, we could. but you have to ask yourself, would that matter? It's not that these scene exceptions are updated every hour.

fernandog:
@p0psicles yes, it was just a confirm that current time = last update time. No biggy

medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,160 @@
...
+
+class SceneExceptionOperationHandler(BaseRequestHandler):
+    """Scene Exception request handler."""
+
+    def set_default_headers(self):
+        """Set default CORS headers."""
+        super(SceneExceptionOperationHandler, self).set_default_headers()
+        self.set_header('Access-Control-Allow-Methods', 'PUT')
+
+    def put(self):
```
OmgImAlexis: Should this not be POST to be similar to how @p0psicles @ratoaq2 and I discussed?

POST is for making new resources where as PUT is for replacing a resource.

medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,160 @@
...
+
+        exceptions = [{'exception_id': row[0],
+                       'indexer': row[1],
+                       'indexer_id': row[2],
+                       'show_name': row[3],
+                       'season': row[4],
+                       'custom': row[5]}
+                      for row in exceptions]
+
+        self.api_finish(data={"exceptions": exceptions, "last_updates": get_last_updates()})
```
OmgImAlexis: All JSON returned should be in camelCase.

medusa/server/api/v2/scene_exception.py
p0psicles: I changed to put, on rats examples.

OmgImAlexis: @ratoaq2 any reason for this?

medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,160 @@
...
+        :param season:
+        """
+        exception_id = self._parse(kwargs.pop('row_id'))
+        indexer = self.get_query_argument('indexer', None)
+        indexer_id = self.get_query_argument('indexer_id', None)
+        season = self.get_query_argument('season', None)
+        detailed = self.get_query_argument('detailed', 'true').lower()
+
+        if detailed == 'false':
+            return self.api_finish(data={"last_updates": get_last_updates()})
```
ratoaq2: GET sceneexception should always return sceneexception. Detailed will return with all info. Non detailed should return with just `header` info

medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def get(self, *args, **kwargs):
+        """Query scene_exception information.
+
+        :param show_indexer_id: The indexer id. Like 1 for tmdb and 3 for tvmaze.
+        :param show_id:
+        :type show_indexer_id: str
+        :param season:
+        """
+        exception_id = self._parse(kwargs.pop('row_id'))
```
ratoaq2: When your requesting a resource by its id, you should just return it. Do not return a singleton list, but just the resource itself

medusa/server/api/v2/scene_exception.py
ratoaq2: And it should be a simple list of sceneexception. Here it's a combination of sceneexception and timestamps

ratoaq2: Do you need to have timestamp per indexer? otherwise you could just use last modified http header

medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,160 @@
...
+                            b', season = ?'
+                            b', custom = 1'
+                            b' WHERE exception_id = ?',
+                            [data.get('indexer'),
+                             data.get('indexer_id'),
+                             data.get('show_name'),
+                             data.get('season'),
+                             exception_id])
+        if cache_db_con.connection.total_changes - last_changes == 1:
+            return self.api_finish(status=201)
```
ratoaq2: Http 201 created. In this case it was updated. either you return 200 with the exception itself in the http body, or just (preferably) 204 http no content

medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,160 @@
...
+        last_changes = cache_db_con.connection.total_changes
+        cache_db_con.action(b'INSERT INTO scene_exceptions'
+                            b' (indexer, indexer_id, show_name, season, custom) '
+                            b' values (?,?,?,?,1)',
+                            [data.get('indexer'),
+                             data.get('indexer_id'),
+                             data.get('show_name'),
+                             data.get('season')])
+        if cache_db_con.connection.total_changes - last_changes > 0:
+            return self.api_finish(status=204)
```
ratoaq2: You're creating a resource. You should return it back in the response since new fields could have been populated (auto generated ids, etc). 
Either you return 201 created with http header with the url location for the resource or http 200 with the resource itself in the response body

medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,160 @@
...
+                            b', custom = 1'
+                            b' WHERE exception_id = ?',
+                            [data.get('indexer'),
+                             data.get('indexer_id'),
+                             data.get('show_name'),
+                             data.get('season'),
+                             exception_id])
+        if cache_db_con.connection.total_changes - last_changes == 1:
+            return self.api_finish(status=201)
+        return self.api_finish(status=400, error="Could not update.")
```
ratoaq2: If resource is not found, 404 should be returned

medusa/server/api/v2/scene_exception.py
```diff
@@ -0,0 +1,160 @@
...
+        :type exception_id: str
+        """
+        exception_id = self._parse(kwargs.pop('row_id'))
+
+        cache_db_con = db.DBConnection('cache.db')
+        last_changes = cache_db_con.connection.total_changes
+        cache_db_con.action(b'DELETE FROM scene_exceptions WHERE exception_id = ?', [exception_id])
+        if cache_db_con.connection.total_changes - last_changes > 0:
+            return self.api_finish(status=204)
+        return self.api_finish(status=400, error="Failed to delete.")
```
ratoaq2: If resource is not found, 404 should be returned

medusa/server/api/v2/scene_exception.py
ratoaq2: And if the resource is not found, 404 should be returned

medusa/server/core.py
```diff
@@ -23,22 +23,40 @@ def get_apiv2_handlers(base):
...
 
     return [
-        (r'{base}/show(?:/{show_id}(?:/{ep_id})?(?:/{query})?)?/?'.format(base=base, show_id=show_id, ep_id=ep_id, query=query), ShowHandler),
-        (r'{base}/config(?:/{query})?/?'.format(base=base, query=query), ConfigHandler),
-        (r'{base}/log(?:/{log_level})?/?'.format(base=base, log_level=log_level), LogHandler),
-        (r'{base}/authenticate(/?)'.format(base=base), AuthHandler),
-        (r'{base}/asset(?:/{asset_group})(?:/{query})?/?'.format(base=base, asset_group=asset_group, query=query_extended), AssetHandler),
-        (r'{base}(/?.*)'.format(base=base), NotFoundHandler)
+        # All operations endpoints should be defined first.
+        (r'{base}/sceneexception/operation'.format(base=base),
```
ratoaq2: I need to think about this and I'll come back to you. Operations on a single resource or in a resource list. This was it's on a resource list

medusa/server/api/v2/scene_exception.py
p0psicles: Ok, so when I need the last_updates. I pass with detailed. And I return both?

p0psicles: what do you mean with: `Non detailed should return with just header info`

medusa/server/api/v2/scene_exception.py
p0psicles: Then i'd prefer to return the resource with a 200.

medusa/server/api/v2/scene_exception.py
p0psicles: I have the lastupdates stored per source. Also when one source is down for ex. The user can see this, because two have latest date, and one is lagging 3 days behind. So maybe that's a resource in itself? Like `sceneexceptionupdate/xem` `sceneexceptionupdate/custom` `sceneexceptionupdate/anidb`?
But where do you put the operation?

fernandog:
Implements https://github.com/pymedusa/Medusa/issues/410

medusa/server/api/v2/scene_exception.py
labrys: JSON has no requirement that keys be camelCased.  Since our API keys directly reference a Python back-end they should remain snake_cased.

medusa/server/api/v2/scene_exception.py
labrys: Actually you can use PUT or POST to create resources.  PUT **MUST** be idempotent, POST is **NOT** idempotent.

medusa/server/api/v2/scene_exception.py
labrys: If it is always expected to be a single item, then yes it should return a single item and not a list containing an item.  If a list of items is expected, it should always return a list, even if it only contains a single item, (or no items).

medusa/server/api/v2/scene_exception.py
OmgImAlexis: @labrys PUT should not be used as the main way of adding new resources.

medusa/server/api/v2/scene_exception.py
OmgImAlexis: camelCase is what we have as our standard over all of the API. Please use it.

medusa/server/api/v2/scene_exception.py
labrys: http://dret.typepad.com/dretblog/2011/11/creating-resources-with-get-put.html
"creating resources the RESTful way can be done with either PUT or POST, with PUT expecting that the URI of the created resource is known in advance (by the client issuing the request)"  

http://restcookbook.com/HTTP%20Methods/put-vs-post/
"The HTTP methods POST and PUT aren't the HTTP equivalent of the CRUD's create and update. They both serve a different purpose. It's quite possible, valid and even preferred in some occasions, to use PUT to create resources, or use POST to update resources."

https://jcalcote.wordpress.com/2008/10/16/put-or-post-the-rest-of-the-story/
"Create can be implemented using an HTTP PUT, if (and only if) the payload of the request contains the full content of the exactly specified URL."

labrys: https://tools.ietf.org/html/rfc2616#section-9.6
> The PUT method requests that the enclosed entity be stored under the
>    supplied Request-URI. If the Request-URI refers to an already
>    existing resource, the enclosed entity SHOULD be considered as a
>    modified version of the one residing on the origin server. If the
>    Request-URI does not point to an existing resource, and that URI is
>    capable of being defined as a new resource by the requesting user
>    agent, the origin server can create the resource with that URI. If a
>    new resource is created, the origin server MUST inform the user agent
>    via the 201 (Created) response. If an existing resource is modified,
>    either the 200 (OK) or 204 (No Content) response codes SHOULD be sent
>    to indicate successful completion of the request.

medusa/server/api/v2/scene_exception.py
```diff
@@ -16,20 +16,67 @@ def get_last_updates():
...
+class SceneExceptionTypeHandler(BaseRequestHandler):
+    """Scene Exception type request handler."""
+
+    def set_default_headers(self):
+        """Set default CORS headers."""
+        super(SceneExceptionTypeHandler, self).set_default_headers()
+        self.set_header('Access-Control-Allow-Methods', 'GET')
+
+    def get(self, exception_type):
+        if exception_type is not None and exception_type not in ['medusa', 'xem', 'anidb']:
```
ratoaq2: I would go with `custom` instead of using the application name (medusa)

labrys: While we're talking about naming exceptions, instead of `SceneExceptions` what do you think about calling them `NamingExceptions`?  They don't only apply to scene, and they aren't exceptions to Scene, so it might be a little more appropriate.

medusa/server/api/v2/scene_exception.py
```diff
@@ -16,20 +16,67 @@ def get_last_updates():
...
+        super(SceneExceptionAllTypeOperationHandler, self).set_default_headers()
+        self.set_header('Access-Control-Allow-Methods', 'POST')
 
     def post(self):
-        """Start fetch retrieving scene name exceptions."""
+        """Start scene exceptions type operation. To refresh all scene exception types."""
+        json_body = json_decode(self.request.body)
+
+        if json_body.get('type', '') == 'REFRESH':
+            retrieve_exceptions(force=True, exception_type=None)
```
OmgImAlexis: This starts an update, right? Should this not return something like in progress?

@ratoaq2 not sure what return we'd use for something that isn't "complete".

medusa/server/api/v2/scene_exception.py
p0psicles: I'd like to reserve a type for manual added exceptions. Thought I could use custom for that. Then use Medusa for ours.

p0psicles: I think changing to naming exception is fine. But then should be changed in a big bang. Db, api, code, comments, wiki. So don't think this Pr is really a good location to discuss that.

medusa/server/api/v2/scene_exception.py
ratoaq2: Is it asynchronous? If so, 202 ACCEPTED should be used. Everything else looks good.

p0psicles: It's a synchronous backend process.

static/js/manage/manage-searches.js
```diff
@@ -2,11 +2,19 @@ MEDUSA.manage.manageSearches = function() {
     // Get total number current scene exceptions per source. Will request medusa, xem and anidb name exceptions.
     var updateExceptionTable = function(exceptions) {
         var status = $('#sceneExceptionStatus');
-        var medusaException = _.find(exceptions.data, function(obj) { return obj.id === 'medusa' })
+        var medusaException = _.find(exceptions.data, function(obj) {
```
OmgImAlexis: Why not use the built in js filter function?

```
exceptions.data.filter(function(obj) {
    return obj.id === 'medusa';
});
```

OmgImAlexis:
Since this was swallowed as an outdated commit.


Why not use the built in js filter function instead of the `_.find()`?
```
exceptions.data.filter(function(obj) {
    return obj.id === 'medusa';
});
```

p0psicles:
@labrys can you update your review? Lot's has changed and discussed since then.

medusa/server/api/v2/scene_exception.py
p0psicles: Put in a spinner to get some visual feedback.

p0psicles:
If all good I'll merge this.

PR #2290, opened 2017-02-24
Fix viewlog objects not working and improve mobile view

fernandog:
Fix inverted order in mobile when pulling right
and double row in my desktop

No one noticed this filter logs since 19 Dec 2016 ?! lol

image

image


fernandog:
@Thraxis can you review this please?

Issue #2281, opened 2017-02-23
Small div in backlogOverview


fernandog:
buttons are using two rows:

![image](https://cloud.githubusercontent.com/assets/2620870/23269063/23f7d842-f9ce-11e6-9d39-b9c2d271dd7a.png)

PR #2294, opened 2017-02-25
Fix button and keys in two rows

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2281

@thraxis

image


fernandog:
@duramato @OmgImAlexis worth adding another filter for status: All | Wanted | Quality ?

Issue #2214, opened 2017-02-18
Adding show from recommend shows without root directory configured...


labrys:
... causes warning in logs but still shows ADDED 

### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:** Develop
**OS:** Win 10
**What you did:**  On fresh install with no root directory configured yet, go to Add From Recommended and add a show
**What happened:**  Series shows as added but show does not get added to the show list.
**What you expected:**  Prompt to configure a root directory before allowing show selection or at least a visible warning that show failed to add.  Currently only warning is from warning pill which does not update until after a new page load.
**Logs:**
```
NOT APPLICABLE
```

PR #2296, opened 2017-02-25
Don't let add show when no default root folder

fernandog:
Also fix notification message

Fixes https://github.com/pymedusa/Medusa/issues/2214

### DIFF
--- a/medusa/server/web/home/add_shows.py
+++ b/medusa/server/web/home/add_shows.py
@@ -462,6 +462,7 @@ class HomeAddShows(Home):
         if not location:
             logger.log(u'There was an error creating the show, '
                        u'no root directory setting found', logger.WARNING)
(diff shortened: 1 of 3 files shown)


Thraxis:
Why are you blocking access to those sections.  The Add New Show and Add existing show pages both already require you to set a show path before you can submit to add the new show.

The add recommended pages have a section that could be used to set path, what is needed is the trigger to disable / enable the buttons on the recommended page when a defaut root dir is not available

fernandog:
No default root folder set as issue says

fernandog:
Anyway its an idea. Ofc can change

Thraxis:
See https://github.com/pymedusa/Medusa/pull/2305

fernandog:
@Thraxis closing this

PR #2300, opened 2017-02-26
Updating guessit and removing unneeded customizations

ratoaq2:
- Updating to lastest development guessit version and removed all customizations that are not needed anymore.


fernandog:
approved

PR #2303, opened 2017-02-26
will fix multiple dirs removed when delete pressed for RootDirs

Thraxis:
root-dirs.js was included both in main.mako and in addShows_newShow.make and addShows_trendingShows.mako causing the delete event to fire multiple times.

### DIFF
--- a/static/js/root-dirs.js
+++ b/static/js/root-dirs.js
@@ -99,6 +99,7 @@ $(document).ready(function() {
         $('#rootDirText').change();
         console.log('rootDirText: ' + $('#rootDirText').val());
     }
+
     function addRootDir(path) {
         if (path.length === 0) {
             return;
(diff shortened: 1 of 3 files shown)


fernandog:
@Thraxis if you remove the JS, user won't be able to add/remove new root while adding new show right?


image

fernandog:
i reproduced the issue. It only removes from the select in the UI. not from config.

fernandog:
this behavior only happens while adding shows.
I deleted a root folder in config/general and it only deleted the one I selected

fernandog:
seems `$('#deleteRootDir').on('click', function() {` is triggered twice

Thraxis:
Well if you read the PR you would see that root-dirs.js is included in main.mako which makes it globally available.  It is also included in the two files I removed it from.  Thus it would make the call twice.  removing the two un-needed includes fixes the issue.

fernandog:
Ok got it. Sorry

PR #2306, opened 2017-02-27
Create Episode object with status

fernandog:
Will create the Episode object with current status and not UNKNOWN, so when we update episode with indexer info, status is correct and it won't cause issues in the search threads

I have a theory that while show is updating, it has status UNKNOWN and everything can happen while in this status if a search thread is running and find a release for that ep

ping @pymedusa/developers 
what do you think?


fernandog:
i couldn't reproduce the status UNKNOWN

Issue #2310, opened 2017-02-27
Error updating: KeyError: 'seasonnumber


fernandog:
2017-02-27 03:01:16 ERROR    SHOWUPDATER :: [ecfd7b7] Problem running show_updater, Indexer TVDBv2 seems to be having issues while trying to get updates for show Fear the Walking Dead. Cause: 'seasonnumber'
Traceback (most recent call last):
  File "/home/osmc/Medusa/medusa/show_updater.py", line 130, in run
    updated_seasons = t.get_last_updated_seasons([show.indexerid], last_update, update_max_weeks)
  File "/home/osmc/Medusa/medusa/indexers/tvdbv2/tvdbv2_api.py", line 596, in get_last_updated_seasons
    total_updates.append(int(episode['seasonnumber']))
KeyError: 'seasonnumber'

fernandog:
@p0psicles will send full logs soon. Any idea?

ratoaq2:
I had the same issue. I wasn't able to submit the logs

fernandog:
im trying to fix the submit issue now

p0psicles:
Same error with just that show?

fernandog:
https://gist.github.com/fernandog/0100931e4f0683add75765959a25b353

result from: https://api.thetvdb.com/series/290853/episodes'`

no `seasonnumber`there but there is a `airedSeason`

ratoaq2:
Yeah, same error, same show

fernandog:
We found out that show has some ep with airedSeason = null.

We will ignore those type of episodes (incomplete) in show update

PR #2321, opened 2017-02-28
Added an additional check for updating a show through the showUpdater…

p0psicles:
…, for shows that have episodes with missing season or episode number.

Fixes https://github.com/pymedusa/Medusa/issues/2310

Apparently this can happen for at least tvdb, when users do not properly add show information.

* Made the warning a little bit more informative. Through providing the user with some advice, on how to resolve the warning.

### DIFF
--- a/medusa/indexers/tvdbv2/tvdbv2_api.py
+++ b/medusa/indexers/tvdbv2/tvdbv2_api.py
@@ -300,7 +300,11 @@ class TVDBv2(BaseIndexer):
                 seasnum, epno = cur_ep.get('seasonnumber'), cur_ep.get('episodenumber')
 
             if seasnum is None or epno is None:
-                logger.warning('An episode has incomplete season/episode number (season: %r, episode: %r)', seasnum, epno)
+                logger.warning('This episode has incomplete information. The season or episode number '
+                               '(season: %r, episode: %r) is missing. '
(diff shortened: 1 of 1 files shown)


fernandog:
@p0psicles i don't think warning is the way to go. Those ep don't even appear in the tvdb website, so there is nothing user can do about

i would just ignore them

p0psicles:
Yes it's related to that. But I do think we should provide the user with some info if they really want to resolve it for the show. Because it's still a fact that it's an issue with tvdb show info.

fernandog:
@p0psicles my point it: they doesn't appear in http://thetvdb.com/?tab=seasonall&id=290853&lid=7

so how user will report it?

p0psicles:
I don't know. That's their problem, if they want to resolve it. Maybe contact someone on the forums? Send them a mail. I honestly don't care. We just inform the user that this is not something we can fix for them. But that's it's data corruption on tvdb's end.

fernandog:
i still couldn't reproduce the issue with other shows.

p0psicles:
Thats great right?  Would be troublesome if you could reproduce this with multiple shows. We just warn for these edge cases.

p0psicles:
We don't want this Pr?

fernandog:
@p0psicles sure we want. otherwise we will get traceback and show won't update

why are you asking ?

PR #2323, opened 2017-02-28
2) Make one snatch per show-season-episode

fernandog:
Fixes duplicates snatches for same show-season-episode
It's the best solution I could find without breaking other stuff

Also we also do the same with PROPERs (only snatch a proper if current ep status is DOWNLOADED)

The only disavantages is that preferred snatch will take a little more time.
BUT as daily searcher is 30min, in most cases, Medusa will PP the allowed quality, and in next daily, the preferred will be snatched.

Example: 10Mbps connection (1.25 mb/s), will transfer in 30min(1800s), 2.25 GB.
Which is necessary to download most qualities.

Worst cases, it will snatch in the second dailysearcher

@pymedusa/developers i won't try to fix this issue in another way.

### DIFF
--- a/medusa/common.py
+++ b/medusa/common.py
@@ -613,7 +613,7 @@ class Quality(object):
 
         #  Can't be SNATCHED_BEST because the quality is already final (unless user changes qualities).
         #  All other status will return false: IGNORED, SKIPPED, FAILED.
-        if cur_status not in (WANTED, DOWNLOADED, SNATCHED, SNATCHED_PROPER, SNATCHED_BEST):
+        if cur_status not in (WANTED, DOWNLOADED):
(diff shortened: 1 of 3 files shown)


labrys:
I've already given my opinion on this in slack

fernandog:
@labrys you said it will take longer to snatch and i wrote in the PR it will take only ONE daily searcher to snatch preferred.  

So what is the another disadvantage in your opinion ?

OmgImAlexis:
@fernandog this makes Medusa a LOT slower to get better quality files. I've voting to close this as this isn't the right fix.

fernandog:
A lot slower how exactly?

Issue #2326, opened 2017-03-02
Episode information shows as undefined


medariox:
![image](https://snapr.pw/i/9e0d87acf5.png)

fernandog:
@OmgImAlexis 

![image](https://cloud.githubusercontent.com/assets/2620870/23508304/cbdaabd2-ff2f-11e6-84fb-fc2d70796b45.png)

```
Traceback (most recent call last):
  File "/home/osmc/Medusa/lib/tornado/web.py", line 1443, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/home/osmc/Medusa/medusa/server/api/v2/show.py", line 68, in get
    return self._handle_episode(tv_show, ep_id, query)
  File "/home/osmc/Medusa/medusa/server/api/v2/show.py", line 110, in _handle_episode
    return self._handle_detailed_episode(tv_episode, query)
  File "/home/osmc/Medusa/medusa/server/api/v2/show.py", line 140, in _handle_detailed_episode
    data = tv_episode.to_json()
  File "/home/osmc/Medusa/medusa/tv/episode.py", line 714, in to_json
    ).isoformat('T')
TypeError: isoformat() argument 1 must be char, not unicode
```

fernandog:
btw that episode status were UNAIRED. no airdate

fernandog:
also this should show the error but its not:
plot-tooltip.js
```javascript
                    }, function(xhr, status, error) {
                        // Upon failure... set the tooltip content to the status and error value
                        qt.set('content.text', status + ': ' + error);
```

fernandog:
@ratoaq2 can we log by default only the 4xx and 5xx status from TORNADO?

ratoaq2:
Tornado might use different log levels for successful calls and these ones. If so, yes we can.

fernandog:
@ratoaq2 

```
2017-03-02 13:07:10 ERROR    TORNADO :: [51ff22a] Uncaught exception GET /api/v2/show/tvdb269578/s4e9/description (127.0.0.1)
HTTPServerRequest(protocol='http', host='localhost:8081', method='GET', uri='/api/v2/show/tvdb269578/s4e9/description',.....)
Traceback (most recent call last):
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\web.py", line 1443, in _execute
    result = method(*self.path_args, **self.path_kwargs)
[...]

PR #2327, opened 2017-03-02
Fixes TypeError: isoformat() argument 1 must be char, not unicode

fernandog:
fixes https://github.com/pymedusa/Medusa/issues/2326

TODO:

- [x] Fix JS status/error as undefined
- [ ] Double check if anything else needs `byte`  

image


fernandog:
@medariox @OmgImAlexis need review & approve please

PR #2328, opened 2017-03-02
Don't delete extracted media file if it was postponed

fernandog:
When we call process_media it can postpone the PP of a media file because it doesn't have subtitles
so after that, we delete all rar content files (nfo, srt) that wasn't listed as wanted "associated files"
This fixes the deletion of the media file when media was postponed

2017-03-02 01:32:04 DEBUG    FINDSUBTITLES :: [32b3c7d] No subtitles associated. Postponing the post-process of this file: Show.S06E16.720p.HDTV.x264-AVS.mkv
2017-03-02 01:32:04 DEBUG    FINDSUBTITLES :: [32b3c7d] Deleting file: Show.S06E16.720p.HDTV.x264-AVS.mkv


fernandog:
someone approve this please?

PR #2332, opened 2017-03-03
Log/style adapter

labrys:

This implements a `BraceMessage` for lazily formatting a brace-formatted string.  It also implements a `BraceAdapter` to adapt a logger to use brace-formatted strings lazily.

Example usage:
```python
import logging
from medusa.logger.adapter.style import BraceAdapter

log = logging.getLogger(__name__)
log.addHandler(logging.NullHandler())
log = BraceAdapter(log)

log.info('The {subject} {predicate}', {'subject': 'cake', 'predicate': 'is a lie'})
log.error('Response received: {}', "I'm sorry Dave, I'm afraid I can't do that.")
```

You can also create a `BraceMessage` in advance for later usage.  This is also helpful if you only want to use lazy braced messages with a standard logger without the `BraceAdapter`.
```python
msg = BraceMessage('I have {x} ideas. They all point to {outcome}.', x='a million`, outcome='certain death')
log.critical(msg)
```

This pattern can be used to modify a message for repeated use, with or without modification:
```python
msg = BraceMessage('{who} {does} {what}', who='Los locos', does='kick', what='your ass')
log.info(msg)
msg.kwargs.update({'what': 'your face'})
log.warning(msg)
msg.msg += ' {where}'
msg.kwargs.update({'what': 'your balls', 'where': 'INTO OUTER SPACE!'})
log.critical(msg)
```

### DIFF
--- a/medusa/browser.py
+++ b/medusa/browser.py
@@ -1,29 +1,14 @@
 # coding=utf-8
-# Author: Nic Wolfe <nic@wolfeden.ca>
-#
 
-#
-# This file is part of Medusa.
-#
-# Medusa is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# Medusa is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with Medusa. If not, see <http://www.gnu.org/licenses/>.
 """Browser module."""
 from __future__ import unicode_literals
 
+import logging
 import os
 import string
 
-from . import logger
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
 
 
 # adapted from http://stackoverflow.com/questions/827371/is-there-a-way-to-list-all-the-available-drive-letters-in-python/827490
@@ -112,7 +97,7 @@ def list_folders(path, include_parent=False, include_files=False):
     try:
         file_list = get_file_list(path, include_files)
     except OSError as e:
-        logger.log('Unable to open %s: %s / %s' % (path, repr(e), str(e)), logger.WARNING)
+        log.warning('Unable to open %s: %s / %s', path, repr(e), str(e))
         file_list = get_file_list(parent_path, include_files)
 
     file_list = sorted(file_list,
--- a/medusa/clients/nzb/nzbget.py
+++ b/medusa/clients/nzb/nzbget.py
@@ -1,64 +1,50 @@
 # coding=utf-8
-# Author: Nic Wolfe <nic@wolfeden.ca>
-
-#
-# This file is part of Medusa.
-#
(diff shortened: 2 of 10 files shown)


medusa/generic_queue.py
```diff
@@ -1,26 +1,10 @@
...
-#
-# You should have received a copy of the GNU General Public License
-# along with Medusa. If not, see <http://www.gnu.org/licenses/>.
 
+import logging
 import datetime
 import threading
 
-from . import logger
+log = logging.getLogger()
```
p0psicles: This rule is incorrect. Already tried disabling it. Really annoying

medusa/clients/nzb/nzbget.py
```diff
@@ -1,64 +1,50 @@
...
+from medusa.logger.adapters.style import BraceAdapter
 from medusa.helper.common import try_int
 
 from six.moves.http_client import socket
 from six.moves.xmlrpc_client import ProtocolError, ServerProxy
 
 
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+log = BraceAdapter(log)
```
ratoaq2: The plan is to have an utility method to return this log pre configured so we don't need to repeat this everywhere, right?

medusa/browser.py
```diff
@@ -112,7 +97,7 @@ def list_folders(path, include_parent=False, include_files=False):
     try:
         file_list = get_file_list(path, include_files)
     except OSError as e:
-        logger.log('Unable to open %s: %s / %s' % (path, repr(e), str(e)), logger.WARNING)
+        log.warning('Unable to open %s: %s / %s', (path, repr(e), str(e)))
```
p0psicles: Shouldn't this be rewritten to the default logging format? Like '%s %s %s', value 1, value2, value3

medusa/generic_queue.py
labrys: I know it is, I'm going to fix the regex for it, but haven't gotten to that yet.

medusa/clients/nzb/nzbget.py
labrys: @ratoaq2 I would likely implement it as a classmethod of `BraceAdapter` so you could do `BraceAdapter.getLogger(__name__)` but was more concerned about functionality before convenience methods.

medusa/clients/nzb/nzbget.py
```diff
@@ -80,15 +66,16 @@ def testNZB(host, username, password, use_https):
     return NZBConnection(url)
 
 
-def sendNZB(nzb, proper=False):  # pylint: disable=too-many-locals, too-many-statements, too-many-branches, too-many-return-statements
+def sendNZB(nzb, proper=False):
```
labrys: Handled in another PR

medusa/browser.py
labrys: Yes it should, just missed removing the extra parentheses.

medusa/clients/nzb/nzbget.py
labrys: Personally i would opt for the following style instead of a convenience method.  I wrote the other style so it was obvious what was being done.
```
log = BraceAdapter(logging.getLogger(__name__))
log.logger.addHandler(logging.NullHandler)
```
That's only slightly less convenient than the following and doesn't require any additional method
```
log = BraceAdapter.get_logger(__name__)
log.logger.addHandler(logging.NullHandler())
```

PR #2343, opened 2017-03-04
Log/helpers

labrys:

Merge after #2332


medusa/databases/__init__.py
```diff
@@ -1,20 +1,3 @@
 # coding=utf-8
```
p0psicles: I think that if we're removing the headers. We should at least add the module docstrings. Whay you guys think?

medusa/helper/common.py
```diff
@@ -310,7 +296,7 @@ def try_int(candidate, default_value=0):
         return int(candidate)
     except (ValueError, TypeError):
         if candidate and ("," in candidate or "." in candidate):
-            logger.error(u"Failed parsing provider. Traceback: %r" % traceback.format_exc())
+            log.error(u"Failed parsing provider. Traceback: %r" % traceback.format_exc())
```
p0psicles: Are we going to accept non-lazy logging? I think this should then at least be changed to: 
`log.error(u"Failed parsing provider. Traceback: %r", traceback.format_exc())`

medusa/databases/__init__.py
labrys: We need to be adding module docstrings regardless.  We also need to add a license file.

medusa/helper/common.py
labrys: You're right, this should be fixed.

medusa/helpers/externals.py
```diff
@@ -41,8 +27,9 @@ def trakt_request(api, url):
                 app.TRAKT_ACCESS_TOKEN = api.access_token
                 app.TRAKT_REFRESH_TOKEN = api.refresh_token
                 app.instance.save_config()
-        except (TokenExpiredException, TraktException) as e:
-            logger.info(u'Could not use Trakt to enrich with externals. Cause: {cause}', cause=e)
+        except (TokenExpiredException, TraktException) as error:
+            log.info(u'Could not use Trakt to enrich with externals: {0}',
+                     error)
```
fernandog: we should log only the e.message and not entire error object
At least its the "default" for most of all code

medusa/helpers/externals.py
```diff
@@ -104,14 +95,17 @@ def get_externals(show=None, indexer=None, indexed_show=None):
...
             try:
                 new_show_externals.update(t.get_id_by_external(**new_show_externals))
-            except IndexerException as e:
-                logger.warning(u'Error getting external ids for other indexer {indexer_name} with cause {cause}',
-                               indexer_name=indexerApi(show.indexer).name, cause=e.message)
+            except IndexerException as error:
+                log.warning(
+                    u'Error getting external ids for other'
+                    u' indexer {name}: {reason}',
+                    {'name': indexerApi(show.indexer).name, 'reason': error.message})
```
fernandog: like here

medusa/helpers/externals.py
labrys: I think consistency is important so I would generally agree, but imo it depends on the intent of the error logging.  My intent with the PR is not to interpret the original intent but to adapt it to the new style in order to finish the transition from the old logger.

Issue #2349, opened 2017-03-05
GIT alert in alerts.mako disappear after you change any config setting, save and reload page


fernandog:
Reproduce: add any wrong folder to git_path, save, restart. git warning will appear in the top of the screen
Go to general settings, do any change, save. reload page. warning disappears

No idea why this happens.

![image](https://cloud.githubusercontent.com/assets/2620870/23591987/d4559e24-01d8-11e7-9aca-95788efa3ac1.png)

PR #2350, opened 2017-03-05
fix disappearing git alert, allows git_path change without restart

Thraxis:
- fix git alert disappearing on config change #2349
- allows git_path change while application is running

### DIFF
--- a/medusa/config.py
+++ b/medusa/config.py
@@ -302,7 +302,9 @@ def change_GIT_PATH():
     Force a run to clear or set any error messages.
     """
     app.version_check_scheduler = None
-    app.version_check_scheduler = scheduler.Scheduler(CheckVersion(), cycleTime=datetime.timedelta(hours=app.UPDATE_FREQUENCY), threadName="CHECKVERSION", silent=False)
+    app.version_check_scheduler = scheduler.Scheduler(CheckVersion(),
+            cycleTime=datetime.timedelta(hours=app.UPDATE_FREQUENCY),
+            threadName="CHECKVERSION", silent=False)
     app.version_check_scheduler.enable = True
     app.version_check_scheduler.start()
     app.version_check_scheduler.forceRun()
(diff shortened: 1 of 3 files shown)


medusa/server/web/config/general.py
```diff
@@ -99,6 +99,7 @@ def saveGeneral(self, log_dir=None, log_nr=5, log_size=1, web_port=None, notify_
         app.GIT_RESET = config.checkbox_to_value(git_reset)
         app.GIT_RESET_BRANCHES = helpers.ensure_list(git_reset_branches)
         app.GIT_PATH = git_path
+        config.change_GIT_PATH()
```
fernandog: @thraxis I would add a IF here if git path changed to avoid checking updates every config save. What do you think?

fernandog:
Thanks for this PR but I didnt get why warning disappear. I didnt change git path while medusa was running. I already started without git path . Why it happens?

Thraxis:
without fixing the ability to change git path while medusa was running there was no way to properly fix the git alert issue it would either disappear when it shouldn't or not disappear when it should.

medusa/config.py
```diff
@@ -294,6 +296,19 @@ def change_VERSION_NOTIFY(version_notify):
         app.version_check_scheduler.forceRun()
 
 
+def change_GIT_PATH():
```
Thraxis: Yeah... not sure what the best solution for this would be as I followed the function naming pattern used by every other function in that file.

medusa/config.py
```diff
@@ -25,8 +25,8 @@
 from six import iteritems
 from six.moves.urllib.parse import urlunsplit, uses_netloc
 from . import app, common, db, helpers, logger, naming, scheduler
-from . version_checker import CheckVersion
 from .helper.common import try_int
+from . version_checker import CheckVersion
```
OmgImAlexis: Should this not be `.verson_checker`?

fernandog:
@Thraxis i found the reason why is not showing after config save.... may i commit ?

PR #2361, opened 2017-03-07
Assume file already PPed if errno 2 (File not found)

fernandog:
Fixes: https://github.com/pymedusa/Medusa/issues/2314

@duramato @OmgImAlexis @medariox

### DIFF
--- a/medusa/helpers/__init__.py
+++ b/medusa/helpers/__init__.py
@@ -416,7 +416,7 @@ def make_dirs(path):
             try:
                 logger.debug(u"Folder {path} didn't exist, creating it", path=path)
                 os.makedirs(path)
-            except (OSError, IOError) as e:
+            except (OSError, IOError, WindowsError) as e:
                 logger.error(u"Failed creating {path} : {error!r}", path=path, error=e)
                 return False
 
@@ -442,7 +442,7 @@ def make_dirs(path):
                     # do the library update for synoindex
                     from medusa import notifiers
                     notifiers.synoindex_notifier.addFolder(sofar)
-                except (OSError, IOError) as e:
+                except (OSError, IOError, WindowsError) as e:
                     logger.error(u'Failed creating {path} : {error!r}', path=sofar, error=e)
                     return False
 
--- a/medusa/helpers/__init__.py
+++ b/medusa/helpers/__init__.py
@@ -416,7 +416,7 @@ def make_dirs(path):
             try:
                 logger.debug(u"Folder {path} didn't exist, creating it", path=path)
                 os.makedirs(path)
-            except (OSError, IOError, WindowsError) as e:
+            except (OSError, IOError) as e:
                 logger.error(u"Failed creating {path} : {error!r}", path=path, error=e)
                 return False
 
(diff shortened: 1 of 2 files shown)


medariox:
We shouldn't assume. Post-processing also doesn't need a queue, it just shouldn't run more than once, that is what causes that kind of errors. This can very easily be achieved once the process_tv refactor is merged.

fernandog:
We need PP queue because multiples PP can run at same time. Causing Omg issue

What did you change in the code to prevent that?

duramato:
IMO it should be a file queue that gets implemented to prevent multi processes at the same time. Multiple PP runs are perfectly normal.

fernandog:
What I assume and @duramato agrees with me is that we used to have a file and it doesnt exit anymore its because it was already PPed by other thread

medariox:
Why are multiple PPs at the same time normal? Quite the contrary. If a PP is already running, another one shouldn't be started, multiple runs also make little sense as a PP run takes care of all the files that need to be processed. So there is no real reason to make a queue. You want your files processed ASAP and all of them. The only issue is if 2 or more run at the same time.

duramato:
I use nzbtomedia for torrents and usenet in an idependant manner. If both finish at the same time and start PP what do you propose it should do? To me it should pp one then the other.

medariox:
It should wait for the next chance to process that is already user configurable.

duramato:
Didn't understood? what is already configurable?

ratoaq2:
This is the queue with max size one. 

you get a pp request and start pp, in the meanwhile you get another pp request, and after that another one. After the first one finishes, you expect only one more run, for the files that were moved to the folder after first had already started. 

So, it's a queue with max size 1

medariox:
The auto PP is user configurable. So the question is if it really makes sense to add a max 1 queue (as rato said) when we already have a setting that can be set with frequent intervals.

duramato:
We aren't only talking about auto PP, there is Manual PP too (and 3rd party scripts who call it)

fernandog:
I think is pretty safe to assume that if file is not there anymore, it was already processed. 

Do you prefer RAISE with traceback and able to submit an error or a WARNING?

medusa/helpers/__init__.py
```diff
@@ -416,7 +416,7 @@ def make_dirs(path):
             try:
                 logger.debug(u"Folder {path} didn't exist, creating it", path=path)
                 os.makedirs(path)
-            except (OSError, IOError) as e:
+            except (OSError, IOError, WindowsError) as e:
                 logger.error(u"Failed creating {path} : {error!r}", path=path, error=e)
```
ratoaq2: This way we'll always log as error and it will be submitable by the user. That's the place that it should delegate to exception handler. All these catches should delegate to exception handler

labrys: `WindowsError` is a subclass of `OSError` and is thus not needed here.

fernandog:
@labrys @ratoaq2 reverted

anyway first commit will fix @labrys issue

OmgImAlexis:
Personally I think for now catching the error so we don't get tracebacks would be best as we know the issue to a point and don't need users getting tracebacks for something we are already looking into.

fernandog:
Ok. Develop already catches the error. Closing this one

PR #2364, opened 2017-03-08
 Show manual search results in API v2

fernandog:
@OmgImAlexis @ratoaq2 

Tested:

Manual episode search:
http://localhost:8081/api/v2/series/tvdb257655/episode/s05e17/releases?api_key=****

Manual season search:
http://localhost:8081/api/v2/series/tvdb257655/season/s05/releases?api_key=****

Screenshot of result:

image

### DIFF
--- a/medusa/server/api/v2/episode.py
+++ b/medusa/server/api/v2/episode.py
@@ -55,6 +55,8 @@ class EpisodeHandler(BaseRequestHandler):
         if path_param:
             if path_param == 'metadata':
                 data = episode.metadata() if episode.is_location_valid() else {}
+            elif path_param == 'releases':
+                data = episode.get_manual_search_results()
             elif path_param in data:
                 data = data[path_param]
             else:
--- a/medusa/server/api/v2/season.py
+++ b/medusa/server/api/v2/season.py
@@ -1,0 +1,57 @@
+# coding=utf-8
+"""Request handler for series and season."""
+
+from medusa.server.api.v2.base import BaseRequestHandler
+from medusa.server.api.v2.series import SeriesHandler
+from medusa.tv.episode import Episode, EpisodeNumber
+from medusa.tv.series import Series, SeriesIdentifier
+
+
+class SeasonHandler(BaseRequestHandler):
+    """Season request handler."""
+
+    #: parent resource handler
+    parent_handler = SeriesHandler
+    #: resource name
+    name = 'season'
+    #: identifier
+    identifier = ('season_slug', r'[\w-]+')
+    #: path param
+    path_param = ('path_param', r'\w+')
+    #: allowed HTTP methods
+    allowed_methods = ('GET', )
+
+    def get(self, series_slug, season_slug, path_param):
+        """Query season information.
+
+        :param series_slug: series slug. E.g.: tvdb1234
+        :param path_param:
+        """
+        series_identifier = SeriesIdentifier.from_slug(series_slug)
+        if not series_identifier:
+            return self._bad_request('Invalid series slug')
+
+        series = Series.find_by_identifier(series_identifier)
+        if not series:
+            return self._not_found('Series not found')
+
+        # By design, manual season search uses the episode 01 of season as base
+        episode_number = EpisodeNumber.from_slug(season_slug + 'e01')
+        if not episode_number:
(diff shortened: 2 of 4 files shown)


medusa/server/api/v2/show.py
```diff
@@ -86,6 +90,18 @@ def _handle_detailed_show(self, tv_show, query):
                 new, existing = tv_show.get_backlogged_episodes(allowed_qualities=allowed_qualities,
                                                                 preferred_qualities=preferred_qualities)
                 data = {'new': new, 'existing': existing}
+            elif query == 'manualSearch':
+                season = self._parse(self.get_argument('season', default=None), int)
```
ratoaq2: I know this is just an experiment, but manual search is not a show property or sub resource. We should use normal rest query capabilities. 
Probably `/show/tvdb1234/s03/releases?a=b` when searching releases for a given season, and `/show/tvdb1234/s01e02/releases?c=d` when searching releases for a given episode

fernandog: Ok!

fernandog: @ratoaq2 what should I do then?

fernandog:
@ratoaq2 @OmgImAlexis where should this be added?

OmgImAlexis:
This isn't right at all...? @ratoaq2

p0psicles:
Your right OMG. But you can't blame fernandog. It's not easy to implement this restful. And explaining it costs allot of time. I'm also still learning so I can also not coach this Pr.

medusa/tv/episode.py
```diff
@@ -974,6 +975,15 @@ def delete_episode(self):
 
         raise EpisodeDeletedException()
 
+    def get_manual_search_results(self, manual_search_type='episode', show_all_results=0, perform_search=0):
```
OmgImAlexis: This should be returning all releases for an episode or season. This should have no coupling to manual search as it should also return cached results from other types of searches like daily, etc.

medusa/server/core.py
```diff
@@ -36,6 +40,9 @@ def get_apiv2_handlers(base):
         # /api/v2/series/tvdb1234/episode
         EpisodeHandler.create_app_handler(base),
 
+        # /api/v2/series/tvdb1234/season
+        SeasonHandler.create_app_handler(base),
```
OmgImAlexis: This should have been added separately, @ratoaq2 any reason why this wasn't added before now? Didn't we support seasons before now?

medusa/server/api/v2/season.py
```diff
@@ -0,0 +1,57 @@
...
+        series_identifier = SeriesIdentifier.from_slug(series_slug)
+        if not series_identifier:
+            return self._bad_request('Invalid series slug')
+
+        series = Series.find_by_identifier(series_identifier)
+        if not series:
+            return self._not_found('Series not found')
+
+        # By design, manual season search uses the episode 01 of season as base
+        episode_number = EpisodeNumber.from_slug(season_slug + 'e01')
```
OmgImAlexis: Just because this was "by design" once before doesn't mean it needs to be like this now.

medusa/server/api/v2/episode.py
```diff
@@ -55,6 +55,8 @@ def get(self, series_slug, episode_slug, path_param):
         if path_param:
             if path_param == 'metadata':
```
OmgImAlexis: Should the releases not be sent back with the metadata? Since it's coming from the cache I can't see any performance downside to this.

medusa/server/api/v2/season.py
fernandog: everything in medusa is "episode" based. can't change that unless start code medusa from zero.

medusa/tv/episode.py
fernandog: that code returns all that. but by design manual search was designed to be episode search or season search. As both does different stuff as the cached result from episode is different from season result

OmgImAlexis: Why not fix that here as well then?

medusa/server/api/v2/season.py
OmgImAlexis: It can be changed just like it was added in the first place.

fernandog:
will continue locally. thanks for all the feedback

PR #2367, opened 2017-03-08
Improve match of snatched NZB items in snatchSelection using size

fernandog:
Fixes the issue where N releases names from same provider (with same release name) were marked as SNATCHED but user had only one SNATCH

@duramato as we talked

TODO:

- [x] Uncomment minor version increment

image

### DIFF
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -21,7 +21,7 @@ MIN_DB_VERSION = 40  # oldest db version we support migrating from
 MAX_DB_VERSION = 44
 
 # Used to check when checking for updates
-CURRENT_MINOR_DB_VERSION = 5
+CURRENT_MINOR_DB_VERSION = 6
 
 
 class MainSanityCheck(db.DBSanityCheck):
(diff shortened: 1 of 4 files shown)


fernandog:
@duramato can you test it too?  it will need size now to "match"

fernandog:
@duramato still waiting

medusa/history.py
```diff
@@ -26,7 +26,7 @@
 
 
 def _logHistoryItem(action, showid, season, episode, quality, resource,
-                    provider, version=-1, proper_tags='', manually_searched=False, info_hash=None, resource_size=None):
+                    provider, version=-1, proper_tags='', manually_searched=False, info_hash=None, resource_size=-1):
```
labrys: Default should be `None` any for any values that are not determined.

medusa/server/web/home/handler.py
```diff
@@ -1259,6 +1262,7 @@ def titler(x):
                 i['status'], i['quality'] = Quality.split_composite_status(i['action'])
                 i['action_date'] = sbdatetime.sbfdatetime(datetime.strptime(str(i['date']), History.date_format), show_seconds=True)
                 i['resource_file'] = os.path.basename(i['resource'])
+                i['resource_size'] = pretty_file_size(i['resource_size']) if i['resource_size'] > -1 else 'N/A'
```
labrys: `i['resource_size'] = pretty_file_size(i['resource_size']) if i['resource_size'] else 'N/A'`

medusa/databases/main_db.py
```diff
@@ -559,3 +559,22 @@ def execute(self):
...
+        """
+        Test if the version is at least 44.6
+        """
+        return self.connection.version >= (44, 6)
+
+    def execute(self):
+        backupDatabase(self.connection.version)
+
+        log.info(u"Adding column resource_size in history")
+        if not self.hasColumn("history", "resource_size"):
```
labrys: Column name should be simply `size` for consistency with other columns... for example we have `provider` not `resource_provider` and `date` not `resource_date`.

medusa/history.py
fernandog: size in medusa db and code is default to -1

medusa/databases/main_db.py
duramato: imo anything only for a release should be `resource_*`, to be descriptive for anyone checking the db

Issue #2369, opened 2017-03-08
Ugly Mass update notification when updating shows


fernandog:
Maybe don't need to show all that info

![image](https://cloud.githubusercontent.com/assets/2620870/23717745/cf67af1e-0413-11e7-84a4-a0873a54011a.png)

OmgImAlexis:
It'd be better to either not show it, split it up or provide a summary notification.

For example "Updating 100 shows..."

PR #2376, opened 2017-03-09
Fix ugly Mass update notification

fernandog:
@OmgImAlexis

Fixed https://github.com/pymedusa/Medusa/issues/2369

image


OmgImAlexis:
@p0psicles I wonder how we should be sending this as it'd be nice for the websocket to be able to get a JSON object instead of just a string like this.

Issue #1980, opened 2017-01-16
[APP SUBMITTED]: gaierror: [Errno -2] Name or service not known


duramato:
### INFO
**Python Version**: `2.7.9 (default, Sep 17 2016, 20:26:04) [GCC 4.9.2]`
**Operating System**: `Linux-4.4.38-v7+-armv7l-with-debian-8.0`
**Locale**: `UTF-8`
**Branch**: [develop](../tree/develop)
**Database**: `44.4`
**Commit**: pymedusa/Medusa@49af1f0a36e3072f0ba468428f07d6c6b72e73fe
**Link to Log**: https://gist.github.com/14455435c7b93d06b4311aeee6229b1b
### ERROR
<pre>
2017-01-16 09:19:26 ERROR    FINDSUBTITLES :: [49af1f0] Unable to find subtitles for A Series of Unfortunate Events S01E03. Error: error -2 : Name or service not known
Traceback (most recent call last):
  File "/home/pi/Medusa/<a href="../blob/49af1f0a36e3072f0ba468428f07d6c6b72e73fe/medusa/subtitles.py#L1031">medusa/subtitles.py</a>", line 1031, in run
    episode_object.download_subtitles()
  File "/home/pi/Medusa/<a href="../blob/49af1f0a36e3072f0ba468428f07d6c6b72e73fe/medusa/tv.py#L2141">medusa/tv.py</a>", line 2141, in download_subtitles
    notifiers.notify_subtitle_download(self.pretty_name(), subtitle_list)
  File "/home/pi/Medusa/<a href="../blob/49af1f0a36e3072f0ba468428f07d6c6b72e73fe/medusa/notifiers/__init__.py#L82">medusa/notifiers/__init__.py</a>", line 82, in notify_subtitle_download
    n.notify_subtitle_download(ep_name, lang)
  File "/home/pi/Medusa/<a href="../blob/49af1f0a36e3072f0ba468428f07d6c6b72e73fe/medusa/notifiers/nma.py#L25">medusa/notifiers/nma.py</a>", line 25, in notify_subtitle_download
    message=ep_name + ": " + lang)
  File "/home/pi/Medusa/<a href="../blob/49af1f0a36e3072f0ba468428f07d6c6b72e73fe/medusa/notifiers/nma.py#L62">medusa/notifiers/nma.py</a>", line 62, in _sendNMA
    response = p.push(application=title, event=event, description=message, priority=nma_priority, batch_mode=batch)
[...]

PR #2377, opened 2017-03-10
Catch requests error while send subtitle notifications

fernandog:
@duramato @ratoaq2

Fixes https://github.com/pymedusa/Medusa/issues/1980

### DIFF
--- a/medusa/tv/episode.py
+++ b/medusa/tv/episode.py
@@ -89,6 +89,8 @@ from medusa.scene_numbering import (
 )
 from medusa.tv.base import TV
 
+from requests.exceptions import RequestException
+
 import shutil_custom
 
 try:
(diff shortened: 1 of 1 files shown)


medusa/tv/episode.py
```diff
@@ -269,7 +271,10 @@ def download_subtitles(self, lang=None):
                         ep=(episode_num(self.season, self.episode) or
                             episode_num(self.season, self.episode, numbering='absolute')))
 
-            notifiers.notify_subtitle_download(self.pretty_name(), subtitle_list)
+            try:
+                notifiers.notify_subtitle_download(self.pretty_name(), subtitle_list)
```
p0psicles: Honestly I don't know how a notification can ever throw an exception. That smells like bad design.
As a notification is only needed, when you have something a notify on. Maybe using some kind of callback, can fix this.

p0psicles: Never mind. I thought it was pNotification.

medariox:
Was this tested at least once? RequestException doesn't catch socket.gaierror.

labrys:
@fernandog when you're not sure you can easily test 
```python
>>> try:
	raise socket.gaierror
except requests.exceptions.RequestException:
	pass

  File "<pyshell#19>", line 2, in <module>
    raise socket.gaierror
socket.gaierror
```

PR #2378, opened 2017-03-10
Fix wrapping in errorlogs

fernandog:
@Thraxis @p0psicles

~~I don't get why text is wrapping. Any idea?~~
~~I tried several things but i can't fix it~~

~~In viewlogs.mako this doesn't happen~~

image

### DIFF
--- a/views/errorlogs.mako
+++ b/views/errorlogs.mako
@@ -13,7 +13,9 @@ pre {
 </style>
 </%block>
 <%block name="content">
+
 <%
+    from mako.filters import html_escape
     if logLevel == logger.WARNING:
         errors = classes.WarningViewer.errors
         title = 'WARNING logs'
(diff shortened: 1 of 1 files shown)


p0psicles:
Maybe I'm not the visual minded. But why would I want to have fanart on the log page?

fernandog:
@p0psicles because we have in viewlogs. 

So or we add to errorlogs or we remove from viewlogs

p0psicles:
Personally I don't think it's needed for the log page. Only distracts.

fernandog:
ok, but anyway the issue about wrapping still exists

also this one: https://github.com/pymedusa/Medusa/issues/2379

fernandog:
@wimpyrbx any idea why its wrapping?

wimpyrbx:
@fernandog  will have a look later today when all IRL tasks are done ;)

OmgImAlexis:
@fernandog going to have to agree with @p0psicles I don't think when diagnosing issues we should have fan art in our faces. Can we just keep the other fixes in this PR but remove the fanart?

PR #2380, opened 2017-03-10
Add list_associated_files test

fernandog:
Need to switch base after `rewrite-process-tv` gets merged as I'm not allowed to PR to that branch =)

~~ratoaq2 confirm i can't run two tests?~~

~~I tried to use the existing structure from `test_should_process`  to avoid call `test_path = create_structure(p['path'], structure=p['structure'])` twice by creating another function called `def test_list_associated_files` but didn't worked~~

~~is there a way ?~~

### DIFF
--- a/tests/test_list_associated_files.py
+++ b/tests/test_list_associated_files.py
@@ -27,9 +27,8 @@ class TestPostProcessor(PostProcessor):
 
 
 @pytest.mark.parametrize('p', [
-    {  # p0: Process file and no associated files
+    {  # p0: No associated files. Wrong basename
         'path': 'media/postprocess/',
-        'expected': True,
         'structure': (
             'bow.514.hdtv-lol[ettv].mkv',
             'bow.514.hdtv-lol.srt',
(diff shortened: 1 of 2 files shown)


tests/test_list_associated_files.py
```diff
@@ -0,0 +1,165 @@
...
+def test_list_associated_files(p, create_structure):
+    """Run the test."""
+    # Given
+    test_path = create_structure(p['path'], structure=p['structure'])
+    path = os.path.join(test_path, os.path.normcase(p['path']))
+    media_file = os.path.join(path, p['structure'][0])
+    expected_associated_files = p['expected_associated_files']
+    subtitles_only = p.get('subtitles_only', False)
+    subfolders = p.get('subfolders', False)
+    app.ALLOWED_EXTENSIONS = p['allowed_extensions']
```
ratoaq2: This should be  monkeypatched to avoid  interference with other tests. Test  should change state and leave it behind. 

`monkeypatch.setattr(app, 'NAME', value)`

PR #2383, opened 2017-03-12
move mako into ui dir

OmgImAlexis:
@labrys @p0psicles


fernandog:
@OmgImAlexis i would create a new PR instead of fix conflicts...

PR #2409, opened 2017-03-14
add more cats

OmgImAlexis:
@fernandog


fernandog:
@duramato why we have diff categories per Season|RSS|Episode?

Can't we have just self.categories = {c2': 1, 'c7': 1, 'c24': 1, etc....}

OmgImAlexis:
Season is there so episodes can be searched without season packs being returned in the results.

PR #2410, opened 2017-03-14
Log modified values in Series object and imdb_info

fernandog:
It helps see what was updated to DB
```
SHOWQUEUE-UPDATE :: [8fbc3d8] Series object keys that were modified: {'last_update_indexer': (736401, 736395)}
SHOWQUEUE-UPDATE :: [8fbc3d8] IMDB object keys that were modified: {'votes': (37124, 36202), 'last_update': (736401, 736395), 'title': (u'American Crime Story', u'The People v. O.J. Simpson: American Crime Story')}
```
@p0psicles


fernandog:
Today we do a INSERT in DB no matter what.
We could check if we have modified values and only then do the INSERT/UPDATE

Issue #1868, opened 2016-12-30
Reverse Proxy Not Working


Smithoo4:
I am trying to set up nginx as a reverse proxy with let’s encrypt so I can remotely access Medusa/Sickrage. When I try to access Medusa/Sickrage through the revers proxy Chrome will not load the page given a message "This page is trying to load scripts from unauthenticated sources". If I tell Chrome to load the unsafe script I get a message “Unable to connect to Medusa!” and them a blank page. 

I can still access Medusa/Sickrage locally by its IP with no issue. I have also gotten the reverse proxy to work for Plex, CouchPotato, Transmission and Ubooquity with little to no issue.

In terms of settings up the reverse proxy I have basically followed the guides below:

http://www.htpcguides.com/secure-nginx-reverse-proxy-with-lets-encrypt-on-ubuntu-16-04-lts/

http://www.htpcguides.com/configure-sickrage-reverse-proxy-nginx/

The big differences is I currently have nginx running on different box for testing and am using static IP instead of 127.0.0.1 in the config and I have not set user name and password for Medusa/Sickrage. I don't think that should make a diffrences but please correct me if I am wrong.

A copy of my Medusa/Sickrage config.ini can be found here:

http://pastebin.com/fbehWp0N

Medusa Info:		
Branch: master
Commit: 		677b47ac1f6a75abc71c4472f3fb9c4eecb6413f
Version: 		v0.1.9
Database: 		44.3
Python Version:	2.7.12 (default, Nov 7 2016, 11:55:55) [GCC 6.2.1 20160830]
SSL Version:		OpenSSL 1.0.2j 26 Sep 2016
OS:			Linux-4.8.13-1-ARCH-x86_64-with-glibc2.2.5
Locale:		en_CA.UTF-8

I am not sure if this is an Medusa/Sickrage or a issue with how I have setup nginx, but any help would be appreciated. Please let me know if there is any other info need.

Thank you for your help and this great project.

OmgImAlexis:
You may want to reset your API key since you just posted it here.

With the errors coming up does it state which scripts are being loaded from "unauthenticated sources"?

Smithoo4:
Good point about the API key, I just rest it. I did not even think about that.

The link below is an output from chrome when it give the message "This page is trying to load scripts from unauthenticated sources".

http://pastebin.com/mJTqKE2j

This link is when I tell chrome to load the unsafe script:

http://pastebin.com/X54BdjCP

[...]

PR #2413, opened 2017-03-14
Refactoring api v2

ratoaq2:
- Removed regex hell
- Stick to standards and our conventions: `/resource/{id}/subresource/{id}`
- Resource names should be singular
- Resource handlers should handle common scenarios: not found, bad request, ok, created, no content, etc
- Resource handlers shouldn't have business logic. E.g.: SeriesHandler should use tv.Series object to retrieve and store data.

Reference implementation: SeriesHandler, EpisodeHandler
TODO:
- [x] Move backlogged and archiveEpisodes to somewhere else
- [x] Remove ShowHandler
- [x] Refactor sceneexceptions
- [x] Fix flake issues
- [x] Fix tests
- [x] Fix UI / javascript
- [x] Update wiki
- [x] Update api v2 docs
- [x] Create tests for all endpoints


OmgImAlexis:
@ratoaq2 checkout [dredd](https://github.com/apiaryio/dredd). I've already added it we just need to enable the tests. It'll make sure we have a valid API by matching our API with our [api-description file](https://github.com/pymedusa/Medusa/blob/master/api-description.yml).

ratoaq2:
The PR is done.
Review and testing is needed.

Next: I'll update the wiki

Tests still will take a while. I'll do them piece by piece in separate PRs

fernandog:
added this to my branch

fernandog:
@ratoaq2 can you solve conflicts please?

medusa/essentials/dictionary.py
```diff
@@ -0,0 +1,28 @@
+# coding=utf-8
+"""Extended dictionaries."""
+
+from collections import OrderedDict
+
+
+def not_none(x):
+    """Return true is x is not none."""
```
labrys: Return true "if" x is not none

medusa/essentials/dictionary.py
```diff
@@ -0,0 +1,28 @@
...
+    return x is not None
+
+
+def not_empty(x):
+    """Return true is x is not none and not empty."""
+    return x is not None and x != ''
+
+
+class OrderedPredicateDict(OrderedDict):
+    """Ordered dictionary that only accept keys that matches the predicate."""
```
labrys: Why does this need to be an OrderedDict? Dicts are better performance so unless maintaining insertion order is absolutely necessary dict should be preferred.

labrys: Also instead of a predicate, the more pythonic way is to just subclass and extend the appropriate methods (e.g. setitem, missing, etc).

medusa/essentials/dictionary.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, predicate=not_empty, *args, **kwargs):
+        """Constructor."""
+        super(OrderedPredicateDict, self).__init__(*args, **kwargs)
+        self.predicate = predicate
+
+    def __setitem__(self, key, value, dict_setitem=dict.__setitem__):
```
labrys: Passing a setitem method from a different class is also an antipattern.

fernandog:
@ratoaq2 archive working  `Successfully archived episodes` and some other stuff we talked

what else do you think a test is mandatory?

ratoaq2:
I'm still finishing tests for http 4xx with dredd

ratoaq2:
Wiki was updated as well.

Wiki: https://github.com/pymedusa/Medusa/wiki/API-v2
Docs: http://docs.medusaapi.apiary.io/#

OmgImAlexis:
@ratoaq2 while you're still doing this what are your thoughts on moving from /authenticate from returning a string to returning a JSON object with the token as well as the expiry length so the dev doesn't have to decrypt the JWT just to see how long it's got until it expires?

Something like this.
```json
{
    "token": "TOKEN_HERE",
    "exp": 3600
}
```

labrys:
@OmgImAlexis personally i think the token by itself is sufficient, decoding the jwt to validate the acquired claims is fairly standard.

fernandog:
Fixes: https://github.com/pymedusa/Medusa/issues/1868

p0psicles:
Needs upmerge!

PR #2415, opened 2017-03-15
Fix wrong type while saving to DB. Indexer is TEXT, not INT

fernandog:
Indexer in DB is TEXT and we are saving as INT

FINDSUBTITLES :: [bde37e2] Episodes object keys that were modified: {'subtitles_searchcount': (1, 0), 'subtitles_lastsearch': ('2017-03-15 01:46:16', u'0001-01-01 00:00:00'), 'indexer': (1, u'1')}

subtitles_lastsearch is wrong but Im not sure where to fix

### DIFF
--- a/medusa/tv/episode.py
+++ b/medusa/tv/episode.py
@@ -865,7 +865,7 @@ class Episode(TV):
                         b'  manually_searched = ? '
                         b'WHERE '
                         b'  episode_id = ?',
-                        [self.indexerid, self.indexer, self.name, self.description, ','.join(self.subtitles),
+                        [self.indexerid, str(self.indexer), self.name, self.description, ','.join(self.subtitles),
                          self.subtitles_searchcount, self.subtitles_lastsearch, self.airdate.toordinal(), self.hasnfo,
                          self.hastbn, self.status, self.location, self.file_size, self.release_name, self.is_proper,
                          self.show.indexerid, self.season, self.episode, self.absolute_number, self.version,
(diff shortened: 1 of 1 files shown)


fernandog:
@pymedusa/developers anyone has a better solution?

keep in mind that table `indexer_mapping` stores as INT. so we can't change type while loading from DB

medariox:
Didn't really have the time to look much into it, but why can't we just remove the int cast here? https://github.com/pymedusa/Medusa/blob/9f7717e166ab32a32e390022b49108fd642ad6b4/medusa/tv/episode.py#L401

fernandog:
Because all medusa code uses it as INT also indexer_mapping table

@p0psicles what do you suggest?

medariox:
So why don't we make the indexer column an int instead? I mean, what is this PR going to solve? Sqlite does an implicit cast already.

fernandog:
Obj is dirry when not needed

fernandog:
@ratoaq2 but DB is text and we cant change that now

ratoaq2:
Db is text and it shouldn't. That's not a reason to  make all other parts wrong as well. 

Why can't we change it? 
And if  it's not possible to change, then  we should  make sure that we  convert it to int  in every read

fernandog:
We convert to int in load_from_db.
Im only converting to STR right before inserting into DB

ratoaq2:
If it's already  int just after you load, what's the issue you're trying to fix? 
In python objects you should only have ints, since you load them from db and convert them to int

labrys:
Type-casting should occur on load from the database since SQLite does not enforce types.

fernandog:
@labrys i see but as I said in the PR, in `load_from_db` we convert indexer into INT. 

What im doing is just make TEXT again before saving to DB

Issue #2418, opened 2017-03-15
"Use initial indexer set to" option is ignored with "Add Existing Shows"


YogSottot:
### INFO
**Python Version**: `2.7.12 (default, Dec 12 2016, 13:32:24) [GCC 4.9.3]`
**Operating System**: `Linux-4.9.13-gentoo-nvidia`
**Locale**: `UTF-8`
**Branch**: [master](../tree/master)
**Database**: `44.5`
**Commit**: pymedusa/Medusa@a0116154f262786d6f5274459cb8f9cdb23012cc

### ERROR
In settings i set: "Use initial indexer set to" to tvmaze. 
![tvmaze](https://i.imgur.com/59m2H3A.png)

But when i use "Add Existing Shows", it set it to tvdbv2 by default. 
![tvmaze](https://i.imgur.com/CiWwjx0.png)

With "Add New Show" it works as expected.
![tvmaze](https://i.imgur.com/HH5PV8V.png)

PR #2419, opened 2017-03-15
Fix not using default indexer in mass add shows if no indexer found in metadata

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2418


fernandog:
PR is simple. Can someone review please?

PR #2420, opened 2017-03-15
TRAKT doesn't support TVmaze. Add support to TMDB

fernandog:
Related to https://github.com/pymedusa/Medusa/issues/2402

TODO:
- [x] Don't let add same show with different indexer

@p0psicles @Thraxis


fernandog:
Last commit fixes the issue i had that I didn't recognized i had a show already added

medusa/__main__.py
```diff
@@ -777,6 +777,9 @@ def initialize(self, console_logging=True):
             app.TRAKT_SYNC = bool(check_setting_int(app.CFG, 'Trakt', 'trakt_sync', 0))
             app.TRAKT_SYNC_REMOVE = bool(check_setting_int(app.CFG, 'Trakt', 'trakt_sync_remove', 0))
             app.TRAKT_DEFAULT_INDEXER = check_setting_int(app.CFG, 'Trakt', 'trakt_default_indexer', 1)
+            if app.TRAKT_DEFAULT_INDEXER == 3:
```
medariox: Instead of doing this, it should be made non-selectable in the UI.

medusa/show/show.py
```diff
@@ -67,7 +67,8 @@ def find(shows, indexer_id, indexer=None):
             return None
 
         indexer_ids = [indexer_id] if not isinstance(indexer_id, list) else indexer_id
-        results = [show for show in shows if (indexer is None or show.indexer == indexer) and show.indexerid in indexer_ids]
+        results = [show for show in shows if show.imdb_id == indexer_id and indexer == 10 or
```
medariox: Can you clarify this? It's not clear what the point of indexer ```== 10``` is.

medusa/__main__.py
fernandog: I already did that.
Some users have the wrong indexer in the UI. This code fix it

medusa/show/show.py
fernandog: IMDB

views/config_notifications.mako
```diff
@@ -1565,7 +1565,12 @@
                                         <span class="component-desc">
                                             <select id="trakt_default_indexer" name="trakt_default_indexer" class="form-control input-sm">
                                                 % for indexer in indexerApi().indexers:
-                                                <option value="${indexer}" ${'selected="selected"' if app.TRAKT_DEFAULT_INDEXER == indexer else ''}>${indexerApi().indexers[indexer]}</option>
+                                                    <%
+                                                        name = indexerApi().indexers[indexer]
```
medariox: You are calling the class every time to do this.
Instead save indexerApi().indexers to a variable and use that in the loop.

medusa/trakt_checker.py
```diff
@@ -421,6 +422,10 @@ def fetch_trakt_shows(self):
                 else:
                     show_name = show_obj['title']
 
+                if Show.find(app.showList, show_obj['imdb_id'], EXTERNAL_IMDB):
+                    logger.log(u"Show already added in your library: {0}".format(show_obj['title']), logger.DEBUG)
```
medariox: Single quotes.

medusa/show/show.py
medariox: Can you add a comment please? Also we should really try to avoid this excessive amount of code in list comprehensions.

views/config_notifications.mako
fernandog: Ops

medusa/__main__.py
medariox: This is the kind of code that gets forgotten at some time and we wonder why we always have indexer 1 selected instead of 3. Why not add a warning saying that it should be changed, just for 1 release and remove it after that?

fernandog: Dont worry. Trakt dev will let me know when tvmaze is added. Then I will remove this

fernandog: Until user remove it will keep erroring. Warning a good solution

medusa/show/show.py
p0psicles: Instead of 10, pls add constants.

medusa/__main__.py
p0psicles: Pls use constants?

medusa/show/show.py
fernandog: Cant use here. Loop import

fernandog:
With this last commit:

2017-03-16 12:01:14 DEBUG    TRAKTCHECKER :: [b4a8a54] Show already added in your library: 3% with indexer: IMDB
2017-03-16 12:01:14 DEBUG    TRAKTCHECKER :: [b4a8a54] Show already added in your library: The Night Of with indexer: TVDB

p0psicles:
I don't fully understand why your approaching it like this.
You want to check if a specific show from trakt, already exists in the library?

In that case why not use the helper check_existing_shows.
https://github.com/pymedusa/Medusa/blob/master/medusa/helper/externals.py#L123

imdb is also handled there, as many others.

fernandog:
@p0psicles I did that first
I need to create a show object and load_from_indexer to get externals (which need indexers available)

But Trakt already give us all ids : tvrage, tmdb, imdb and tvdb. Its a lot faster

p0psicles:
I'm not totally convinced on the changes made to Show.find(). As I don't like the hacking of the imdb comparison there. But rest of PR is okay. As it's confined to the already messy trakt code.

PR #2435, opened 2017-03-19
Fix all externals being lost in the INSERT - except the last one

fernandog:
TODO: 

- [ ] Remove comment in minor increase version

@p0psicles

```sql
SELECT indexer_id, count(mindexer) from indexer_mapping
GROUP by indexer_id
```

will return only one row per indexer_id. when it should return at least 4 like mine SS after i modified table:

image

### DIFF
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -608,10 +608,11 @@ class Tmdb(BaseIndexer):
                 if result.get('tv_results') and result['tv_results'][0]:
                     # Get the external id's for the passed shows id.
                     externals = self.tmdb.TV(result['tv_results'][0]['id']).external_ids()
+                    externals['tmdb_id'] = result['tv_results'][0]['id']
+
                     externals = {external_id: external_value
                                  for external_id, external_value
                                  in externals.items()
                                  if external_value and external_id in ['tvrage_id', 'imdb_id', 'tvdb_id']}
-                    externals['tmdb_id'] = result['tv_results'][0]['id']
                     return externals
         return {}
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -594,7 +594,7 @@ class Tmdb(BaseIndexer):
         return show_season_updates
 
     def get_id_by_external(self, **kwargs):
-        """Search tvmaze for a show, using an external id.
+        """Search tmdb for a show, using an external id.
 
         Accepts as kwargs, so you'l need to add the externals as key/values.
         :param tvrage_id: The tvrage id.
@@ -608,11 +608,10 @@ class Tmdb(BaseIndexer):
                 if result.get('tv_results') and result['tv_results'][0]:
                     # Get the external id's for the passed shows id.
                     externals = self.tmdb.TV(result['tv_results'][0]['id']).external_ids()
-                    externals['tmdb_id'] = result['tv_results'][0]['id']
-
                     externals = {external_id: external_value
                                  for external_id, external_value
                                  in externals.items()
                                  if external_value and external_id in ['tvrage_id', 'imdb_id', 'tvdb_id']}
+                    externals['tmdb_id'] = result['tv_results'][0]['id']
                     return externals
         return {}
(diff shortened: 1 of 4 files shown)


fernandog:
With second commit:

image

fernandog:
image

fernandog:
@p0psicles ?

p0psicles:
Need to reproduce this myself. As I tested the indexer_mapping table when I worked on the externals. And didn't notice this back then.

medusa/indexers/tmdb/tmdb.py
```diff
@@ -603,11 +603,10 @@ def get_id_by_external(self, **kwargs):
                 if result.get('tv_results') and result['tv_results'][0]:
                     # Get the external id's for the passed shows id.
                     externals = self.tmdb.TV(result['tv_results'][0]['id']).external_ids()
-                    externals['tmdb_id'] = result['tv_results'][0]['id']
-
                     externals = {external_id: external_value
                                  for external_id, external_value
                                  in externals.items()
                                  if external_value and external_id in ['tvrage_id', 'imdb_id', 'tvdb_id']}
+                    externals['tmdb_id'] = result['tv_results'][0]['id']
```
fernandog: well, i figure it out later that moving this is not needed because only in tvmaze we check the `mapping` list.  and the externals['tvmaze_id'] need to be done after that list compreh.

fernandog:
maybe mindexer_id don't need to be PK right? only mindexer

fernandog:
@p0psicles please? we are wasting all calls to externals with this

p0psicles:
Hmm seems this is how the INSERT OR IGNORE works. I'd actually prefer to change this to just do a query and then optional an insert. Don't know how the rest feels about this? @labrys @medariox  ?

fernandog:
@p0psicles yes, but as the other rows aren't PK, it will always overrides the others indexers

medariox:
What is the reason to make indexer_id and indexer PKs in the first place?

p0psicles:
I think because we want to use INSERT OR IGNORE. That uses columns that are primary key, insert or ignore it. Under water some sort of constraint exception is used, to ignore it.

@fernandog I suggest to change this to a simple Select (check if the row is already in the table) and if not, insert it. I prefer that over the INSERT OR IGNORE "hack".

fernandog:
but why can't we make PK all columns? 
I talked to a colleague (DBA) from work and he said this is the best solution

With your solution we will double the "access" to DB for every show update

Thraxis:
I can't imagine ANY DBA recommending that every column in a table be made PK that makes no sense at all. 

Add  an auto-number id column as primary key, create a unique constraint on the 4 columns

```CREATE TABLE [indexer_mapping2](
    [id] INTEGER PRIMARY KEY AUTOINCREMENT, 
    [indexer_id] INTEGER, 
    [indexer] NUMERIC, 
    [mindexer_id] INTEGER, 
    [mindexer] NUMERIC, 
    UNIQUE([indexer_id], [indexer], [mindexer_id], [mindexer]));```

The INSERT or IGNORE should work then

**Comment by fernandog** [comment]
@Thraxis thanks, anything else should be changed in the sqls to query the table?

**Comment by labrys** [comment]
@Thraxis In a relational db an association table is a common design pattern in which you associate two items in a separate table for a many-to-many relationship.  For such a table having both items as a composite natural key is an intentional and quite functional implementation.  In this instance the foreign keys for the association are both composite natural keys so having a 4-column pk is not an anti-pattern.  

The anti-pattern would be to have columns included in the primary key that aren't required to uniquely identify a row.

Examples of this can be seen in **SQL and Relational Theory, 3rd Edition** and **Relational Theory for Computer Professionals**

**Comment by fernandog** [comment]
@Thraxis can't we make this instead of row id?

"CREATE UNIQUE INDEX idx_indexer_id ON indexer_mapping(indexer_id);"

**Comment by labrys** [comment]
I would also create a (non-unique) index on the indexer and indexer_id for efecient lookup. Also a reverse-index could be useful if reverse-mappings are desired from the database.

File: medusa/tv/series.py [hunk 2]
```diff
@@ -900,13 +900,16 @@ def _load_externals_from_db(self, indexer=None, indexer_id=None):
                b'WHERE (indexer = ? AND indexer_id = ?) '
                b'OR (mindexer = ? AND mindexer_id = ?)')
 
-        results = main_db_con.select(sql, [indexer, indexer_id, indexer, indexer_id])
-
+        sql_results = main_db_con.select(sql, [indexer, indexer_id, indexer, indexer_id])
+        results = [dict(result) for result in sql_results]
```
fernandog: For debugging as I cant see the contents of sqlite row.
I can revert it

Issue #1729, opened 2016-12-02
AttributeError: 'NoneType' object has no attribute 'indexer'


fernandog:
@p0psicles 

@ratoaq2 I only saw this issue in log, didn't got any traceback. know why?

```
2016-12-02 19:05:00 INFO     FINDPROPERS :: [Rarbg] :: [1d6c65b] Searching for any new PROPER releases from Rarbg
2016-12-02 19:05:01 DEBUG    FINDPROPERS :: [Rarbg] :: [1d6c65b] GET URL: http://torrentapi.org/pubapi_v2.php?get_token=get_token&app_id=medusa&format=json [Status: 200]
2016-12-02 19:05:01 DEBUG    FINDPROPERS :: [Rarbg] :: [1d6c65b] Search mode: Episode
2016-12-02 19:05:01 DEBUG    FINDPROPERS :: [Rarbg] :: [1d6c65b] Unknown exception while searching for propers in Rarbg, skipping: 'NoneType' object has no attribute 'indexer'
2016-12-02 19:05:01 DEBUG    FINDPROPERS :: [Rarbg] :: [1d6c65b] Traceback (most recent call last):
  File "C:/Users/Fernando/Documents/Medusa\medusa\search\proper.py", line 137, in _get_proper_results
    cur_propers = cur_provider.find_propers(recently_aired)
  File "C:/Users/Fernando/Documents/Medusa\medusa\providers\generic_provider.py", line 148, in find_propers
    for item in self.search(search_strings[0], ep_obj=episode_obj):
  File "C:/Users/Fernando/Documents/Medusa\medusa\providers\torrent\json\rarbg.py", line 99, in search
    search_params['search_tvdb'] = self._get_tvdb_id()
  File "C:/Users/Fernando/Documents/Medusa\medusa\providers\generic_provider.py", line 486, in _get_tvdb_id
    return self.show.indexerid if self.show.indexer == INDEXER_TVDBV2 else self.show.externals.get('tvdb_id')
AttributeError: 'NoneType' object has no attribute 'indexer'
```

p0psicles:
Aaah yes, that was the proper issue. Forgot about that one!

fernandog:
when debugging, in rarbg.py, the self is the show and self.show.indexer = 1

but when it go to generic_provider, show is none:
![image](https://cloud.githubusercontent.com/assets/2620870/20850624/40f60488-b8c3-11e6-9963-b7366a14b9dd.png)

[...]

PR #2439, opened 2017-03-21
Update adba to medusa-1.0.0

labrys:
Update adba to [adba-medusa-1.0.0](https://github.com/pymedusa/adba) for py3 compatibility.


Thraxis:
I just took a quick look through...  Why was the downloading of the updated ScudLee XML files removed?

labrys:
@p0psicles unintended.. i just removed the old folder and replaced with the updated lib folder.  I'll add it back but IMO if its not part of the original lib it shouldn't go in the libs folder.

labrys:
Oops meant to tag that @Thraxis

p0psicles:
Ideally no your right, we shouldn't place monkey patched libs in the lib folder. But if we remove adba, then there are at least another dozen which should be moved to where??

fernandog:
i think we should use this PR and fix the https://github.com/pymedusa/Medusa/issues/1729#issuecomment-283417088 as @medariox said

medariox:
@labrys 
What's the status here? Needs review/testing? What about the removed XML files?

labrys:
@medariox Adding the removed XML to medusa and having them loaded i think was the last thing this was waiting on.  I think they should be added external to the lib personally.  Thoughts?

p0psicles:
I think providing a caching location for the xml would be the cleanest solution.

medariox:
@labrys 
Sorry for the late answer, I totally forgot about the mention.
The problem I see here is not much about the missing XML files per se, but rather about the fact that you removed the whole download/update logic of those files. We don't want to be manually updating those every time (they seem to get updates quite often). IMO, since we already forked this lib and changed it quite a bit (and the changes won't be added upstream), we should keep the update mechanism. Possibly not storing the XMLs in the lib folder, but rather in the cache folder like @p0psicles suggested.

PR #2443, opened 2017-03-21
Fix history size comparison and add size to downloaded status

fernandog:
@medariox please test when you can

### DIFF
--- a/medusa/history.py
+++ b/medusa/history.py
@@ -84,7 +84,7 @@ def log_snatch(searchResult):
                         provider, version, proper_tags, manually_searched, info_hash, size)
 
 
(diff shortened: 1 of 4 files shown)


medariox:
Have this running locally now and row highlight seems to work again.

Column is still ugly:
_Too small:_
image
_Missing border:_
image

I will leave this running tonight, so we can see if the missing size is also still an issue.

fernandog:
@medariox ok. 

about the "ugly". it depends on the release name size. If is too long, it will need more space.

medariox:
If you look at the whole table, you can see that "size" has completely unproportional borders:
image

And there is still the missing bottom border.

fernandog:
@medariox i only added column. just like the others

maybe @Thraxis can help understand why

PR #2444, opened 2017-03-21
Feature/tvdbv2 jwt hook

p0psicles:

This PR will add the following improvements:
* tvdb apiv2 lib's urllib3 replaced with requests.
* for jwt authentication, use a JWT hook. tnx @labrys.
* lazy authentication. Meaning, api token bearer will only be retrieved when a call is made to tvdb. With this we can make sure we always can create a TVDB indexer instance.

TODO:
- [x] ~~Remove the exception handling throughout Medusa, where we initiate the tvdb indexer instance. Like here: https://github.com/pymedusa/Medusa/blob/develop/medusa/show_updater.py#L67~~
Unfortunatly we cannot remove these yet, as tmdb also does api initialization in the init.
- [x] Include changes in pyMedusa/tvdbv2
- [x] Fix trakt recommended shows, as it's using tvdb for posters. And that doesn't work anymore.
- [x] Implement timeout based caching on the trakt recommended shows, for the posters that tvdb does not currently have. Meaning if for a trakt show, i couldn't find a tvdb poster. It's added to a list for 3 days. Then for these 3 days it will not attempt looking for it on tvdb.


lib/tvdbapiv2/rest.py
```diff
@@ -51,125 +31,50 @@
...
-                                              fields=query_params,
-                                              headers=headers)
-        except urllib3.exceptions.SSLError as e:
+                r = self.session.request(method, url, params=query_params, headers=headers)
+
+        except RequestException as e:
             msg = "{0}\n{1}".format(type(e).__name__, str(e))
             raise ApiException(status=0, reason=msg)
 
         r = RESTResponse(r)
```
labrys: First I see no use for the `RESTResponse` class at all secondly `r` is referenced before assignment so this should all be in the `else` clause of the `try`.

lib/tvdbapiv2/rest.py
```diff
@@ -51,125 +31,50 @@
...
 
-        # In the python 3, the response.data is bytes.
-        # we need to decode it to string.
-        #if sys.version_info > (3,):
-        r.data = r.data.decode('utf8')
-
         # log response body
         logger.debug("response body: %s" % r.data)
```
labrys: Logging the entire response content is a bad idea.

labrys:
The `GET`, `PUT`, etc methods should all be lowercase.

lib/tvdbapiv2/configuration.py
```diff
@@ -20,8 +20,7 @@
 
 from __future__ import absolute_import
 
-import base64
-from requests.packages import urllib3
+from requests.packages.urllib3.util import make_headers
 
 try:
     import httplib
```
labrys: `from six.moves.http_client import HTTPConnection`

labrys:
`api_client.ApiClient.set_default_headers` is unused and I don't see a reason to ever have it since you can do `ApiClient.default_headers[name] = value` directly.

lib/tvdbapiv2/api_client.py
```diff
@@ -67,22 +67,26 @@ class ApiClient(object):
...
-    def __init__(self, host=None, header_name=None, header_value=None, cookie=None):
+    def __init__(self, host=None, cookie=None, session=None, user_agent=None, api_key=None):
 
         """
         Constructor of the class.
         """
-        self.rest_client = RESTClientObject()
+        self.session = session or requests.session()
+        if api_key:
+            self.session.auth = TVDBAuth(api_key)
```
labrys: Some form of authentication is required for `TVDBAuth`, either a token or an apikey.

p0psicles: Ok, I understand that I can throw an exception when when we don't have an api_key passed. But should I also make it check for token?

lib/tvdbapiv2/configuration.py
p0psicles: tnx

lib/tvdbapiv2/rest.py
```diff
@@ -51,125 +31,50 @@
...
-        """
-        method = method.upper()
-        assert method in ['GET', 'HEAD', 'DELETE', 'POST', 'PUT', 'PATCH', 'OPTIONS']
-
-        if post_params and body:
-            raise ValueError(
-                "body parameter cannot be used with post_params parameter."
-            )
+    def __init__(self, session=None):
+        self.session = session
```
p0psicles: Session is always passed on from api_client. Then I think `session=None` should be changed to `session`.

lib/tvdbapiv2/api_client.py
labrys: Yes since you can auth with one or the other.

lib/tvdbapiv2/rest.py
```diff
@@ -51,125 +31,50 @@
...
 
-        post_params = post_params or {}
-        headers = headers or {}
-
-        if 'Content-Type' not in headers:
-            headers['Content-Type'] = 'application/json'
+    def request(self, method, url, query_params=None, headers=None, body=None, post_params=None):
```
p0psicles: I have to as some of them can't be passed on by *args. Like query_params and post_params. Or maybe i'm misunderstanding.

lib/tvdbapiv2/rest.py
labrys: personally I don't see a problem with either way

lib/tvdbapiv2/rest.py
```diff
@@ -51,125 +31,50 @@
...
             else:
-                r = self.pool_manager.request(method, url,
-                                              fields=query_params,
-                                              headers=headers)
-        except urllib3.exceptions.SSLError as e:
+                r = self.session.request(method, url, params=query_params, headers=headers)
+
+        except RequestException as e:
             msg = "{0}\n{1}".format(type(e).__name__, str(e))
             raise ApiException(status=0, reason=msg)
```
p0psicles: The ApiException and the choice to check for the status 200..206, has been made by the swagger compiler.

Regarding `RequestException.response.status`, I don't know if I can do that. As a RequestException does not necessarily have a response object.

p0psicles: Not saying you don't have good pointers there. I'll look into implementing the raise_for_status.

lib/tvdbapiv2/rest.py
p0psicles: Yeah good point. I'll remove that.

lib/tvdbapiv2/rest.py
p0psicles: It's swagger compiled. I'll see if I can get rid of it. The less code the better.

p0psicles:
I'll remove `api_client.ApiClient.set_default_headers`. If it's not used sure. But still we should keep in mind that the lib can be used without providing it a session object. for those, someone might need to modify a header value.

lib/tvdbapiv2/rest.py
labrys:
```
@property
def status_code(self):
    try:
         return self.response.status_code
    except AttributeError:  # i think this is the correct exception
         return 0
```

labrys: Rarely should a `RequestException` from a response ever not have a response object.  If you don't know if the exception is from a response then thats a different story.

labrys:
That's why the `session or requests.Session()` so a session is always present.  Also without `session` all your `self.session.request` would raise exceptions.

p0psicles:
Aah yes your right. You can always adjust session post api initializing

lib/tvdbapiv2/rest.py
labrys: Well since you have already changed the swagger compiled code (by adding the sessions, etc) I see no reason in retaining this.

lib/tvdbapiv2/rest.py
```diff
@@ -51,125 +31,50 @@
...
-        post_params = post_params or {}
-        headers = headers or {}
-
-        if 'Content-Type' not in headers:
-            headers['Content-Type'] = 'application/json'
+    def request(self, method, url, query_params=None, headers=None, body=None, post_params=None):
 
         try:
             # For `POST`, `PUT`, `PATCH`, `OPTIONS`
             if method in ['POST', 'PUT', 'PATCH', 'OPTIONS']:
```
p0psicles: I think I changed everything except the if .. else. I don't really see the need for changing this right now. Can you live with that?

medusa/metadata/generic.py
```diff
@@ -805,7 +805,7 @@ def _season_banners_dict(self, show_obj, season):
             return result
 
         if ('seasonwide' not in indexer_show_obj['_banners'] or
-                'original' not in indexer_show_obj['_banners']['season'] or
+                'original' not in indexer_show_obj['_banners']['seasonwide'] or
```
p0psicles: this one popped up.

lib/tvdbapiv2/api_client.py
```diff
@@ -67,52 +61,36 @@ class ApiClient(object):
...
         Constructor of the class.
         """
         self.session = session or requests.session()
-        if api_key:
-            self.session.auth = TVDBAuth(api_key)
+        if not api_key and not token:
+            raise MissingAuthentication('Please provide an api_key or token to authenticate.')
+
+        try:
+            self.session.auth = TVDBAuth(api_key=api_key, token=token)
```
p0psicles: Apparently the TVDBAuth is not fully lazy. As when you disconnect your network connection, this is still raising exceptions.

medusa/name_parser/parser.py
```diff
@@ -105,7 +105,8 @@ def _parse_string(self, name):
                 )
 
             if season_number is None or not episode_numbers:
-                logger.debug('Show {name} has no season or episodes, using indexer...', name=result.show.name)
+                log.debug('Series {name} has no season or episodes, using indexer',
+                          {'name': result.show.name})
```
fernandog: Missing space to show indexer name right?

p0psicles: Huh? Missing space where?

fernandog: never mind... i thought you were adding the name of indexer in the end

lib/tvdbapiv2/api_client.py
labrys: Lazy does not mean it won't try to auth if the network is down. It means it won't try to auth until the session is used. If the session keeps getting used while the connection is down, it will still keep trying to auth.

lib/tvdbapiv2/api_client.py
```diff
@@ -67,52 +61,36 @@ class ApiClient(object):
...
-    def __init__(self, host=None, cookie=None, session=None, user_agent=None, api_key=None):
+    def __init__(self, host=None, cookie=None, session=None, api_key=None, token=None):
 
         """
         Constructor of the class.
         """
         self.session = session or requests.session()
-        if api_key:
-            self.session.auth = TVDBAuth(api_key)
+        if not api_key and not token:
```
labrys: `if not (api_key or token)`  It only requires one or the other, not both.  The purpose of the token is if you already have a previous token-auth you wish to reuse.

lib/tvdbapiv2/api_client.py
p0psicles: But thats when a call is being made. Here it's trying to Auth when the Api client is initialized. So session is nog doing a get or post. I would not expect it to try to authenticate at that point.

labrys:
@p0psicles Merging dev into this branch pollutes the commit history since this is a lib and lib changes should not be affected by local changes.  The PR to merge the libs changes into dev would be the place to handle any subsequent changes in Medusa.  This is another reason that I think lib changes should be made in the upstream lib first (particularly since this is our own lib).

p0psicles:
That won't go here. As I'm also making integration related changes. That's also the reason I'm pulling in dev. To be able to test any integration issues with the lib. We can make all kinds of changes to the lib. But then every time I discover integration issues. We can again patch up the lib, and merge it again into Medusas code base. That also pollutes commit history.

labrys:
@p0psicles Actually it doesn't because then it serves to document in the lib the reasons for the changes, as when all this gets pushed back to the lib, all the reasons for the changes in the lib won't be attached to their changes.  You have to think of libs as completely separate entities that need to maintain their own identity.  I know why you merged dev in this PR, my point is to reinforce why I think this should have been PRs to the lib instead.  Also this could still have been tested simultaneously in a Medusa branch that wasn't intended to be merged to identify issues that would need to be addressed in the lib.

lib/tvdbapiv2/api_client.py
```diff
@@ -122,7 +100,7 @@ def __call_api(self, resource_path, method,
         if path_params:
             path_params = self.sanitize_for_serialization(path_params)
             for k, v in iteritems(path_params):
-                replacement = quote(str(self.to_path_value(v)))
+                replacement = (str(self.to_path_value(v)))
```
p0psicles: Why can't we have str here?

lib/tvdbapiv2/api_client.py
```diff
@@ -67,39 +61,28 @@ class ApiClient(object):
...
 
         """
         Constructor of the class.
         """
-        self.rest_client = RESTClientObject()
-        self.default_headers = {}
-        if header_name is not None:
-            self.default_headers[header_name] = header_value
+        self.session = session or requests.session()
+        if not api_key and not token:
```
p0psicles: @labrys

lib/tvdbapiv2/rest.py
```diff
@@ -51,7 +51,10 @@ def request(self, method, url, query_params=None, headers=None, body=None, post_
 
         except RequestException as error:
             msg = "{0}\n{1}".format(type(error).__name__, str(error))
-            raise ApiException(status=0, reason=msg)
+            status = 0
+            if hasattr(error, 'response'):
+                status = error.response.status_code
```
labrys:
```python
status = 0
try:
    status = error.response.status_code
except AttributeError:
    pass    
finally:
    raise ApiException(status, msg)
```

Issue #2429, opened 2017-03-19
PP moves files to SEEDING dir before season pack finishes PPing


OmgImAlexis:
@fernandog

fernandog:
thanks, will fix it monday

fernandog:
Sorry about delay but i didn't forget about this

PR #2445, opened 2017-03-21
Fix moving torrent file when PP first item of season pack

fernandog:
Also with this if the connection to client fails in next PP it will try again.

This is the only solution I could make it work
I will do more tests to make sure it works properly

I store as:
```python
 { 
'3259e1ed9d2364623078311ab4c0dd0b56d910d5': ['Show.S01E01', 'Show.S01E02'],
'a732116764caa9042d7a077b8473142e35dd561e': ['Show.S02E03']
}
```
So i can log the release names associated to that hash

@OmgImAlexis 

Fixes https://github.com/pymedusa/Medusa/issues/2429


fernandog:
worked fine:

```
2017-03-21 16:55:31 WARNING  Thread-22 :: [adc909b] Moved torrent from '[u'Legion.S01E04.1080p.AMZN.WEBRip.DD5.1.x264-KiNGS']' with hash: 8fba81332dbdd1a8a76cc6aed051d7f53dea03fd to: '/media/SAMSUNG/media/downloaded/seeding'
```

fernandog:
@OmgImAlexis did you test?

PR #2446, opened 2017-03-21
Add missing status from indexers

fernandog:
@p0psicles 

```python
import urllib2
import json

status_list = []
for i in range(1, 10):
    url = "http://api.tvmaze.com/shows?page={}".format(i)
    content = urllib2.urlopen(url=url).read()
    json_data = json.loads(content)
    for i in json_data:
        status = i.get('status', None)
        if status:
            status_list.append(str(status))

print set(status_list)


```
set(['In Development', 'Ended', 'Running', 'To Be Determined'])

### DIFF
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -236,7 +236,7 @@ class MainSanityCheck(db.DBSanityCheck):
             'in production': 'Continuing',
             'pilot': 'Continuing',
             'returning series': 'Continuing',
-            'cancelled': 'Continuing',
+            'cancelled': 'Ended',
             '': 'Unknown',
         }
 
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -230,6 +230,7 @@ class MainSanityCheck(db.DBSanityCheck):
             'pilot rejected': 'Ended',
             'canceled': 'Ended',
             'ended': 'Ended',
+            'to be determined': 'Continuing',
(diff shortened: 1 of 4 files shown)


fernandog:
This is for TMDB:
"Planned" is for any TV show that are planned but not yet in production.
"In Production" is for any TV show that are in production (i.e. from the first days of filming until the series premiere).
"Pilot" is for things like Amazon pilots or TV pilots that were never picked up to series.
"Returning Series" is used from the first episode to the last episode of any TV show.
"Ended"/"Cancelled" are used after the end of a series. (Miniseries are usually marked as Ended when they end.)

https://www.themoviedb.org/talk/58ad201d9251410794002a78

fernandog:
we should remove this code from all mako partials:
https://github.com/pymedusa/Medusa/blob/develop/views/partials/home/simple.mako#L169-L174

medusa/databases/main_db.py
```diff
@@ -230,6 +230,13 @@ def fix_indexer_show_statues(self):
             'pilot rejected': 'Ended',
             'canceled': 'Ended',
             'ended': 'Ended',
+            'to be determined': 'Continuing',
+            'running': 'Continuing',
+            'planned': 'Continuing',
+            'in production': 'Continuing',
+            'pilot': 'Continuing',
+            'returning series': 'Continuing',
+            'cancelled': 'Continuing',
```
p0psicles: Should be ended?

PR #2447, opened 2017-03-22
Fix wrong log formatting

fernandog:
prevent snatches

### DIFF
--- a/medusa/clients/torrent/generic.py
+++ b/medusa/clients/torrent/generic.py
@@ -242,7 +242,7 @@ class GenericClient(object):
 
         if not self.auth:
             if not self._get_auth():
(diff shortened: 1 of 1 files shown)


medusa/clients/torrent/generic.py
```diff
@@ -238,11 +238,11 @@ def send_torrent(self, result):
         """
         r_code = False
 
-        log.log('Calling {name} Client', {'name': self.name})
+        log.debug('Calling {name} Client', {'name': self.name})
 
         if not self.auth:
             if not self._get_auth():
-                log.log('{name}: Authentication Failed', {'name': self.name})
+                log.debug('{name}: Authentication Failed', {'name': self.name})
```
ratoaq2: Why debug here?

PR #2448, opened 2017-03-22
Fix new log error when logging a dict/json

fernandog:
```
File "C:/Medusa\medusa\init\logconfig.py", line 79, in __str__
   return result.format(*self.args, **self.kwargs) if self.args or self.kwargs else result
KeyError: u'"arguments"'
Logged from file style.py, line 63
```

image

also this: https://pymedusa.slack.com/files/wimpyrbx/F4N1CRZKP/gistfile1.txt from @wimpyrbx 

@ratoaq2 @labrys

### DIFF
--- a/medusa/init/logconfig.py
+++ b/medusa/init/logconfig.py
@@ -76,7 +76,7 @@ class BraceMessage(object):
         :rtype: str
         """
         result = text_type(self.fmt)
-        return result.format(*self.args, **self.kwargs) if self.args or self.kwargs else result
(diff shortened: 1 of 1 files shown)


fernandog:
doesn't work. it breaks the logging format

fernandog:
this is why it breaks qbitorrent:

logging the response that has the `hash` key
```
[f260199] qbittorrent: Requested a POST connection to http://localhost:xxxx/command/resume with Params: None Data: {u'hash': u'D916407A32D2EA8955A'}
```

PR #2451, opened 2017-03-22
Fix TypeError when updates_api returns None. Fixes #2424

### LINKED ISSUE #2424: [APP SUBMITTED]: TypeError: 'NoneType' object is not iterable
### INFO
**Python Version**: `2.7.9 (default, Sep 17 2016, 20:26:04) [GCC 4.9.2]`
**Operating System**: `Linux-4.4.38-v7+-armv7l-with-debian-8.0`
**Locale**: `UTF-8`
**Branch**: [develop](../tree/develop)
**Database**: `44.5`
**Commit**: pymedusa/Medusa@a70a6df42ec915a51409a059df3a58f56c1b2bd7
**Link to Log**: https://gist.github.com/839dc72eb02a30a96952cd7150164207
### ERROR
<pre>
2017-03-17 02:01:22 ERROR    SHOWUPDATER :: [a70a6df] Problem running show_updater, Indexer TVDBv2 seems to be having issues while trying to get updates for show Hannibal. Cause: 'NoneType' object is not iterable.
  File "/home/pi/Medusa/<a href="../blob/a70a6df42ec915a51409a059df3a58f56c1b2bd7/medusa/show_updater.py#L86">medusa/show_updater.py</a>", line 86, in run
    update_max_weeks)
  File "/home/pi/Medusa/<a href="../blob/a70a6df42ec915a51409a059df3a58f56c1b2bd7/medusa/indexers/tvdbv2/tvdbv2_api.py#L565">medusa/indexers/tvdbv2/tvdbv2_api.py</a>", line 565, in get_last_updated_series
    last_update_ts = max(x.last_updated for x in updates)
TypeError: 'NoneType' object is not iterable
</pre>
---

medariox:
As discussed in https://github.com/pymedusa/Medusa/issues/2424

### DIFF
--- a/medusa/indexers/tvdbv2/tvdbv2_api.py
+++ b/medusa/indexers/tvdbv2/tvdbv2_api.py
@@ -562,9 +562,10 @@ class TVDBv2(BaseIndexer):
         try:
             while updates and count < weeks:
                 updates = self.updates_api.updated_query_get(from_time).data
-                last_update_ts = max(x.last_updated for x in updates)
-                from_time = last_update_ts
-                total_updates += [int(_.id) for _ in updates]
+                if updates is not None:
+                    last_update_ts = max(x.last_updated for x in updates)
+                    from_time = last_update_ts
+                    total_updates += [int(_.id) for _ in updates]
                 count += 1
         except (ApiException, MaxRetryError, RequestError) as e:
             raise IndexerUnavailable('Error connecting to Tvdb api. Caused by: {0!r}'.format(e))


fernandog:
@medariox if updates is None, should we count +1?

medariox:
To my understanding, yes - I'm sure @p0psicles knows better.

fernandog:
Checking the code what I understood is that we only update count when we get results from API

If we are getting a None is because response from API is None. So we shouldnt increment count

```python
    def updated_query_get(self, from_time, **kwargs):
        """
        Returns an array of series that have changed in a maximum of one week blocks since the provided `fromTime`.\n\n\nThe user may specify a `toTime` to grab results for less than a week. Any timespan larger than a week will be reduced down to one week automatically.
```

fernandog:
@medariox notice that it returns the response:


```python
        response = self.api_client.call_api(resource_path, method,
                                            path_params,
                                            query_params,
                                            header_params,
                                            body=body_params,
                                            post_params=form_params,
                                            files=files,
                                            response_type='UpdateData',
                                            auth_settings=auth_settings,
                                            callback=params.get('callback'))
        return response
```

medariox:
It doesn't really matter, as the while stops anyway - I would still count it, as it can causes an endless loop otherwise. I think we should raise something like ```ApiException```.

PR #2452, opened 2017-03-22
Fixing new log style conflicting with legacy

ratoaq2:
- TODO: Replace legacy log style in order to completely drop it


fernandog:
@wimpyrbx @ripper1967 please update

PR #2455, opened 2017-03-23
Improve "manual search provider" feature description

fernandog:
as discussed
I added the manual search provider feature. this was the concept and scope of it when i designed

@OmgImAlexis better?


OmgImAlexis:
This doesn't fix it. Let someone like @medariox or @ratoaq2 that actually understands the issue look into this. You're still not understanding it.

If a MANUAL action by a user is triggered we should be using the MANUAL search providers. This has nothing todo with the name or description and this PR just makes it worse IMO.

duramato:
The thing is this setting was never to be for anything else other than the new manual search , there are N reasons for it to be separate from the forced search.

OmgImAlexis:
This just makes it worse since people are now going to be thinking there's some new feature they don't know about... @fernandog why do you have to keep rushing features and discussions? Just give people a second to think of ways to fix things before rushing into merging. FFS.

fernandog:
this feature is in develop for like 1-1,5 years
I don't get what you are saying about rushing.

Yes, we have a new feature: manual search! That is one of the reasons medusa was started

Issue #2459, opened 2017-03-24
KeyError: u'season'


fernandog:
```
2017-03-24 13:09:12 DEBUG    SHOWQUEUE-UPDATE :: [b4a8a54] Using DVD ordering.
Exception in thread SHOWQUEUE-UPDATE:
Traceback (most recent call last):
  File "C:\Python27\lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "C:/Users/fernando/Documents/Medusa\medusa\show_queue.py", line 722, in run
    self.show.load_from_indexer()
  File "C:/Users/fernando/Documents/Medusa\medusa\tv\series.py", line 1211, in load_from_indexer
    indexed_show = self.indexer_api[self.indexerid]
  File "C:/Users/fernando/Documents/Medusa\medusa\indexers\indexer_base.py", line 254, in __getitem__
    self._get_show_data(key, self.config['language'])
  File "C:/Users/fernando/Documents/Medusa\medusa\indexers\tvmaze\tvmaze_api.py", line 432, in _get_show_data
    self._get_episodes(tvmaze_id, specials=False, aired_season=None)
  File "C:/Users/fernando/Documents/Medusa\medusa\indexers\tvmaze\tvmaze_api.py", line 255, in _get_episodes
    use_dvd = cur_ep['dvd_season'] is not None and cur_ep['dvd_episodenumber'] is not None
KeyError: u'dvd_season'
```
ping @p0psicles

duramato:
### INFO
**Python Version**: `2.7.9 (default, Sep 17 2016, 20:26:04) [GCC 4.9.2]`
**Operating System**: `Linux-4.4.38-v7+-armv7l-with-debian-8.0`
**Locale**: `UTF-8`
**Branch**: [develop](../tree/develop)
**Database**: `44.8`
**Commit**: pymedusa/Medusa@8d7cd6a79c064398accc80244c2317d3996fa11a
**Link to Log**: https://gist.github.com/407c9834f03c14799f89e63ac22b15ee
### ERROR
<pre>
2017-04-07 15:59:37 ERROR    SHOWQUEUE-REFRESH :: [8d7cd6a] 275788: Error while refreshing show Dude, You're Screwed. Error: season
Traceback (most recent call last):
  File "/home/pi/Medusa/<a href="../blob/8d7cd6a79c064398accc80244c2317d3996fa11a/medusa/show_queue.py#L632">medusa/show_queue.py</a>", line 632, in run
[...]

PR #2460, opened 2017-03-24
 Fix KeyError: u'dvd_season' while updating show in TVmaze with dvd_order enabled

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2459

Add dvd_order indicator to mass update
Improve logs by removing the "[" and "["
Fix displayshow issue


fernandog:
@medariox @duramato any of you have dvd order enabled for any show?

duramato:
I don't
It will work the same way as scene numbering.

PR #2462, opened 2017-03-25
Fix sab api-key

labrys:
Fix sab api key not being initialized until a proper auth is added.


fernandog:
how does it worked since?

labrys:
I'm not sure what you're asking, but during the sab rewrite I pushed the sab apikey from the request to the session initialization, but the sab module is loaded before the apikey is loaded from the config so the api key is None.  This PR pushes the API auth back into the request temporarily until I write a proper auth method.

fernandog:
I asked how it always worked. But you just explained

PR #2464, opened 2017-03-25
Fix encoding error in notifications

labrys:
Fix charmap cant encode character during notification.


fernandog:
How does it fixes if you only changed to singlo quotes?

labrys:
Line 223 and 227 were changed to unicode strings.

PR #2466, opened 2017-03-25
Fix BraceMessage dict logging

labrys:
Fixes brace message for dict logging.

### DIFF
--- a/medusa/logger/adapters/style.py
+++ b/medusa/logger/adapters/style.py
@@ -27,7 +27,10 @@ class BraceMessage(object):
                 args = []
                 kwargs = self.args[0]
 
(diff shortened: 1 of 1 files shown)


fernandog:
@labrys what this PR fixes?

ping @ratoaq2

fernandog:
@labrys what this PR fixes?

labrys:
```python
        try:
            return self.msg.format(*args, **kwargs)
        except IndexError:
            return self.msg.format(kwargs)
```
IndexError which happens when a trying to log a dict as positional arg:
```python
log.warning('Look at my dict: {}' {'large': True})
```

PR #2468, opened 2017-03-25
Store N days of already processed propers. Not 100 items

fernandog:
If user have few providers, it will use few items
If has lot of providers, it will have more items.

It uses PROPER_SEARCH_DAYS

### DIFF
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -186,12 +186,16 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
         sorted_propers = sorted(propers.values(), key=operator.attrgetter('date'), reverse=True)
         final_propers = []
 
-        # Keep only last 100 items of processed propers:
-        self.processed_propers = self.processed_propers[-100:]
+        # Keep only items from last 3 days in processed propers:
+        latest_proper = datetime.datetime.now() - datetime.timedelta(days=3)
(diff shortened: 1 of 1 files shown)


medusa/search/proper.py
```diff
@@ -186,12 +186,16 @@ def _get_proper_results(self):  # pylint: disable=too-many-locals, too-many-bran
         sorted_propers = sorted(propers.values(), key=operator.attrgetter('date'), reverse=True)
         final_propers = []
 
-        # Keep only last 100 items of processed propers:
-        self.processed_propers = self.processed_propers[-100:]
+        # Keep only items from last 3 days in processed propers:
+        latest_proper = datetime.datetime.now() - datetime.timedelta(days=3)
```
p0psicles: I thought we already had this configurable. Can't we use that instead of 3 days?

fernandog: @p0psicles this is a new setting. the PROPER_SEARCH_DAYS (used for search)

this setting is to CLEAN the already_processed  propers we store to avoid to guessit over and over.

p0psicles: OK I understand

fernandog: @p0psicles what i have doubt is if we should align this 3 days with PROPER_SEARCH_DAYS

p0psicles: Well you'll only have a change on a recurring proper for the duration that you search for that proper. So imo it would make sense.

Issue #2471, opened 2017-03-26
TypeError: Unsupported header value None


fernandog:
2017-03-26 10:21:22 ERROR    TORNADO :: [d958c98] Uncaught exception
Traceback (most recent call last):
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\http1connection.py", line 238, in _read_message
    delegate.finish()
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\httpserver.py", line 289, in finish
    self.delegate.finish()
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\web.py", line 2022, in finish
    self.execute()
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\web.py", line 2042, in execute
    **self.handler_kwargs)
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\web.py", line 181, in __init__
    self.clear()
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\web.py", line 285, in clear
    self.set_default_headers()
  File "C:/Users/Fernando/Documents/Medusa\medusa\server\api\v2\base.py", line 80, in set_default_headers
    self.set_header('X-Medusa-Server', app.APP_VERSION)
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\web.py", line 329, in set_header
    self._headers[name] = self._convert_header_value(value)
  File "C:\Users\Fernando\Documents\Medusa\lib\tornado\web.py", line 361, in _convert_header_value
    raise TypeError("Unsupported header value %r" % value)
TypeError: Unsupported header value None
ERROR:tornado.application:Uncaught exception

PR #2472, opened 2017-03-26
Fix TypeError: Unsupported header value None

fernandog:
It only get updated in github update

Fixes https://github.com/pymedusa/Medusa/issues/2471

@ratoaq2 @OmgImAlexis

### DIFF
--- a/medusa/server/api/v2/auth.py
+++ b/medusa/server/api/v2/auth.py
@@ -17,7 +17,7 @@ class AuthHandler(BaseRequestHandler):
     def set_default_headers(self):
         """Set default CORS headers."""
         super(AuthHandler, self).set_default_headers()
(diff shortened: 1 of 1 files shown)


medusa/server/api/v2/auth.py
```diff
@@ -17,7 +17,7 @@ class AuthHandler(BaseRequestHandler):
     def set_default_headers(self):
         """Set default CORS headers."""
         super(AuthHandler, self).set_default_headers()
-        self.set_header('X-Medusa-Server', app.APP_VERSION)
+        self.set_header('X-Medusa-Server', app.APP_VERSION if app.APP_VERSION else 'not available')
```
ratoaq2: You should only add the header if there's a version.

PR #2474, opened 2017-03-26
Move all_possible_names to Series object

fernandog:
Just moved code to the series object


labrys:
@fernandog Can we wait a little bit on changes like this to give some others a chance to review and provide feedback.

For example:
L946 in generic.py `show.get_all_possible_names` is either missing its parentheses or is a property.  If its a property then its improperly named as `get` is an action.  Instead it should have been something like `show.all_names'.  (imo it should be a property)

Im guessing you missed the parentheses because in generic_provider `episode.show.get_all_possible_names` is called as a function.  Also thats a really awkward method of getting to the information.  Instead a pattern like this would be more clear:

```python
# get the series from the episode
series = episode.show

# iterate through the names
for name in series.names:
    ...

# iterate through season names
for name in series.season_names:
    ...

# iterate through scene season names
for name in series.scene_season_names:
    ...
``

PR #2478, opened 2017-03-27
Fix dvd order log message

fernandog:
@medariox

### DIFF
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -270,9 +270,9 @@ class Tmdb(BaseIndexer):
             else:
                 seasnum, epno = cur_ep.get('seasonnumber'), cur_ep.get('episodenumber')
                 if self.config['dvdorder']:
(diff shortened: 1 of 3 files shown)


medusa/indexers/tmdb/tmdb.py
```diff
@@ -270,9 +270,9 @@ def _get_episodes(self, tmdb_id, specials=False, aired_season=None):  # pylint:
             else:
                 seasnum, epno = cur_ep.get('seasonnumber'), cur_ep.get('episodenumber')
                 if self.config['dvdorder']:
-                    logger.warning('Episode doest not have DVD ordering available (season: %s, episode: %s). '
+                    logger.warning("Episode doesn't have DVD order available (season: %s, episode: %s). "
```
OmgImAlexis: Should you not be using single quotes and just escape it?

fernandog: its an alternative yes. but its hard to find the log message while debugging, so i prefer do this

p0psicles: PEP8 conventions prefer alternating to single quotes then to escape.

PR #2479, opened 2017-03-27
Warn user that rename files will broken symlinks

fernandog:
image

### DIFF
--- a/views/testRename.mako
+++ b/views/testRename.mako
@@ -16,7 +16,7 @@
     % if app.PROCESS_METHOD == 'symlink':
 <div class="text-center">
 <div class="alert alert-danger upgrade-notification hidden-print" role="alert">
-    <span>WARNING: Your process method is SYMLINK. If you rename files, links will broken and you won't be able to seed</span>
+    <span>WARNING: Your current process method is SYMLINK. Renaming these files will break all symlinks in your seeding directory for this show</span>
 </div>
 </div>
     % endif
--- a/views/testRename.mako
+++ b/views/testRename.mako
@@ -16,7 +16,7 @@
     % if app.PROCESS_METHOD == 'symlink':
 <div class="text-center">
 <div class="alert alert-danger upgrade-notification hidden-print" role="alert">
-    <span>WARNING: Your current process method is SYMLINK. Renaming these files will break all symlinks in your seeding directory for this show</span>
+    <span>WARNING: Your current process method is SYMLINK. Renaming these files will break all symlinks created in Post-Processor for this show</span>
 </div>
 </div>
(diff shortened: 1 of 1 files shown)


fernandog:
@OmgImAlexis 

One of the ideas is to loop in all media/links files in PP/Seed folder and check link target is the same as the file we are renaming, if so, `unlink` and create link again.

My only concern is if user does a mass rename... or rename a show with like 20 seasons
It will be too heavy right?

ping @pymedusa/developers. any of you have another idea to not break links when renaming files already in library?

OmgImAlexis:
I'd go for the creating new links instead of just a warning.

Also it should read something like `WARNING: Your current process method is SYMLINK. Renaming these files will break all symlinks in your seeding directory for this show.`

fernandog:
@OmgImAlexis yes, read my second message in this PR. Above your comment

OmgImAlexis:
If you think about it creating symlinks isn't that resource intensive and really doesn't take all that long. For example even if it takes 1 second per item on a show like "Doctor Who (2005)" that'd only take about 4 minutes and I can't see it taking that long when it doesn't have to move any files around, just re-create symlinks.

views/testRename.mako
```diff
@@ -13,6 +13,13 @@
 <script type="text/javascript" src="js/test-rename.js"></script>
 </%block>
 <%block name="content">
+    % if app.PROCESS_METHOD == 'symlink':
+<div class="text-center">
+<div class="alert alert-danger upgrade-notification hidden-print" role="alert">
+    <span>WARNING: Your current process method is SYMLINK. Renaming these files will break all symlinks in your seeding directory for this show</span>
```
duramato: Don't put "seeding directory" a user can have symlinks for nzb.

fernandog: Isnt a seeding directory?

fernandog: Got it.

So what is your suggestion?

views/testRename.mako
```diff
@@ -16,7 +16,7 @@
     % if app.PROCESS_METHOD == 'symlink':
 <div class="text-center">
 <div class="alert alert-danger upgrade-notification hidden-print" role="alert">
-    <span>WARNING: Your current process method is SYMLINK. Renaming these files will break all symlinks in your seeding directory for this show</span>
+    <span>WARNING: Your current process method is SYMLINK. Renaming these files will break all symlinks created in Post-Processor for this show</span>
```
OmgImAlexis: @fernandog this doesn't really work either. It should state Seeding or post processor or something else maybe as this makes it look like it won't affect the seeding dir.

PR #2481, opened 2017-03-27
Improve Traktor exception handler

fernandog:
@p0psicles now I'm getting the correct message:

2017-03-27 18:05:10 DEBUG    TRAKTCHECKER :: [99dae51] Unable to retrieve shows from Trakt collection. Error: Trakt may have some issues and it's unavailable. Try again later please

### DIFF
--- a/lib/traktor/__init__.py
+++ b/lib/traktor/__init__.py
@@ -17,7 +17,10 @@
 
 from __future__ import absolute_import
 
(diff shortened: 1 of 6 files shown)


fernandog:
should we log the in the `except` if we already raises a custom exception with a message?

p0psicles:
Ik would say no need to log the same message twice.

fernandog:
@p0psicles finished this PR. when you have time can you check it? thanks

Issue #1432, opened 2016-11-01
[APP SUBMITTED]: IOError: [Errno 28] No space left on device


duramato:
### INFO
**Python Version**: `2.7.6 (default, Jun 22 2015, 18:00:18) [GCC 4.8.2]`
**Operating System**: `Linux-4.2.0-42-generic-i686-with-Ubuntu-14.04-trusty`
**Locale**: `None`
**Branch**: [develop](../tree/develop)
**Database**: `44.2`
**Commit**: pymedusa/Medusa@1d9e07aa7e8f1ae0bdc3f8cc8d0dc798b93e1b74
**Link to Log**: https://gist.github.com/4948fc9420a8b5fd93a152c75cf7a51c
### ERROR
<pre>
2016-11-01 15:42:23 ERROR    POSTPROCESSOR :: [1d9e07a] Unable to write file to /media/extssd/Shows/Gotham/Season 03/Gotham.S03E07.Mad.City.Red.Queen.1080p.HDTV.x264-DIMENSION.nfo - are you sure the folder is writable? error 28 : No space left on device
Traceback (most recent call last):
  File "/home/home/SickRage/<a href="../blob/1d9e07aa7e8f1ae0bdc3f8cc8d0dc798b93e1b74/medusa/metadata/generic.py#L450">medusa/metadata/generic.py</a>", line 450, in write_ep_file
    data.write(nfo_file, encoding='UTF-8')
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 820, in write
    serialize(write, self._**********, encoding, qnames, namespaces)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 939, in _serialize_xml
    _serialize_xml(write, e, encoding, qnames, None)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 939, in _serialize_xml
    _serialize_xml(write, e, encoding, qnames, None)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 944, in _serialize_xml
    write(_escape_cdata(elem.tail, encoding))
IOError: [Errno 28] No space left on device
</pre>

---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

duramato:
@ratoaq2 shouldn't this be a thing of the past now? lol

ratoaq2:
Fixed by #1437

duramato:
### INFO
[...]

Issue #1219, opened 2016-10-03
SSLError: ('The read operation timed out',)


fernandog:
@ratoaq2 is there something we can do about this?

```
2016-10-03 01:30:30 WARNING  FINDSUBTITLES :: [c750397] Unexpected error in provider 'opensubtitles'
Traceback (most recent call last):
  File "/home/osmc/SickRage/lib/subliminal/core.py", line 118, in list_subtitles_provider
    return self[provider].list_subtitles(video, provider_languages)
  File "/home/osmc/SickRage/lib/subliminal/providers/opensubtitles.py", line 265, in list_subtitles
    query=query, season=season, episode=episode, tag=os.path.basename(video.name))
  File "/home/osmc/SickRage/lib/subliminal/providers/opensubtitles.py", line 180, in f_retry
    return f(*args, **kwargs)
  File "/home/osmc/SickRage/lib/subliminal/providers/opensubtitles.py", line 216, in query
    response = checked(self.server.SearchSubtitles(self.token, criteria))
  File "/usr/lib/python2.7/xmlrpclib.py", line 1233, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1591, in __request
    verbose=self.__verbose
  File "/usr/lib/python2.7/xmlrpclib.py", line 1273, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1303, in single_request
    response = h.getresponse(buffering=True)
  File "/usr/lib/python2.7/httplib.py", line 1111, in getresponse
    response.begin()
  File "/usr/lib/python2.7/httplib.py", line 444, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python2.7/httplib.py", line 400, in _read_status
    line = self.fp.readline(_MAXLINE + 1)
  File "/usr/lib/python2.7/socket.py", line 476, in readline
    data = self._sock.recv(self._rbufsize)
  File "/usr/lib/python2.7/ssl.py", line 714, in recv
    return self.read(buflen)
  File "/usr/lib/python2.7/ssl.py", line 608, in read
    v = self._sslobj.read(len or 1024)
SSLError: ('The read operation timed out',)
```

duramato:
### INFO
**Python Version**: `2.7.9 (default, Sep 17 2016, 20:26:04) [GCC 4.9.2]`
[...]

PR #2486, opened 2017-03-28
Subtitle broad exception remove

fernandog:
@ratoaq2 let it crash so we know what errors could happen and where

as you commented in here: https://github.com/pymedusa/Medusa/issues/1219#issuecomment-282399236

and here: https://github.com/pymedusa/Medusa/issues/1432

### DIFF
--- a/medusa/subtitles.py
+++ b/medusa/subtitles.py
@@ -774,7 +774,7 @@ def delete_unwanted_subtitles(dirpath, filename):
         return
 
     code = filename.rsplit('.', 2)[1].lower().replace('_', '-')
-    language = from_code(code, unknown='') or from_ietf_code(code, unknown='und')
+    language = from_code(code, unknown='') or from_ietf_code(code)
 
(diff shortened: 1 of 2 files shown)


medusa/subtitles.py
```diff
@@ -439,12 +439,6 @@ def download_subtitles(tv_episode, video_path=None, subtitles=True, embedded_sub
     except IOError as error:
         if 'No space left on device' in ex(error):
             logger.warning(u'Not enough space on the drive to save subtitles')
-        else:
-            logger.warning(traceback.format_exc())
-    except Exception as error:
-        logger.debug(u'Exception: %s', error)
-        logger.info(u'Error occurred when downloading subtitles for: %s', video_path)
-        logger.error(traceback.format_exc())
```
fernandog: @ratoaq2 maybe we can remove the IOError too?

ratoaq2: You should remove it. The current way you're just swalling the IOError exception without logging it.

fernandog: removed

PR #2492, opened 2017-03-29
Minor Propers fix

fernandog:
- Line too long
- Broad catch
- Referenced before assignement


medusa/search/proper.py
```diff
@@ -421,7 +427,7 @@ def _get_last_proper_search():
 
         try:
             last_proper_search = datetime.date.fromordinal(int(sql_results[0][b'last_proper_search']))
```
p0psicles: How should this raise an exception?

fernandog: int() - ValueError
sql_results[0] - Index Error
sql_results[0][b'last_proper_search'] - KeyError

Issue #2427, opened 2017-03-18
add recomended shows imdbs popular


ripper1967:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:**  current dev
**OS:**Windows 10
**What you did:**
**What happened:**
**What you expected:**
**Logs:**
```
PASTE LOGS HERE
```
![add recomebded shows](https://cloud.githubusercontent.com/assets/7034303/24073979/fdbfe9ec-0bd6-11e7-93dc-ba15cc7e71d7.png)

fernandog:
@duramato can you help here?

recommendations\imdb.py:
```python
        soup = BeautifulSoup(response.text, 'html5lib')
        results = soup.find('div', class_='lister-list')
        rows = results.find_all('div', class_='lister-item mode-advanced')
```

results is None causing the issue. Maybe class got renamed or else

https://gist.github.com/fernandog/6e5a9dc901d23464dd4e5478986449ce

duramato:
I can try and make this use  imdbpie

fernandog:
@ripper1967 please update

PR #2497, opened 2017-03-31
Use imdb pie to get popular shows

fernandog:
Fixes:
https://github.com/pymedusa/Medusa/issues/2427

@duramato beat you lol

Current problem is: for each show we find in imdb, we hit tvdb and tvmaze to get tvdb.
popular shows returns 50 shows. So, do the maths


medusa/server/web/home/add_shows.py
```diff
@@ -7,6 +7,8 @@
 import os
 import re
 
+import requests
```
p0psicles: No need to import requests, if you only need RequestException.

medusa/server/web/home/add_shows.py
```diff
@@ -308,7 +310,7 @@ def popularShows(self):
 
         try:
             recommended_shows = ImdbPopular().fetch_popular_shows()
-        except (StandardError, requests.RequestException) as e:
+        except (RequestException, socket.gaierror, StandardError) as e:
```
p0psicles: why do you need socket.gaierror?

fernandog: @p0psicles because of this errors: https://github.com/pymedusa/Medusa/issues/1980

p0psicles: Hmm that's raised by pyNMA. What does that have to do with fetching the imdb shows? I think that requests already handles that. Reason why pyNMA raises that, is because they directly use httplib.

I Don't see a good reason to start handling that exception everywhere??

fernandog: @p0psicles that can happen to any connection for any URL.

fernandog: if you read the issue 1980, you will see that request doesn't handle it

p0psicles: Show we the error where it's raised by requests? I'm not interested if RequestsExceptions handles it or not. I know it won't handle that. I'm interested if it can be raised by a request.session.get()..

p0psicles: As explained. The requests lib, swallows the socket.gaierror through `except socket.error`.
https://github.com/kennethreitz/requests/blob/master/requests/adapters.py#L485

p0psicles:
Then we should only hit the indexer_mapping for already known mappings.

PR #2506, opened 2017-03-31
Standardize logging

labrys:

### DIFF
--- a/medusa/name_parser/parser.py
+++ b/medusa/name_parser/parser.py
@@ -146,10 +146,10 @@ class NameParser(object):
 
             for episode_number in episode_numbers:
                 s = season_number
-                error = episode_number
+                episode = episode_number
 
                 if result.show.is_scene:
-                    (s, error) = scene_numbering.get_indexer_numbering(
+                    (s, episode) = scene_numbering.get_indexer_numbering(
                         result.show.indexerid,
                         result.show.indexer,
                         season_number,
@@ -157,9 +157,9 @@ class NameParser(object):
                     )
                     log.debug(
                         'Scene numbering enabled series {name}, using indexer numbering: {ep}',
-                        {'name': result.show.name, 'ep': episode_num(s, error)}
+                        {'name': result.show.name, 'ep': episode_num(s, episode)}
                     )
-                new_episode_numbers.append(error)
+                new_episode_numbers.append(episode)
                 new_season_numbers.append(s)
 
         elif result.show.is_anime and result.is_anime:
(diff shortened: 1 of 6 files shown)


medusa/name_parser/parser.py
```diff
@@ -117,35 +119,52 @@ def _parse_string(self, name):
...
+                    log.warning(
+                        'Indexer exception: {indexer_api.name}: {error}',
+                        {'indexer_api': indexer_api, 'error': error.message}
+                    )
                     episode_numbers = []
 
             for episode_number in episode_numbers:
                 s = season_number
-                e = episode_number
+                error = episode_number
```
ratoaq2: Why error?

medusa/name_parser/parser.py
```diff
@@ -117,35 +119,52 @@ def _parse_string(self, name):
...
                     episode_numbers = []
-                except IndexerError as e:
-                    logger.warning('Unable to contact {indexer_api.name}: {e}', indexer_api=indexer_api, e=e.message)
+                except IndexerError as error:
+                    log.warning(
+                        'Unable to contact {indexer_api.name}: {error}',
+                        {'indexer_api': indexer_api, 'error': error.message}
+                    )
                     episode_numbers = []
-                except IndexerException as e:
```
fernandog: are we changing default `as e`  as we agreed in the past?

labrys: We agreed in the past that single letter variables like `e` should be avoided.

medusa/name_parser/parser.py
labrys: messed up refactor... was refactoring `Exception as e` and somehow that got changed too

medusa/name_parser/parser.py
fernandog: where is documented?

medusa/name_parser/parser.py
labrys: `e = episode_number` overwrites the existing `e` in the scope from `Exception as e` so auto-refactor refactored both

medusa/name_parser/parser.py
labrys: @fernandog my second comment to @ratoaq2 is exactly why names like `e` are bad...  its too easy to end up clobbering an existing variable and its not descriptive of the content... just like `e` for episode is bad too.

medusa/name_parser/parser.py
```diff
@@ -156,53 +175,59 @@ def _parse_string(self, name):
...
-                new_episode_numbers.append(e)
-                new_season_numbers.append(s)
+                new_episode_numbers.append(episode)
+                new_season_numbers.append(season)
 
         # need to do a quick sanity check heregex.  It's possible that we now have episodes
         # from more than one season (by tvdb numbering), and this is just too much
         # for the application, so we'd need to flag it.
-        new_season_numbers = list(set(new_season_numbers))  # remove duplicates
+        new_season_numbers = sorted(set(new_season_numbers))  # remove duplicates
```
p0psicles: Okay I checked this. Was concerned at first but it returns a list. So seems safe.

medusa/name_parser/parser.py
```diff
@@ -264,11 +293,13 @@ def assert_supported(result):
         :type result: ParseResult
         """
         if not result.show:
-            raise InvalidShowException('Unable to match {result.original_name} to a show in your database. '
+            raise InvalidShowException('Unable to match {result.original_name} to a series in your database. '
                                        'Parser result: {result}'.format(result=result))
```
p0psicles: Nice

medusa/name_parser/rules/rules.py
```diff
@@ -1220,8 +1222,8 @@ def when(self, matches, context):
                     to_remove.extend(matches.named('title', predicate=lambda match: match.value != values[0].value))
                     continue
 
-                logger.info(u"Guessed more than one '%s' for '%s': %s",
-                            name, matches.input_string, u','.join(unique_values), exc_info=False)
+                log.info(u"Guessed more than one '%s' for '%s': %s",
```
p0psicles: Let's use {0} for Positional argument substitution for Medusa code okay? The we limit the amount of log patterns.

labrys: @p0psicles I left this unchanged intentionally, as there was only a single log line I figured that importing the `BraceAdapter` may not be necessary.  Thoughts?

p0psicles: Okay I can see that.

PR #2507, opened 2017-03-31
Standardize logging

labrys:

### DIFF
--- a/medusa/metadata/wdtv.py
+++ b/medusa/metadata/wdtv.py
@@ -1,35 +1,23 @@
 # coding=utf-8
-# Author: Nic Wolfe <nic@wolfeden.ca>
-#
-# This file is part of Medusa.
-#
-# Medusa is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# Medusa is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with Medusa. If not, see <http://www.gnu.org/licenses/>.
 
 import datetime
+import logging
 import os
 import re
 
-from .. import helpers, logger
-from ..helper.common import dateFormat, episode_num as ep_num, replace_extension
-from ..indexers.indexer_api import indexerApi
-from ..indexers.indexer_exceptions import IndexerEpisodeNotFound, IndexerSeasonNotFound
-from ..metadata import generic
+from medusa import helpers
+from medusa.helper.common import dateFormat, episode_num as ep_num, replace_extension
+from medusa.indexers.indexer_api import indexerApi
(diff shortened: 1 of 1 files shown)


medusa/metadata/wdtv.py
```diff
@@ -162,12 +150,11 @@ def get_season_poster_path(show_obj, season):
...
         if not season_dir:
-            logger.log(u'Unable to find a season directory for season {season_num}'.format
-                       (season_num=season), logger.DEBUG)
+            log.debug(u'Unable to find a season directory for season {0}', season)
             return None
 
-        logger.log(u'Using {path}/folder.jpg as season dir for season {season_num}'.format
-                   (path=season_dir, season_num=season), logger.DEBUG)
+        log.debug(u'Using {path}/folder.jpg as season dir for season {x}',
+                  {'path': season_dir, 'x': season})
```
medariox: Same as other PRs, try to avoid x.

medusa/metadata/wdtv.py
```diff
@@ -193,10 +180,12 @@ def _ep_data(self, ep_obj):
             try:
                 my_ep = my_show[ep_to_write.season][ep_to_write.episode]
             except (IndexerEpisodeNotFound, IndexerSeasonNotFound):
-                logger.log(u'Unable to find episode {ep_num} on {indexer}... '
-                           u'has it been removed? Should I delete from db?'.format
-                           (ep_num=ep_num(ep_to_write.season, ep_to_write.episode),
-                            indexer=indexerApi(ep_obj.show.indexer).name))
+                log.info(
+                    u'Unable to find episode {x} on {indexer}... has it been removed? Should I delete from db?', {
+                        'x': ep_num(ep_to_write.season, ep_to_write.episode),
```
medariox: Here too.

medusa/metadata/wdtv.py
labrys: I chose `x` for things like that because its a common convention to represent a number as `x` and as `x` isnt actually a variable it doesnt pollute the local namespace with one-letter variables

also imo its clearer to read then `with {number} seeders` or `with {seeders} seeders` or in this case `for season {season}`

as proof that its not actually a variable try this:  `'This is a {variable test}'.format(**{'variable test': 'success!'})`
it works even though `variable test` isnt a valid variable due to the space

medariox: I would agree with you if it was being used in a context where we want to represent a mathematical variable. In that case, ```x``` would fit perfectly. In this context, I don't think it's the best fit as we are just representing a number. I also think it fits better to the example you made. Calling it ```number``` makes most sense to me and imo it's also clearer:
```
for season {number}
```
```
log.debug(u'Using {path}/folder.jpg as season dir for season {number}'
```
In the end it's mostly a metter of personal preference in this case, which is why I didn't request changes.

labrys: My main reasoning was for locations like the seeders example where you can't use `number` since there's more than one number involved:

Compare:
'Downloading {name} with {seeders} seeders and {leechers} leechers from {provider}'
'Downloading {name} with {x} seeders and {y} leechers from {provider}'

I think anyone reading that without any other context would understand that `x` is the number of seeders and `y` is the number of leechers.

medariox: In that case ```x``` and ```y``` make much more sense indeed. I think we should decide depending on the context.

Issue #1975, opened 2017-01-14
Changing scene episode numbers after a show refresh, will keep reusing "wrong" guessit cached name parsing


p0psicles:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:**
**OS:**
**What you did:**
* Refreshed a show dir with the show Ultimate Spider-Man season 3.
* noticed that xem is wrong and is mapping the episode 18 -> indexer ep 15.
* fixed the scene column episode numbers, changing 18 back to 3x15.
* did a rescan
**What happened:**
My downloaded episode 18 is still mapped to episode 15.
**What you expected:**
It to be put in the row of episode 18.
**Logs:**
Not really needed as I've already found the issue. Which we just need to discuss and find a possible solution for.
**Workaround:**
Restart medusa, and refresh.

The reason why it keeps parsing the ep 18 to ep 15, is because the parse result is cached. It was cached when the scene column was incorrectly configured by xem. Rebooting empties the cache, and makes it parse the episode using the new manually configured episode mapping.

A possible solution could be:
1. Keeping track of the timestamp parse cached results.
2. Then check keeping track of changes made to the episode in tv_episodes table. As when you override the scene episode column, changes are made to that table. We would need to add a column to track the timestamp.
3. When a cached parse is older then the newest change of one of the episodes for that season, re-parse it. It needs to be a little complex, as we don't know if we can trust the episode of the cached parse. Season would most likely be ok.

Other solution:
1. Empty parse cache every 24 hours.

fernandog:
forcing show update doesn't invalidate cache?

p0psicles:
No it doesn't. Show updater, doesn't do anything with the name cache.

fernandog:
@ratoaq2 can we call something like this to remove cached parsed results  when user changes scene numbering?

```python
    def remove(self, indexer, indexerid):
        for cached_name, cached_parsed_result in self.cache.items():
            if cached_parsed_result.show.indexer == indexer and cached_parsed_result.show.indexerid == indexerid:
                self.cache.pop(cached_name)
                logger.debug("Removing cached parse result for show ID: '{id}'", id=indexerid)
```

ratoaq2:
That can be a bit inefficient, right? How many items usually there are in the cache?

fernandog:
no idea. how to make it efficient?
the main goal here is to invalidate/remove "outdated" cached itesm

[...]

PR #2508, opened 2017-03-31
Fix reusing "wrong" guessit cached name parsing after new numbering

fernandog:
Fixes: https://github.com/pymedusa/Medusa/issues/1975

@p0psicles @duramato @ratoaq2

```
[b4a8a54] Removed parsed cached result for release: Bull.2016.S01E17.1080p.HDTV.X264-DIMENSION[rartv]
[b4a8a54] Removed parsed cached result for release: Bull.S01E17.REPACK.720p.HDTV.x264-AVS[rartv]
[b4a8a54] Removed parsed cached result for release: Bull.S01E17.HDTV.x264-SVA[rartv]
```

### DIFF
--- a/medusa/name_parser/parser.py
+++ b/medusa/name_parser/parser.py
@@ -500,7 +500,7 @@ class NameParserCache(object):
         """Remove cache item given indexer and indexer_id."""
         if not indexer or not indexer_id:
             return
-        to_remove = (cached_name for cached_name, cached_parsed_result in self.cache.items() if
+        to_remove = (cached_name for cached_name, cached_parsed_result in self.cache.iteritems() if
                      cached_parsed_result.show.indexer == indexer and cached_parsed_result.show.indexerid == indexer_id)
         for item in to_remove:
             self.cache.popitem(item)
--- a/medusa/name_parser/parser.py
+++ b/medusa/name_parser/parser.py
@@ -26,6 +26,8 @@ from medusa.indexers.indexer_exceptions import (
 )
 from medusa.logger.adapters.style import BraceAdapter
 
+from six import iteritems
+
 
 log = BraceAdapter(logging.getLogger(__name__))
 log.logger.addHandler(logging.NullHandler())
@@ -500,7 +502,7 @@ class NameParserCache(object):
         """Remove cache item given indexer and indexer_id."""
         if not indexer or not indexer_id:
             return
-        to_remove = (cached_name for cached_name, cached_parsed_result in self.cache.iteritems() if
+        to_remove = (cached_name for cached_name, cached_parsed_result in iteritems(self.cache) if
                      cached_parsed_result.show.indexer == indexer and cached_parsed_result.show.indexerid == indexer_id)
         for item in to_remove:
             self.cache.popitem(item)
--- a/medusa/name_parser/parser.py
+++ b/medusa/name_parser/parser.py
@@ -504,7 +504,7 @@ class NameParserCache(object):
                      cached_parsed_result.show.indexer == indexer and cached_parsed_result.show.indexerid == indexer_id)
         for item in to_remove:
             self.cache.popitem(item)
-            logger.debug('Removed parsed cached result for release: {release}'.format(release=item))
+            log.debug('Removed parsed cached result for release: {release}'.format(release=item))
 
 
 name_parser_cache = NameParserCache()
--- a/medusa/name_parser/parser.py
+++ b/medusa/name_parser/parser.py
@@ -252,6 +252,10 @@ class NameParser(object):
 
         return result
 
+    def unparse(self, indexer, indexer_id):
+        """Unparse all names from given indexer and indexer_id."""
+        name_parser_cache.remove(indexer, indexer_id)
+
     def parse(self, name, cache_result=True):
         """Parse the name into a ParseResult.
 
@@ -490,6 +494,16 @@ class NameParserCache(object):
         if name in self.cache:
             log.debug('Using cached parse result for {name}', {'name': name})
             return self.cache[name]
+
+    def remove(self, indexer, indexer_id):
+        """Remove cache item given indexer and indexer_id."""
+        if not indexer or not indexer_id:
+            return
+        to_remove = (cached_name for cached_name, cached_parsed_result in self.cache.items() if
+                     cached_parsed_result.show.indexer == indexer and cached_parsed_result.show.indexerid == indexer_id)
+        for item in to_remove:
+            self.cache.popitem(item)
+            logger.debug('Removed parsed cached result for release: {release}'.format(release=item))
 
 
 name_parser_cache = NameParserCache()
--- a/medusa/name_parser/parser.py
+++ b/medusa/name_parser/parser.py
@@ -252,8 +252,9 @@ class NameParser(object):
 
         return result
 
-    def unparse(self, indexer, indexer_id):
-        """Unparse all names from given indexer and indexer_id."""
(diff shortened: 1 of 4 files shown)


medusa/server/web/home/handler.py
```diff
@@ -1571,6 +1571,14 @@ def editShow(self, show=None, location=None, allowed_qualities=None, preferred_q
                     logger.log("Unable to refresh show '{show}': {error}".format
                                (show=show_obj.name, error=e.message), logger.WARNING)
 
+            # Check if we should erase parsed cached results f or that show
+            do_erase_parsed_cache = False
+            for item in ['scene', 'anime', 'sports', 'air_by_date', 'dvd_order']:
+                if getattr(show_obj, item) != ast.literal_eval(item):
```
fernandog: @ratoaq2  @p0psicles getting `ValueError: malformed string` , do you know why?

fernandog:
@duramato now its ok, i did a quick test here

can you confirm?

medusa/name_parser/parser.py
```diff
@@ -224,6 +224,10 @@ def _parse_string(self, name):
 
         return result
 
+    def unparse(self, indexer, indexer_id):
```
p0psicles: 'Unparse' doesn't make any sense. Better to just use clear_parsed_name_cache(). Or something similar.

medusa/server/web/home/handler.py
```diff
@@ -1571,6 +1571,14 @@ def editShow(self, show=None, location=None, allowed_qualities=None, preferred_q
                     logger.log("Unable to refresh show '{show}': {error}".format
                                (show=show_obj.name, error=e.message), logger.WARNING)
 
+            # Check if we should erase parsed cached results f or that show
+            do_erase_parsed_cache = False
+            for item in ['scene', 'anime', 'sports', 'air_by_date', 'dvd_order']:
+                if getattr(show_obj, item) != ast.parse(item):
```
p0psicles: Can you explain me what your doing here with ast.parse()?
I've tried it, and it's emptying cache always.

medusa/server/web/home/handler.py
```diff
@@ -1652,9 +1660,21 @@ def editShow(self, show=None, location=None, allowed_qualities=None, preferred_q
...
+            # Erase parsed cached names as we are changing scene numbering
+            show_obj.flush_episodes()
+            show_obj.erase_cached_parser()
+
+            # Need to refresh show as we updated scene numbering or changed show format
+            try:
+                app.show_queue_scheduler.action.refreshShow(show_obj)
+            except CantRefreshShowException as e:
+                errors += 1
+                logger.log("Unable to refresh show '{show}'. You must manually refresh!. Error: {error}".format
```
p0psicles: Can we change this into a little bit more subtle message?

medusa/server/web/home/handler.py
```diff
@@ -1652,9 +1660,21 @@ def editShow(self, show=None, location=None, allowed_qualities=None, preferred_q
                 logger.log("Unable to force an update on scene numbering for show '{show}': {error}".format
                            (show=show_obj.name, error=e.message), logger.WARNING)
 
-            # Must erase cached results when toggling scene numbering
+            # Must erase cached DB results when toggling scene numbering
             self.erase_cache(show_obj)
 
+            # Erase parsed cached names as we are changing scene numbering
+            show_obj.flush_episodes()
```
p0psicles: Why are we clearing out the episode objects for the show object here?

medusa/tv/series.py
```diff
@@ -335,6 +335,10 @@ def flush_episodes(self):
                 self.episodes[cur_season][cur_ep] = None
                 del my_ep
 
+    def erase_cached_parser(self):
+        """Erase parsed results because user changed xem numbering."""
```
p0psicles: I don't think this log message is correct anymore. As we're erasing it for multiple reasons now right?

fernandog: Yes. Right

medusa/server/web/home/handler.py
fernandog: @p0psicles this checks

 if show_obj.scene != scene

So if user change any those vars, i set tje flag to clean

medusa/name_parser/parser.py
fernandog: Ok

medusa/server/web/home/handler.py
fernandog: Because user change the show format. So all loaded show needs to be updated again.

medusa/server/web/home/handler.py
fernandog: Like how?

medusa/server/web/home/handler.py
p0psicles: Okay. But it's not working

medusa/server/web/home/handler.py
fernandog: Like here: https://github.com/pymedusa/Medusa/blob/develop/medusa/scene_numbering.py#L225

fernandog:
@p0psicles already working. just need your reply in my comments to your review

fernandog:
@duramato did you test this?

medusa/name_parser/parser.py
```diff
@@ -461,6 +466,16 @@ def get(self, name):
             logger.debug("Using cached parse result for '{name}'", name=name)
             return self.cache[name]
 
+    def remove(self, indexer, indexer_id):
+        """Remove cache item given indexer and indexer_id."""
+        if not indexer or not indexer_id:
+            return
+        to_remove = (cached_name for cached_name, cached_parsed_result in self.cache.items() if
```
ratoaq2: It's still looping through all entries. Probably the data structure of the cache should be different where you could access all items that relates to a specific indexer / indexer id

fernandog: need to loop as there is no other way. I can't `pop` a item without knowing the name of it
Also I'm using generator

I don't see why is bad to loop. 
How often a user change scene numbering? once a week or few days?
This is a task that will only run when user does some action. Not a scheduled action

medariox: .items() returns a list, so the generator is a bit useless here. Use .iteritems() instead.

fernandog: @ratoaq2 how to proceed about this? use as I proposed or do you want to change self.cache structure?

fernandog: @ratoaq2 what do you think?

fernandog:
change to 0.1.13 as I'm still waiting @ratoaq2 feedback

medusa/server/web/home/handler.py
p0psicles: `Please manually trigger a full show refresh`

medusa/name_parser/parser.py
```diff
@@ -500,7 +500,7 @@ def remove(self, indexer, indexer_id):
         """Remove cache item given indexer and indexer_id."""
         if not indexer or not indexer_id:
             return
-        to_remove = (cached_name for cached_name, cached_parsed_result in self.cache.items() if
+        to_remove = (cached_name for cached_name, cached_parsed_result in self.cache.iteritems() if
```
fernandog: @medariox just that?

p0psicles: Nah, best way of using iteritems, is using it from six.
```
from six import iteritems
to_remove = (cached_name for cached_name, cached_parsed_result in iteritems(self.cache) if ..
```

PR #2511, opened 2017-04-01
Standardize logging

labrys:


medusa/search/core.py
```diff
@@ -163,10 +152,10 @@ def snatch_episode(result):
                 client = torrent.get_client_class(app.TORRENT_METHOD)()
                 result_downloaded = client.send_torrent(result)
             else:
-                logger.log(u"Torrent file content is empty", logger.WARNING)
+                log.warning(u'Torrent file content is empty')
                 result_downloaded = False
     else:
-        logger.log(u"Unknown result type, unable to download it (%r)" % result.resultType, logger.ERROR)
+        log.error(u'Unknown result type, unable to download it: {!r}', result.resultType)
```
p0psicles: Don't yo also need a positional argument here? {0!r} ?

medusa/search/core.py
```diff
@@ -253,7 +245,7 @@ def pick_best_result(results, show):  # pylint: disable=too-many-branches
     """
     results = results if isinstance(results, list) else [results]
 
-    logger.log(u'Picking the best result out of ' + str([x.name for x in results]), logger.DEBUG)
+    log.debug(u'Picking the best result out of {0}', [x.name for x in results])
```
p0psicles: I assume the logger is going to be okay with getting a list as argument?

medusa/search/core.py
```diff
@@ -443,27 +440,27 @@ def search_for_needed_episodes():
...
-            logger.log(u"Authentication error: " + ex(e), logger.ERROR)
+        except AuthException as error:
+            log.error(u'Authentication error: {0}', ex(error))
             continue
-        except Exception as e:
-            logger.log(u"Error while searching " + cur_provider.name + u", skipping: " + ex(e), logger.ERROR)
-            logger.log(traceback.format_exc(), logger.DEBUG)
+        except Exception as error:
+            log.error(u'Error while searching {0}, skipping: {1}', cur_provider.name, ex(error))
+            log.debug(traceback.format_exc())
```
p0psicles: I think it would be nicer to combine these two to: log.exception().

medusa/search/core.py
```diff
@@ -540,44 +538,44 @@ def search_providers(show, episodes, forced_search=False, down_cur_quality=False
...
-                               (cur_provider.name, ex(e)), logger.WARNING)
+            except Exception as error:
+                if 'ECONNRESET' in error or (hasattr(error, 'errno') and error.errno == errno.ECONNRESET):
+                    log.warning(u'Connection reseted by peer while searching {0}. Error: {1!r}',
+                                cur_provider.name, ex(error))
                 else:
-                    logger.log(u"Unknown exception while searching %s. Error: %r" %
-                               (cur_provider.name, ex(e)), logger.ERROR)
-                    logger.log(traceback.format_exc(), logger.DEBUG)
+                    log.error(u'Unknown exception while searching {0}. Error: {1!r}',
```
p0psicles: Same here. Use log.exception().

medusa/search/core.py
labrys: The index is optional, but our policy is to include it, nice catch.

medusa/search/core.py
labrys: Yes it will print the repr as normal.  Only a `collections.Mapping` descendant (like `dict`) as a single param require special treatment because it will try to unpack them first as kwargs.

medusa/search/core.py
labrys: I think the intent here was to only show a traceback for debug purposes but always show the error.

fernandog: Yes. The traceback will be in uploaded log when user submit the error

p0psicles:
Can you fix the continuation lines. The  it's okay.

labrys:
Yeah ill fix those now

medusa/search/core.py
```diff
@@ -75,44 +64,44 @@ def _download_result(result):
...
         try:
-            with open(file_name, 'w') as fileOut:
+            with open(file_name, u'w') as fileOut:
                 fileOut.write(result.extraInfo[0])
 
             helpers.chmod_as_parent(file_name)
 
         except EnvironmentError as e:
-            logger.log(u"Error trying to save NZB to black hole: " + ex(e), logger.ERROR)
+            log.error(u'Error trying to save NZB to black hole: {0}', ex(e))
```
ratoaq2: It's there a reason to not drop `ex()` function?

medusa/search/core.py
ratoaq2: And why not calling debug with exc_info?

p0psicles: Then you can also call. Log.exception(). It's the same.

medusa/search/core.py
fernandog: all code i change i remove the ex()

labrys: @ratoaq2 AFAIK there's no reason to keep `ex()`

labrys:
`traceback.format_exc()` presents a formatted version of `exc_info`.  Since we don't need to do any manipulation of the `exc_info`, the `traceback.format_exc()` should be correct.

labrys:
@fernandog I'm not sure what your logic is in the last PR, please explain

PR #2517, opened 2017-04-01
Standardize logging

labrys:

### DIFF
--- a/medusa/tv/cache.py
+++ b/medusa/tv/cache.py
@@ -1,35 +1,22 @@
 # coding=utf-8
-# Author: Nic Wolfe <nic@wolfeden.ca>
-#
-# This file is part of Medusa.
-#
-# Medusa is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# Medusa is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


medusa/tv/cache.py
```diff
@@ -124,9 +106,9 @@ def __init__(self, provider_id):
                     b'    time NUMERIC)'
                 )
         except Exception as error:
-            logger.log('Error while searching {provider_id}, skipping: {e!r}'.
-                       format(provider_id=provider_id, e=error), logger.DEBUG)
-            logger.log(traceback.format_exc(), logger.DEBUG)
+            log.debug('Error while searching {provider_id}, skipping: {error!r}',
+                      {'provider_id': provider_id, 'error': error})
+            log.debug(traceback.format_exc())
```
ratoaq2: Why not using exc_info?

ratoaq2: And combine both debug calls

labrys: tbh there's lots of logs I would have changed, but I was more concerned about changing format than rewriting them all.  You're right tho, this should probably use exc_info and combine.

ratoaq2: If rewrite is not in the scope then it's fine, given the quantity of changes

labrys: Actually thinking about it I would probably stick with `traceback.format_exc()` but I would also still combine them.  `traceback.format_exc()` prints out the `exc_info` formatted and since we have no need to change the formatting (such as stripping linebreaks, etc) its probably the best option.

labrys: Even though rewrite wasn't really the intent, when better to look at these log messages than now when they are being reviewed? :)

PR #2519, opened 2017-04-01
Standardize logging

labrys:


medusa/search/manual.py
```diff
@@ -177,12 +172,13 @@ def collect_episodes_from_search_thread(show):
...
 
-def get_provider_cache_results(indexer, show_all_results=None, perform_search=None, **search_show):  # pylint: disable=too-many-locals,unused-argument
+def get_provider_cache_results(indexer, show_all_results=None, perform_search=None, show=None,
+                               season=None, episode=None, manual_search_type=None, **search_show):
     """Check all provider cache tables for search results."""
-    show = search_show.get('show')
-    season = search_show.get('season')
-    episode = search_show.get('episode')
-    manual_search_type = search_show.get('manual_search_type')
+    show = show
```
ratoaq2: These 4 lines can be removed now

labrys: I'm sorry, which 4 lines? The `search_show` kwargs?

ratoaq2: show = show
episode = episode
Etc

fernandog: If we remove the var.get we need to check if we have all values, if not return

labrys: @fernandog the change to kw args from kwargs using `get` changes nothing as far as that's concerned.

fernandog: @labrys if we call that method without season or show or episode, it will raise a traceback right?

labrys: Not sure if inside the method it will (would have to look at the code) but calling the method itself would not cause a traceback because default values are provided.

fernandog: ok

ratoaq2: My comment is that these four assignments have no effect.

PR #2550, opened 2017-04-01
Fix spelling

labrys:


ripper1967:
Edge produced the Same error, i even tried a working rss feed i know that
is good  http://extratorrent.cc/rss.xml?type=today&cid=8 . i know we have
extratorrent already just wanted to try with a good working rss feed, that
as well created the same error,

PR #2561, opened 2017-04-01
More fix to RSS torrent

fernandog:
@labrys tested with: https://thepiratebay.org/rss/top100/208

as i said, first commit of `bugfix/add-custom-provider` works for me
Proof:

image


fernandog:
@labrys, 

@duramato doesn't have a rss torrent with proxy. it's up to you know

PR #2564, opened 2017-04-04
Improve already processed propers code check and fix unhashable type: 'dict'

fernandog:
Don't need to keep comparing previous with new processed list. Just need to not check it

self.processed_propers i now is a list of dict with dates, so i need to check if proper is already in the list.

Tested:
- 2 normals runs (second run it didn't re-processed the proper)
- 1 normal run then 1 forced run then 1 normal run. In the second run it reprocessed again (forced) and in the last one it didn't because it was already processed in the first run

Also fixes this that happens in a forced proper search
```
  File "/home/osmc/Medusa/medusa/search/proper.py", line 95, in run
    current_processed_propers.extend(set(self.processed_propers).difference(set(current_processed_propers)))
TypeError: unhashable type: 'dict'
```

### DIFF
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -47,6 +47,7 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
         """Initialize the class."""
         self.amActive = False
         self.processed_propers = []
+        self.ignore_processed_propers = False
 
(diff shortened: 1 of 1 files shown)


fernandog:
Fix the force after this PR: https://github.com/pymedusa/Medusa/pull/2468

PR #2567, opened 2017-04-04
Fix add custom provider

fernandog:
@labrys your commits + mine


labrys:
@fernandog You already merged that with #2561

fernandog:
@labrys that PR was against YOUR branch

PR #2569, opened 2017-04-05
Fix log history not using correct action for PROPERs

fernandog:
Old issue since Sickrage

Also fix all user history

This is the proof the query works
```sql
SELECT action, substr(action, 1, length(action) - 2) || '09' as new_action
FROM history
WHERE action LIKE "%02" and proper_tags is not ""
```

ping @labrys

### DIFF
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -20,7 +20,7 @@ MIN_DB_VERSION = 40  # oldest db version we support migrating from
 MAX_DB_VERSION = 44
 
 # Used to check when checking for updates
-CURRENT_MINOR_DB_VERSION = 8
+CURRENT_MINOR_DB_VERSION = 9
 
 
 class MainSanityCheck(db.DBSanityCheck):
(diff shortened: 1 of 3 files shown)


labrys:
We should deprecate the `Snatched Proper` status code `09` for 2 reasons.  First it has never been fully implemented and no current code relies on it.  Secondly we now store proper tags in DB so:
`Snatched + Has Proper Tags == Snatched Proper`

fernandog:
@labrys we use in PP to check if episode status is PROPER and a lot of other places. I wouldn't touch that now as it may break stuff that would need a lot of testing to make sure it works.

labrys:
@fernandog my point is we shouldn't be adding more methods that use `09` status when `SNATCHED` + `PROPER TAGS` is a more reliable check.

fernandog:
in the new DB/quality model we can ditch it. I sew no point removing it now.
as for today, history is wrong. This PR only fixes it

PR #2570, opened 2017-04-05
Fix not showing correct default indexer

fernandog:
@ripper1967 please try this. Worked for me

### DIFF
--- a/views/home_massAddTable.mako
+++ b/views/home_massAddTable.mako
@@ -38,7 +38,7 @@
         <td align="center">
             <select name="indexer">
                 % for curIndexer in indexerApi().indexers.items():
-                    <option value="${curIndexer[0]}" ${('', 'selected="selected"')[curIndexer[0] == indexer] or curIndexer[0] == app.INDEXER_DEFAULT}>${curIndexer[1]}</option>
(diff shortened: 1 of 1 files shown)


fernandog:
Really? a quick and simple fix? waiting approval....

views/home_massAddTable.mako
```diff
@@ -38,7 +38,7 @@
         <td align="center">
             <select name="indexer">
                 % for curIndexer in indexerApi().indexers.items():
-                    <option value="${curIndexer[0]}" ${('selected="selected"', '')[curIndexer[0] == indexer] or curIndexer[0] == app.INDEXER_DEFAULT}>${curIndexer[1]}</option>
+                    <option value="${curIndexer[0]}" ${('', 'selected="selected"')[curIndexer[0] == indexer] or curIndexer[0] == app.INDEXER_DEFAULT}>${curIndexer[1]}</option>
```
labrys: As mentioned in IRC this should be written as `x if condition else y` instead of `(y, x)[condition]`.  It's clearer and less error prone.

labrys:
Even quick fixes can sometimes be improved upon 😃

PR #2581, opened 2017-04-11
Don't log as ERROR an OSError

fernandog:
Don't log as ERROR an OSError

Related to: https://github.com/pymedusa/Medusa/issues/2576

@ratoaq2


fernandog:
@ratoaq2 yeah, i also changed Exception to OSError those that handle with write/read files

fernandog:
@p00chie this is safe to be merged

PR #2582, opened 2017-04-11
Detect if new releases were added to the LTV .RAR

fernandog:
Uses the archive name to download archive again as LTV append new release groups to it

@ratoaq2

Logs:
2017-04-11 17:16:07 DEBUG    Thread-24 :: [b4a8a54] Releases found in cache but they are outdated
2017-04-11 17:16:18 DEBUG    Thread-24 :: [b4a8a54] Downloading archive 58cc028f6

From LTV:

### Debug
image

### DIFF
--- a/lib/subliminal/providers/legendastv.py
+++ b/lib/subliminal/providers/legendastv.py
@@ -47,6 +47,9 @@ timestamp_re = re.compile(r'(?P<day>\d+)/(?P<month>\d+)/(?P<year>\d+) - (?P<hour
 #: Cache key for releases
 releases_key = __name__ + ':releases|{archive_id}'
 
+#: Cache key for archive name
+archive_name_key = __name__ + ':archive_name|{archive_id}'
+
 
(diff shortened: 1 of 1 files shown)


lib/subliminal/providers/legendastv.py
```diff
@@ -416,6 +431,9 @@ def query(self, language, title, season=None, episode=None, year=None):
                     # cache the releases
                     region.set(releases_key.format(archive_id=a.id), releases)
 
+                    # cache the subtitle archive name
+                    region.set(archive_name_key.format(archive_id=a.id), a.name)
```
ratoaq2: We shouldn't have two different caches: one for release names and another for archive name. 
Probably the only thing you need to change is the cache key to use archive id and name

fernandog:
@ratoaq2 if this is ok, i will change base branch to the subliminal. agree?

fernandog:
will keep the branch if someone wants to test it

PR #2583, opened 2017-04-12
Feature/apiv2 refactoring episode numbering

labrys:
@ratoaq2 Here's the first part of my episode numbering refactor for you to look at


medusa/tv/episode.py
```diff
@@ -87,20 +84,209 @@
...
+        """
+        self.season = kwargs.pop('season', NotImplemented)
+        self.episode = episode
+
+
+    @property
+    def season(self):
+        return NotImplemented
+
+    @season.setter
```
ratoaq2: Why not make this object immutable? I would prefer to not change its content after creation

medusa/tv/episode.py
```diff
@@ -87,20 +84,209 @@
...
+        """Check episodes for inequality."""
+        return not self == other
+
+    def __iter__(self):
+        for attr in self._attrs:
+            yield getattr(self, attr)
+
+    def __bool__(self):
+        """Check boolean value of episode numbering."""
+        return all(self)
```
ratoaq2: I assume if all attributes are zero this still returns true

medusa/tv/episode.py
```diff
@@ -87,20 +84,209 @@
...
+        return cls(**cls.parse(numbering))
+
+
+class AbsoluteNumber(EpisodeNumber):
+    """Season-less episode number."""
+
+    _attrs = ['episode']
+
+    regex = re.compile(
+        r'(?:s(?P<season>\d{1,4}))?(?:e?(?P<episode>\d+))',
```
ratoaq2: In the end, this parent class knows about its subclass. It would be good to have no season in this class

medusa/tv/episode.py
labrys: Sorry that was supposed to be `any(self)`.  But (with any) this would only return true if at least a single attribute was truthy (e.g. not zero, none, etc)

medusa/tv/episode.py
labrys: The idea here is to minimize the number of regex's required to reduce parsing.  The parent class doesn't really know anything about its child class, but it does know that Seasons exist and that it should not have any.

medusa/tv/episode.py
labrys: It is immutable.  Just because the season has a setter, does not mean it can actually be set.  Essentially this is here to raise an appropriate exception (`InvalidNumbering`) if a season is attempted to be set.  This handles a 'season = None` from the default regex.

medusa/tv/episode.py
```diff
@@ -87,20 +84,209 @@
...
+    @air_date.setter
+    def air_date(self, value):
+        if not isinstance(value, (date, datetime)):
+            value = datetime.strptime(value, self.date_fmt)
+        with supress(AttributeError):
+            value = value.date()
+        setattr(self, '_air_date', value)
+
+    @classmethod
+    def parse(cls, number, fmt=None):
```
labrys: I'm still undecided if I want to allow a `fmt` argument here (as you could always change the date_fmt in the class or instance).

medusa/tv/episode.py
```diff
@@ -87,20 +84,209 @@
     get_scene_numbering,
     xem_refresh,
 )
-from medusa.tv.base import TV
-
-import shutil_custom
+from medusa.tv.base import Identifier, TV
```
labrys: Leaving this here for now as I may implement an EpisodeIdentifier class that contains series ID info and episode numbering.

medusa/tv/episode.py
```diff
@@ -87,20 +84,209 @@
...
+                    value,
+                )
+            )
+
+    def __str__(self):
+        """Magic method."""
+        return 'e{0:02d}'.format(self.episode)
+
+    @classmethod
+    def parse(cls, number, regex=None):
```
labrys: Same as for the `fmt` argument above.

medusa/tv/episode.py
```diff
@@ -87,20 +84,209 @@
...
+
+
+    @property
+    def season(self):
+        return NotImplemented
+
+    @season.setter
+    def season(self, value):
+        try:
+            assert value in (None, NotImplemented)
```
labrys: Will likely change this to a similar check (and maybe change the base exception for InvalidNumbering).  This was just a quick way to check it while testing.

labrys:
Also, I think the default constructor for `AirByDateNumbering` should be `year, month, day` and not a datetime.  Maybe add a `AirByDateNumbering.from_date()` as alternate constructors for `datetime.datetime` and `datetime.date`.

labrys:
I am considering exposing `strftime` for `AirByDateNumbering` and implementing a `format` method for absolute and relative (e.g. for `1x01` style output).

ratoaq2:
I suggest to close this one

PR #2584, opened 2017-04-12
Add missing button to force Trakt sync

fernandog:
@p0psicles your code was there but the button not.

Also i fixed the remove from collection as it wasn't using status like the add to collection

### DIFF
--- a/medusa/trakt_checker.py
+++ b/medusa/trakt_checker.py
@@ -209,7 +209,7 @@ class TraktChecker(object):
         """Remove episode from trakt collection.
 
         For episodes that no longer have a media file (location)
-        :param filter_show: optional. Only remove from given shows
+        :param filter_show: optional. Only remove episodes from trakt collection for given shows
         """
         if app.TRAKT_SYNC_REMOVE and app.TRAKT_SYNC and app.USE_TRAKT:
 
--- a/medusa/trakt_checker.py
+++ b/medusa/trakt_checker.py
@@ -208,7 +208,7 @@ class TraktChecker(object):
     def remove_episode_trakt_collection(self, filter_show=None):
         """Remove episode from trakt collection.
 
-        For episodes that no longer has media file (location)
+        For episodes that no longer have a media file (location)
         :param filter_show: optional. Only remove from given shows
         """
         if app.TRAKT_SYNC_REMOVE and app.TRAKT_SYNC and app.USE_TRAKT:
@@ -257,7 +257,7 @@ class TraktChecker(object):
     def add_episode_trakt_collection(self):
         """Add all existing episodes to Trakt collections.
 
-        For episodes that has media file (location)
+        For episodes that have a media file (location)
         """
         if app.TRAKT_SYNC and app.USE_TRAKT:
 
--- a/medusa/trakt_checker.py
+++ b/medusa/trakt_checker.py
@@ -206,7 +206,11 @@ class TraktChecker(object):
                 logger.log(u"Synced Trakt collection", logger.DEBUG)
 
     def remove_episode_trakt_collection(self, filter_show=None):
-        """Remove episode from trakt collection that no longer has media file (location)."""
+        """Remove episode from trakt collection.
+
+        For episodes that no longer has media file (location)
+        :param filter_show: optional. Only remove from given shows
+        """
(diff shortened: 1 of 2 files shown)


medusa/trakt_checker.py
```diff
@@ -206,19 +206,21 @@ def sync_library(self):
                 logger.log(u"Synced Trakt collection", logger.DEBUG)
 
     def remove_episode_trakt_collection(self, filter_show=None):
-        """Remove episode from trakt collection."""
+        """Remove episode from trakt collection that no longer has media file (location)."""
```
p0psicles: You don't need to explain the method in detail in your summary. If you really want to explain conditions in your method, then provide that on a separate line, with one white line in between. Or just add a comment above the condition.

medusa/trakt_checker.py
```diff
@@ -249,15 +251,14 @@ def remove_episode_trakt_collection(self, filter_show=None):
                                    (error=e.message), logger.INFO)
 
     def add_episode_trakt_collection(self):
-        """Add all existing episodes to Trakt collections."""
+        """Add all existing episodes to Trakt collections that has media file (location)."""
```
p0psicles: Same here.

medusa/trakt_checker.py
```diff
@@ -396,7 +397,7 @@ def add_episode_watchlist(self):
                                    (error=e.message), logger.INFO)
 
     def add_show_watchlist(self):
-        """Add show to Trakt watchlist."""
+        """Add show to Trakt watchlist if is in show list."""
```
p0psicles: Same here.

medusa/trakt_checker.py
fernandog: ok. will change it

medusa/trakt_checker.py
fernandog: ok. will change it

medusa/trakt_checker.py
```diff
@@ -206,19 +206,25 @@ def sync_library(self):
                 logger.log(u"Synced Trakt collection", logger.DEBUG)
 
     def remove_episode_trakt_collection(self, filter_show=None):
-        """Remove episode from trakt collection."""
+        """Remove episode from trakt collection.
+
+        For episodes that no longer has media file (location)
```
p0psicles: Have a..

medusa/trakt_checker.py
```diff
@@ -206,19 +206,25 @@ def sync_library(self):
...
+        :param filter_show: optional. Only remove from given shows
```
p0psicles: `Only remove from given shows` this log is not really explaining anything.

medusa/trakt_checker.py
```diff
@@ -249,15 +255,17 @@ def remove_episode_trakt_collection(self, filter_show=None):
                                    (error=e.message), logger.INFO)
 
     def add_episode_trakt_collection(self):
-        """Add all existing episodes to Trakt collections."""
+        """Add all existing episodes to Trakt collections.
+
+        For episodes that has media file (location)
```
p0psicles: have a..

medusa/trakt_checker.py
```diff
@@ -206,19 +206,25 @@ def sync_library(self):
...
                             b'e.season, e.episode, e.status ' \
-                            b'FROM tv_episodes AS e, tv_shows AS s WHERE s.indexer_id = e.showid and e.location = ""'
+                            b'FROM tv_episodes AS e, tv_shows AS s WHERE s.indexer_id = e.showid and e.location = "" ' \
+                            b'AND e.status in ({0})'.format(','.join(selection_status))
             if filter_show:
                 sql_selection += b' AND s.indexer_id = ? AND e.indexer = ?'
                 params = [filter_show.indexerid, filter_show.indexer]
 
-            sql_result = main_db_con.select(sql_selection, params)
+            sql_result = main_db_con.select(sql_selection, Quality.DOWNLOADED + Quality.ARCHIVED + params)
```
p0psicles: did you test this? As I don't think this will work. You can only use list + list here.

fernandog: @p0psicles yes, Quality.DOWNLOADED and Quality.ARCHIVED are lists

p0psicles: Ooh okay, then no prob. Thought they where constants.

PR #2586, opened 2017-04-12
Fix showing provider image when Release group is equal to provider name

fernandog:
@duramato

image


duramato:
No icons are showing now:

img

PR #2588, opened 2017-04-13
Use self.release_name to remove from failed.db history

fernandog:
Also move it to the end when PP is ok as we can't only remove from failed.db history table when it PP ok
Works fine with torrents. I tested a few

@medariox instead of the other PR I closed

@duramato need to do some more tests with nzb


fernandog:
@medariox need to adjust the test

labrys:
@OmgImAlexis I've already given my reasons for disagreeing with this.  We can discuss it in slack if you would like me to cover them again.

fernandog:
about this, in failed.db, the column is `release` so I'm only using the official release name stored in self.release_name and removing the code 

image

fernandog:
Another issue i found:

the row in failed.db history table uses the "torrent name": `Show.S01E01.720p.HDTV.X264-DIMENSION[rarbg]`
After failed renaming it becomes: `Show-S02E11-720p-HDTV-X264-DIMENSION-RARBG-`

Mind the ending "-" ....

In PP, it uses the media file: `Show.S01E01.720p.HDTV.X264-DIMENSION.mkv`, so self.release_name = `Show.S01E01.720p.HDTV.X264-DIMENSION`. After failed renaming, it becomes:  `Show-S02E11-720p-HDTV-X264-DIMENSION`

So in this case, the row in failed.db, history table won't be removed as:
`Show-S02E11-720p-HDTV-X264-DIMENSION` != `Show-S02E11-720p-HDTV-X264-DIMENSION-RARBG-`

maybe the self.release_name must first use the folder name as release_name, then the filename
But don't know if this breaks something. maybe @medariox can tell

fernandog:
@medariox still waiting you inputs as https://github.com/pymedusa/Medusa/pull/2588#issuecomment-302491269

fernandog:
no reply. no interest. closing. logged as known issue

PR #2591, opened 2017-04-14
Fix error trying to add show: invalid literal for int() with base 10: ''

labrys:

### DIFF
--- a/medusa/tv/series.py
+++ b/medusa/tv/series.py
@@ -1223,7 +1223,7 @@ class Series(TV):
         self.classification = getattr(indexed_show, 'classification', 'Scripted')
         self.genre = getattr(indexed_show, 'genre', '')
         self.network = getattr(indexed_show, 'network', '')
-        self.runtime = int(getattr(indexed_show, 'runtime', 0))
+        self.runtime = int(getattr(indexed_show, 'runtime', 0) or 0)
 
         # set the externals, using the result from the indexer.
         self.externals = {k: v for k, v in getattr(indexed_show, 'externals', {}).items() if v}


medusa/tv/series.py
```diff
@@ -1223,7 +1223,7 @@ def load_from_indexer(self, tvapi=None):
         self.classification = getattr(indexed_show, 'classification', 'Scripted')
         self.genre = getattr(indexed_show, 'genre', '')
         self.network = getattr(indexed_show, 'network', '')
-        self.runtime = int(getattr(indexed_show, 'runtime', 0))
+        self.runtime = int(getattr(indexed_show, 'runtime', 0) or 0)
```
fernandog: Cant have runtime of 0

In the past used to be try_int with 60 as default.
@p0psicles did you changed that?

labrys: Using fake values for stuff (like 60) is a bad idea.  Personally I'd prefer None to 0 for a situation where the value is not present, but as a runtime shouldn't ever be 0, 0 vs None has little bearing here.

fernandog: not fake. show has a runtime right? but its not there. Default show run time is 40 for drama and 20 for commedy. Don't see why is wrong

labrys: Because you don't know the actual time.  For logic like when to search you can easily do `self.runtime or 60` when the runtime is falsey (i.e. `None` or `0`).  Conversely there's no easy way to do the reverse.  If you assign a runtime of `60` the only way to know its a fake time (or default time, if you prefer) is to check the indexer again.

PR #2592, opened 2017-04-14
Don't allow user submit error with EpisodeNotFoundException

fernandog:
@p0psicles

### DIFF
--- a/medusa/tv/series.py
+++ b/medusa/tv/series.py
@@ -1048,8 +1048,9 @@ class Series(TV):
                     if not cur_ep:
                         raise EpisodeNotFoundException
                 except EpisodeNotFoundException:
-                    logger.warning(u'{indexerid}: Episode not found for this show. Please full update show. Skipping {filepath}',
-                                   indexerid=self.indexerid, filepath=filepath)
+                    logger.warning(u"{indexerid}: Episode not found for this show: '{show}'. "
+                                   u'Please run a full update. Skipping file: {filepath}',
+                                   indexerid=self.indexerid, show=self.name, filepath=filepath)
                     continue
 
             else:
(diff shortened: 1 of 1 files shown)


medusa/tv/series.py
```diff
@@ -1048,8 +1048,8 @@ def make_ep_from_file(self, filepath):
                     if not cur_ep:
                         raise EpisodeNotFoundException
                 except EpisodeNotFoundException:
-                    logger.error(u'{indexerid}: Unable to figure out what this file is, skipping {filepath}',
-                                 indexerid=self.indexerid, filepath=filepath)
+                    logger.warning(u'{indexerid}: Episode not found for this show. Please full update show. Skipping {filepath}',
```
duramato: `Please full update show` ?

fernandog: @duramato sorry?

duramato: Doesn't sound right to me

labrys: `Unable to determine episode for this file.  Please run a full update on the series.`

fernandog: Will say again. We KNOW which ep is. Please read slack history

labrys: Please read slack history `Can't we add a message that makes sense before merging it?`

fernandog: we just add it

labrys: And merged it without a follow-up review

labrys: What was missed was imprecise wording: `Please run a full update.` 
This is ambiguous and could be taken to mean a full library update instead of the intended series update.  On a large library that could mean the difference between a few minutes to update to several hours (or even more) to update.

Issue #2571, opened 2017-04-06
testRename: renaming files deletes associated nfo files.


wimpyrbx:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:** develop
**OS:** windows
**What you did:** rename files
**What happened:** files renamed but nfo's were deleted
**What you expected:** for nfo's not to be deleted, but renamed

when doing a rescan after the rename the nfos are generated again.

fernandog:
@wimpyrbx debug logs?

wimpyrbx:
```
2017-04-07 00:48:40 DEBUG    SHOWQUEUE-REFRESH :: [51ad817] 305719: Creating episode from: D:\TV\COMEDY SERIES\Big Little Lies\Season 01\Big Little Lies - S01E07 - You Get What You Need (1080p.WEB-DL).mkv
2017-04-07 00:48:40 DEBUG    SHOWQUEUE-REFRESH :: [51ad817] 305719: Creating episode object from D:\TV\COMEDY SERIES\Big Little Lies\Season 01\Big Little Lies - S01E07 - You Get What You Need (1080p.WEB-DL).mkv
2017-04-07 00:48:40 DEBUG    SHOWQUEUE-REFRESH :: [51ad817] Matched release D:\TV\COMEDY SERIES\Big Little Lies\Season 01\Big Little Lies - S01E07 - You Get What You Need (1080p.WEB-DL).mkv to a series in your database: Big Little Lies
2017-04-07 00:48:40 DEBUG    SHOWQUEUE-REFRESH :: [51ad817] Parsed D:\TV\COMEDY SERIES\Big Little Lies\Season 01\Big Little Lies - S01E07 - You Get What You Need (1080p.WEB-DL).mkv into title: Big Little Lies, season: 1, episode: [7], episode_title: You Get What You Need, screen_size: 1080p, format: WEB-DL, container: mkv, mimetype: video/x-matroska, type: episode, parsing_time: 0.0549998283386, total_time: 0.0549998283386, absolute_episode: [], quality: 1080p WEB-DL
2017-04-07 00:48:40 DEBUG    SHOWQUEUE-REFRESH :: [51ad817] 305719: D:\TV\COMEDY SERIES\Big Little Lies\Season 01\Big Little Lies - S01E07 - You Get What You Need (1080p.WEB-DL).mkv parsed to Big Little Lies S01E07
[...]

PR #2598, opened 2017-04-16
Don't delete files in manual renamer

fernandog:
The list associated files will only delete in PP

The code in `list_associated_files` that create the files to be deleted checks `if app.MOVE_ASSOCIATED_FILES:` and this is a PP setting
So, that can be only executed in PP

@wimpyrbx can you test it too please?
ping @duramato 

Fixes: https://github.com/pymedusa/Medusa/issues/2571


fernandog:
@medariox has a better suggestion @wimpyrbx lets wait it

PR #2602, opened 2017-04-17
Trakt: Create episode object because we need the actual status (from db)

fernandog:
Create episode object because we need the actual status (from db)
Otherwise we will read default blank episode status (UNKNOWN)

This can happen if show never got updated
So self.episodes would be empty and the episode object in trakt would be None

### DIFF
--- a/medusa/trakt_checker.py
+++ b/medusa/trakt_checker.py
@@ -31,9 +31,10 @@ from .search.queue import BacklogQueueItem
 from .show.show import Show
 
 
-def set_episode_to_wanted(show, s, e):
+def set_episode_to_wanted(show, season, episode):
     """Set an episode to wanted, only if it is currently skipped."""
-    ep_obj = show.get_episode(s, e, no_create=True)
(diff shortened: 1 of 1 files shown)


p0psicles:
@fernandog you implemented this here: https://github.com/pymedusa/Medusa/commit/a6af9b96f10e053cfb6045c1d390db6413af37d4

What was the reason you changed this in that commit?

fernandog:
@p0psicles the no_create=True is only used in trakt and in another place. After debugging med issue i found out that no_create param doesn't load (my idea was to avoid re-load, that's why i used it) the episode from database but if show we never updated, the ep obj will never be created. I'm my tests the show was always updated so I didn't found this issue

p0psicles:
Does the param have any use then? Shouldn't we also remove it?

fernandog:
@p0psicles i already asked @ratoaq2 if we can deprecated/remove it

p0psicles:
@fernandog please don't use bugfix/ for branches your merging into develop. /bugfix and /hotfix are kind of reserved for other purposes.

PR #2609, opened 2017-04-19
Normalize show status while loading from indexer

fernandog:
@medariox @p0psicles

### DIFF
--- a/medusa/indexers/indexer_config.py
+++ b/medusa/indexers/indexer_config.py
@@ -51,6 +51,28 @@ EXTERNAL_MAPPINGS = {EXTERNAL_IMDB: 'imdb_id', EXTERNAL_ANIDB: 'anidb_id',
 # trakt indexer name vs Medusa indexer
 TRAKT_INDEXERS = {'tvdb': INDEXER_TVDBV2, 'tmdb': INDEXER_TMDB, 'imdb': EXTERNAL_IMDB, 'trakt': EXTERNAL_TRAKT}
 
+STATUS_MAP = {
+    'returning series': 'Continuing',
+    'canceled/ended': 'Ended',
+    'tbd/on the bubble': 'Continuing',
+    'in development': 'Continuing',
+    'new series': 'Continuing',
+    'never aired': 'Ended',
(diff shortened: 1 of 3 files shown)


medusa/indexers/indexer_config.py
```diff
@@ -51,6 +51,28 @@
...
+    'pilot rejected': 'Ended',
+    'canceled': 'Ended',
+    'ended': 'Ended',
+    'to be determined': 'Continuing',
+    'running': 'Continuing',
+    'planned': 'Continuing',
+    'in production': 'Continuing',
+    'pilot': 'Continuing',
+    'cancelled': 'Ended',
+    '': 'Unknown',
```
p0psicles: I don't think you need this. This is redundant now.

medusa/tv/series.py
```diff
@@ -548,6 +549,13 @@ def release_required_words(self):
         """Return release ignore words."""
         return [v for v in (self.rls_require_words or '').split(',') if v]
 
+    @staticmethod
+    def normalize_status(series_status):
+        """Return a normalized status given current indexer status."""
+        if not series_status:
+            return 'Unknown'
```
p0psicles: Or we use this as default unknown status. Or we put it in the STATUS_MAP. So this could be a little bit more DRY.

fernandog: But if series_status is None I can't make it lower() as i do in the next line

So what to do?

medusa/tv/series.py
```diff
@@ -548,6 +549,11 @@ def release_required_words(self):
         """Return release ignore words."""
         return [v for v in (self.rls_require_words or '').split(',') if v]
 
+    @staticmethod
+    def normalize_status(series_status):
+        """Return a normalized status given current indexer status."""
+        return STATUS_MAP.get(series_status.lower()) if series_status else 'Unknown'
```
medariox: This will return None if the status is not in the map. Are you sure this is wanted?

medariox:
This causes all TheTVDB shows to be set to Unknown (try a force update). @fernandog

PR #2617, opened 2017-04-20
Fix limetorrents daily

p0psicles:

### DIFF
--- a/medusa/providers/torrent/html/limetorrents.py
+++ b/medusa/providers/torrent/html/limetorrents.py
@@ -56,8 +56,9 @@ class LimeTorrentsProvider(TorrentProvider):
         self.urls = {
             'update': urljoin(self.url, '/post/updatestats.php'),
             'search': urljoin(self.url, '/search/tv/{query}/'),
-            'rss': urljoin(self.url, '/browse-torrents/TV-shows/date/{page}/'),
-            'tv': urljoin(self.url, '/browse-torrents/TV-shows/'),
+            # Original rss feed url, temporary offline. Replaced by the main Tv-show page.
+            # 'rss': urljoin(self.url, '/browse-torrents/TV-shows/date/{page}/'),
+            'rss': urljoin(self.url, '/browse-torrents/TV-shows/'),
         }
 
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/html/limetorrents.py
```diff
@@ -57,6 +57,7 @@ def __init__(self):
             'update': urljoin(self.url, '/post/updatestats.php'),
             'search': urljoin(self.url, '/search/tv/{query}/'),
             'rss': urljoin(self.url, '/browse-torrents/TV-shows/date/{page}/'),
+            'tv': urljoin(self.url, '/browse-torrents/TV-shows/'),
```
p0psicles: @duramato why introducing tv? Are we switching from rss to tv?

medusa/providers/torrent/html/limetorrents.py
```diff
@@ -96,7 +97,8 @@ def search(self, search_strings, age=0, ep_obj=None):
 
                     search_url = self.urls['search'].format(query=search_string)
                 else:
-                    search_url = self.urls['rss'].format(page=1)
+                    search_url = self.urls['tv']
+                    #search_url = self.urls['rss'].format(page=1)
```
p0psicles: Why comment it here? I would just change it.

duramato: I have hopes this is just something temporary, so i'm leaving it here to change back.

medusa/providers/torrent/html/limetorrents.py
duramato: rss was a bad variable name, it's the same the only diff is pages aren't working besides the 1st

p0psicles: TV or rss is just a convention we use? We have lots of 'rss' URLs that work perfectly.

duramato: `rss` here was reffering to it beeing an url for daily, just that. It isn't an xml rss or anything like that. I used `tv` becouse it is the page of latest tv shows.

PR #2650, opened 2017-04-21
Fix not changing poster in home

fernandog:
@wimpyrbx @OmgImAlexis


fernandog:
Already started to create the changelog:
https://github.com/pymedusa/medusa.github.io/pull/50

Issue #2612, opened 2017-04-19
Privacy setting note doesn't match the options in the listbox


fernandog:
Also afaik we could change it without restart just like enabling subliminal/tornado logs:
```python
# Reconfigure the logger
logger.reconfigure()
```

![image](https://cloud.githubusercontent.com/assets/2620870/25197823/c3f0c28c-251b-11e7-8370-b96d6ab53b1d.png)

fernandog:
in common we have a lot more levels (and no docs about what they mean)

```python
privacy_levels = {
    'absurd': 9002,
    'stupid': 9001,  # it's over 9000!
    'max': 9000,
    'high': 30,
    'normal': 20,
    'low': 10,
    'disabled': 0
}
```

in mako we have only: 
```mako
% for privacy_level in ['high', 'normal', 'low', ]:
```

and somehow even we reconfigure log in save_general, changing levels doesn't apply to logs
I tested doing manual searches and checking if user and password appear in the provider URLS

fernandog:
https://github.com/pymedusa/Medusa/pull/494#issuecomment-220771599

labrys:
The other levels were undocumented for a reason.  `disabled` is for dev usage only.  `stupid` was a test for excessive censoring, and `absurd` was a joke level that censored absolutely everything, so I'm not surprised that that breaks the logger.

fernandog:
_`disabled` is for dev usage only._

![image](https://cloud.githubusercontent.com/assets/2620870/25206269/48235b66-253d-11e7-9b1c-025ebc01ad2e.png)

labrys:
Yes it was discussed in IRC to make it unavailable.  After the feature was coded, the UI was then updated with commit f98245b.  The note should never have been in the UI.

labrys:
No need to assign him, just remove the warning.  The decision in IRC to not enable it in the UI was that if a user wants to make a dangerous change like that then they can manually edit the config.  If the teams wants to allow users to disable censoring, my recommendation is to then prevent auto-reporting of issues while disabled.

PR #2651, opened 2017-04-21
Log privacy: 'disabled' option not available

fernandog:
Changing log level doesn't need restart as rato said

Fixes: https://github.com/pymedusa/Medusa/issues/2612

### DIFF
--- a/views/config_general.mako
+++ b/views/config_general.mako
@@ -642,7 +642,9 @@
                                     <span>
                                         Set the level of log-filtering.
                                         Normal (default).
+                                        <div class="clear-left">
                                         <br><b>NOTE:</b> Low: only passwords and api keys. Normal: also hide usernames. High: also hide hosts
+                                        </div>
                                     </span>
                                 </span>
                             </label>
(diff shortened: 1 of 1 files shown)


labrys:
I think the level descriptions should be more generalized, so if we change what gets filtered at a certain level the note doesn't need to be changed.

views/config_general.mako
```diff
@@ -642,8 +642,7 @@
                                     <span>
                                         Set the level of log-filtering.
                                         Normal (default).
-                                        <br>NOTE: A restart may be required to take effect.
-                                        <br>WARNING: Setting to "DISABLED" will show sensitive information such as passwords in the logs!
+                                        <br><b>NOTE:</b> Low: only passwords and api keys. Normal: also hide usernames. High: also hide hosts
```
labrys: I think the level descriptions should be more generalized, so if we change what gets filtered at a certain level the note doesn't need to be changed.

fernandog: Right down your suggestion

PR #2653, opened 2017-04-21
Change web root in config general

fernandog:
@medariox @OmgImAlexis 

also log the web root in the log

### DIFF
--- a/medusa/server/core.py
+++ b/medusa/server/core.py
@@ -202,8 +202,11 @@ class AppWebServer(threading.Thread):  # pylint: disable=too-many-instance-attri
             protocol = 'http'
             self.server = HTTPServer(self.app)
 
-        logger.log('Starting Medusa on {scheme}://{host}:{port}/'.format
-                   (scheme=protocol, host=self.options['host'], port=self.options['port']))
+        logger.log('Starting Medusa on {scheme}://{host}:{port}{web_root}'.format
+                   (scheme=protocol,
+                    host=self.options['host'],
+                    port=self.options['port'],
+                    web_root=self.options['web_root']))
 
         try:
             self.server.listen(self.options['port'], self.options['host'])
--- a/medusa/server/core.py
+++ b/medusa/server/core.py
@@ -202,7 +202,7 @@ class AppWebServer(threading.Thread):  # pylint: disable=too-many-instance-attri
             protocol = 'http'
             self.server = HTTPServer(self.app)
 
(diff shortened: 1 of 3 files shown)


medusa/server/core.py
```diff
@@ -202,8 +202,11 @@ def run(self):
             protocol = 'http'
             self.server = HTTPServer(self.app)
 
-        logger.log('Starting Medusa on {scheme}://{host}:{port}/'.format
-                   (scheme=protocol, host=self.options['host'], port=self.options['port']))
+        logger.log('Starting Medusa on {scheme}://{host}:{port}{web_root}'.format
```
OmgImAlexis: Does this not need `/` before and after the `{web_root}` or does `web_root` include a starting and ending one?

views/config_general.mako
```diff
@@ -458,6 +458,19 @@
...
                         </div>
+                        <div class="field-pair">
+                            <label for="web_root">
+                                <span class="component-title">HTTP web root</span>
+                                <span class="component-desc">
+                                    <input type="text" name="web_root" id="web_root" value="${app.WEB_ROOT}" class="form-control input-sm input300" autocomplete="no" />
+                                    blank = disabled
+                                <div class="clear-left">
+                                <p>Set a base URL, for use in reverse proxies.</p>
+                                <p><b>Note:</b> Must restart to have effect. Mind that old URL will not work.</p>
```
OmgImAlexis: "Mind that old URL will not work."

This needs to be reworded.

medusa/server/core.py
fernandog: @OmgImAlexis medusa auto add the `/` to web_root

views/config_general.mako
duramato: "Keep in mind that any previously configured base URLs won't work, after this change."

fernandog: suggestion?

medusa/server/core.py
fernandog: `INFO     TORNADO :: [d958c98] Starting Medusa on http://0.0.0.0:8081/medusa`

```python
        # web root
        if self.options['web_root']:
            app.WEB_ROOT = self.options['web_root'] = ('/' + self.options['web_root'].lstrip('/').strip('/')
```

OmgImAlexis: All good, just thought I'd check. 👍

fernandog:
@medariox any comments?

PR #2656, opened 2017-04-22
Fix episodes being converted to bare archived

fernandog:
and in boot, converted to ARCHIVED (UNKNOWN)
`Found X shows with bare archived status, attempting automatic conversion`
very very old issue


labrys:
Use `int` not `try_int` and get rid of default values.

PR #2658, opened 2017-04-22
Feature/deprecate show

labrys:
This deprecates `Episode.show` in favor of `Episode.series`.

The `show` attribute is used dynamically in a lot of places, so a quick refactor will miss many usages.

This is a safe change as the deprecation factory will still allow usage of the deprecated `show` attribute, while raising a warning (not logging a warning) for missed refactors with the stack trace to allow for easy replacement.

To display the warnings run python with the `-Wd` argument: e.g. `python -Wd start.py`


tests/conftest.py
```diff
@@ -142,7 +142,7 @@ def create(indexer=INDEXER_TVDBV2, indexerid=0, lang='', quality=Quality.UNKNOWN
 def create_tvepisode(monkeypatch):
     def create(show, season, episode, filepath='', **kwargs):
```
medariox: What about the other ```show```s in this file? Should probably be replaced as well.

labrys: Most likely, but my first local attempt I did some mass replaces assuming that all `show`s in a file that were named similar and found that wasn't the case.  For these commits so far, I have restricted changes to those logged during actual usage.  I figured questionable usage could come in either later commits or, preferably, later PRs.

OmgImAlexis: As long as everything is working adding more commits to change vars used in methods should be fine since they don't really matter what they're called as long as they're called the same thing through out the method.

PR #2660, opened 2017-04-23
Add fallback for BraceMessage formatting

labrys:
If string formatting of BraceMessage fails, fall back to the `repr` to log the message and log a traceback of the reason.

### DIFF
--- a/medusa/logger/adapters/style.py
+++ b/medusa/logger/adapters/style.py
@@ -8,6 +8,8 @@ import logging
 
 from six import text_type
 
+log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


fernandog:
@labrys did a test here and it doesn't log the "location" of the wrong log:

```
2017-04-24 11:17:08 ERROR    SEARCHQUEUE-DAILY-SEARCH :: [1a97a77] BraceMessage string formatting failed. Using representation instead.
  File "/home/osmc/Medusa/medusa/logger/adapters/style.py", line 34, in __str__
    return self.msg.format(*args, **kwargs)
KeyError: 'test'
```

PR #2667, opened 2017-04-24
In forced daily search, updade RSS even not needed episodes

fernandog:


p0psicles:
Sorry, but  I have no idea what your trying to do here.

fernandog:
If user doesn't have a wanted episode. we don't update RSS from providers in auto daily search

by doing a forced daily search, it will update all RSS items from providers and help manual search recently snatched episodes by getting all recent results instead of running N manual searches per episode, so N x X providers hits per episode

duramato:
This is a needed change. I tried to make @fernandog 's comment more clear to understand.

p0psicles:
Ok then the term forced is a little off imo. As your not forcing the manual search. But you trying to force the caching. So I would start changing that.

fernandog:
@p0psicles while running a forced daily, i will also force the rss cache update. got it ? I can't force only the cache

labrys:
I don't have the slightest clue what was meant by the comment.  However I am a proponent of caching all items from all searches of all types.  It's much easier and quicker to check our cache for wanted items/qualities than to do provider requests and parsing.

medusa/search/core.py
```diff
@@ -400,7 +400,7 @@ def wanted_episodes(show, from_date):
     return wanted
 
 
-def search_for_needed_episodes():
+def search_for_needed_episodes(force):
```
labrys: Force should be an optional parameter that defaults to False.

labrys: Parameters like force should be optional and should default to False.

fernandog:
` It's much easier and quicker to check our cache for wanted items/qualities than to do provider requests and parsing.`

we already do that

labrys:
Read the whole comment.
>  However I am a proponent of caching all items from all searches of all types
We don't do that.

fernandog:
Daily, Backlog and Manual Search

```
DEBUG    SEARCHQUEUE-BACKLOG-311945 :: [Rarbg] :: [c8cc65a] Adding item from search to cache: Bull.2016.S01E22.720p.HDTV.x264-AVS[rartv]
DEBUG    SEARCHQUEUE-BACKLOG-311945 :: [NZBGeek] :: [c8cc65a] Adding item from search to cache: Bull.2016.S01E22.720p.HDTV.x264-AVS-BUYMORE
080p.BluRay.x264.DHD
```

labrys:
>  However I am a proponent of caching **>>all<<** items from all searches of all types

We don't cache all items.

fernandog:
which search is missing except Backlog, Daily and Manual search?!

fernandog:
If you are talking to cache all items, even unwanted ones, we already decided a long time ago that doing that is not good.

labrys:
Which was the entire point of my comment:

>  However **I am a proponent of caching all items from all searches of all types**

fernandog:
I see but that should be discussed elsewhere. 
Anyway PR is concluded

fernandog:
@medariox which test do you expect me to add?

p0psicles:
I've looked at this. And what this should do, is always run through all the providers RSS feeds.
As opposed to what we do now. Only run through them, if there is something needed. I'm for this.

But, why are you adding a chain of force=True params, when it's always true? There are no situations where you are not adding the force=True right?

fernandog:
@p0psicles 
My goal is when user hit the button "**Force daily**", it will "run through all the providers RSS feeds" even if not needed/wanted episodes (we don't update provider rss feeds if nothing is wanted)

p0psicles:
I don't think that's what you realized. It will now force always. Also for scheduled daily runs

fernandog:
forcing:

2017-06-08 15:14:53 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [cffd483] Search mode: RSS
2017-06-08 15:14:53 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [Rarbg] :: [cffd483] GET URL: http://torrentapi.org/pubapi_v2.php?get_token=get_token&app_id=medusa&format=json [Status: 200]
2017-06-08 15:14:52 INFO     SEARCHQUEUE-DAILY-SEARCH :: [cffd483] Using daily search providers
2017-06-08 15:14:52 DEBUG    SEARCHQUEUE-DAILY-SEARCH :: [cffd483] Seeing if we need anything from ****

PR #2668, opened 2017-04-25
Remove another broad exception for subtitles

fernandog:
https://github.com/pymedusa/Medusa/issues/1432

@duramato


medariox:
Only catch ValueError, instead of removing the whole try .. except.

fernandog:
@medariox i think we don' need valueError because we have this on the first line:

`mode = 'downloading' if picked_id else 'searching'`
so `mode` can be only those two
agree?

medariox:
That's your code, right? Why did you add it at that time? I suppose there was a reason for it.

fernandog:
@medariox checked original PR and i can't remember why. I rewrote that code several times. I had 68 commits in the original PR

But i know for sure we will never reach that else

PR #2673, opened 2017-04-26
Don't allow submit error when can't refresh show because of indexer issues

fernandog:
Add missing exception

@p0psicles @duramato 

Fixes: https://github.com/pymedusa/Medusa/issues/2669


fernandog:
Thanks

PR #2676, opened 2017-04-26
Possible fix to deleting all network timezones and logging all as ERROR

fernandog:
The only way I could reproduce the issue is when `iteritems(d)` is empty

@p0psicles


medusa/network_timezones.py
```diff
@@ -67,17 +67,26 @@ def update_network_dict():
     network_list = dict(cache_db_con.select('SELECT * FROM network_timezones;'))
 
     queries = []
-    for network, timezone in iteritems(d):
+    remote_network_list = iteritems(d)
```
p0psicles: Why not directly check d?

fernandog: To understand better the code imo
Can change if you think is better

fernandog: @p0psicles also i use in two places so there is no reason to use iteritems(d) in both

p0psicles: I still don't fully understand this change. In which situation does it get to this point? I think only if an empty json is returned. As else it would have hit line 63.

So if in some situation we want to return if no result is found. Then we should just add the check at line 65. Then we also don't need to connect to db and list all existing timezones right?

Then I would just check on `if not d: ..`. But maybe i'm missing something.

fernandog: Sure.
Its not a concluded PR. I still cant reproduce the issue
Point is to discuss solutions.
Thanks

p0psicles: Ok then just leave the `for network, timezone in remote_network_list:` And add the log with the return at line 65.

PR #2690, opened 2017-05-03
[Hotfix] Use T411 new domain

fernandog:
again another domain change
@duramato


IDerr:
Yes exactly :D

lilp69:
Still not merge commit for new t411 address?

IDerr:
@lilp69 Seems to work on develop for the moment thanks to redirections, I think

p0psicles:
We're not going to hotfix this. If needed we look into alternative solution

IDerr:
When the .ai domain will be down, you'll have to hotfix this.

Issue #2696, opened 2017-05-05
Scaning/Refining video in the _unpack folder causing MalformedMKVError


fernandog:
It should ignore `_unpack` folder

```
2017-05-05 02:12:38 DEBUG    FINDSUBTITLES :: [36052ba] Scanning video u'eac78f3f87564f4a8069530a8c22824a.mkv' in u'/media/SAMSUNG/media/downloaded/series/Show.S01E20.1080p.AMZN.WEBRip.DD.5.1.x264-RTN-Scrambled/_unpack'
2017-05-05 02:12:41 DEBUG    FINDSUBTITLES :: [36052ba] Refining video with metadata
2017-05-05 02:12:41 WARNING  FINDSUBTITLES :: [36052ba] Failed to refine video
Traceback (most recent call last):
  File "/home/osmc/Medusa/lib/subliminal/core.py", line 543, in refine
    refiner_manager[refiner].plugin(video, **kwargs)
  File "/home/osmc/Medusa/lib/subliminal/refiners/metadata.py", line 32, in refine
    mkv = MKV(f)
  File "/home/osmc/Medusa/lib/enzyme/mkv.py", line 56, in __init__
    raise MalformedMKVError('Parsing error: %s' % e)
MalformedMKVError: Parsing error: Less than 1 bytes read (0)
```

medariox:
Fixed with https://github.com/pymedusa/Medusa/pull/2698

PR #2698, opened 2017-05-05
Fix scaning/refining video in the _unpack folder causing MalformedMKVError

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2696

### DIFF
--- a/medusa/subtitles.py
+++ b/medusa/subtitles.py
@@ -811,7 +811,7 @@ class SubtitlesFinder(object):
 
         run_post_process = False
         for root, _, files in os.walk(app.TV_DOWNLOAD_DIR, topdown=False):
-            # Must ignore files that are being unpacked by NZB client
+            # Skip folders that are being used for unpacking
             if u'_UNPACK' in root.upper():
                 continue
             for filename in sorted(files):
--- a/medusa/subtitles.py
+++ b/medusa/subtitles.py
@@ -811,6 +811,9 @@ class SubtitlesFinder(object):
 
         run_post_process = False
         for root, _, files in os.walk(app.TV_DOWNLOAD_DIR, topdown=False):
+            # Must ignore files that are being unpacked by NZB client
+            if u'_UNPACK' in root.upper():
+                continue
             for filename in sorted(files):
                 # Delete unwanted subtitles before downloading new ones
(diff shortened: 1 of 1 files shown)


medusa/subtitles.py
```diff
@@ -811,6 +811,9 @@ def subtitles_download_in_pp():  # pylint: disable=too-many-locals, too-many-bra
 
         run_post_process = False
         for root, _, files in os.walk(app.TV_DOWNLOAD_DIR, topdown=False):
+            # Must ignore files that are being unpacked by NZB client
+            if u'_UNPACK' in root.upper():
```
medariox: How can root have _UNPACK as  it's app.TV_DOWNLOAD_DIR?

fernandog: Did you see the issue? 
I didn't understand your doubt.

medariox: Yes, root will never be _UNPACK so this needs to be removed.

fernandog: Check line 878.
Rato did that code.
I did the same

medariox: Yes, there it makes sense because root can actually be _UNPACK, but here it is app.TV_DOWNLOAD_DIR, so it will never be _UNPACK. Can't explain it any better.

fernandog:
```python
        for root, _, files in os.walk(app.TV_DOWNLOAD_DIR, topdown=False):
            # Must ignore files that are being unpacked by NZB client
            logger.info('Checking if root has UNPACK: {root}'.format(root=root))
            if u'_UNPACK' in root.upper():
                continue
```
2017-05-06 08:53:16 INFO     FINDSUBTITLES :: [ce27a23] Checking if root has UNPACK: /media/SAMSUNG/media/downloaded/series/Bull.S01E20.Make.Me.1080p.AMZN.WEBRip.DD.5.1.x264-RTN-Scrambled/_UNPACK
2017-05-06 08:53:16 INFO     FINDSUBTITLES :: [ce27a23] Checking if root has UNPACK: /media/SAMSUNG/media/downloaded/series/Bull.S01E20.Make.Me.1080p.AMZN.WEBRip.DD.5.1.x264-RTN-Scrambled

medariox: Ok

PR #2693, opened 2017-05-04
Add HorribleSubs torrent provider

medariox:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

PR #2700, opened 2017-05-05
Add AniDex torrent provider

medariox:


medariox:
To be merged after: https://github.com/pymedusa/Medusa/pull/2693

medusa/providers/torrent/html/anidex.py
```diff
@@ -145,13 +149,13 @@ def parse(self, data, mode):
...
-                        logger.log('Found result: {0} with {1} seeders and {2} leechers'.format
-                                   (title, seeders, leechers), logger.DEBUG)
+                        log.debug('Found result: {0} with {1} seeders and {2} leechers',
+                                  title, seeders, leechers)
 
                     items.append(item)
                 except (AttributeError, TypeError, KeyError, ValueError, IndexError):
-                    logger.log('Failed parsing provider. Traceback: {0!r}'.format
-                               (traceback.format_exc()), logger.ERROR)
+                    log.error('Failed parsing provider. Traceback: {0!r}',
```
p0psicles: I prefer to use log.exception() in these situations. @labrys what you think?

labrys: Agreed.  Its much cleaner to do `log.exception('Failed parsing provider')`.

medariox: It's the same for me, but we should decide what to use. I saw @labrys still replacing it with log.error?

labrys: For my replaces its mostly just mass replaces with regex to deprecate the old brace message format so we can remove it.  Not spending much effort on content of messages.

PR #2705, opened 2017-05-06
Fix undesired words

labrys:
Undesired words short-cuts check too early causing releases with undesired words to be selected.


fernandog:
@labrys this is causing an issue. You failed to see that when best_result = cur_result, this `if any(ext in best_result.name.lower() for ext in undesired_words) and not any(ext in cur_result.name.lower() for ext in undesired_words):` will always be `False`

labrys:
simple just add `if best != current` to the check
the other option would be to count number of undesired words and `current <= best`
additionally we should probably consider short-circuiting (i.e. `continue`) any time `best_result = current_result`

labrys:
@fernandog I responded hastily, that's correct and that's the way it should work.

You are correct that if `best_result == current_result` it will be false.
It should be, if they are equal then there is no reason to assign `current_result` to `best_result`.

Now it could be optimized to skip the check when `best_result == current_result` but it wouldn't change the logic.

PR #2710, opened 2017-05-07
Streamline newpct title processing

labrys:


medariox:
I noticed that the search engine of Newpct is terrible. It can return hundreds of unrelated results. We should think about adding a hard-coded limit (e.g. 50 results).

labrys:
Are the earlier results better than later results?

PR #2778, opened 2017-05-09
Fix manage torrents not working because of mixed content blocking

fernandog:
Also fix security issue as we are storing the user and password in the html and we were using iframe

With this PR, manage torrents page will open a new tab

Tested with: Deluge, uTorrent and Transmission


fernandog:
i think it need some improved to avoid calling `get_manage_torrents_url()` every page load

I don't have any idea how to

medusa/helpers/__init__.py
```diff
@@ -1871,3 +1871,25 @@ def get_title_without_year(title, title_year):
     if year in title:
         title = title.replace(year, '')
     return title
+
+
+def get_manage_torrents_url():
+    """Ger the manage torrents URL."""
```
labrys: Typo

fernandog:
@labrys as i said above, with actual solution it loads in every page load, maybe need to move back to handler but don't know how to make a link in the mako that calls the handler

@OmgImAlexis ideas?

fernandog:
Ok, i figure it out

fernandog:
@OmgImAlexis replying from slack: yes, its a redirect now. As PR body, we store user/pass in mako/html now which is totally wrong.

PR #2788, opened 2017-05-13
Make parse_pubdate more generic

medariox:

TODO:
- [x] Add test

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -14,7 +14,7 @@ from itertools import chain
 from os.path import join
 from random import shuffle
 
-from dateutil import parser
+from dateutil import parser, tz
 from pytimeparse import parse
 
 from medusa import (
@@ -54,7 +54,6 @@ from medusa.scene_exceptions import get_scene_exceptions
 from medusa.show.show import Show
 
 from requests.utils import add_dict_to_cookiejar
-from pytz import timezone
 
 log = BraceAdapter(logging.getLogger(__name__))
 log.logger.addHandler(logging.NullHandler())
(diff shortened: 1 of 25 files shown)


medusa/providers/generic_provider.py
```diff
@@ -53,6 +54,7 @@
 from medusa.show.show import Show
 
 from requests.utils import add_dict_to_cookiejar
+from pytz import timezone
```
labrys: Is there a reason to select pytz here instead of dateutil since we already use dateutil timezones?

medusa/providers/generic_provider.py
```diff
@@ -480,6 +482,31 @@ def search(self, search_params, age=0, ep_obj=None):
         """Search the provider."""
         return []
 
+    @staticmethod
+    def parse_pubdate(pubdate, human_time=False, tz=None):
```
labrys: If we really want to make this more generic, does this need to be a part of providers?  We could move it out of providers altogether and just call it parse_date

medusa/providers/generic_provider.py
medariox: Changed to dateutil.

medusa/providers/generic_provider.py
medariox: I'd like to keep it in providers for now, as I don't really see any use outside of it in the near future.

medusa/providers/torrent/html/horriblesubs.py
```diff
@@ -102,7 +102,7 @@ def parse(self, data, mode):
                         if mode != 'RSS':
                             # keep the date and strip the rest
                             pubdate_raw = row.find('td', class_='rls-label').get_text()[1:9]
-                            pubdate = self.parse_pubdate(pubdate_raw)
+                            pubdate = self.parse_pubdate(pubdate_raw, timezone='America/Los_Angeles')
```
fernandog: @medariox dumb question: can we get that while reading response from server?

medariox: Only if the server provides it (almost never).

fernandog: what about using the server time (gmt) vs now() utc to convert the pubdate into user localtime?

medariox: https://tools.ietf.org/html/rfc7231#section-7.1.1.2

It's not the server time.

PR #2789, opened 2017-05-13
Add provider shanaproject.

p0psicles:
Note! Provider does not support:
* Seeders
* Leechers
* Size is not always available.
* Search by season and episode.

I know we maybe not need to add this provider. But I was almost finished with this provider. so thought let's add it as a PR non-the-less.


medusa/providers/torrent/html/shanaproject.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        Search a provider and parse the results.
+
+        :param search_strings: A dict with mode (key) and the search value (value)
+        :param age: Not used
+        :param ep_obj: Not used
+        :returns: A list of search results (structure)
+        """
+        results = []
+        if self.show and not self.show.is_anime:
```
p0psicles: @medariox you didn't add this for the AniDex provider. Was that on purpose?

medariox: Nope, forgot to add it.
Also forgot about:
```
self.anime_only = True
```

medusa/providers/torrent/html/shanaproject.py
```diff
@@ -0,0 +1,186 @@
...
+log = BraceAdapter(logging.getLogger(__name__))
+log.logger.addHandler(logging.NullHandler())
+
+
+class ShanaProjectProvider(TorrentProvider):
+    """ShanaProject Torrent provider."""
+
+    def __init__(self):
+        """Initialize the class."""
+        super(self.__class__, self).__init__('ShanaProject')
```
labrys: Need to avoid copy-pasta errors.  The providers should never have had this change merged as its breaks the init.  When resolving the super call it starts at the subclass instead of the intended starting position causing an infinite recursion error.  Proper super would be `super(MyClass, self).__init__()`.

labrys: Need to avoid copy-pasta errors.  The providers should never have had this change merged as its breaks the init.  When resolving the super call it starts at the subclass instead of the intended starting position causing an infinite recursion error.  Proper super would be `super(MyClass, self).__init__()`.

medusa/providers/torrent/html/shanaproject.py
```diff
@@ -0,0 +1,186 @@
...
+                        continue
+
+                    download_url = urljoin(self.url, download_url)
+
+                    # Provider does not support seeders or leechers.
+                    seeders = -1
+                    leechers = -1
+
+                    torrent_size = first_cell.find('div', {'class': 'release_size'}).get_text()
+                    size_regex = re.compile('([\d.]+)(.*)')
```
labrys: Compiling a regex is used to create a regex that will be used repeatedly later on.  By compiling it here you are not only creating the new compiled once per method, but once per row iterated through.  This is less efficient than just calling `re` directly on each pass.  Ideally the re.compiled regex would be at the module or class level.

medusa/providers/torrent/html/shanaproject.py
```diff
@@ -0,0 +1,190 @@
...
+        self.anime_only = True
+
+        # Torrent Stats
+        self.minseed = None
+        self.minleech = None
+
+        # Cache
+        self.cache = tv.Cache(self, min_time=20)
+
+    def search(self, search_strings, age=0, ep_obj=None):
```
labrys: Yes.  If the params were called by keyword instead of positionally it would break.

p0psicles: @medariox how come in your anidex pr it didn't complain about this?

medariox: @p0psicles 
About what?

p0psicles: Codacy. I ignored the warning. About the parameters.

PR #2793, opened 2017-05-14
Add sqlalchemy model for recommended shows.

p0psicles:

End goal of this Pr is to add the data modal needed to store the recommended shows. This will enable us to periodically check for new recommended shows in the background. Making it possible to load the recommended shows from local storage. 

Storing the recommended shows in db will also make it possible to recommend shows to users based on what the user already has in library.

### DIFF
--- a/medusa/model/recommended.py
+++ b/medusa/model/recommended.py
@@ -1,0 +1,91 @@
+
+import datetime
+from sqlalchemy import create_engine, ForeignKey
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy import Boolean, Column,  DateTime, Integer, String
+from sqlalchemy.orm import relationship, sessionmaker
+from sqlalchemy.sql import func
+
+from medusa.indexers.indexer_config import (EXTERNAL_ANIDB, EXTERNAL_IMDB, EXTERNAL_TRAKT, INDEXER_TMDB,
+                                            INDEXER_TVDBV2, INDEXER_TVMAZE)
+
+engine = create_engine('sqlite:///:memory:', echo=True)
+
+Base = declarative_base()
+
+
+class Indexer(Base):
+    __tablename__ = 'indexer'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    name = Column(String(255))
+    mapped_to = Column(String(255))
+    identifier = Column(String(255))
+    indexer_id = Column(Integer)
+    is_indexer = Column(Boolean)
+
+
+class RecommededShow(Base):
+    __tablename__ = 'recommended'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    title = Column(String(255))
+    show_id = Column(String(255))
+    image_href = Column(String(255))
+    image_src = Column(String(255))
+    cached_image = Column(String(255))
+    votes = Column(Integer)
+    rating = Column(String(255))
+    already_added = Column(Boolean)
+    created = Column(DateTime(timezone=True), server_default=func.now())
+    updated = Column(DateTime(timezone=True), onupdate=func.now())
+    rec_show_source_id = Column(Integer, ForeignKey('indexer.indexer_id'))
+    default_indexer_id = Column(Integer, ForeignKey('indexer.indexer_id'))
+
+    rec_show_source = relationship("Indexer", foreign_keys=[rec_show_source_id])
+    default_indexer = relationship("Indexer", foreign_keys=[default_indexer_id])
+
+# initial loading of indexer table
+
+indexers = [
+    {'id': INDEXER_TVDBV2,
+     'name': 'TVDBv2',
+     'mapped_to': 'tvdb_id',
+     'identifier': 'tvdb',
+     'is_indexer': True},
+    {'id': INDEXER_TVMAZE,
+     'name': 'TVMaze',
+     'mapped_to': 'tvmaze_id',
+     'identifier': 'tvmaze',
+     'is_indexer': True},
+    {'id': INDEXER_TMDB,
+     'name': 'TMDB',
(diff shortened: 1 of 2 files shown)


p0psicles:
@labrys i've create a PR so I can get your feedback as early as possible. As now I can still move stuff around.

labrys:
Any SQLA models should have Alembic support from the get-go.

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    name = Column(String(255))
+    mapped_to = Column(String(255))
+    identifier = Column(String(255))
+    indexer_id = Column(Integer)
+    is_indexer = Column(Boolean)
+
+
+class RecommededShow(Base):
+    __tablename__ = 'recommended'
```
labrys: `recommended` is a poor choice of terms... it should be `recommendations` or `recommended_series`.  Personally my preference is `recommendations` as then its not just specifically series, it could also be movies in the future.

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+
+Base = declarative_base()
+
+
+class Indexer(Base):
+    __tablename__ = 'indexer'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    name = Column(String(255))
+    mapped_to = Column(String(255))
```
labrys: remove this

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+    identifier = Column(String(255))
+    indexer_id = Column(Integer)
+    is_indexer = Column(Boolean)
```
labrys: If this is an indexer table then `is_indexer` should always be true.  Remove this.

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class Indexer(Base):
+    __tablename__ = 'indexer'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    name = Column(String(255))
+    mapped_to = Column(String(255))
+    identifier = Column(String(255))
+    indexer_id = Column(Integer)
```
labrys: You have an ID, why is there also an indexer_id? Remove this

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+Base = declarative_base()
+
+
+class Indexer(Base):
+    __tablename__ = 'indexer'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    name = Column(String(255))
+    mapped_to = Column(String(255))
+    identifier = Column(String(255))
```
labrys: You already have a name, whats the purpose of an identifier?  If you mean a slug, slug would be a better term.

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+    indexer_id = Column(Integer)
+    is_indexer = Column(Boolean)
+
+
+class RecommededShow(Base):
+    __tablename__ = 'recommended'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
```
labrys: This whole table should be really simple.. an source (e.g. indexer) and an id (the id from the indexer).  The rest of the data should be a relational series table contains info about a series.

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+     'name': 'TVMaze',
+     'mapped_to': 'tvmaze_id',
+     'identifier': 'tvmaze',
+     'is_indexer': True},
+    {'id': INDEXER_TMDB,
+     'name': 'TMDB',
+     'mapped_to': 'tmdb_id',
+     'identifier': 'tmdb',
+     'is_indexer': True},
+    {'id': EXTERNAL_IMDB,
```
labrys: Remove.  This is should be done through a mapping table.

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+     'name': 'IMDB',
+     'mapped_to': 'imdb_id',
+     'identifier': 'imdb',
+     'is_indexer': False},
+    {'id': EXTERNAL_ANIDB,
```
labrys: same.

medusa/model/recommended.py
```diff
@@ -0,0 +1,91 @@
...
+     'name': 'AniDB',
+     'mapped_to': 'anidb_id',
+     'identifier': 'anidb',
+     'is_indexer': False},
+    {'id': EXTERNAL_TRAKT,
```
labrys: same

medusa/model/recommended.py
p0psicles: We could also use it for sources that do not act as indexer. Like trakt and Imdb. I need a way to reference those.

medusa/model/recommended.py
p0psicles: Can you reference an example?

medusa/model/recommended.py
p0psicles: If I'm doing that I could normalize 5 more levels deep. That would really be counter productive in this case. Therefor I'd rater limit it to store recommendations as being its own entity.

labrys: This will be easier to explain in slack.

medusa/model/recommended.py
labrys: This too will be easier to explain in slack.

medusa/model/recommended.py
labrys: They are still indexers in that they provide show information, even though they aren't being used to create series.  The database shouldnt need to know implementation details.

PR #2797, opened 2017-05-15
Pass str (.message) in stead of object.

p0psicles:


p0psicles:
It seems that when logging an object, it's raising exception. Dont know if this can be fixed centralized?

labrys:
@p0psicles It can be fixed, but wanted to remove old style logger first before I fixed it... right now they have some conflicts by having them both and I'd rather not try to work around the old one while fixing the new.

p0psicles:
@labrys any reason to not accept this as a fix for this case?

labrys:
No not at all.

PR #2802, opened 2017-05-18
Always calculate memory usage in footer

fernandog:


fernandog:
somehow when I tested I didn't updated (using psutil) after several page refreshes.
Closing but i will keep monitoring

PR #2808, opened 2017-05-19
Fix labels in AniDex provider. Fixes #2806

### LINKED ISSUE #2806: Anidex Exception
On develop, I'm getting this exception because they don't always post a torrent file but only a magnet link. But I don't if magnets links are fully supported by Medusa.
``` shell 
2017-05-19 09:18:34 ERROR    SEARCHQUEUE-DAILY-SEARCH :: [AniDex] :: [c6cfffd] Failed parsing provider. Traceback: 'Traceback (most recent call last):\n  File "/opt/Medusa/medusa/providers/torrent/html/anidex.py", line 120, in parse\n    download_url = cells[labels.index(\'Torrent\')].a.get(\'href\')\nAttributeError: \'NoneType\' object has no attribute \'get\'\n'
  File "/opt/Medusa/medusa/providers/torrent/html/anidex.py", line 120, in parse
    download_url = cells[labels.index('Torrent')].a.get('href')
AttributeError: 'NoneType' object has no attribute 'get'
```

medariox:

### DIFF
--- a/medusa/providers/torrent/html/anidex.py
+++ b/medusa/providers/torrent/html/anidex.py
@@ -107,9 +107,10 @@ class AniDexProvider(TorrentProvider):
                 log.debug('Data returned from provider does not contain any torrents')
                 return items
 
-            table_spans = table_header.find_all('span')
-            # Skip 'Likes' to have the same amount of cells and labels
-            labels = [label.get('title') for label in table_spans if label.get('title') != 'Likes']
+            table_ths = table_header.find_all('th')
+            # [u'Category', u'', u'Filename', u'Comments', u'Torrent', u'Magnet',
+            #  u'File size', u'Age', u'Seeders', u'Leechers', u'Completed']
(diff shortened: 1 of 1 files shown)


duramato:
I just noticed the provider has both magnets and torrents, it is missing something like this 
https://github.com/pymedusa/Medusa/blob/b4d77f7f4bb3a0a07818a49884498e8e97ac7e3e/medusa/providers/torrent/rss/extratorrent.py#L129

p0psicles:
If this is something generic. It should be moved out of the individual provider, and moved into TorrentProvider.

medariox:
AniDex hosts the torrents directly, so we don't have to rely on 3rd party indexers. They also always include a download link with their torrents, unlike the magnets, that can be missing.

duramato:
We only want torrent files for black hole, magnets are preferred because of proxies. Its explained in the comment.

PR #2812, opened 2017-05-19
Implemented searching anime shows, without passing an episode or season

p0psicles:
…on number.

Used a fallback, where when it did not get results using the default search_strings, it will fallback to a search without ep/season.

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -96,6 +96,7 @@ class GenericProvider(object):
         self.show = None
         self.supports_absolute_numbering = False
         self.supports_backlog = True
+        self.anime_search_show_name = True
         self.url = ''
         self.urls = {}
         # Ability to override the search separator. As for example anizb is using '*' instead of space.
@@ -205,6 +206,12 @@ class GenericProvider(object):
             for search_string in search_strings:
                 # Find results from the provider
                 items_list += self.search(search_string, ep_obj=episode)
+
+            if not items_list:
+                # Find results from the provider using a fall-back search string.
+                search_string['Episode'] = search_string['EpisodeFallback']
+                for search_string in search_strings:
+                    items_list += self.search(search_string, ep_obj=episode)
 
         if len(results) == len(episodes):
             return results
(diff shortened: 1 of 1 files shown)


fernandog:
@p0psicles nice one

About this methods, RARBG for example allow us to search using tvdb id, but today we do N searches because of all name exceptions, when we could do only one request.

Related:
https://github.com/pymedusa/Medusa/pull/2467
and here I reverted:
https://github.com/pymedusa/Medusa/pull/2473

PR body explain my reasons
Maybe that issue can help you change those methods or not. 
Just telling one issue I had in my plans to fix someday

p0psicles:
I don't understand. Rarbg is working fine now right?

p0psicles:
Only thing I still need to add is add a _create_default_eepisode_string.

fernandog:
hum yes. but we are doing unnecessary requests:

For example show name is ABC and it was added in medusa with tvdb ID 1234. 
Scene names are ABCD and ABCDE

We do this searches in rarbg:  
&tvbid=1234&search=ABC+S01E01
&tvbid=1234&search=ABCD+S01E01
&tvbid=1234&search=ABCDE+S01E01

When in fact it could do only this one:
&tvbid=1234&search=S01E01


Understood?

Stonedestroyer:
One suggestion is to add episodes as 
Show - 01 rather then what it is doing now. 
 `SEARCHQUEUE-MANUAL-13814 :: [ebdacab] Unable to find manual results for: Sousei no Onmyouji - 000 - The Destined Two - Boy Meets Girl`

p0psicles:
I'm going to modify this PR slightly. As in only keeping the refactor. And not adding any new features.

medusa/providers/generic_provider.py
```diff
@@ -45,7 +46,6 @@
     InvalidShowException,
     NameParser,
 )
-from medusa.scene_exceptions import get_season_scene_exceptions
```
p0psicles: If anyone is wondering why i've changed this..
You can't mock it in pytest when you import the method directly.

PR #2816, opened 2017-05-20
new logging: cleaned up api/v1/core.py

p0psicles:
* Replaced double quotes with single, where possible.
* Removed ex.
* Replaced %s string parsing.
* Removed string concatenations with formats.

### DIFF
--- a/medusa/server/api/v1/core.py
+++ b/medusa/server/api/v1/core.py
@@ -157,7 +157,7 @@ class ApiHandler(RequestHandler):
                 out = '{0}({1});'.format(callback, out)  # wrap with JSONP call if requested
         except Exception as error:  # if we fail to generate the output fake an error
             log.exception(u'API :: Traceback')
-            out = '{"result": "{0}", "message": "error while composing output: {1!r"}'.format(
+            out = '{"result": "{0}", "message": "error while composing output: {1!r}"}'.format(
                 result_type_map[RESULT_ERROR], error
             )
         return out
--- a/medusa/server/api/v1/core.py
+++ b/medusa/server/api/v1/core.py
@@ -65,7 +65,7 @@ from tornado.web import RequestHandler
 log = BraceAdapter(logging.getLogger(__name__))
 log.logger.addHandler(logging.NullHandler())
 
-indexer_ids = ['indexerid', 'tvdbid', 'tvmazeid', 'tmdbid']
+INDEXER_IDS = ('indexerid', 'tvdbid', 'tvmazeid', 'tmdbid')
 
 # basically everything except RESULT_SUCCESS / success is bad
 RESULT_SUCCESS = 10  # only use inside the run methods
@@ -125,7 +125,7 @@ class ApiHandler(RequestHandler):
         try:
             out_dict = _call_dispatcher(args, kwargs)
         except Exception as e:  # real internal error oohhh nooo :(
-            log.exception(u'API :: {0}', e)
+            log.exception(u'API :: {0!r}', e)
             error_data = {
                 'error_msg': e,
                 'args': args,
@@ -186,7 +186,7 @@ class ApiHandler(RequestHandler):
                 if len(cmd.split('_')) > 1:
                     cmd, cmd_index = cmd.split('_')
 
-                log.debug(u'API :: {0}: cur_kwargs', cmd, cur_kwargs)
+                log.debug(u'API :: {0}: {1}', cmd, cur_kwargs)
                 if not (cmd in ('show.getbanner', 'show.getfanart', 'show.getnetworklogo', 'show.getposter') and
                         multi_commands):  # skip these cmd while chaining
                     try:
@@ -315,7 +315,7 @@ class ApiCall(ApiHandler):
         if len(self._missing) == 1:
             msg = 'The required parameter: {0!r} was not set'.format(self._missing[0])
         else:
-            msg = 'The required parameters: {0!r} where not set"'.format("','".join(self._missing))
+            msg = 'The required parameters: {0!r} were not set'.format("','".join(self._missing))
         return _responds(RESULT_ERROR, msg=msg)
 
     def check_params(self, args, kwargs, key, default, required, arg_type, allowed_values):
@@ -325,11 +325,11 @@ class ApiCall(ApiHandler):
         """
 
         # auto-select indexer
-        if key in indexer_ids:
+        if key in INDEXER_IDS:
             if 'tvdbid' in kwargs:
                 key = 'tvdbid'
 
-            self.indexer = indexer_ids.index(key)
+            self.indexer = INDEXER_IDS.index(key)
 
         missing = True
         org_default = default
@@ -1637,7 +1637,7 @@ class CMD_SearchIndexers(ApiCall):
                     continue
 
                 for cur_series in api_data:
-                    results.append({indexer_ids[_indexer]: int(cur_series['id']),
+                    results.append({INDEXER_IDS[_indexer]: int(cur_series['id']),
                                     'name': cur_series['seriesname'],
                                     'first_aired': cur_series['firstaired'],
                                     'indexer': int(_indexer)})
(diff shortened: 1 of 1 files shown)


p0psicles:
Sorry for the big PR. I've also replaced the quotes while I was at it.
You'll have to do some searching to really review the logging changes.

medusa/server/api/v1/core.py
```diff
@@ -61,7 +62,10 @@
 from six import iteritems, text_type
 from tornado.web import RequestHandler
 
-indexer_ids = ["indexerid", "tvdbid", "tvmazeid", "tmdbid"]
+log = BraceAdapter(logging.getLogger(__name__))
+log.logger.addHandler(logging.NullHandler())
+
+indexer_ids = ['indexerid', 'tvdbid', 'tvmazeid', 'tmdbid']
```
medariox: Maybe capitalize to make clear this is a constant.

medusa/server/api/v1/core.py
```diff
@@ -116,19 +120,19 @@ def get(self, *args, **kwargs):
...
 
             _call_dispatcher = profile(_call_dispatcher, immediate=True)
-            del kwargs["profile"]
+            del kwargs['profile']
 
         try:
             out_dict = _call_dispatcher(args, kwargs)
         except Exception as e:  # real internal error oohhh nooo :(
-            logger.log(u"API :: " + ex(e), logger.ERROR)
+            log.exception(u'API :: {0}', e)
```
medariox: {0!r} instead?

medusa/server/api/v1/core.py
```diff
@@ -145,16 +149,17 @@ def _out_as_image(self, _dict):
...
             if callback:
-                out = callback + '(' + out + ');'  # wrap with JSONP call if requested
-        except Exception as e:  # if we fail to generate the output fake an error
-            logger.log(u"API :: " + traceback.format_exc(), logger.DEBUG)
-            out = '{"result": "%s", "message": "error while composing output: %s"}' % \
-                  (result_type_map[RESULT_ERROR], ex(e))
+                out = '{0}({1});'.format(callback, out)  # wrap with JSONP call if requested
+        except Exception as error:  # if we fail to generate the output fake an error
+            log.exception(u'API :: Traceback')
+            out = '{{"result": "{0}", "message": "error while composing output: {1!r}"}}'.format(
```
medariox: Shouldn't this be a dict?

medusa/server/api/v1/core.py
```diff
@@ -163,25 +168,25 @@ def call_dispatcher(self, args, kwargs):  # pylint:disable=too-many-branches
...
             for cmd in commands:
                 cur_args, cur_kwargs = self.filter_params(cmd, args, kwargs)
 
-                if len(cmd.split("_")) > 1:
-                    cmd, cmd_index = cmd.split("_")
+                if len(cmd.split('_')) > 1:
+                    cmd, cmd_index = cmd.split('_')
 
-                logger.log(u"API :: " + cmd + ": cur_kwargs " + str(cur_kwargs), logger.DEBUG)
+                log.debug(u'API :: {0}: cur_kwargs', cmd, cur_kwargs)
```
medariox: Missing {1} ?

medusa/server/api/v1/core.py
```diff
@@ -281,36 +286,36 @@ def return_help(self):
...
+            msg = self._help['desc']
         return _responds(RESULT_SUCCESS, self._help, msg)
 
     def return_missing(self):
         if len(self._missing) == 1:
-            msg = "The required parameter: '" + self._missing[0] + "' was not set"
+            msg = 'The required parameter: {0!r} was not set'.format(self._missing[0])
         else:
-            msg = "The required parameters: '" + "','".join(self._missing) + "' where not set"
+            msg = 'The required parameters: {0!r} where not set"'.format("','".join(self._missing))
```
medariox: Something went wrong here.

medusa/server/api/v1/core.py
p0psicles: It's parsed into json. I'm rewriting using. Format now. Haven't looked into rewriting it to parse from dict.

medusa/server/api/v1/core.py
p0psicles: Yes good one

medusa/server/api/v1/core.py
p0psicles: Yes

medusa/server/api/v1/core.py
p0psicles: I think only the double quote at the end shouldn't be there. I'll do some tests.

labrys: Also "were not set" not "where not set"

medusa/server/api/v1/core.py
labrys: If its a constant it should be a tuple.

fernandog:
so no one will review&approve this PR @p0psicles took a lot of effort to make?
tons of log PR were merged in this is stalled?

medusa/server/api/v1/core.py
medariox: What I mean is:
_Before:_
```
'{"result": "%s", "message": "error while composing output: %s"}'
```
_After_
```
'{{"result": "{0}", "message": "error while composing output: {1!r}"}}'
```
Why the additional curly braces?

medusa/server/api/v1/core.py
```diff
@@ -163,25 +168,25 @@ def call_dispatcher(self, args, kwargs):  # pylint:disable=too-many-branches
             or calls the TVDBShorthandWrapper when the first args element is a number
             or returns an error that there is no such cmd
         """
-        logger.log(u"API :: all args: '" + str(args) + "'", logger.DEBUG)
-        logger.log(u"API :: all kwargs: '" + str(kwargs) + "'", logger.DEBUG)
+        log.debug(u'API :: all args: {0!r}', args)
+        log.debug(u'API :: all kwargs: {0!r}', kwargs)
```
medariox: Shouldn't this stay str instead of repr?

p0psicles: It will place it between single quotes.

medariox: Okay. I still think repr should be used for object representation. If we really need to have the single quotes (do we need them anyway?), we can use double quotes for the outer quotes.

p0psicles: I kind of followed labrys lead here. But I'm fine with either way.

medusa/server/api/v1/core.py
```diff
@@ -192,11 +197,11 @@ def call_dispatcher(self, args, kwargs):  # pylint:disable=too-many-branches
                         elif _is_int(cmd):
                             cur_out_dict = TVDBShorthandWrapper(cur_args, cur_kwargs, cmd).run()
                         else:
-                            cur_out_dict = _responds(RESULT_ERROR, "No such cmd: '" + cmd + "'")
+                            cur_out_dict = _responds(RESULT_ERROR, 'No such cmd: {0!r}'.format(cmd))
                     except ApiError as error:  # Api errors that we raised, they are harmless
-                        cur_out_dict = _responds(RESULT_ERROR, msg=ex(error))
+                        cur_out_dict = _responds(RESULT_ERROR, msg=error)
```
medariox: msg expects a string, so maybe ```msg=error.message```?

medusa/server/api/v1/core.py
```diff
@@ -192,11 +197,11 @@ def call_dispatcher(self, args, kwargs):  # pylint:disable=too-many-branches
...
                 else:  # if someone chained one of the forbidden commands they will get an error for this one cmd
-                    cur_out_dict = _responds(RESULT_ERROR, msg="The cmd '" + cmd + "' is not supported while chaining")
+                    cur_out_dict = _responds(RESULT_ERROR, msg='The cmd {0!r} is not supported while chaining'.fomrat(cmd))
```
medariox: ```format``` instead of ```fomrat``` (I blame rato for this one)

labrys: 'The cmd `fomrat` is not supported while chaining'
Or any other time really

medariox: I think this is the only critical change needed. After that we can start testing it. @p0psicles?

p0psicles: Next on my list

medusa/server/api/v1/core.py
p0psicles: Because I'm using format now instead of %s. Format uses {} as special chars. So you have to escape them.

p0psicles:
Anything that I still need to look at? @medariox @labrys ?

PR #2819, opened 2017-05-22
Vue changes

OmgImAlexis:
This adds the navbar, footer and a few blank components to build from.

All pages that currently aren't implemented should be linking to to-be-implemented instead of their own pages to keep clutter down.

```html
<router-link :to="{ name: 'to-be-implemented' }"></router-link>
```

A note to anyone adding in actions for Vuex please make sure all actions return a promise as they should all be chainable.


OmgImAlexis:
@p0psicles thoughts on what we can do about the `build.js` being changed every time we change anything with vue?

vue/src/api/api.js
```diff
@@ -1,41 +1,37 @@
 import axios from 'axios';
 
 const api = axios.create({
-    baseURL: '/api/v2/',
-    timeout: 10000,
+    baseURL: 'http://localhost:8081/api/v2/',
```
p0psicles: This can be moved back to being relative?

OmgImAlexis: Yep, I moved it to that since I was testing this via caddy and nginx as well as `/vue/` in Medusa.

OmgImAlexis:
I'll update that today.

OmgImAlexis:
Just waiting on https://github.com/strongloop/fsevents/pull/174 to fix yarn install.

p0psicles:
I think that has been merged.

OmgImAlexis:
@p0psicles did you want to review this or shall we just merge once I get travis passing?

p0psicles:
I've already went through it once. But to have it properly reviewed you'll need someone else, that has more experience in vue and js in general. So I say let's go with it if Travis gives the go. 

Like I've mentioned before. We should go to a situation where we have a vue scaffolding setup, where everyone can add components, directives, filters.. Etc. without needing to be aware of any custom code we've added. Or then we need to add it to standards & guidelines.

OmgImAlexis:
This is still showing up as well as issues about optional deps, this will be fixed soon and doesn't cause any issues they're just warnings.

```
SyntaxError
    at XMLHttpRequest.open (/home/travis/build/pymedusa/Medusa/vue/node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:486:15)
    at dispatchXhrRequest (/home/travis/build/pymedusa/Medusa/vue/node_modules/axios/lib/adapters/xhr.js:45:13)
    at xhrAdapter (/home/travis/build/pymedusa/Medusa/vue/node_modules/axios/lib/adapters/xhr.js:12:10)
    at dispatchRequest (/home/travis/build/pymedusa/Medusa/vue/node_modules/axios/lib/core/dispatchRequest.js:52:10)
    at process._tickCallback (internal/process/next_tick.js:103:7)
```

OmgImAlexis:
@p0psicles tests are passing again.

p0psicles:
Going through it I noticed there isn't almost a single line of comment written. I think we should really start setting the example here. What do you think?

OmgImAlexis:
What exactly should I be commenting though? If you understand vue then it's exactly the same as almost every example app that is shown when people are learning so I figures there's almost no point right now as we don't really have anything that isn't documented in the vue and vuex docs.

labrys:
Considering the team is primarily not front end devs, comments would be very helpful.  Besides "If you understand `X language/library/app` then ..." could be said of almost anything.   For example:
```python
df['result'] = df['result'].map(lambda x: x.lstrip('+-').rstrip('aAbBcC'))
```
If you understand Python and Pandas you know exactly what thats doing.

OmgImAlexis:
@labrys no that's completely different, the code I've added for vue and vuex is exactly the same code in their documentation guide. The same guide I'd expect anyone working on vue to read.

SInce it's already documented I don't see the point of essentially rewriting it.

labrys:
Actually its not.  The code I posted is from the Pandas documentation.  Out of context and to someone that's not familiar its all greek to them.

labrys:
For example you have this comment:
```
// https://stackoverflow.com/a/41434763/2311366
```
If that link for some reason became invalid, what was the purpose of it?  If it was worth looking up on stackoverflow it probably deserves an actual comment as well.  For example something like:
```

PR #2827, opened 2017-05-23
Notify on leaked episodes

fernandog:
This is a **POC** showing how to notify users when a leaked episode is detected

The downside of it is that RSS cache must be updated always, even not wanted episodes
For now I only get unaired episodes in the next 7 days. I see no reason to have a leak before that. Maybe in the future add a setting for this too but it may slow down the DAILYSEARCH thread

Only NMA notifier is supported for now
If others want to commit to add other notifiers feel free

How to test:
- Edit your cache DB
- Pick one DAILY provider and pick a show with unaired episode
- Edit column episodes and release name to the episode number of a unaired episode
- Force start daily


duramato:
Nothing should be named `leaked`. It only found a result `pre air`

p0psicles:
Few questions. 
1. How do you prevent from sending multiple notifications for the same pre-release?
2. Why searching from cache, and not just send a notifications when a result from a daily search is stored in cache?

fernandog:
@p0psicles 

1) For now,  it will notify for every leaked release for every provider and it will keep nofitying. Yes, I know this is not good. 
2) Because we can't know if that episode is wanted or unaired while adding in the cache. At least I remember its how is done without looking at the code

fernandog:
maybe to prevent multiple notifications, we could add a new column in cache, to flag we already notified that release

p0psicles:
If it is a new cache item, we could do a quick search on tv_episode table, and check the status for the matching episode? that way you dont need to add a column to the cache table.

fernandog:
yes, but there is a lot of other checks we should do like qualities (allowed and preferred, ignored words, etc) and current status
I don't think doing this while adding to cache is a good idea

p0psicles:
Then don't check for those. Maybe only check the default status. The others are of importance for automated searches. We're talking about an exception here, which we notify the user on. We notify once per release, and then the user can make his own decision based on the release itself.?

medusa/tv/series.py
```diff
@@ -2160,11 +2162,20 @@ def want_episode(self, season, episode, quality, forced_search=False, download_c
...
             notifiers.notify_leaked(notify_message)
+
+            # Update notify flag
+            main_db_con.action(b'UPDATE tv_episodes '
+                               b'SET notified_earlier_release = 1 '
+                               b'WHERE '
+                               b'  showid = ? '
+                               b'  AND season = ? '
+                               b'  AND episode = ?', [self.indexerid, season, episode])
+
```
fernandog: my only concern about this is that if we need to update episode object loaded from DB or not
No idea

p0psicles: Don't understand.

fernandog:
problem is that we are notifying for unwanted quality. Imagine I only want web-dl and its leaked as SD. At least I wouldn't want it

Anyway, its a POC. You all decided what to do with it

medusa/tv/series.py
fernandog: @p0psicles for tv_episodes we usually do `load_from_db` and load episode info into memory. Doing that UPDATE, i don't know if I also need to update the loaded episode from memory with the `notified_earlier_release = 1`

p0psicles:
Doesn't matter. We're only notifying the user. We're not snatching anything. Maybe a user want to watch it 'early' in SD. User can make still make that decision.

fernandog:
@p0psicles if that is the case, ditch this PR

fernandog:
keep the branch

PR #2831, opened 2017-05-27
Feature/parsedatetime

p0psicles:


tests/providers/test_generic_provider.py
```diff
@@ -21,7 +21,7 @@
     },
     {  # p1: date and time
         'pubdate': '2017-05-18 15:00:15',
-        'expected': datetime(2017, 5, 18, 15, 0, 15, tzinfo=tz.gettz('UTC'))
+        'expected': datetime(2017, 5, 18, 15, 0, 15, tzinfo=tz.tzlocal())
```
p0psicles: @medariox this is where I got really confused. Why would we want to use a default timezone of UTC, if we can also use the users, local timezone?

tests/providers/test_generic_provider.py
```diff
@@ -39,23 +39,48 @@
...
+    actual = sut.parse_pubdate(parsed_date, provided_timezone=tzone, human_time=ht)
 
     # Calculate the difference for human date comparison
-    if ht:
-        actual = (datetime.now(tz.tzlocal()) - actual).seconds
-
-    # Then
-    assert expected == actual
+    if ht and actual:
+        actual = int((datetime.now(tz.tzlocal()) - actual).total_seconds())
```
p0psicles: here you where already using local timezone.

PR #2833, opened 2017-05-28
Added aliasses for 'right now' 'just now'.. etc.

p0psicles:
* Fixed issue where .seconds() on 24 hours ago would test as 0. Because .seconds resets after one day. Used total_seconds() instead.
* Added test for failing assertion on human times.

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -493,7 +493,7 @@ class GenericProvider(object):
 
         :returns: a datetime object or None
         """
-        now_alias = ['right now', 'just now', 'now']
+        now_alias = ('right now', 'just now', 'now')
 
         # This can happen from time to time
         if pubdate is None:
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -493,6 +493,8 @@ class GenericProvider(object):
 
         :returns: a datetime object or None
         """
+        now_alias = ['right now', 'just now', 'Now']
+
         # This can happen from time to time
         if pubdate is None:
             log.debug('Skipping invalid publishing date.')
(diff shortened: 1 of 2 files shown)


tests/providers/test_generic_provider.py
```diff
@@ -54,8 +84,8 @@ def test_parse_pubdate(p):
     actual = sut.parse_pubdate(parsed_date, human_time=ht, timezone=tzone)
 
     # Calculate the difference for human date comparison
-    if ht:
-        actual = (datetime.now(tz.tzlocal()) - actual).seconds
+    if ht and actual:
+        actual = int((datetime.now(tz.tzlocal()) - actual).total_seconds())
```
p0psicles: I needed to use total_seconds(). As when using .seconds. It would result in `0` for exactly one day ago / 24 hours ago.

medusa/providers/generic_provider.py
```diff
@@ -493,15 +493,20 @@ def parse_pubdate(pubdate, human_time=False, timezone=None):
 
         :returns: a datetime object or None
         """
+        now_alias = ['right now', 'just now', 'Now']
```
ratoaq2: Probably it would be better to have all of them lowercase and use re.I to ignore case in the regex. This could be a tuple instead of a list as well

p0psicles: In the other branch where I did a more extensive refactor, I made the list extensible through param. I kind of copied the code, but didn't copy the extensibility. But in this case the list doesn't need to be mutable. So I can change that with the benefit of a small performance increase.

But to challenge that, I don't think there is really allot of performance increase in comparing a list of three strings.

tests/providers/test_generic_provider.py
```diff
@@ -42,20 +42,50 @@
...
+        'expected': 0,  # difference in seconds
+        'human_time': True
+    },
+    {  # p9: human now variant
+        'pubdate': 'Now',
+        'expected': 0,  # difference in seconds
+        'human_time': True
+    },
+    {  # p10: invalid human time string
+        'pubdate': 'This is not a valid hum readable date format!',
```
duramato: `hum`

p0psicles: Should be an invalid string anyway.

PR #2834, opened 2017-05-29
LTV:  Fix not finding new subtitles and empty archive for vip/donors

fernandog:
Also add some logs to better debug the search

All known issues and discussed with @ratoaq2 a couple of months ago

### DIFF
--- a/medusa/subtitle_providers/legendastv.py
+++ b/medusa/subtitle_providers/legendastv.py
@@ -274,11 +274,11 @@ class LegendasTVProvider(Provider):
         """
         logger.info('Getting archives for title %d and language %d', title_id, language_code)
         archives = []
-        page = 1
+        page = 0
         while True:
             # get the archive page
-            url = self.server_url + 'util/carrega_legendas_busca_filme/{title}/{language}/-/{page}'.format(
-                title=title_id, language=language_code, page=page)
+            url = self.server_url + 'legenda/busca/-/{language}/-/{page}/{title}'.format(
+                language=language_code, page=page, title=title_id)
             r = self.session.get(url)
             r.raise_for_status()
 
@@ -308,6 +308,8 @@ class LegendasTVProvider(Provider):
                     raise ProviderError('Archive timestamp is in the future')
 
                 # add archive
+                logger.info('Found archive for title %d and language %d at page %s: %s',
+                            title_id, language_code, page, archive)
                 archives.append(archive)
 
             # stop on last page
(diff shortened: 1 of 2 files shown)


fernandog:
@ratoaq2 as we talked, I removed those commit as my goal was only to get a review of the code itself

medusa/subtitle_providers/legendastv.py
```diff
@@ -51,6 +51,9 @@
 #: Cache key for releases
 releases_key = __name__ + ':releases|{archive_id}|{archive_name}'
 
+# Override so we can detect earlier if new shows were added to LTV
+SHOW_EXPIRATION_TIME = timedelta(hours=12).total_seconds()
```
ratoaq2: Really not sure about this one. Better to not change it in the scope of this PR

medusa/subtitle_providers/legendastv.py
```diff
@@ -246,7 +249,7 @@ def search_titles(self, title):
                     if match:
                         title['season'] = int(match.group('season'))
                     else:
-                        logger.warning('No season detected for title %d', title_id)
+                        logger.debug('No season detected for title id: %d and name: %s', title_id, name)
```
ratoaq2: If this PR were without the log message changes, then we could easily PR upstream. 
This warning shouldn't change to debug

fernandog: I only found all this issues in LTV because of the logs.
We discard/continue loop without logging. Its hard to debug without it.


IMO there is no point showing as warning as there is nothing to do about it. Why warn user about that? Debug is fine

medusa/subtitle_providers/legendastv.py
fernandog: All subtitles from seasons premier wont be found because of this cache..

This is a critical issue IMO

p0psicles:
Closed as there is no future for this Pr here. Changes for upstream need to be done upstream. What's left we can discuss in new pr.

fernandog:
@p0psicles there is no changes upstream.

fernandog:
I'm still waiting @ratoaq2 because code skips subtitles and doesn't log it so we don't know why it was skipped. Talked to him in slack and add comment here but still no replied

p0psicles:
Please Pr upstream, not here. Rato was clear. It's not getting merged into Medusa like how it's now.

fernandog:
@p0psicles explained in slack. this PR is not in a upstream file. Its our custom file

fernandog:
@ratoaq2 now is synced with upstream PR with the requested changes

fernandog:
@ratoaq2 removed the ltv maintenance commit as per the other upstream PR

Issue #2596, opened 2017-04-15
[APP SUBMITTED]: TypeError: int() argument must be a string or a number, not 'NoneType'


labrys:
### INFO
**Python Version**: `2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)]`
**Operating System**: `Windows-10-10.0.14393`
**Locale**: `cp1252`
**Branch**: [file-size-log](../tree/file-size-log)
**Database**: `44.8`
**Commit**: pymedusa/Medusa@5db71ce532f073b0cd9807865388a1db74ca2388
**Link to Log**: https://gist.github.com/bd56768c9b46611944e846beb60378e2
### ERROR
<pre>
2017-04-15 08:52:38 ERROR    POSTPROCESSOR :: [5db71ce] Exception generated: int() argument must be a string or a number, not 'NoneType'
Traceback (most recent call last):
  File "H:\usr\local\medusa\medusa\scheduler.py", line 103, in run
    self.action.run(self.force)
  File "H:\usr\local\medusa\medusa\auto_post_processor.py", line 57, in run
    process_tv.ProcessResult(app.TV_DOWNLOAD_DIR, app.PROCESS_METHOD).process(force=force)
  File "H:\usr\local\medusa\medusa\process_tv.py", line 162, in process
    ignore_subs=ignore_subs)
  File "H:\usr\local\medusa\medusa\process_tv.py", line 336, in process_files
    self.process_media(path, self.video_files, force, is_priority, ignore_subs)
  File "H:\usr\local\medusa\medusa\process_tv.py", line 577, in process_media
    self.result = processor.process()
  File "H:\usr\local\medusa\medusa\post_processor.py", line 1075, in process
    if int(ep_obj.season) > int(max_season[0][0]):
TypeError: int() argument must be a string or a number, not 'NoneType'
</pre>
---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

medariox:
The show has no seasons?

labrys:
No it was caused by a race condition when a series was removed, it crashed the PP though.

duramato:
### INFO
**Python Version**: `2.7.9 (default, Sep 17 2016, 20:26:04) [GCC 4.9.2]`
**Operating System**: `Linux-4.4.38-v7+-armv7l-with-debian-8.0`
**Locale**: `UTF-8`
[...]

PR #2836, opened 2017-05-29
Fix TypeError: int() argument must be a string or a number, not 'NoneType'

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2596#issuecomment-301526508

### DIFF
--- a/medusa/search/manual.py
+++ b/medusa/search/manual.py
@@ -205,8 +205,8 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
             [cur_provider.get_id()]
         )
         columns = [i[1] for i in main_db_con.select("PRAGMA table_info('{0}')".format(cur_provider.get_id()))] if table_exists else []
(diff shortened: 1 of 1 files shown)


medusa/search/manual.py
```diff
@@ -205,8 +205,8 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
             [cur_provider.get_id()]
         )
         columns = [i[1] for i in main_db_con.select("PRAGMA table_info('{0}')".format(cur_provider.get_id()))] if table_exists else []
-        minseed = int(cur_provider.minseed) if hasattr(cur_provider, 'minseed') else -1
-        minleech = int(cur_provider.minleech) if hasattr(cur_provider, 'minleech') else -1
+        minseed = int(cur_provider.minseed) if hasattr(cur_provider, 'minseed') and cur_provider.minseed  else -1
```
p0psicles: i think it's  better to change this with: `.. if getattr(cur_provider, 'minseed', None)`

fernandog: ok!

PR #2840, opened 2017-05-30
Fix warnings about deprecated show

fernandog:


fernandog:
no one is getting this warnings?!

Issue #2652, opened 2017-04-21
Not detecting episodes


bnevets27:
Branch: master
Commit: a0116154f262786d6f5274459cb8f9cdb23012cc
OS:Linux

Not showing episode in medusa. 
```
# guessit: 2.1.2.dev0  rebulk: 0.8.3.dev0
? /tv/Inventions that Shook the World/Season 01/Inventions that Shook the World - S01E01 - The 1900s.wtv
: title: Inventions that Shook the World
  season: 1
  episode: 1
  episode_title: The 1900s wtv
  type: episode
  parsing_time: 0.158881902695
```
Debug log:
```
2017-04-21 19:12:16 DEBUG    SHOWQUEUE-REFRESH :: [a011615] 258368: Not changing current status 'Unaired' based on file: /tv/Inventions that Shook the World/Season 01/Inventions that Shook the World - S01E01 - The 1900s.wtv. Reason: New file is the same as current file
2017-04-21 19:12:16 DEBUG    SHOWQUEUE-REFRESH :: [a011615] 258368: Setter sets location to /tv/Inventions that Shook the World/Season 01/Inventions that Shook the World - S01E01 - The 1900s.wtv
2017-04-21 19:12:16 DEBUG    SHOWQUEUE-REFRESH :: [a011615] 258368: /tv/Inventions that Shook the World/Season 01/Inventions that Shook the World - S01E01 - The 1900s.wtv parsed to Inventions that Shook the World S01E01
2017-04-21 19:12:16 DEBUG    SHOWQUEUE-REFRESH :: [a011615] Using cached parse result for '/tv/Inventions that Shook the World/Season 01/Inventions that Shook the World - S01E01 - The 1900s.wtv'
2017-04-21 19:12:16 DEBUG    SHOWQUEUE-REFRESH :: [a011615] 258368: Creating episode object from /tv/Inventions that Shook the World/Season 01/Inventions that Shook the World - S01E01 - The 1900s.wtv
2017-04-21 19:12:16 DEBUG    SHOWQUEUE-REFRESH :: [a011615] 258368: Creating episode from: /tv/Inventions that Shook the World/Season 01/Inventions that Shook the World - S01E01 - The 1900s.wtv
```

medariox:
The issue is here: https://github.com/pymedusa/Medusa/blob/master/medusa/tv/series.py#L1190-L1194

If the quality is Unknown the episode won't be refreshed.

fernandog:
By design.

[...]

PR #2841, opened 2017-05-30
Don't assume same_file if size is 0 (unable to get size)

fernandog:
~~Possible fix to: https://github.com/pymedusa/Medusa/issues/2652~~

From his logs:
`SHOWQUEUE-REFRESH :: [a011615] 258368: Not changing **current status 'Unaired'** based on file: /tv/Inventions that Shook the World/Season 01/Inventions that Shook the World - S01E01 - The 1900s.wtv`
**Reason: New file is the same as current file**

if status is UNAIRED, then episode object file size is 0, and if we can't get a size, size will also be 0, thus the reason above showing both have the same size (which is wrong)

He copied the file to show folder without using the PP and hit re-scan files

### DIFF
--- a/medusa/tv/series.py
+++ b/medusa/tv/series.py
@@ -1317,9 +1317,13 @@ class Series(TV):
 
                 with cur_ep.lock:
                     old_size = cur_ep.file_size
+
+                    # Setting a location to cur_ep, we will get the size of the filepath
                     cur_ep.location = filepath
+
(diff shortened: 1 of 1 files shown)


fernandog:
@medariox even the size can be unrelated in that issue, we can't assume the flag `same_size` as true if size is 0.

fernandog:
thanks. i can't merge btw...

Issue #2830, opened 2017-05-27
Medusa not finding season packs


famulor:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:** Master/9c19b10731765197136e042f30ebca793414cac1
**OS:** Windows
**What you did:** Searched for season packs
**What happened:** Nothing. Skips to episode search
**What you expected:** Finding season packs :D 
**Logs:**

Btw im not sure thats the log you want but its the only one in /data/logs

Logs: [application2.txt](https://github.com/pymedusa/Medusa/files/1033528/application2.txt)

medariox:
Issue: Backlog search for series (Show S02) ignores results which should be kept.

fernandog:
@medariox are you going to debug this?

p0psicles:
Is this for a specific search. Or every season pack search?

medariox:
@fernandog 
No, I was only adding more information to the issue.

@p0psicles 
It's for every season pack search.

famulor:
https://gist.github.com/famulor/36ebee8705c1ee5e330e333d8075b0c5 @fernandog

PR #2843, opened 2017-05-30
Fix season pack forced search

fernandog:
Fixes: https://github.com/pymedusa/Medusa/issues/2830

Also fixes the providers hammering in a season search.
In a season search it searches for same string (Show Sxx) N times (where N is the number of episodes for that season)

Logs:
```
SEARCHQUEUE-BACKLOG-257655 :: [MoreThanTV] :: [b4a8a54] All episodes in this season are needed, downloading torrent Show.S05.1080p.WEBRip.DD5.1.x264-P2P
SEARCHQUEUE-BACKLOG-257655 :: [b4a8a54] Downloading Show.S05.1080p.WEBRip.DD5.1.x264-P2P with 9 seeders and 1 leechers and size 100.48 GB from MoreThanTV
```

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -186,7 +186,8 @@ class GenericProvider(object):
         results = {}
         items_list = []
 
-        for episode in episodes:
+        # Need to be sorted because in a season search we only need episode 01 to get season search string
+        for episode in sorted(episodes, key=lambda k: k.episode):
             if not manual_search:
                 cache_result = self.cache.search_cache(episode, forced_search=forced_search,
                                                        down_cur_quality=download_current_quality)
@@ -207,6 +208,10 @@ class GenericProvider(object):
             for search_string in search_strings:
                 # Find results from the provider
                 items_list += self.search(search_string, ep_obj=episode)
+
+            # In season search, we can't loop in episodes lists as we only need one episode to get the season string
+            if search_mode == 'sponly':
+                break
 
         if len(results) == len(episodes):
             return results
@@ -379,8 +384,9 @@ class GenericProvider(object):
                           search_result.name, search_result.url)
                 continue
 
-            if not manual_search:
+            if not (manual_search or search_mode == 'sponly'):
                 # The second check, will loop through actual_episodes and check if there's anything useful in it.
+                # If is a season pack search, we only have the season, not the episodes.
                 if not search_result.check_episodes_for_quality(forced_search, download_current_quality):
                     log.debug('Ignoring result {0}', search_result.name)
                     continue
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -323,12 +323,6 @@ class GenericProvider(object):
                             )
                             search_result.result_wanted = False
                             continue
-
-                    # We've performed some checks to decided if we want to continue with this result.
-                    # If we've hit this, that means this is not an air_by_date and not a sports show. And it seems to be
-                    # a valid result. Let's store the parsed season and episode number and continue.
-                    search_result.actual_season = search_result.parsed_result.season_number
-                    search_result.actual_episodes = search_result.parsed_result.episode_numbers
                 else:
                     # air_by_date or sportshow.
                     search_result.same_day_special = False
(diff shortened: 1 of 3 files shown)


fernandog:
user tested and reported it's working

medusa/providers/generic_provider.py
```diff
@@ -379,8 +384,9 @@ def find_search_results(self, show, episodes, search_mode, forced_search=False,
                           search_result.name, search_result.url)
                 continue
 
-            if not manual_search:
+            if not (manual_search or search_mode == 'sponly'):
```
medariox: This will need to be adapted/extended to check for the season quality.

fernandog: @medariox Did you tried or just looked the code?

p0psicles:
Your referring to a class, not a single method.
The reason why I did that refactor (as you could also read in the PR, that was responsible for that change) was to 1. Make it better readabe. 2. Make sure all results where added to cache, in stead of a the results that where at the time not usable. 3. Make sure we reduce the amount of name_parse actions. 4. store parse results to the release object, so we could more easily re-use them. Well those are some of the reasons that come to mind.

I tested some season packs back then. That seemed to work. I also asked super and you to test it, and from what I remember everything worked.

But of course we could have missed something. The change you made hear feels like a workaround. So I suggest to do some more looking into, before changing the status of this PR to concluded.

One last thing. Don't wave away a feedback. That's not how we do things here. I agree with @medariox.

p0psicles:
IF there is not a reason for the quality check (could be), then we should at least not add exceptions. We should address the issue.

fernandog:
I wont explain again about quality check as I proved with logs that it still does. 
If this solution is not good too bad
Feel free to expand/adapt and check twice the quality

p0psicles:
Digging a little deeper it seems that we use two quality gates, for determining if we want to accept a Search result. Where the search result can result in downloading a new episode or replacing one.

## Per provider
**Summary**
The first gate, is one used to filter out releases. It's run per provider. Where a provider can do a number of searches. All search results are added, and looped through in: find_search_results().

At this stage, additionally results that meet the criteria as well as the results that do not fit the searched episode or season (or other criteria). But do belong to a show, that's in the users library, will be added to the cache.

**Quality checks:**
* if the searched item is wanted. Only applies to Episodes. As season pack searches, include multiple episodes.
* if we have enough information from the parse result
* if the search results show / season / episode match the searched ep. Also accounts for scene numbering / scene naming.
* if the result is within the allowed_qualities, preferred qualities.
* if one of the resulting episodes matches an episodes in db. And if that episode has a status of WANTED.
* if one the resulting episodes, needs replacing. For example, we have downloaded an episode in 720p, and this result is in 1080p. (only applies to episodes and multi-eps). Not season packs!

##Selecting the best result
**Summary**
Out of the thinned out results from all providers, again we are selecting the best result.

**Quality checks:**
* if the result is within the allowed_qualities, preferred qualities.
* For anime shows, when enabled, if the release group is valid (in whitelist, not blacklist).
* For torrent results, when enabled, result leechers / seeders match.
* if enabled, result does not have any ignored words.
* if enabled, result has a required word.
* if the result matches any of the "badReleaseName" keywords, like 'dirfix', 'nfofix'. (Don't know how usefull this is actually).
* If failed download enabled (and a size is available). Check if the search result, was downloaded before and failed.
* And last it filters out the best quality available. Including: propers, 'internals', x264 over xvid, undesired words.

p0psicles:
@fernandog I tried documenting the search quality gates. And as you can see, some have a little overlap. Like checking if the search result meets the quality in the allowed_qualities, preferred qualities lists.

But some checks are not duplicate. So we should think about what belongs where. And how this fits in evaluating the season pack results.

fernandog:
Nice job.
Agree
In a season pack we already loop th all episodes and make sure we want all.
This is done in pick best result.

When I did the quality PR I fixed some of overlaps. Maybe I missed that one in find search results (imo that method doesnt need to check quality.  pick best result that should do It imo)

To filter a quality we need to check all available qualities and then decide. We cant decide in a provider scope

p0psicles:
I think that next thing we should do, is check if we can remove the check on the episodes, as opposed to build in an exception for checking the season packs, in the per provider quality gates.

fernandog:
@p0psicles I have a backup of you commit

medusa/providers/generic_provider.py
```diff
@@ -318,12 +323,6 @@ def find_search_results(self, show, episodes, search_mode, forced_search=False,
                             )
                             search_result.result_wanted = False
                             continue
-
-                    # We've performed some checks to decided if we want to continue with this result.
-                    # If we've hit this, that means this is not an air_by_date and not a sports show. And it seems to be
-                    # a valid result. Let's store the parsed season and episode number and continue.
-                    search_result.actual_season = search_result.parsed_result.season_number
-                    search_result.actual_episodes = search_result.parsed_result.episode_numbers
```
fernandog: @p0psicles we already set that in line 273

PR #2844, opened 2017-05-31
LTV: Fix KeyError in custom provider

fernandog:
I don't have any missed subtitles so only noticed it now

@ratoaq2 @duramato

```
2017-05-31 09:37:02 WARNING  FINDSUBTITLES :: [68a0049] Unexpected error in provider 'legendastv', discarding it
  File "/home/osmc/Medusa/lib/subliminal/core.py", line 171, in download_subtitle
    self[subtitle.provider_name].download_subtitle(subtitle)
  File "/home/osmc/Medusa/lib/subliminal/core.py", line 64, in __getitem__
    raise KeyError
KeyError
```

Proof:
image


fernandog:
@ratoaq2 thanks for replying. The 'subliminal sync' PR is not in master.

Also I'm doing another commit into this PR

fernandog:
@ratoaq2 
should we add something like this to explain users about the "legendastv2" ?

image

medariox:
@fernandog 
I would remove the note about "Legendatv2". The users don't need to know about that.

fernandog:
removed noted

PR #2845, opened 2017-05-31
Add TheWiz custom provider for Hebrew subtitles

fernandog:
Code from @ofir123 that PRed against Subliminal

After Subliminal merge that PR we can remove the custom provider


fernandog:
@medariox like this ?

PR #2855, opened 2017-06-02
Fix undesired words when the first item has undesired words

fernandog:
In the first loop, best_result = cur_result, so if the first item has a undesired word, we don't log as undesired release when we should

Undesired words: DRACULA AND P2P

**Without the fix:**
SEARCHQUEUE-BACKLOG-257655 :: [MoreThanTV] :: [b4a8a54] Picking the best result out of [u'Show.S05.1080p.WEB-DL.DD5.1.H264-DRACULA', u'Show.S05.1080p.WEBRip.DD5.1.x264-P2P']
SEARCHQUEUE-BACKLOG-257655 :: [MoreThanTV] :: [b4a8a54] Unwanted release Show.S05.1080p.WEBRip.DD5.1.x264-P2P (contains undesired word(s))

**With the fix:**
SEARCHQUEUE-BACKLOG-257655 :: [MoreThanTV] :: [b4a8a54] Picking the best result out of [u'Show.S05.1080p.WEB-DL.DD5.1.H264-DRACULA', u'Show.S05.1080p.WEBRip.DD5.1.x264-P2P']
**SEARCHQUEUE-BACKLOG-257655 :: [MoreThanTV] :: [b4a8a54] Unwanted release Show.S05.1080p.WEB-DL.DD5.1.H264-DRACULA (contains undesired word(s))**
SEARCHQUEUE-BACKLOG-257655 :: [MoreThanTV] :: [b4a8a54] Unwanted release Show.S05.1080p.WEBRip.DD5.1.x264-P2P (contains undesired word(s))

### DIFF
--- a/medusa/search/core.py
+++ b/medusa/search/core.py
@@ -327,7 +327,8 @@ def pick_best_result(results, show):  # pylint: disable=too-many-branches
             elif u'xvid' in best_result.name.lower() and u'x264' in cur_result.name.lower():
                 log.info(u'Preferring {0} (x264 over xvid)', cur_result.name)
                 best_result = cur_result
-            if any(ext in best_result.name.lower() for ext in undesired_words) and not any(ext in cur_result.name.lower() for ext in undesired_words):
+            if any(ext in best_result.name.lower() for ext in undesired_words) and \
+                    (best_result == cur_result or not any(ext in cur_result.name.lower() for ext in undesired_words)):
                 log.info(u'Unwanted release {0} (contains undesired word(s))', cur_result.name)
                 best_result = cur_result
 


labrys:
The first result is never undesired.  Remember undesired words is not ignored words, a release is allowed to have undesired words.  A release without undesired words is preferred over a release that has them. 
 Until an iteration hits a result without undesired words, undesired words are perfectly fine.  As soon as you get a result without undesired words, all candidates after that would then be required to be free of all undesired words.

fernandog:
I think my English is bad: as I stated in the PR " we don't log as undesired release when we should"

If you check my logs without the fix **we don't know release had undesired words**

labrys:
If you only want to log if a release has undesired words, that doesn't need to change the logic of the check.  Instead simplify the whole thing like so:
```python
best_result_undesired = any(ext in best_result.name.lower() for ext in undesired_words)
current_result_undesired = any(ext in current_result.name.lower() for ext in undesired_words)
if best_result_undesired and not current_result_undesired:
```

fernandog:
so why preferred words doesn't have that long check then?

labrys:
It does, only it's just a half-check, it just checks if the `current_result` has preferred words.
```python
if any(ext in cur_result.name.lower() for ext in preferred_words):
```
Personally I think that should be corrected to:
```python
if any(ext in cur_result.name.lower() for ext in preferred_words) and not any(ext in best_result.name.lower() for ext in preferred_words):
```

labrys:
Also I would probably turn the check into a generator function:
```python
def contains_words(item, words):
    item = item.lower()
    for word in words:
        if word in item:
            yield word
```
This would generate words as they are found in the target item.  Thus you could either stop on the first found or iterate through the full list to generate a list of all words in the  item.
Example usage:
```python
if any(contains_words(best_result.name, undesired_words)) and not any(contains_words(cur_result.name, undesired_words)):
```
or
```python
best_undesired = list(contains_words(best_result.name, undesired_words))
cur_undesired = list(contains_words(best_result.name, undesired_words))
if len(best_undesired) < len(cur_undesired):
    log.debug('Preferring result {cur_result.name} with undesired words {cur_undesired} over {best_result.name} with undesired words {best_result.undesired})
    best_result = cur_result
```

fernandog:
What if two releases have both one undesired words? Or one has two and other has three?

labrys:
Current check doesn't care (and never has) my last suggestion handles that with `len(undesired)`.

labrys:
If I was going to use `len(found_words)` for the check i would probably make `found_words` a set.

fernandog:
I will close this as current code is bad and need some complex changes to me

PR #2858, opened 2017-06-03
Feature/contains words

labrys:
Lay groundwork for updating words checks (ignored, required, preferred, undesired).

### DIFF
--- a/medusa/show/naming.py
+++ b/medusa/show/naming.py
@@ -16,7 +16,7 @@ from medusa.name_parser.parser import InvalidNameException, InvalidShowException
 from six import string_types
 
 log = BraceAdapter(logging.getLogger(__name__))
-log.logger.addHandler(logging.NullHandler)
+log.logger.addHandler(logging.NullHandler())
 
 
 resultFilters = [
--- a/medusa/show/naming.py
+++ b/medusa/show/naming.py
@@ -2,6 +2,8 @@
 
 """Series naming helpers for selecting results."""
 
+from __future__ import unicode_literals
+
 import logging
 import fnmatch
 import os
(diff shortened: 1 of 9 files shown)


medusa/show/naming.py
```diff
@@ -0,0 +1,159 @@
...
+]
+
+if hasattr('General', 'ignore_und_subs') and app.IGNORE_UND_SUBS:
+    resultFilters.append('sub(bed|ed|pack|s)')
+
+if hasattr('General', 'ignored_subs_list') and app.IGNORED_SUBS_LIST:
+    resultFilters.append('(' + app.IGNORED_SUBS_LIST.replace(',', '|') + ')sub(bed|ed|s)?')
+
+
+def contains_words(item, words, strict=True):
```
p0psicles: I think you should add unit tests for these.

medusa/show/naming.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # select strict or lenient method for the test
+    item_contains = _strict if strict else _lenient
+
+    for word in words:
+        if item_contains(word):
+            yield word
+
+
+def contains_at_least_one_word(name, words):
```
p0psicles: I think you should add unit tests for these.

labrys: This will be deleted.  Its the old method that I'm replacing.

medusa/show/naming.py
p0psicles: Also going through the function. I can't really justify returning a generator in stead of a list. So that's adding a little complexity without any obvious reason.

labrys: Generator instead of function is so you can choose how to use the return value.  For example if you want a set instead of a list:
```python
words = set(contains_words(item, words))
```
If you want to short-circuit early to avoid processing all words (e.g. break on first match)
```
has_words = any(contains_words(item, words))
```
If you want to compare words in sequence between two items:
```python
item1_word = contains_words(item1, words)
item2_word = contains_words(item2, words)
if next(item1_word) and not next(item2_word):
    ...
```
etc.

Issue #2859, opened 2017-06-04
Medusa Not Detecting Quality


bnevets27:
The following file is marked as unknown quality and therefore not added to medusa even though there is SDTV quality in the filename.

Season 01/Bosch - S01E01 - Tis the Season [SDTV][SickRage].mp4
 
I also think medusa should add files without known quality. The files would be labeled unknown. This part ties in with this other issue: https://github.com/pymedusa/Medusa/issues/2652

Commit: 2a21fd430ad688cd633f8929d2fb90cd78362e9e
Version: v0.1.13

```
2017-06-04 13:35:43 DEBUG    SHOWQUEUE-ADD :: [2a21fd4] 277928: Creating episode from: /tv/Bosch/Season 01/Bosch - S01E01 - Tis the Season [SDTV][SickRage].mp4
2017-06-04 13:35:43 DEBUG    SHOWQUEUE-ADD :: [2a21fd4] 277928: Creating episode object from /tv/Bosch/Season 01/Bosch - S01E01 - Tis the Season [SDTV][SickRage].mp4
2017-06-04 13:35:43 DEBUG    SHOWQUEUE-ADD :: [2a21fd4] Scene numbering enabled series Bosch using indexer numbering: S01E01
2017-06-04 13:35:43 DEBUG    SHOWQUEUE-ADD :: [2a21fd4] Converted parsed result /tv/Bosch/Season 01/Bosch - S01E01 - Tis the Season [SDTV][SickRage].mp4 into format: tv, title: Bosch, season: 1, episode: [1], episode_title: Tis the Season, release_group: SickRage, container: mp4, mimetype: video/mp4, type: episode, parsing_time: 0.0819518566132, total_time: None, absolute_episode: [], quality: Unknown
2017-06-04 13:35:43 DEBUG    SHOWQUEUE-ADD :: [2a21fd4] Matched release /tv/Bosch/Season 01/Bosch - S01E01 - Tis the Season [SDTV][SickRage].mp4 to a series in your database: Bosch
[...]

PR #2866, opened 2017-06-08
Allow refresh of UNKNOWN qualities

fernandog:
Related to: https://github.com/pymedusa/Medusa/issues/2859

tests may fail but I will handle it later

### DIFF
--- a/medusa/tv/series.py
+++ b/medusa/tv/series.py
@@ -1217,7 +1217,7 @@ class Series(TV):
             if new_quality != Quality.UNKNOWN:
                 return True, 'New file has different name from the database but has valid quality.'
             else:
-                return False, 'New file has UNKNOWN quality'
+                return True, 'New file has different name from the database and an UNKNOWN quality.'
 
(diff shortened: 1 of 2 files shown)


p0psicles:
I'm okay with the change. But I can't really oversee  the consequences. @labrys @medariox any good reason not to merge this?

medusa/tv/series.py
```diff
@@ -1217,7 +1217,7 @@ def should_refresh_file(cur_status, same_file, check_quality_again, anime, filep
             if new_quality != Quality.UNKNOWN:
                 return True, 'New file has different name from the database but has valid quality.'
             else:
-                return False, 'New file has UNKNOWN quality'
+                return True, 'New file has different name from the database and an UNKNOWN quality.'
```
p0psicles: @fernandog shouldn't we make this consistent with the should_replace? As now, when I have an UNKNOWN quality file downloaded locally, it will not upgrade to an "allowed" quality if found.

Issue #2879, opened 2017-06-13
Cannot delete custom newznab provider


Kilay:
### Javascript error when trying to delete a custom newznab provider:

**Branch/Commit:** master

**OS:** Server : Linux (Synology) - Client : Windows 10, Chrome

**What you did:**
From Config -> Search Providers -> Configure Custom Newznab Providers, I select a custom provider and click on delete

**What happened:**
I get the following javascript error

**What you expected:**
Deletion of the provider

**Logs from Chrome developer console :**
```
Uncaught TypeError: $(...).removeOption is not a function
    at ma.fn.init.$.fn.deleteProvider (config-providers.js:114)
    at HTMLInputElement.<anonymous> (config-providers.js:472)
    at HTMLInputElement.dispatch (vender.min.js?14072:2)
    at HTMLInputElement.q.handle (vender.min.js?14072:2)
```

fernandog:
Fixed in develop

PR #2882, opened 2017-06-13
Fix TypeError: $(...).removeOption is not a function

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2879


fernandog:
@p0psicles can this be in this release?

fernandog:
@OmgImAlexis why I can't do this?

✖  114:35  Unexpected string concatenation of literals.  no-useless-concat
✖  130:38  Unexpected string concatenation of literals.  no-useless-concat

OmgImAlexis:
@fernandog why would you need to concat two strings? It warns you since it makes no sense todo so.

PR #2883, opened 2017-06-14
Feature/fix proper search

p0psicles:

* Removed the Proper class. And used SearchResult class in stead.
* Added the search_type attribute to the SearchResult class. This can be used to keep track of the search, that resulted in the search result.
Added PROPER search_type.
* Moved the search types to search/__init__.py
* Removed show parameter from pick_best_result, as the passed search_result object, should always have the show object.
* Refactor proper.py to make sure it's utilizing the SearchResult class.
* Refactor daily search, to make sure that results from cache, are utilizing the SearchResult class.
* Fixed a logging bug, which was swallowing exceptions generate through search.
* Fixed shanaproject daily search

### DIFF
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -327,7 +327,7 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
             # then add it to our list of propers
             if best_result.indexerid != -1 and (
                 best_result.indexerid, best_result.actual_season, best_result.actual_episodes[0]
-            ) not in map(operator.attrgetter('indexerid', 'season', 'episode'), final_propers):
+            ) not in map(operator.attrgetter('indexerid', 'actual_season', 'actual_episode'), final_propers):
                 logger.log('Found a desired proper: {name}'.format(name=best_result.name))
                 final_propers.append(best_result)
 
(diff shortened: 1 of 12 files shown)


medusa/tv/cache.py
```diff
@@ -502,50 +504,54 @@ def find_needed_episodes(self, episode, forced_search=False,
...
             # build a result object
-            title = cur_result[b'name']
-            url = cur_result[b'url']
+            search_result.name = cur_result[b'name']
+            search_result.url = cur_result[b'url']
 
             log.debug(
                 '{id}: Using cached results from {provider} for series {show_name!r} episode {ep}', {
-                    'id': ep_obj.series.indexerid,
+                    'id': search_result.episodes[0].series.indexerid,
```
fernandog: @p0psicles why we need the `[0]` now?

p0psicles: As the ep_obj was kinda misused. It was the last looped episode object instance, that was misused in the rest of the code. I made it more explicit, in adding a list of epiosde_objects to the searchr_result.episodes variable (which for every item from cache is always 1). And then just use the first one.

fernandog: what if a cached result is multi-ep?

medusa/search/proper.py
```diff
@@ -201,53 +200,54 @@ def _get_proper_results(self):  # pylint: disable=too-many-locals, too-many-bran
...
+            cur_proper.quality = cur_proper.parse_result.quality
             cur_proper.content = None
-            cur_proper.proper_tags = parse_result.proper_tags
+            cur_proper.proper_tags = cur_proper.parse_result.proper_tags
+
+            # filter release, in this case, it's just a quality gate. As we only send one result.
+            best_result = pick_best_result(cur_proper)
 
-            # filter release
-            best_result = pick_best_result(cur_proper, parse_result.show)
```
fernandog: @p0psicles can we change the log message below this code `Rejected proper due to release filters` be changed to like `Rejected proper:` as we log the reason why it was rejected in pick_best_result and its not only because of release filters. can be rejected because of quality too

p0psicles: yeah we can

medusa/tv/cache.py
p0psicles: We didn't support that before. We don't now.

PR #2890, opened 2017-06-14
Fix knowit update

ratoaq2:

### DIFF
--- a/medusa/common.py
+++ b/medusa/common.py
@@ -433,12 +433,13 @@ class Quality(object):
         if not height:
             return Quality.UNKNOWN
 
+        height = int(height.magnitude)
+
         # TODO: Use knowledge information like 'resolution'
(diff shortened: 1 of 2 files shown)


medusa/common.py
```diff
@@ -438,6 +438,7 @@ def quality_from_file_meta(file_path):
         bluray = re.search(r"blue?-?ray|hddvd|b[rd](rip|mux)", base_filename, re.I) is not None
         webdl = re.search(r"web.?dl|web(rip|mux|hd)", base_filename, re.I) is not None
 
+        height = height.magnitude
```
labrys: This should probably moved up to just after `if height:` and should probably be type-cast to `int`.  Also for knowit, I would recommend a naive version that doesn't require the  pint library or an option to disable it.

Issue #2893, opened 2017-06-16
Bug? Location dropdown /home/ not working


wimpyrbx:
latest develop.

![image](https://user-images.githubusercontent.com/13415240/27241166-f298811c-52d7-11e7-91f5-52fbba3f23b9.png)

this solution that was added not long ago is not working. it always resets to "All Folders"

fernandog:
@OmgImAlexis 

since when this started?

`2017-06-20 17:23:41 WARNING  TORNADO :: [97e2852] Config patch ignored {u'selectedRootIndex': u'0'}`

fernandog:
@OmgImAlexis why this is not working anymore?

![image](https://user-images.githubusercontent.com/2620870/27357599-32cb25fa-55e9-11e7-9b12-9232f7bcc09d.png)

PR #2902, opened 2017-06-20
Fix config patch ignored {u'selectedRootIndex': u'0'}. Needs to be int

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2893

The value of the select is the index number of the root dirs list

Also fixes:
image

When selected root index is 0 it would set as None


fernandog:
Thanks

PR #2904, opened 2017-06-22
Use aliases while manual searching for subtitles

fernandog:
Fixes an old issue when it doesn't find subtitles because show in provider uses a different name

Examples:
`TURN` and `Turn: Washington's Spies`
`DC’s Legends of Tomorrow` and `Legends of Tomorrow`
`Marvels Agents of SHIELD` vs `Agents of SHIELD`
`Marvel's Jessica Jones` vs `Jessica Jones`
*Most Marvel's shows

It will also work for releases with translated releases like `La Lista Negra` for `The Blacklist`

@ratoaq2 @duramato

### DIFF
--- a/medusa/subtitles.py
+++ b/medusa/subtitles.py
@@ -286,6 +286,17 @@ def score_subtitles(subtitles_list, video):
     return [(s, score) for (s, score, provider_order) in result]
 
 
+def get_subtitles_using_alias(tv_episode, pool, video, languages, subtitles_list, scored_subtitles):
+    """List and score subtitles using series alias."""
+    for alias in tv_episode.series.aliases:
+        logger.debug(u'Searching subtitles using alias: %s', alias)
+        video.series = alias
+        subtitles_list.extend(pool.list_subtitles(video, languages))
+        # Calculate the score with new video
+        scored_subtitles.extend(score_subtitles(set(subtitles_list), video))
+    return subtitles_list, scored_subtitles
+
+
 def list_subtitles(tv_episode, video_path=None, limit=40):
     """List subtitles for the given episode in the given path.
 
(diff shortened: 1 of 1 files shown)


duramato:
Those are imdb alias, there are foreign ones. Scene exceptions aren't used?

fernandog:
@duramato as we talked the aka is in `imdb_info` and as the new imdb lib we don't have `akas`  anymore,

the aliases are the scene exceptions:
```python
    @property
    def aliases(self):
        """Return series aliases."""
        return self.exceptions or get_scene_exceptions(self.indexerid, self.indexer)
```

ratoaq2:
I don't want to be the bad guy... but this is called `brute force`

For each single episode for each single series for each subliminal provider you will execute multiple searches with all known names of that series. This is how websites decide to blacklist certain clients.

fernandog:
@ratoaq2 but we cache with dogpile right?

besides that, what do you propose to fix this issue?

medariox:
@fernandog 
The solution to this "problem" is to use subtitle providers that match subtitles with hashes (e.g. OpenSubtitles), instead of the file names. The most that can be done is asking the providers to add this functionality to their site/API.

fernandog:
We hit providers in the same way.
Most dont have API and we scrap then.
How with subs is different?

medariox:
@fernandog 
For providers we have no choice - there are different groups that release with different names, they are completely independent from the providers.
With subtitle providers the situation is completely different, we have many ways to tackle this issue:
1) Ask the providers to name their subtitles properly (as the original release)
2) Ask the providers to use hashes to find subtitles
3) Use subtitle providers that already support hashes
4) Rename releases like the subtitle provider expects them to be named

fernandog:
makes no sense what you said. translate to me @duramato lol

fernandog:
@ratoaq2 what now? only in manual search. ping @duramato - as least we can manual search and snatch it

ratoaq2:
Yore issue is that you shouldn't trigger list subtitles for every alias for every provider. 

Thesubdb uses only hash, and you will trigger the search n times. Other providers will return the correct subtitles in the first try, not needing to search further. 
Even for legendastv,  why searching TBBT if The Big Bang Theory already returned the correct results? Most of the aliases searches will never return the correct results. They are not the rule, but the exception. Why searching `Turn: Washington's Spies` in legendastv to get s01e02 if for s01e01 only `turn` returned subs? There's no reason to search for `Turn: Washington's Spies` for every single episode in legendastv

fernandog:
@ratoaq2 i get your point. so what should we do?

PR #2907, opened 2017-06-23
Snooze instead of sleep

labrys:
Use non-blocking `tornado.gen.sleep` instead of `time.sleep`.

### DIFF
--- a/medusa/db.py
+++ b/medusa/db.py
@@ -21,10 +21,11 @@ import os.path
 import re
 import sqlite3
 import threading
-import time
 import warnings
 
 from six import text_type
+import tornado.gen
+
 from . import app, logger
 from .helper.exceptions import ex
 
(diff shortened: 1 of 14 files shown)


p0psicles:
http://www.tornadoweb.org/en/stable/_modules/tornado/gen.html# sleep. I doubt this will be useful  like this. Reading docs it should be used in coroutines yielding a result.

But you probably have more experience using async then I do. 

```
Note that calling this function on its own does nothing; you must wait on the Future it returns (usually by yielding it).
```

labrys:
Supposedly in threads its supposed to sleep the current thread.  This is just a test to see how it works.

p0psicles:
I think the sleep command tells the event scheduler to wait an additional amount of time before it provides time to the coroutine again. That's why its tightly bound to coroutines.

labrys:
Coroutines do nothing on their own.  A coroutine must be driven.  Without something driving the coroutine it just returns a coroutine object (or in this specific case a future).  Coroutines can drive other coroutines, otherwise you have to drive them yourself (e.g. with a loop).

p0psicles:
Well with a coroutine I mean a driven coroutine.

PR #2908, opened 2017-06-23
LTV provider: Add debug logs when discarding title

fernandog:
@ratoaq2 with the changes you asked. If is ok, I will also PR upstream

### DIFF
--- a/medusa/subtitle_providers/legendastv.py
+++ b/medusa/subtitle_providers/legendastv.py
@@ -246,7 +246,7 @@ class LegendasTVProvider(Provider):
                     if match:
                         title['season'] = int(match.group('season'))
                     else:
-                        logger.warning('No season detected for title %d', title_id)
+                        logger.warning('No season detected for title %d (%s)', title_id, name)
 
             # extract year
             if year:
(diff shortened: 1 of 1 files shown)


medusa/subtitle_providers/legendastv.py
```diff
@@ -436,7 +445,7 @@ def query(self, language, title, season=None, episode=None, year=None):
                 for r in releases:
                     subtitle = LegendasTVSubtitle(language, t['type'], t['title'], t.get('year'), t.get('imdb_id'),
                                                   t.get('season'), a, r)
-                    logger.debug('Found subtitle %r', subtitle)
+                    logger.info('Found subtitle %r', subtitle)
```
ratoaq2: This is the only thing. That should still be debug in subliminal world. 
All the rest is great 👍

fernandog: Ok thanks

fernandog:
@ratoaq2 approve?

fernandog:
@ratoaq2 btw we need to revert the level mappigns for ltv and itasa or proper fix it as we told you in slack

PR #2910, opened 2017-06-23
LTV provider: Fix not finding news seasons (new titles ids)

fernandog:
@ratoaq2
As each season is a new title ID, we must cache title name together with season

If is good, I will also PR upstream

UPDATE:
second commit logs cached titles instead of show nothing
DEBUG    Thread-16 :: [743b471] Found 30 cached titles for TURN


fernandog:
@ratoaq2 about the cached empty result:
`DEBUG    Thread-16 :: [743b471] Found 0 cached titles for turn washington's spies`

Shouldn't we don't cache empty titles results?
Otherwise the cache issue will persist

medusa/subtitle_providers/legendastv.py
```diff
@@ -52,6 +52,12 @@
 releases_key = __name__ + ':releases|{archive_id}|{archive_name}'
 
 
+#: Check if the value should actually be cached in dogpile or not
+def should_cache_titles(titles):
+    """Return False if title is an empty dict to avoid caching it."""
+    return titles != {}
```
ratoaq2: Why not just returning `titles` or `bool(titles)`?

medusa/subtitle_providers/legendastv.py
```diff
@@ -200,17 +206,23 @@ def terminate(self):
 
         self.session.close()
 
-    @region.cache_on_arguments(expiration_time=SHOW_EXPIRATION_TIME)
-    def search_titles(self, title):
+    @region.cache_on_arguments(expiration_time=SHOW_EXPIRATION_TIME, should_cache_fn=should_cache_titles)
+    def search_titles(self, title, season):
         """Search for titles matching the `title`.
 
+        For episodes, each season has it own title, so we must cache with season number
```
ratoaq2: Probably only the first phrase is enough.

medusa/subtitle_providers/legendastv.py
```diff
@@ -349,12 +361,23 @@ def download_archive(self, archive):
 
     def query(self, language, title, season=None, episode=None, year=None):
         # search for titles
-        titles = self.search_titles(sanitize(title))
+        titles = self.search_titles(sanitize(title), season)
+
+        titles_cache_key = __name__ + ':search_titles|{title} {season}'.format(title=title.lower(), season=season)
```
ratoaq2: These 3 lines can be removed, right?

medusa/subtitle_providers/legendastv.py
```diff
@@ -349,12 +361,23 @@ def download_archive(self, archive):
...
 
         # search for titles with the quote or dot character
         ignore_characters = {'\'', '.'}
         if any(c in title for c in ignore_characters):
-            titles.update(self.search_titles(sanitize(title, ignore_characters=ignore_characters)))
+            sanitized_title = sanitize(title, ignore_characters=ignore_characters)
+            additional_titles = self.search_titles(sanitized_title, season)
+            titles.update(additional_titles)
+
+            titles_cache_key = __name__ + ':search_titles|{title} {season}'.format(title=sanitized_title.lower(),
```
ratoaq2: I didn't follow why we need to interact directly with the cache api here

medusa/subtitle_providers/legendastv.py
```diff
@@ -352,10 +364,21 @@ def query(self, language, title, season=None, episode=None, year=None):
         sanitized_title = sanitize(title)
         titles = self.search_titles(sanitized_title)
```
ratoaq2: You should pass `season` here, right?

fernandog: rebase issue

medusa/subtitle_providers/legendastv.py
```diff
@@ -52,6 +52,12 @@
 releases_key = __name__ + ':releases|{archive_id}|{archive_name}'
 
 
+#: Check if the value should actually be cached in dogpile or not
+def should_cache_titles(titles):
```
ratoaq2: In the end this could be a simple lambda :

```
@region.cache_on_arguments(expiration_time=SHOW_EXPIRATION_TIME, should_cache_fn=lambda value: value)
```

That makes even simpler, right?

PR #2925, opened 2017-06-29
LTV: Use alias as fallback to original series name

fernandog:
Fixes an old issue when it doesn't find subtitles because show in provider uses a different name

Examples:
`TURN` and `Turn: Washington's Spies`
`DC’s Legends of Tomorrow` and `Legends of Tomorrow`
`Marvels Agents of SHIELD` vs `Agents of SHIELD`
`Marvel's Jessica Jones` vs `Jessica Jones`
*Most Marvel's shows

It will also work for releases with translated releases like `La Lista Negra` for `The Blacklist`

@ratoaq2 @duramato

PR is in sync with upstream

### DIFF
--- a/medusa/subtitle_providers/legendastv.py
+++ b/medusa/subtitle_providers/legendastv.py
@@ -122,8 +122,8 @@ class LegendasTVSubtitle(Subtitle):
         # episode
         if isinstance(video, Episode) and self.type == 'episode':
             # series
-            if video.series and (sanitize(video.series) in (
-                    sanitize(name) for name in [self.title] + video.alternative_series)):
+            if video.series and (sanitize(self.title) in (
+                    sanitize(name) for name in [video.series] + video.alternative_series)):
                 matches.add('series')
 
             # year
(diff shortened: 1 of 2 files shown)


fernandog:
@ratoaq2 waiting your approval

medusa/subtitle_providers/legendastv.py
```diff
@@ -122,7 +122,8 @@ def get_matches(self, video, hearing_impaired=False):
         # episode
         if isinstance(video, Episode) and self.type == 'episode':
             # series
-            if video.series and sanitize(self.title) == sanitize(video.series):
+            if video.series and (sanitize(video.series) in (
```
ratoaq2: This is wrong. It's inverted self e video

medusa/subtitle_providers/legendastv.py
```diff
@@ -136,7 +137,8 @@ def get_matches(self, video, hearing_impaired=False):
         # movie
         elif isinstance(video, Movie) and self.type == 'movie':
             # title
-            if video.title and sanitize(self.title) == sanitize(video.title):
+            if video.title and (sanitize(video.title) in (
```
ratoaq2: Same here

fernandog:
@ratoaq2 done. don't know why I did that mistake while cherry-picking your commit

PR #2930, opened 2017-06-30
LTV: Don't log unknown exception on 503 Service Unavailable

fernandog:
@ratoaq2 this change is part of a upstream PR


fernandog:
@ratoaq2 please?

medusa/subtitle_providers/legendastv.py
```diff
@@ -477,3 +477,12 @@ def download_subtitle(self, subtitle):
 
         # extract subtitle's content
         subtitle.content = fix_line_ending(subtitle.archive.content.read(subtitle.name))
+
+
+def raise_for_status(r):
+    # When site is under maintaince and http status code 200.
+    if 'Em breve estaremos de volta' in r.text or r.status_code == 503:
+        # For now raise Timeout until the upstream PR gets merged that catches ServiceUnavailable
+        raise Timeout
```
ratoaq2: This is misleading and an unnecessary hack

fernandog: what do you suggest as we can't raise a ServiceUnavailable?

fernandog: What is the requested change? Replace Timeout for what?

ratoaq2: The requested change is to fix the issue properly.

Don't expect always a solution from my side. I'm not working on this and I not in front of my ide with the system running to evaluate the possibilities. 
What I know is that a timeout is completely different to 503 or service unavailable. And all of this is just a nasty workaround because of the way we handle logging and user notifications.

fernandog: I couldn't find a way to avoid to catch it. 
Another solution is to raise `ProviderError` but that will show a traceback with `Unexpected error in provider`

PR #2932, opened 2017-07-01
Add requirements.txt for vanilla libraries

medariox:

This PR adds requirements.txt with the libraries that we currently use in their vanilla state (unmodified).
The goal is not to add automated installer capabilities (yet), but rather to organize our current libraries.
- The ext folder contains all vanilla libraries and their dependencies
- All libraries in requirements.txt have been updated to their latest versions
- Libraries in the lib folder are customized (can't be updated via pip or similar)
- Unused libraries have been removed

_About modified libraries:_
Most of these should be either replaced, or forked with our changes.
- adba [outdated, no py3 compatibility, waiting for [update](https://github.com/pymedusa/Medusa/pull/2439)]
- bencode [outdated, should be replaced]
- fake_useragent [move additional files to different place]
- libgrowl [outdated, should be replaced]
- pymediainfo [version needs to be hardcoded, needs workaround]
- pythontwitter [outdated, should be replaced]
- pytimeparse [code modified, needs fork]
- pytvmaze [outdated, code modified, needs fork]
- rtorrent [code modified, needs fork]
- send2trash [outdated, code modified, needs fork]
- shutil_custom [modified, evaluate if we need changes, needs fork]
- simpleanidb [modified, needs fork]
- synchronousdeluge [outdated, should be replaced]
- tmdbsimple [outdated, modified, needs fork and update]
- traktor [needs own repo]
- tvdbapiv2 [modified, needs repo update]
- unrar2 [modified, replace with rarfile]

_Vanilla libraries from GitHub:_
- babelfish [contains a bugfix, unreleased on pypi]
- cfscrape [versions > 1.7.1 require nodejs]
- feedparser [develop contains many fixes/improvements]
- python-oauth2 [contains fixes, adds py3 support]
- subliminal [already using develop, adds providers, fixes]

_Future goals:_
- Reduce libraries in lib folder as much as possible
- Replace/extend libraries that don't support py3
- Remove ext folder and replace with an automated setup
- Setup some kind of package monitoring (e.g. requires.io) in the meanwhile


fernandog:
@medariox will need to force a "git clean -df lib"

sharkykh:
@medariox 
Why did you remove `backports.ssl-match-hostname`? I say that because [you seem to allow running on Python versions under 2.7.9](https://github.com/pymedusa/Medusa/blob/4e7d532506cc81eb86bc4ced155c1ae1027e7c7c/medusa/init/__init__.py#L32-L35) (obviously `>= 2.7.0`)

medariox:
@sharkykh 
I've always been under the impression that we want to support Python >= 2.7.9. Our readme even [states Python 2.7.10](https://github.com/pymedusa/Medusa/blob/master/readme.md#dependencies), but tbh I'm not sure what's the exact plan. If there are good reasons to keep supporting such old Python versions (Python 2.7.8 is 3 years old), we can add it back of course. One of our goals is to have Python 3 compatibility in the future, and trying to support ancient Python versions sounds a bit like a contradiction to me. Not to mention that Python 2.7.9 contains critical security fixes. Users should be motivated to update to a more recent Python version.

sharkykh:
@medariox I don't think Medusa or "the previous one" really support any Python version below **2.7.10**.
And if your goal is to support Python 3, you (as a team, I mean) should explicitly **drop the support** for any Python version **below 2.7.10**.
I think it's really easy to update Python on **any platform** (not saying from experience, though).

labrys:
@medariox @sharkykh >= 2.7.9 was the target due to some NASs.  Many of those now have later versions available (and I've heard mix reports on updating).  However most of those same NASs do support at least some version of python 3.

p0psicles:
This is not something that's easily reviewable. Also there shouldn't been made any changes that have allot of impact. As I understand it's just allot of ordening. Shall we continue with merging this one?

medariox:
> As I understand it's just allot of ordening.

Is that the conclusion you have come to after reading the PR description?

p0psicles:
And after going through the commit description yes.

duramato:
For the libraries we are using the latest git version we can use that in the requirements.txt rather than the pypi version

medariox:
This PR is not a simple "moving around".
- Updates 28 libs (including essential ones like Tornado) to their latest versions
- Makes them updatable with a one line command
- Removes plenty of unused libs
- Keeps all libs that are custom and still need changes in the /lib folder
It is a major step into py3 support (as we can actually know which libraries still aren't compatible) and being able to unbundle the external libraries in the futures (with a package manager).

medariox:
@duramato 
That's already the case: https://github.com/pymedusa/Medusa/blob/use-requirements/requirements.txt

sharkykh:
@medariox imo you shouldn't use branch names, but rather commit hashes or tags. So you have control over what version is installed.
And not for that, then just so you always have the packages actual version/commit hash written down.

fernandog:
about this `subliminal [already using develop, adds providers, fixes]`:

all patches are inside medusa now, we can use upstream lib without patching it.

p0psicles:
My preference would be using tags as much as possible.

sharkykh:
@p0psicles Tags are usually interchangeable with PyPI releases, and PyPI's better and faster than installing from VCS. But commit hashes are the best specification for a package version installed from a VCS source.

medariox:
Tags are almost never available or they point to an extremely outdated commit, so they are not really a doable option. I agree we should use commit hashes instead of branches (like I did for cfscrape), thanks for pointing that out @sharkykh

labrys:
My opinion is to use tags in most cases.  The only use-case that I see for using commit hashes is for developers or targeting a specific issue for a user.  Proper tagging covers PyPi and covers VCS.

For example, list tags with `git tag -l`.  You can check out a tag with `git checkout tags/<tag_name>` and you can create a new branch from a tag with `git checkout tags/<tag_name> -b <branch_name>`.

medariox:
I've [added the commit hashes](https://github.com/pymedusa/Medusa/blob/use-requirements/requirements.txt) as requested and updated all libs to their latest version.

fernandog:
`git` isn't also a requirement?

medariox:
Do you mean PyGithub? It's already [added](https://github.com/pymedusa/Medusa/blob/use-requirements/requirements.txt#L20) to requirements.

PR #2933, opened 2017-07-01
LegendasTV: Don't discard provider when BadRarFile or BadZipFile

fernandog:
Show a 'Invalid subtitle' warning instead
Without this, until subtitle is manually downloaded, it will keep discarding provider every run

@ratoaq2

### DIFF
--- a/medusa/subtitle_providers/legendastv.py
+++ b/medusa/subtitle_providers/legendastv.py
@@ -170,6 +170,7 @@ class LegendasTVProvider(Provider):
         self.username = username
         self.password = password
         self.logged_in = False
+        self.session = None
 
     def initialize(self):
         self.session = Session()
--- a/medusa/subtitle_providers/legendastv.py
+++ b/medusa/subtitle_providers/legendastv.py
@@ -164,7 +164,7 @@ class LegendasTVProvider(Provider):
     server_url = 'http://legendas.tv/'
 
     def __init__(self, username=None, password=None):
-        if username and not password or not username and password:
+        if any((username, password)) and not all((username, password)):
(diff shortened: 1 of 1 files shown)


lib/subliminal/providers/legendastv.py
```diff
@@ -442,4 +442,9 @@ def download_subtitle(self, subtitle):
             self.download_archive(subtitle.archive)
 
         # extract subtitle's content
-        subtitle.content = fix_line_ending(subtitle.archive.content.read(subtitle.name))
+        try:
+            subtitle.content = fix_line_ending(subtitle.archive.content.read(subtitle.name))
+        except (BadRarFile, BadZipfile):
+            # This will make subtitle.is_valid as False and show a 'Invalid Subtitle' warning
+            subtitle.content = None
```
ratoaq2: Why do you set content to None?

fernandog: Cause none Will make subtitle.is_valid as False. See pytest upstream

ratoaq2: But why set None to a field that's already None?

fernandog: so what to put in the except ?

ratoaq2: A proper log with a proper message so you know what archive is bad

fernandog: logger.error('Bad archive for %s', subtitle.name) ?

fernandog:
@ratoaq2 done. Also i was patching the wrong file lol

fernandog:
@ratoaq2 please?

Issue #2597, opened 2017-04-16
Post Processor hangs when unraring incomplete archive


labrys:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:** ApiV2 refactor
**OS:** Win 10
**What you did:** Triggered Manual PP 
**What happened:**  Manual PP never completed - hung during unrar (rar missing parts)
**What you expected:**  PP to continue
**Logs:**
```
2017-04-16 11:24:49 DEBUG    Thread-34 :: [b02debf] Packed files detected: [u'76370e2041174f0181c16b48d8ec8635.part01.rar', u'9ccc44aafb974edb82bc493119deaea2.part01.rar']
2017-04-16 11:24:49 DEBUG    Thread-34 :: [b02debf] Unpacking archive: 76370e2041174f0181c16b48d8ec8635.part01.rar
2017-04-16 11:24:51 INFO     TORNADO :: [b02debf] 200 GET /ui/get_messages (::1) 2.00ms
INFO:tornado.access:200 GET /ui/get_messages (::1) 2.00ms
2017-04-16 11:24:54 INFO     TORNADO :: [b02debf] 200 GET /ui/get_messages (::1) 2.00ms
INFO:tornado.access:200 GET /ui/get_messages (::1) 2.00ms
2017-04-16 11:24:57 INFO     TORNADO :: [b02debf] 200 GET /ui/get_messages (::1) 2.00ms
INFO:tornado.access:200 GET /ui/get_messages (::1) 2.00ms
2017-04-16 11:25:00 INFO     TORNADO :: [b02debf] 200 GET /ui/get_messages (::1) 2.00ms
INFO:tornado.access:200 GET /ui/get_messages (::1) 2.00ms
2017-04-16 11:25:03 INFO     TORNADO :: [b02debf] 200 GET /ui/get_messages (::1) 2.00ms
INFO:tornado.access:200 GET /ui/get_messages (::1) 2.00ms
```

fernandog:
it should do this before unpack:

```
from rarfile Error
rar = RarFile(path)
try:
    rar.testrar()
except Error:
   log.warning(....)
```
it will raise an exception
http://rarfile.readthedocs.io/en/latest/api.html#rarfile.RarFile.testrar 

maybe @medariox may evaluate the solution

medariox:
We aren't even using rarfile for post-processing.

[...]

PR #2945, opened 2017-07-06
Use rarfile instead of unrar2

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/2597

based on @sharkykh code 

Need more testing
Tested a good rar and a bad rar in Windows

`[b4a8a54] Failed unpacking archive Show.s04e01.1080p.webrip.x264-morose.rar: Not a RAR file`


medariox:
https://rarfile.readthedocs.io/en/latest/faq.html#how-can-i-get-it-work-on-windows

fernandog:
@pymedusa/developers 

To test this PR this is not needed BUT, need to decide which option will be implemented. 

- Add location of unrar.exe to PATH, **or**
- Set rarfile.UNRAR_TOOL to full path of unrar.exe, **or**
- Copy unrar.exe to your program directory, **or**
- Copy unrar.exe to system directory that is in PATH, eg. C:\Windows.

p0psicles:
I would use UNRAR_TOOL. Make it configurable by user. And of not set, try to set it from PATH. But maybe that's already done implicitly.

fernandog:
imho i would send the unrar in the /lib (/ext) folder so it would work out of the box for all users

@ratoaq2 NAS has unrar installed or easy to install? PATH is set?

ratoaq2:
In Linux you just need to have it installed. It works out of  the box with no configuration. My nas has it installed by default

fernandog:
@p0psicles added an option to custom path (like git path)

Tested in Windows with no path set and path set as "C:\Program Files\WinRAR\UnRAR.exe" - worked nice

As soon you set the path, javascript checks if the unrar is valid (I didn't add that code)

views/config_postProcessing.mako
```diff
@@ -200,6 +200,18 @@
                                     <span class="component-desc"><b>NOTE:</b> Only working with RAR archive</span>
                                 </label>
                             </div>
+                            <div id="content_unpack">
```
p0psicles: Change to hyphens. No underscores in id's, names, classes, etc.

fernandog: @p0psicles need to be that way so if will only show that option if the `unpack` is enabled. But somehow is not working. ping @OmgImAlexis

fernandog: Like here: https://github.com/pymedusa/Medusa/blob/develop/views/config_search.mako#L47

fernandog: Fixed. Waiting review now

fernandog:
Tested with a few rared torrents in my production. All good

fernandog:
I really don't know why test is now failing. The commit that I added the test is passing

Also it works on pycharm:

image

fernandog:
@medarios please answer as requested:

https://github.com/pymedusa/Medusa/pull/2945#issuecomment-313542821

I already explained the underscore. Please take a time to read everything.

views/config_postProcessing.mako
fernandog: @medariox You missed this talk

p0psicles: No @fernandog you missed my review comment. `content_unpack` should be `content-unpack`.

fernandog: @p0psicles @medariox so this is wrong? I explained why and know I'm pointing why. If is still wrong need to talk with omg

```javascript
    $('.enabler').each(function() {
        if (!$(this).prop('checked')) {
            $('#content_' + $(this).attr('id')).hide();
        }
    });
```
https://github.com/pymedusa/Medusa/blob/develop/static/js/common/init.js#L247

p0psicles: Aah now I see. Yeah that's a bit of legacy. Your right.

fernandog: thanks. glad we resolved

fernandog:
now we need to decide about unrar in windows. Waiting the anwser to https://github.com/pymedusa/Medusa/pull/2945#issuecomment-313542821

Already did what @p0psicles suggested, any more suggestions?

medariox:
It would make more sense to change it **now** (since we all agree it's wrong) instead of keep adding more incorrectly named ids. Someone will need to do it anyway at some point, right?

fernandog:
Still waiting the answer

medariox:
I would put ```UnRAR.exe``` including its license in ```lib/unrar``` and set that path as rarfile.UNRAR_TOOL. Seems like the best option for now, but that's just my opinion.

fernandog:
Need to decide if we ship unrar as you said or do like pops said (code already implemented)
Can you decide?

p0psicles:
Regarding the snake_case. That will mean allot of changes to js, html/mako and css files. I think that justifies, changing that in a dedicated PR. As we want to keep our PR's small of possible.

Regarding bundling the unrar.exe. I'm not a fan of mixing source code and binaries. So my vote goes, to not bundling it. And make sure we try to "detect" it, using path or expected location. And additionally make it configurable. And with that, try to make this a default / uniform way of handling binary dependencies.

static/js/config/post-processing.js
```diff
@@ -13,7 +13,7 @@ MEDUSA.config.postProcessing = function() { // eslint-disable-line max-lines
         };
     })();
 
-    function is_rar_supported() {
+    function is_rar_supported() { // eslint-disable-line camelcase
```
p0psicles: js should really be camelCase.

static/js/config/post-processing.js
```diff
@@ -14,9 +14,9 @@ MEDUSA.config.postProcessing = function() { // eslint-disable-line max-lines
     })();
 
     function isRarSupported() {
-        $.get('config/postProcessing/isRarSupported', function(data) {
-            if (data !== 'supported') {
-                $('#unpack').qtip('option', {
+        $.post('config/postProcessing/isRarSupported', {unrar_tool: $('#unrar_tool').val()}, function(data) { // eslint-disable-line camelcase
```
p0psicles: Here `config/postProcessing/isRarSupported` is a route. The same as `config/postProcessing/savePostProcessing` and `home/testSABnzbd` these are all in camelCase. So this is correct, or at least following the current convention.

medusa/server/web/config/post_processing.py
```diff
@@ -213,19 +220,13 @@ def isNamingValid(pattern=None, multi=None, abd=False, sports=False, anime_type=
             return 'invalid'
 
     @staticmethod
-    def isRarSupported():
-        """
-        Test Packing Support:
-            - Simulating in memory rar extraction on test.rar file
-        """
-
+    def isRarSupported(unrar_tool):
```
p0psicles: Here `config/postProcessing/isRarSupported` is a route. The same as `config/postProcessing/savePostProcessing` and `home/testSABnzbd` these are all in camelCase. So this is correct, or at least following the current convention.

p0psicles: @medariox what is it that you like to see done different?

medariox: @p0psicles 
This is really no excuse to keep using bad practices. Especially a route should **never** be camel case or even use upper case characters. I thought we wanted to enforce best practices and I don't see how this attitude is coherent with our goals.

p0psicles: Well for starters it may be a bad practice, but it's also a convention we've chosen. One thing I hate more then a bad convention is an implementation that only partially follows a 'bad' convention. So or we do all of them at once, meaning we change the convention for the routes, or we don't.

So if you want to do something about, go ahead. But that does have nothing to do with this PR.

medariox: I completely disagree. I don't believe we aver chose to keep using bad practices, at the contrary, I'm sure we have chosen to start using best practices, we should definitely enforce them. Also, it has everything to do with this PR, as it directly changes code code that can and should be brought to our chosen standards.

p0psicles: So you want to expand a PR that replaces an old RAR dependency with a new one, to a PR that also replaces about 166 routes? Those are not just doing search/replaces. But also all Mako templates, js files need to be changed and tested. And of course updating the documentation, for the new convention. And I bet i'm not overseeing all changes. So we'll probably get some unexpected errors.

medariox: No, I'm talking about this one route used exactly in one place. It's not a new convention, it's bringing old code to our current standards.

p0psicles: No, it's changing one route, which doesn't follow our current convention. And you know what happens after that.. nothing! As nobody will feel him/her self called to change the rest.

medariox: How do you expect anything to change if we ignore the standards we decided to enforce? It's a first step, will be one less method to rename in the future.

OmgImAlexis: I say we change this to use the new standard.

With js linting we have a way to fix issues magically by asking the linter to do it for us. Can we do that with Python?

OmgImAlexis: This should have been added as a new route in the API since you're just fixing an old one let's just leave it as is. @fernandog maybe once I've got some time again I'll help you move this over to the new API v2 section and we can fix the case issues then.

medariox:
That was the smallest issue. The python code is a mess and this shouldn't have been approved nor should it be merged.

p0psicles:
Well as discussed with omg, the only python related issues remaining are the once regarding the route. And there we came to an agreement that, if we want to do this the right way. We should implement the route in apiv2. As we agreed a while back, that when touching rest routes, they should be implemented in apiv2. If there is anything else that should be changed, I expect you to let us know, through a review comment. As you depict it as "the python code is a mess". I expect you have some.

medariox:
I will, don't worry.

medusa/__main__.py
```diff
@@ -2070,6 +2078,51 @@ def shutdown(self, event):
             logging.shutdown()
             os._exit(0)  # TODO: Remove in another PR. There's no need for this one.
 
+    @staticmethod
+    def configure_unrar():
```
medariox: All of this needs to be greatly simplified.

medusa/__main__.py
```diff
@@ -2070,6 +2078,51 @@ def shutdown(self, event):
...
+            else:
+                rarfile._check_unrar_tool()
+        except (rarfile.RarCannotExec, rarfile.RarExecError, OSError):
+            if platform.system() == 'Windows':
+                logger.info('Looking for UNRAR in your installed programs')
+                winrar_path = 'WinRAR\\UnRAR.exe'
+                # Make a set of unique paths to check from existing environment variables
+                check_locations = {
+                    os.path.join(location, winrar_path) for location in (
+                        os.environ.get("ProgramW6432"), os.environ.get("ProgramFiles(x86)"),
```
medariox: Why are double quotes being used here?

medusa/process_tv.py
```diff
@@ -29,14 +29,12 @@
 from medusa.name_parser.parser import InvalidNameException, InvalidShowException, NameParser
 from medusa.subtitles import accept_any, accept_unknown, get_embedded_subtitles
 
+import rarfile
```
medariox: Don't import the whole library, but only what you need.

medusa/process_tv.py
```diff
@@ -432,14 +430,20 @@ def unrar(self, path, rar_files, force=False):
...
-                    rar_handle = RarFile(os.path.join(path, archive))
+                    rar_handle = rarfile.RarFile(os.path.join(path, archive))
+
+                    # Test rar, if fails it will raise an exception
+                    rar_handle.testrar()
 
                     # Skip extraction if any file in archive has previously been extracted
                     skip_extraction = False
                     for file_in_archive in [os.path.basename(each.filename)
                                             for each in rar_handle.infolist()
```
medariox: Why are you using infolist() instead of namelist()?

medusa/server/web/config/post_processing.py
medariox: unrar_tool should be optional as it is not mandatory for the check (e,g, Linux)

medusa/server/web/config/post_processing.py
```diff
@@ -6,12 +6,13 @@
 
 import os
 
+import rarfile
```
medariox: Same as above.

medusa/server/web/config/post_processing.py
```diff
@@ -213,19 +220,13 @@ def isNamingValid(pattern=None, multi=None, abd=False, sports=False, anime_type=
...
-            testing = RarFile(rar_path).read_files('*test.txt')
-            if testing[0][1] == 'This is only a test.':
-                return 'supported'
-            logger.log('Rar Not Supported: Can not read the content of test file', logger.ERROR)
-            return 'not supported'
-        except Exception as msg:
-            logger.log('Rar Not Supported: {error}'.format(error=ex(msg)), logger.ERROR)
-            return 'not supported'
+            rarfile.custom_check(unrar_tool)
+        except (rarfile.RarCannotExec, rarfile.RarExecError, OSError) as error:
```
medariox: Why are you catching OSError here?

medusa/process_tv.py
fernandog: As you didn't changed in last refactor it's good right? You touched this exact code and didn't change it

medusa/__main__.py
fernandog: Why can't I ? 
As I'm a contributor I'm following 
https://github.com/pymedusa/Medusa/blob/master/CONTRIBUTING.md and it says nothing about it. Please update contribution.md updated if that is the case

medusa/__main__.py
fernandog: This is not a proper review. Please explain better

medusa/server/web/config/post_processing.py
fernandog: Not all linux installations have unrar. Can you say with 100% sure ALL distros have?

medusa/process_tv.py
medariox: Unrar2 didn't have namelist(), wake up.

medusa/__main__.py
medariox: We agreed on single quotes long time ago and use it throughout the code (and you know that).

medusa/__main__.py
medariox: This is proper review. I'm not going to do the work for you. If you are not ready to improve your code, I don't see how you expect it to be merged.

medusa/__main__.py
fernandog: please update contribution.md

medusa/process_tv.py
fernandog: why not use infolist? I'm not expert. explain to me please

medusa/__main__.py
fernandog: I'm asking what is wrong, can't you reply?

medusa/__main__.py
medariox: This doesn't need to be added to contributions. It is part of PEP 8.
```
In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it. When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.
```
**Pick a rule and stick to it.**
We chose single quotes as you can see in the rest of the code. There is nothing to add to this.

medusa/process_tv.py
medariox: Read the rarfile documentation.

fernandog: namelist calls infolist. Why do you **prefer** namelist?

If you don't have the patience to do a proper review, how do you expect other devs to join and code? As almost no one else is coding/improving here except @p0psicles 

rarfile.py
```python
    def namelist(self):
        """Return list of filenames in archive.
        """
        return [f.filename for f in self.infolist()]

    def infolist(self):
        """Return RarInfo objects for all files/directories in archive.
        """
        return self._file_parser.infolist()
```

medariox: Because when the API provides a method to do **exactly** what you need, you shouldn't be writing custom code. That is the very basic thought behind using an API. Even people that never coded in their lives would understand that.

fernandog:
@OmgImAlexis please when you have time can you review as we talked?

This PR will stay open until I get reports of successful tests of unpacking

medusa/__main__.py
medariox: I think you should start at the very beginning, maybe something like [this](https://www.amazon.com/Python-Best-Practices-Programming-Code-ebook/dp/B01N4LDZT1) could be a good start and it's almost free. Otherwise try to google about good coding practices in general (not necessarily about Python). How to write simple, clean functions, how to write good comments, readable code, etc.

fernandog: @sharkykh sorry about that. Your code is good and works every well! thanks btw

ratoaq2: Why adding this bunch of logic in main.py?

fernandog:
@ratoaq2 can't reply your review don't know why. 
It was decided to check unrar at boot and disable unpack if not rar not supported

medusa/process_tv.py
fernandog: must be this way. check code

medusa/process_tv.py
```diff
@@ -441,9 +441,9 @@ def unrar(self, path, rar_files, force=False):
 
                     # Skip extraction if any file in archive has previously been extracted
                     skip_extraction = False
-                    for file_in_archive in [os.path.basename(each.filename)
-                                            for each in rar_handle.infolist()
-                                            if not each.isdir()]:
+                    for file_in_archive in [os.path.basename(each)
+                                            for each in rar_handle.namelist()
+                                            if not os.path.isdir(each)]:
```
medariox: This will always return True. You need to check the full path to the file/folder.

medusa/process_tv.py
```diff
@@ -459,9 +459,9 @@ def unrar(self, path, rar_files, force=False):
                     if not skip_extraction:
                         rar_handle.extract(path)
 
-                    for each in rar_handle.infolist():
-                        if not each.isdir():
-                            basename = os.path.basename(each.filename)
+                    for each in rar_handle.namelist():
+                        if not os.path.isfile(each):
```
medariox: Same as above.

medusa/process_tv.py
medariox: Why would that be?

medusa/process_tv.py
fernandog: It was so much better to use infolist that isdir() method.....

fernandog: do you still think infolist is not more useful?

medariox: Considering that we need the file names of videos to check for already_postprocessed(), it makes more sense to use namelist() and check for helpers.is_media_file().

medusa/process_tv.py
```diff
@@ -453,26 +457,17 @@ def unrar(self, path, rar_files, force=False):
                             break
 
                     if not skip_extraction:
-                        rar_handle.extract(path=path, withSubpath=False, overwrite=False)
+                        rar_handle.extract(path)
```
medariox: I'd like to point out two things here:
1) Although this works, according to the documentation .extractall() should be used to extract all files, while .extract() should be used to extract a single file.
2) This modification changes the current behavior. While with unrar2 subfolders were discarded (withSubpath=False), this is not the case anymore.

medusa/process_tv.py
medariox: I'd suggest to do something like this:
```
for file_in_archive in (each
                        for each in rar_handle.namelist()
                        if is_media_file(each)):
```
....
```
if not force and self.already_postprocessed(os.path.basename(file_in_archive)):
```
This would avoid breaking the os.path.isfile() check later on as well.

medusa/process_tv.py
fernandog: from what @duramato said some time ago, the RARed torrents are only for the video file. 
Sample, NFO and everything else is not rared

medariox: Doesn't change anything to what I wrote above.

fernandog: how not ? 
 `While with unrar2 subfolders were discarded (withSubpath=False) this is not the case anymore` : there is no subfolder in a rared torrent

about extract, its inside a a for-loo: `for archive in rar_files:` so we only have one file

medariox: When you code something (this is a general rule), you never do it with the assumption that something that you have no control over will never change. Thing evolve and change constantly. Just because it works today, it doesn't mean it will work tomorrow. This way of thinking is very shortsighted.

medariox: And no, it doesn't change a bit the fact that:
> Although this works, according to the documentation .extractall() should be used to extract all files, while .extract() should be used to extract a single file.

ratoaq2:
There are so many things that are loaded and configured at boot time and they aren't in __main__.py

fernandog:
What's your suggestion?

fernandog:
Won't have time to continue on this one. Too many issues. At least I tried.
If someone wants to pick it up feel free

PR #2948, opened 2017-07-08
Implemented smart caching for imdb recommended lists

p0psicles:
This will make imdb rec load fast after a first load.

Because we recently started to use imdbpy, it needs to load additional information for each show, as opposed to getting it in one request.

This will cache the detailed show info, for a duration of 7 days.

* Refactored the 'old' caching class. And added a caching class which can store a key and value. But does also expire. I could have used existing caching libs for this. But the code I added fit's my purpose perfect. And it's only needed until we properly store this info in db. Maybe dogpile was also an option.
* Cached imdb rec list. So it will be usable after the first load.

### DIFF
--- a/medusa/show/recommendations/__init__.py
+++ b/medusa/show/recommendations/__init__.py
@@ -4,6 +4,10 @@ import time
 from collections import namedtuple
 
 
+class ExpectUniqueKeyException(Exception):
+    """Exception raised when multiple keys are detected, but a unique key is expected."""
+
+
 class ExpiringList(list):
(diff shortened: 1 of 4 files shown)


fernandog:
@p0psicles Nice. Ready to do some tests?

p0psicles:
Yeah it should already work. Only need to write some docstrings.

fernandog:
@p0psicles working nice. Deleted image caches, started medusa and opened imdb rec shows and added a show. Then opened the page again it opened really fast. nice job

Issue #2897, opened 2017-06-19
Bug? Multiple notification popups


wimpyrbx:
![image](https://user-images.githubusercontent.com/13415240/27271548-b52ae348-54c5-11e7-8c54-485e89a2e47c.png)

this happens since a few updates ago on develop

medariox:
Multiple tabs open when it happens? That's because of WebSockets.

IDerr:
Happens too. Like @medariox said, it's because of multiple tabs

sharkykh:
@medariox Try setting `PNotify.desktop.tag` option for each message to a unique id (in JS).
I used a `hash()` on the Notification object in Python and then transferred that to Javascript along with the other params.

wimpyrbx:
i'm pretty sure experienced it with only one tab as well. will update if i see it again.

labrys:
@wimpyrbx I'm pretty sure I've had it with a single tab as well.

labrys:
Actually it appears the double pop-ups I had with a single tab open appear to be duplicate snatches, not duplicate notifications.

fernandog:
@OmgImAlexis when you have time can you try that fix please?

fernandog:
@wimpyrbx confirm fixed?

PR #2949, opened 2017-07-10
Fix to N notifications (N tabs open)

fernandog:
Code from @sharkykh

I did a quick test and it's not working
image

Fixes: https://github.com/pymedusa/Medusa/issues/2897

### DIFF
--- a/static/js/ajax-notifications.js
+++ b/static/js/ajax-notifications.js
@@ -59,6 +59,6 @@ function wsCheckNotifications() {
 $(document).ready(function() {
     wsCheckNotifications();
     if (test) {
-        displayPNotify('error', 'test', 'test<br><i class="test-class">hello <b>world</b></i><ul><li>item 1</li><li>item 2</li></ul>', 'notification-test')
+        displayPNotify('error', 'test', 'test<br><i class="test-class">hello <b>world</b></i><ul><li>item 1</li><li>item 2</li></ul>', 'notification-test');
     }
 });
--- a/static/js/ajax-notifications.js
+++ b/static/js/ajax-notifications.js
@@ -15,10 +15,13 @@ PNotify.prototype.options.styling = 'jqueryui';
 PNotify.prototype.options.width = '340px';
 PNotify.desktop.permission();
 
-function displayPNotify(type, title, message) {
+function displayPNotify(type, title, message, id) {
     new PNotify({ // eslint-disable-line no-new
         type: type,
         title: title,
+        desktop: {
+            tag: id
+        },
         text: String(message).replace(/<br[\s/]*(?:\s[^>]*)?>/ig, '\n')
             .replace(/<[/]?b(?:\s[^>]*)?>/ig, '*')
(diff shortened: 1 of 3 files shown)


fernandog:
@OmgImAlexis not working

image

static/js/ajax-notifications.js
```diff
@@ -39,7 +42,7 @@ function wsCheckNotifications() {
 
         // Add handling for different kinds of events. For ex: {"event": "notification", "data": {"title": ..}}
         if (msg.event === 'notification') {
-            displayPNotify(msg.data.type, msg.data.title, msg.data.body);
+            displayPNotify(msg.data.type, msg.data.title, msg.data.body, msg.data.hash);
         } else {
             displayPNotify('info', '', msg);
```
OmgImAlexis: Does the notification you're testing have `msg.event === 'notification'`?

fernandog: how to check that?

fernandog: image

sharkykh: You can add `console.log(msg.event)` before line 44 so it shows up in the DevTools console

fernandog: here is the issue:

image

sharkykh: What does the response from /ui/get_messages give you? (Open the network tab, then generate a notification from Medusa)

fernandog: I added a break-point here but it's never reach this code:

image


maybe the websockets @p0psicles ?

sharkykh: I just figured it out.
You changed in the wrong location (this is not the websocket implementation)

You need to change it in [`medusa/ui.py`](https://github.com/pymedusa/Medusa/blob/feature/issue_2897/medusa/ui.py) in both [`Notifications.message()`](https://github.com/pymedusa/Medusa/blob/feature/issue_2897/medusa/ui.py#L47-L50) and [`Notifications.error()`](https://github.com/pymedusa/Medusa/blob/feature/issue_2897/medusa/ui.py#L59-L63)
(but you should leave this here, too)

OmgImAlexis: I was about to say that when your comment appeared. 😆

fernandog: worked! thanks

static/js/ajax-notifications.js
```diff
@@ -15,10 +15,13 @@ PNotify.prototype.options.styling = 'jqueryui';
 PNotify.prototype.options.width = '340px';
 PNotify.desktop.permission();
 
-function displayPNotify(type, title, message) {
+function displayPNotify(type, title, message, id) {
```
p0psicles: In python its preferred to not use buildins like `id`. So you can easily solve this, by rename it to notification_id or something like that.

p0psicles: Ow and nice work!

sharkykh: @p0psicles This is Javascript?

p0psicles: ow lol. Your right. Don't know what i've been smoking.

Issue #2950, opened 2017-07-11
Guessit 'format' match issue: folder vs filename


fernandog:
WEB-DL
Folder: WEBRIP

This is becoming quite common for me. @ratoaq2 Can we use the format from file instead of folder?
The subtitles won't match because of this

ping @duramato 

```
python report_guessit.py "/series/Fear.the.Walking.Dead.S03E07.1080p.AMZN.WEBRip.DD5.1.x264-VLAD[rarbg]/Fear.the.Walking.Dead.S03E07.1080p.AMZN.WEB-DL.DD+5.1.H.264-VLAD.mkv"
# guessit: 2.1.3.dev0  rebulk: 0.8.3.dev0
? /series/Fear.the.Walking.Dead.S03E07.1080p.AMZN.WEBRip.DD5.1.x264-VLAD[rarbg]/Fear.the.Walking.Dead.S03E07.1080p.AMZN.WEB-DL.DD+5.1.H.264-VLAD.mkv
: title: Fear the Walking Dead
  season: 3
  episode: 7
  screen_size: 1080p
  streaming_service: Amazon Prime
  format: WEBRip
  audio_codec: AC3
  audio_channels: 5.1
  video_codec: h264
  video_encoder: x264
  release_group: VLAD
  container: mkv
  mimetype: video/x-matroska
  type: episode
  parsing_time: 0.915619850159
```

fernandog:
@rarbg quick question: is the uploaders or you that create the folder name in the RARBG releases?

noticed that the folder is "webrip" and file name is "web-dl".

fernandog:
just noticed now that code is different too: x264 vs H.264

rarbg:
We have a strict naming convention
How guessit treats it is up to guessit

fernandog:
@rarbg i know but I'm asking about the torrent itself, not the guessit result

Check it out: https://rarbg.is/torrent/hnvu3w7#expand

Torrent name: Fear.the.Walking.Dead.S03E07.1080p.AMZN.WEBRip.DD5.1.x264-VLAD
Filename: Fear.the.Walking.Dead.S03E07.1080p.AMZN.WEB-DL.DD+5.1.H.264-VLAD.mkv

duramato:
In this case file name is the correct release name.

rarbg:
There is nothing to check. There are thousands of torrents like that

rarbg:
@duramato 

whats right for us doesnt mean its right for guessit

[...]

PR #2953, opened 2017-07-12
Warn about possible subtitle match

fernandog:
Fix for https://github.com/pymedusa/Medusa/issues/2950

@ratoaq2 @duramato


ratoaq2:
Well, I dislike this feature. 
You have missing subtitles page, manual sub search and here you're just reruning guessit for every sub search

fernandog:
A real example:

WARNING  FINDSUBTITLES :: [7a152d4] Consider doing a manual subtitle search for: TURN.Washingtons.Spies.S04E04.Nightmare.1080p.AMZN.WEB-DL.DDP5.1.H.264-ViSUM.mkv

Notice:
DDP5 vs DD5
WEBRIP vs WEB-DL

Subtitle:
"Turn.S04E04.Nightmare.1080p.AMZN.WEBRip.DD5.1.x264-ViSUM"

fernandog:
@ratoaq2 Im not using guessit. Im using the property `subtitle.get_matches(video)`

medariox:
Isn't this exactly what missed subtitles management is for? Check it after X days and use manual search.

fernandog:
Each episode has it own "3 days" without subs as each one has it own airdate/download date

fernandog:
Also even if after 3 days it may not have subs. This doesn't  search. It only warns about existing subs it already found

medariox:
The point is: if you don't have the subs after X days (looking at the missed subs page), you will know you should try to search with manual subs search, there is no need for such a warning. Additionally, a warning is something that went wrong with the system, it is not suitable for this kind of information. A much better approach would be to integrate some kind of visual notification to the missed subtitles page.

fernandog:
On hold because get_matches is not cached. Need to think about it

fernandog:
@medariox like this? =)

image

PR #2962, opened 2017-07-14
LegendasTV: fix caching titles for wrong season

fernandog:
It should only cache titles that matches give season when episode type
@ratoaq2 if ok, I will commit upstream to the respective PR

### DIFF
--- a/medusa/subtitle_providers/legendastv.py
+++ b/medusa/subtitle_providers/legendastv.py
@@ -203,72 +203,116 @@ class LegendasTVProvider(Provider):
 
         self.session.close()
 
+    @staticmethod
+    def is_valid_title(title, title_id, sanitized_title, season, episode, year):
+        """Check if is a valid title."""
+        sanitized_result = sanitize(title['title'])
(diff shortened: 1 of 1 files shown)


medusa/subtitle_providers/legendastv.py
```diff
@@ -258,6 +258,12 @@ def search_titles(self, title, season):
                     else:
                         logger.warning('No season detected for title %d (%s)', title_id, name)
 
+                # discard title mismatches on season
```
ratoaq2: Now you have parts of the mismatches discard logic in `search_titles` and parts in `query`.

fernandog: Title is associated to the season, so it must be here.

fernandog:
@ratoaq2 I rewrote. I had to rename `title` var because you override the one from argument when you `title = {'type': type_map[source['tipo']]}` - same with new param `year` I added and year from result

p0psicles:
@ratoaq2, as far as I can review it, it looks good. Shall we approve?

PR #2968, opened 2017-07-16
Handle api error

duramato:

### DIFF
--- a/medusa/providers/torrent/json/danishbits.py
+++ b/medusa/providers/torrent/json/danishbits.py
@@ -8,7 +8,7 @@ import logging
 import traceback
 
 from medusa import tv
-from medusa.helper.common import convert_size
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/json/danishbits.py
```diff
@@ -40,6 +40,7 @@ def __init__(self):
 
         # Miscellaneous Options
         self.freeleech = True
+        self.headers.update({'User-Agent': USER_AGENT})
```
p0psicles: I think this won't have any effect. As the self.headers is used to update self.session.headers, before you get to here.
So you should directly update self.session.headers in stead.

p0psicles: For ex:
```self.session.headers['User-Agent'] = USER_AGENT```

PR #2970, opened 2017-07-17
Introduce real configObj lists, as it's supported by configobj

p0psicles:
Introduce real configObj lists, as it's supported by configobj.
* Added the v10 convert method.
* Added check_setting_list, as a helper to get the parsed list from configobj.
* Removed most of the splits used during initialize.
* Changed default global options csv strings, to lists.

# TODO:
- [x] transform newznab providers code
- [x] transform custom torrent providers
- [x] adapt mako pages, with splits().
- [x] codec the config pages
- [x] Fix unit tests, which expect strings (csv)

# Screenees
image


medusa/config.py
```diff
@@ -536,6 +536,20 @@ def clean_url(url):
...
+    values = []
+    if isinstance(value, list):
+        return value
+
+    try:
+        values = value.split(delimiter)
+        if trim:
+            values = [_.strip() for _ in values]
+    except Exception as error:
+        pass
```
p0psicles: Reminder, need to do something with this

fernandog:
nobody can't stop @p0psicles! =)

p0psicles:
Yeah they can. Just remove my rights :-).
But I started this PR, cause I wanted to add a feature, to add a lists of directory names, that are ignored by PP. So I want to be able to use normal lists (no ||| or | or , csv crap), before I can continue on that.

medusa/app.py
```diff
@@ -271,7 +271,7 @@
 TV_DOWNLOAD_DIR = None
 UNPACK = False
 SKIP_REMOVED_FILES = False
-ALLOWED_EXTENSIONS = "srt,nfo,sub,idx"
+ALLOWED_EXTENSIONS = ('srt', 'nfo', 'sub', 'idx')
```
p0psicles: I need to check why I initiated this as a tuple. Think it should be list though.

fernandog:
@p0psicles root dir is ok now

I noticed the provider order in the UI completely lost. 
the config `provider_order` didn't change so I guess is a javascript issue
Also the new `provider_order` have a trailling comma in the end. Don't know if is the cause of the issue

Also the subtitle provider I had set is not enabled.
It used to be `SUBTITLES_SERVICES_ENABLED = 1|0|0|0|0|0|0|0|0|0` 
and now is `SUBTITLES_SERVICES_ENABLED = 1, 0, 0, 0, 0, 0, 0, 0, 0, 0`

fernandog:
About subtitles, I found the issue. It used to be int now is unicode:

image

fernandog:
I **think** it will cause issues with this configs too as it should be considered INT (enabled|disabled) and not unicode:

metadata_kodi = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
metadata_kodi_12plus = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
metadata_mediabrowser = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
metadata_ps3 = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
metadata_wdtv = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
metadata_tivo = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
metadata_mede8er = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0

p0psicles:
Tnx for testing fernandog! I will look into these.

p0psicles:
Some of the notifiers use lists. I can't test these, like Plex and prowl. Can anyone else tests these?

fernandog:
@p0psicles still in progress? I think we can merge this after release branch

medusa/config.py
```diff
@@ -652,6 +676,46 @@ def check_setting_str(config, cfg_name, item_name, def_val, silent=True, censor_
...
+        my_val = default
+        try:
+            config[cfg_name][item_name] = my_val
+        except Exception:
+            config[cfg_name] = {}
+            config[cfg_name][item_name] = my_val
+
+    if privacy_level >= censor_level or (cfg_name, item_name) in iteritems(logger.censored_items):
+        if not item_name.endswith('custom_url'):
+            logger.censored_items[cfg_name, item_name] = my_val
```
p0psicles: Hi @pymedusa/developers i need some help here. Currently my_val is always a string or integer. And that works fine with the censored_items. But with this PR, the value can also be a list. I doubt we can just assign a list to the censored_item dict? So how to go about this?

labrys:
```python
def safe_iter(items):
		return (item for item in items)
	else:
		return [items]

for item in safe_iter(value):
    if not item_name.endswith('custom_url'):
        logger.censored_items[cfg_name, item_name] = value
```

labrys: While the code I posted will add all the values to the censored items, you should consider if items in the list really should go in the censored items.  For example do you really want to censor all instances of 0 or 1?

labrys: Additionally we should store boolean values as True or False and not `0` or `1`

p0psicles: For now it's only used to sensor some tokens. Also your example won't work. As it will overwrite the value for n items in the list.

p0psicles: I have a solution in mind. Just adding the list. And changing the rebuild_censored_list method

labrys: OK, i was under the impression censored_items was a list.

medusa/logger/__init__.py
```diff
@@ -68,7 +70,15 @@
...
-    results = {value for value in itervalues(censored_items) if value}
+    results = set()
+    for value in itervalues(censored_items):
+        if not value:
+            continue
+        if isinstance(value, collections.Iterable) and not isinstance(value, six.string_types):
+            results.update(value)
+        else:
+            results.add(value)
+
```
p0psicles: @labrys this is what it came down to.
Also I noticed that the rebuild_censored_list() function was called for each check_setting_str() and check_setting_list() function. I reduced that to 1 call. Should improve startup times.

medusa/config.py
p0psicles: Regarding storing the boolean values. If I would to implement that, we need to translate that. As now we pass on the values as 0 and 1 through js or post data. We could change it, but it's allot of work in frontend and backend, if done properly.

fernandog:
I wouldn't change that now...

medusa/config.py
labrys: Priority imo would be low, but it more accurately reflects the actual data.

p0psicles:
@pymedusa/developers I can merge this now. But it would be better if we have 2 or 3 devs running this branch, to test of their config is migrated without any issues.

medariox:
Using this branch since yesterday, seems all good for now, great job!

Only thing I noticed:
img

Empty options that expect a list (?) have a hanging `,` - with one item: `word,` - with multiple items `one, two`.

p0psicles:
Yes that's how it's supposed to be. The comma represents a list? So that screenshot looks fine.

medariox:
Oh ok, I didn't know it was supposed to represent a list.

PR #2971, opened 2017-07-17
LegendasTV: return only valid archives for given season and episode

fernandog:
@ratoaq2 this is my idea but as `get_archives` cache expiration is 15minutes (subliminal and medusa) I don't know if is worth as Medusa subtitle search frequency is 1hour, archive cache will be always expired. 

What do you think?

With this PR and the title cache PR, all mismatch logic are not longer in the `def query`


fernandog:
@p0psicles 
This PR is good, but misses review

PR #2993, opened 2017-08-08
Obtain show's countries from tmdb

duramato:

@p0psicles 

image

### DIFF
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -1,5 +1,8 @@
 # coding=utf-8
 
+"""TMDB module."""
+
+
 from __future__ import unicode_literals
 
 import logging
@@ -188,7 +191,7 @@ class Tmdb(BaseIndexer):
         return OrderedDict({'series': mapped_results})['series']
 
     def _get_shows_countries(self, tmdb_id):
-        """Retrieve show's countries from TMDB
+        """Retrieve show's countries from TMDB.
 
         :param tmdb_id: The shows tmdb id
         :return: A string with the show's countries
@@ -379,9 +382,7 @@ class Tmdb(BaseIndexer):
         return _images
 
     def _parse_actors(self, sid):
-        """Parse actors XML.
-
-        """
+        """Parse actors XML."""
         log.debug('Getting actors for {0}', sid)
 
         # TMDB also support passing language here as a param.
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -198,7 +198,7 @@ class Tmdb(BaseIndexer):
         """
         show_info = self._get_show_by_id(tmdb_id)['series']
 
-        return show_info.get('origin_country') or ''
+        return show_info.get('origin_country', '')
 
     def _get_show_by_id(self, tmdb_id, request_language='en'):  # pylint: disable=unused-argument
         """Retrieve tmdb show information by tmdb id, or if no tmdb id provided by passed external id.
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -599,6 +599,6 @@ class Tmdb(BaseIndexer):
                     externals = {tmdb_external_id: external_value
                                  for tmdb_external_id, external_value
                                  in externals.items()
-                                 if external_value and tmdb_external_id in ['tvrage_id', 'imdb_id', 'tvdb_id']}
+                                 if external_value and tmdb_external_id in ['tvrage_id', 'imdb_id', 'tvdb_id', 'tmdb_id']}
                     return externals
         return {}
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -170,7 +170,7 @@ class Tmdb(BaseIndexer):
 
     # Tvdb implementation
     def search(self, series):
-        """Search tmdb.com for the series name.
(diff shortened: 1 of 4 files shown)


medusa/tv/series.py
```diff
@@ -1553,15 +1556,25 @@ def load_imdb_info(self):
...
             return
+            
+        tmdb = Tmdb()
+        tmdb_id = self.externals.get('tmdb_id')
+        country_code = ''
+        country = ''
+        if tmdb_id:
+            show_info = tmdb._get_show_by_id(tmdb_id)
+            country_code = show_info['series']['origin_country']
+            print(country_code)
```
p0psicles: Reminder to remove this.

medusa/tv/series.py
```diff
@@ -1553,15 +1556,25 @@ def load_imdb_info(self):
...
                       {'id': self.indexerid, 'imdb_id': self.imdb_id})
             return
+            
+        tmdb = Tmdb()
+        tmdb_id = self.externals.get('tmdb_id')
+        country_code = ''
+        country = ''
+        if tmdb_id:
+            show_info = tmdb._get_show_by_id(tmdb_id)
+            country_code = show_info['series']['origin_country']
```
p0psicles: Do all shows on tmdb have the country code? What happens if it's not available?

medusa/tv/series.py
```diff
@@ -1553,15 +1556,25 @@ def load_imdb_info(self):
...
 
         self.imdb_info = {
             'imdb_id': imdb_obj.imdb_id,
             'title': imdb_obj.title,
             'year': imdb_obj.year,
             'akas': '',
             'genres': '|'.join(imdb_obj.genres or ''),
-            'countries': '',
-            'country_codes': '',
+            'countries': country,
```
p0psicles: I know it's the easiest solution now, but technically this is not imdb info. And it should not be stored as such, to prevent confusion in future. Just something to keep in mind.

medusa/indexers/tmdb/tmdb.py
```diff
@@ -187,6 +187,16 @@ def search(self, series):
 
         return OrderedDict({'series': mapped_results})['series']
 
+    def _get_shows_countries(self, tmdb_id):
+        """Retrieve show's countries from TMDB
+
+        :param tmdb_id: The shows tmdb id
+        :return: A string with the show's countries
+        """
+        show_info = self._get_show_by_id(tmdb_id)['series']
```
p0psicles: What if the tmdb_id used, is incorrect. The id can come from a different indexer, so there is no guarantee it's correct.

medusa/subtitles.py
```diff
@@ -209,6 +209,20 @@ def from_ietf_code(code, unknown='und'):
         return Language(unknown) if unknown else None
 
 
+def from_code_to_name(code):
```
p0psicles: Where are you using this method?

medusa/tv/series.py
```diff
@@ -1554,14 +1555,21 @@ def load_imdb_info(self):
...
+
         self.imdb_info = {
             'imdb_id': imdb_obj.imdb_id,
             'title': imdb_obj.title,
             'year': imdb_obj.year,
             'akas': '',
             'genres': '|'.join(imdb_obj.genres or ''),
-            'countries': '',
-            'country_codes': '',
+            'countries': '|'.join(countries),
```
p0psicles: Make sure you at least leave some comments, stating where this info is coming from.

medusa/tv/series.py
```diff
@@ -1554,14 +1555,21 @@ def load_imdb_info(self):
                       {'id': self.indexerid, 'imdb_id': self.imdb_id})
             return
 
+        tmdb_id = self.externals.get('tmdb_id')
+        if tmdb_id:
+            country_code = Tmdb()._get_shows_countries(tmdb_id)
+            countries = '|'
+            for country in country_code.split('|'):
+                countries = countries.join(subtitles.name_from_code(country))
```
fernandog: @p0psicles is this efficient? it will keep joining already joined countries

I added generator here but got removed

medusa/indexers/tmdb/tmdb.py
duramato: What do we mean? For all we know it is correct. It is as correct as the imdb_id we also have and use.

medusa/subtitles.py
duramato: It's being used now

fernandog:
Even the TMDB country is being saved as imdb info, i still think this is a valid way to get the info.
Anyone has something against it?

medusa/indexers/tmdb/tmdb.py
p0psicles: Your searching tmdb by tmdb id. Only you are not 100% sure the tmdb id came from tmdb. It's just in theory. In practice  a tmdb could only come from tmdb. 

For eample an imdb id, could come from tvdb (mapped from tvdbid) or tvmaze (mapped from tvdbid or tvmazeid).
If one of them maps it to a wrong id, you could request for an id that does not exist on TMDB.

Does that make sense?

medusa/indexers/tmdb/tmdb.py
```diff
@@ -187,6 +190,16 @@ def search(self, series):
...
 
+    def _get_shows_countries(self, tmdb_id):
+        """Retrieve show's countries from TMDB.
+
+        :param tmdb_id: The shows tmdb id
+        :return: A string with the show's countries
+        """
+        show_info = self._get_show_by_id(tmdb_id)['series']
+
+        return show_info.get('origin_country') or ''
```
p0psicles: `show_info.get('origin_country', '')`

PR #2994, opened 2017-08-08
Add TMDB to externals after filtering. Otherwise info will be lost

fernandog:
The same way we do in tvmaze:
https://github.com/pymedusa/Medusa/blob/develop/medusa/indexers/tvmaze/tvmaze_api.py#L491-L496 @p0psicles

@duramato this is why your tmdb country doesn't work for non tmdb shows


p0psicles:
My commit to supers PR has the same outcome. I'm fine with this one also.

fernandog:
I did this PR because I didn't know if that PR was going to be merged as per your comment as that is not a imdb info but tmdb's

PR #2996, opened 2017-08-09
New logging for scheduler, failed_processor and issue_submitter

fernandog:
small files
I have enough of log changes lol.
A regex replace would be better

I use this but it not always work, specially when logger.INFO is missing:
```regex
(?:\.format)(?:\s)*(?:\()(?:(.*?)(?:=)(.*?))(?:(?:,\s*)(.*?)(?:=)(.*))?(?:\)(?:,\s*)logger\.(?:INFO|DEBUG|WARNING)\))
```
replace with:
`, {'$1': $2, '$3': $4, '$5': $6})`

### DIFF
--- a/medusa/failed_processor.py
+++ b/medusa/failed_processor.py
@@ -33,41 +33,41 @@ class FailedProcessor(object):
 
         :return: True
         """
-        self._log(u'Failed download detected: ({nzb}, {dir})', {'nzb': self.nzb_name, 'dir': self.dir_name})
+        self.log(logger.INFO, u'Failed download detected: ({nzb}, {dir})'.format(nzb=self.nzb_name, dir=self.dir_name))
 
         releaseName = naming.determine_release_name(self.dir_name, self.nzb_name)
         if not releaseName:
-            self._log(u'Warning: unable to find a valid release name.', logger.WARNING)
+            self.log(logger.WARNING, u'Warning: unable to find a valid release name.')
             raise FailedPostProcessingFailedException()
 
         try:
             parsed = NameParser().parse(releaseName)
         except (InvalidNameException, InvalidShowException):
-            self._log(u'Not enough information to parse release name into a valid show. '
+            self.log(logger.WARNING, u'Not enough information to parse release name into a valid show. '
                       u'Consider adding scene exceptions or improve naming for: {release}'.format
-                      (release=releaseName), logger.WARNING)
+                      (release=releaseName))
             raise FailedPostProcessingFailedException()
 
-        self._log(u'Parsed info: {result}'.format(result=parsed), logger.DEBUG)
+        self.log(u'Parsed info: {result}'.format(result=parsed), logger.DEBUG)
 
         segment = []
         if not parsed.episode_numbers:
             # Get all episode objects from that season
-            self._log(u'Detected as season pack: {release}'.format(release=releaseName), logger.DEBUG)
+            self.log(logger.DEBUG, 'Detected as season pack: {release}'.format(release=releaseName))
             segment.extend(parsed.show.get_all_episodes(parsed.season_number))
         else:
-            self._log(u'Detected as single/multi episode: {release}'.format(release=releaseName), logger.DEBUG)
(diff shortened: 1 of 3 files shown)


medusa/failed_processor.py
```diff
@@ -52,30 +44,30 @@ def process(self):
...
         except (InvalidNameException, InvalidShowException):
             self._log(u'Not enough information to parse release name into a valid show. '
-                      u'Consider adding scene exceptions or improve naming for: {release}'.format
-                      (release=releaseName), logger.WARNING)
+                      u'Consider adding scene exceptions or improve naming for: {release}',
+                      {'release': releaseName}, logger.WARNING)
             raise FailedPostProcessingFailedException()
 
-        self._log(u'Parsed info: {result}'.format(result=parsed), logger.DEBUG)
+        self._log(u'Parsed info: {result}', {'result': parsed}, logger.DEBUG)
```
fernandog: I don't need to use brace here as I only need to pass the `message` to self._log

medusa/failed_processor.py
```diff
@@ -37,11 +29,11 @@ def __init__(self, dirName, nzbName):
 
     def process(self):
         """
-        Do the actual work
+        Do the actual work.
 
         :return: True
         """
-        self._log(u'Failed download detected: ({nzb}, {dir})'.format(nzb=self.nzb_name, dir=self.dir_name))
+        self._log(u'Failed download detected: ({nzb}, {dir})', {'nzb': self.nzb_name, 'dir': self.dir_name})
```
p0psicles: @labrys I think I need your help here. What's cleanest way of solving this? I would look into removing that helper.

fernandog: i forget to restore this one. self._log is a wrapper to log, we don't need the brace here

p0psicles: Yes. I know that's wrong. But I'm looking into removing the log helper anyway.

fernandog: oh ok!

labrys: @p0psicles what specifically would you like to solve?

p0psicles: Remove the _log method. Maybe use log adapter?

labrys: A simple log adapter could work, however I don't think one is necessary since it would see such limited use.

In my opinion `self._log` should be renamed `self.log` since it's the method we call directly.  The args/kwargs for `self.log` should get fixed so they match that of `logger.log`.

The `BraceAdapter` should not be used as `self.log` modifies everything regardless of level, so lazy formatting is not necessary.

fernandog: What about the \n that self._log adds?

labrys: @fernandog I'm not proposing changing the functionality of `self._log` just the name.  It appends `\n` to every log line regardless of log level.  I believe that @p0psicles is proposing a log adapter so we can do `self.log.debug(msg)` type logging.  While I would normally agree, the log adapter would have such limited use I think the single 2-line method that currently exists is sufficient and adding an adapter here isn't necessary.

medusa/failed_processor.py
```diff
@@ -37,45 +29,45 @@ def __init__(self, dirName, nzbName):
...
         if segment:
-            self._log(u'Adding this release to failed queue: {release}'.format(release=releaseName), logger.DEBUG)
+            self.log(logger.DEBUG, u'Adding this release to failed queue: {release}'.format(release=releaseName))
             cur_failed_queue_item = FailedQueueItem(parsed.show, segment)
             app.forced_search_queue_scheduler.action.add_item(cur_failed_queue_item)
 
         return True
 
-    def _log(self, message, level=logger.INFO):
+    def log(self, level, message):
```
p0psicles: aah okay, that's why it was private. So we can't do this.

medusa/failed_processor.py
```diff
@@ -33,41 +33,41 @@ def process(self):
...
```
labrys: No the `self.log` in this method should be the one thats private.  I'll PR.

Issue #2983, opened 2017-08-05
Support YggTorrent


IDerr:
Hi,
can you support Yggtorrent,
it's like the new "T411" 

Thanks a lot

p0psicles:
Currently in develop

PR #565, opened 2016-05-09
Fix pymedusa/SickRage/issues/562

duramato:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Always make sure the PR has a clear description of why,how, and what your trying to fix. The same goes for improvements or new features. Also if you refering to an issue, make sure a summary is available in the PR.
- [x] Read [contribution guide](https://github.com/PyMedusa/SickRage/blob/master/.github/CONTRIBUTING.md)

Fixes
https://github.com/pymedusa/SickRage/issues/562
https://github.com/pymedusa/SickRage/issues/563
https://github.com/pymedusa/SickRage/issues/564

PR #2998, opened 2017-08-10
Add provider Yggtorrent

p0psicles:
* Fixed convert_size helper, when passing with sep=None. Regex wasn't working as it should. (need to add unittests for this).

#2983 

For now without support for publishing date, as that's in french.
## TODO: 
- [x] Add unit tests to test te convert_size coverage.


adaur:
I should have posted here as well haha sorry

p0psicles:
What you should have done,  is use Medusa and it's awesome manual search feature that we build for you!

medusa/providers/torrent/html/yggtorrent.py
```diff
@@ -0,0 +1,174 @@
...
+                                      title, seeders)
+                        continue
+
+                    item = {
+                        'title': title,
+                        'link': download_url,
+                        'size': size,
+                        'seeders': seeders,
+                        'leechers': leechers,
+                        'pubdate': None,
```
p0psicles: Yeah date is in french, so I don't know if we should implement now. I think we should have some kind of translation table in place. @duramato do you know of another provider where we do pubdate translation?

duramato: None currently, foreign providers that have pubdate are using the ISO format mostly iir. 
I'd go for a English<->provider language map in a dictionary that could be passed to the date parser function, so the date parser function would only handle English

medusa/helper/common.py
```diff
@@ -205,9 +205,9 @@ def convert_size(size, default=None, use_decimal=False, **kwargs):
             scalar, units = size_tuple[0], size_tuple[1:]
             units = units[0].upper() if units else default_units
         else:
-            regex_units = re.search(r'(\w+)', size, re.IGNORECASE)
-            units = regex_units.group() if regex_units else default_units
-            scalar = size.strip(units)
+            regex_units = re.search(r'(\d*[\\.|\\,]?\d*)(\s?({scale}))'.format(scale='|'.join(scale)), size, re.IGNORECASE)
```
medariox: Shouldn't the character escaping be a single backslash? Also d+ as we want at least one match:
```
(\d+[\.|\,]?\d+)
```

fernandog: Sony need backslash at all. It's using raw: r''

medusa/providers/torrent/html/yggtorrent.py
```diff
@@ -0,0 +1,174 @@
...
+            'submit': ''
+        }
+
+        self.session.post(self.urls['login'], data=login_params)
+        response = self.session.get(self.url)
+        if not response:
+            log.warning('Unable to connect to provider')
+            return False
+
+        if 'Connexion' in response.text:
```
p0psicles: The thing is, I decided to go with Connexion, as that's a broader catch of exceptions. As I don't know if it will only respond to bad username/passwords entered? Maybe an account is disabled or banned, and a different message is shown.

duramato: You can check for `Mon compte` then, if it's login was successful. Do you can return false if incorrect user credentials,  return true if `Mon compte` is present aka login successful, else return false because login wasn't successful for some **other** reason

medusa/helper/common.py
p0psicles: Couldn't I just make it allot simpler by doing this: `([0-9,.]+)(\s?(B|KB|MB|GB|TB|PB))`. What am I missing here?

p0psicles: Using the current regex it won't match: `120,008,12B` which is actually also a valid size. Although I think rather rare. But still..

fernandog: \d : any digit
+: one or N digits
[.|,]?: optional dot or comma

By sewing In providers we remove the thousand separator

fernandog: Like here: https://github.com/pymedusa/Medusa/pull/565/files

p0psicles: Okay sure, then.. `([\d.]+)(\s?(B|KB|MB|GB|TB|PB))` ?
As we shouldn't need to take into account the comma then?

p0psicles:
Only thing left on this PR, is decide on the regex.

medusa/providers/torrent/html/yggtorrent.py
duramato:
```
    def login(self):
        """Login method used for logging in before doing search and torrent downloads."""
        login_params = {
            'id': self.username,
            'pass': self.password,
            'submit': ''
        }

        self.session.post(self.urls['login'], data=login_params)
        response = self.session.get(self.url)
        if not response:
            log.warning('Unable to connect to provider.')
            return False

        if 'Ces identifiants sont invalides' in response.text:
            log.warning('Invalid username or password. Check your settings.')
            return False

        if 'Mon compte' in response.text:
            return True
        
        log.warning('Unable to login to provider.')
        return False
```

This was more of what I had in mind 😇, for one reason, this way you know from the log message that it was actually wrong credentials or some other unexpected issue (like a ban is) with logging in.
What do you think @p0psicles 

@medariox

adaur:
They added a new column recently

p0psicles:
I think u already accounted for that. I did some successful searches today

adaur:
Hm actually their layout is messed up right now (size below seeders, etc)

p0psicles:
Oow OK. Nice timing lol.

PR #3009, opened 2017-08-14
Use medusa safe session exception handler

fernandog:
@p0psicles


fernandog:
@p0psicles is there a reason for clients not use Medusa**Safe**Session?

p0psicles:
If you want to have more control, like read act on certain status codes, you can opt to use the MedusaSession.

medusa/indexers/indexer_config.py
```diff
@@ -1,12 +1,14 @@
...
+"""Indexer config module."""
+
 import re
 
 from .tmdb.tmdb import Tmdb
 from .tvdbv2.tvdbv2_api import TVDBv2
 from .tvmaze.tvmaze_api import TVmaze
 from ..app import BASE_PYMEDUSA_URL
-from ..session.core import MedusaSession
+from ..session.core import MedusaSafeSession
```
p0psicles: I'd rather keep this MedusaSession as long as possible. Any reason you see to change this?
I know we do some exception handling on this. So this could potentially mess that up.

medusa/scene_exceptions.py
```diff
@@ -295,39 +279,31 @@ def _get_custom_exceptions(force):
...
-                                  for season in indexer_exceptions
-                                  for exception in indexer_exceptions[season]]
-                    custom_exceptions[indexer][indexer_id] = alias_list
-            except Exception as error:
-                logger.error(
-                    'Unable to update scene exceptions for {indexer}.'
-                    ' Error: {error}'.format(
-                        indexer=indexer, error=error
+                # When any Medusa Safe session exception, session returns None and then AttributeError when json()
+                jdata = indexerApi(indexer).session.get(location, timeout=60).json()
```
p0psicles: If we use the indexers session (MedusaSession) we need exception handling here.

fernandog: the `RequestException` or the broad one ?

p0psicles: Actually I would create a new session for scene_exceptions. And you could use the MedusaSafeSession for that. It's just to get the scene_exceptions from cdn.pymedusa.com. Don't need to use the indexers sessions for that. You could rename xem_session to session. Change it to MedusaSafeSession. And use that for all the session calls in scene_exceptions.

medusa/scene_exceptions.py
```diff
@@ -15,14 +15,14 @@
...
-from .session.core import MedusaSession
+from .session.core import MedusaSafeSession
 
 logger = logging.getLogger(__name__)
 
 exceptions_cache = defaultdict(lambda: defaultdict(set))
 exceptionLock = threading.Lock()
 
-xem_session = MedusaSession()
+xem_session = MedusaSafeSession()
```
fernandog: @p0psicles this `xem_session` is imported in scene_numbering.py so I prefer to kept the name

medusa/scene_exceptions.py
```diff
@@ -286,7 +286,7 @@ def _get_custom_exceptions(force):
             )
             try:
                 # When any Medusa Safe session exception, session returns None and then AttributeError when json()
-                jdata = indexerApi(indexer).session.get(location, timeout=60).json()
+                jdata = MedusaSafeSession().get(location, timeout=60).json()
```
fernandog: @p0psicles is it ok like this? sure I can use xem_session as it's only for cdn

p0psicles: You can easily reuse session objects. Not really efficient to create a new one each time

fernandog: done

medusa/scene_exceptions.py
fernandog: @medariox that's why I didn't renamed to safe_session as you said..

Should I create another session then?

fernandog:
@p0psicles what to do with this PR? If the MedusaSafe session commit is not good, the other commits are.

p0psicles:
let's just use the safe_session name, and be done with it okay?

fernandog:
done

PR #3014, opened 2017-08-15
Improved the overall UI, by overflowing the tables with a scrollbar

p0psicles:
This will result in only showing a scrollbar for tables that exceed the page viewport.

* Added buttons at the bottom of the page, when a scrollbar is visible, similar to the one that has been implement to scroll all the way up.

Biggest advantage of this is, that we will not have a messed up overal UI, like needing to scroll to the right for the menu button on mobile.

# TODO:
- [x] Implement for displayShow
- [x] Implement for snatchSelection
- [x] Implement for massUpdate
- [x] Test with multiple tables on a page, that exceed the width.
- [ ] ..

image

Maybe I wasn't clear before. But the buttons only _should_ appear when there are scrollbars visible. It's implemented, and works pretty well on screen size changes. There are some small issues to be worked out on screen initialization.

As Fernandog showed me in #3016, we already have class for this, named `horizontal-scroll`.
```css
.horizontal-scroll {
    overflow-x: auto;
}
```

I'll see if I can create something easy applyable and still keep the scroll-left-right buttons.

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -2555,7 +2555,7 @@ a.whitelink {
 }
 
 /* BacklogOverview */
-div#status-summary {
+#status-summary {
     padding-top: 10px;
     padding-bottom: 2px;
 }
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -4289,7 +4289,7 @@ CSS helper classes
 /* =======================================================================
  Scroll to top
 ========================================================================== */
-.scroll-top-wrapper {
+.scroll-wrapper {
     position: fixed;
     opacity: 0;
     visibility: hidden;
@@ -4332,3 +4332,37 @@ CSS helper classes
     display: block;
 }
 */
+
+/* =======================================================================
+ overflow for horizontal scrolling, without breaking ui.
+========================================================================== */
+
+.full-screen-scroll {
+    clear: none;
+}
+
+.scroll-wrap {
+    overflow-x: auto;
+}
+
+div#imageContainer {
+    position: fixed;
+    bottom: 50px;
+    right: 100px;
+}
+
+.scroll-wrapper.show {
+    visibility: visible;
+    cursor: pointer;
+    opacity: 1;
+}
+
+.scroll-wrapper.left {
+    position: fixed;
+    right: 150px;
+}
+
+.scroll-wrapper.right {
+    position: fixed;
+    right: 90px;
+}
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1788,7 +1788,6 @@ schedule.mako
     padding: 2px 5px;
     font-size: 13px;
     font-weight: bold;
-    border-radius: 5px;
 }
 
 .listing-default {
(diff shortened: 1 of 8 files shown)


p0psicles:
@OmgImAlexis what you think of the idea?

OmgImAlexis:
I'm really not likely the scroll bar and the buttons..

p0psicles:
What would be your alternative in making sure, our UI does not break when using large tables?
Because afaik this is the only solution to make it somewhat workable on mobile. Which currently it is not.

p0psicles:
@omgimalexis can you give me a code review an the js/css changes?

static/css/style.css
```diff
@@ -2555,6 +2554,12 @@ a.whitelink {
     color: rgb(255, 255, 255);
 }
 
+/* BacklogOverview */
+div#status-summary {
```
OmgImAlexis: Shouldn't need div here.

static/js/common/init.js
```diff
@@ -24,24 +24,63 @@ MEDUSA.common.init = function() {
...
+            $('.scroll-wrapper.right').addClass('show');
+        } else {
+            $('.scroll-wrapper.left').removeClass('show');
+            $('.scroll-wrapper.right').removeClass('show');
+        }
+    };
+
+    initHorizontalScroll();
+
     $(window).resize(function() {
```
OmgImAlexis: `on('resize'`

static/js/common/init.js
```diff
@@ -24,24 +24,63 @@ MEDUSA.common.init = function() {
...
         $('.backstretch').css('top', backstretchOffset());
+        initHorizontalScroll();
     });
 
     // Scroll Functions
     function scrollTo(dest) {
         $('html, body').animate({scrollTop: $(dest).offset().top}, 500, 'linear');
     }
 
+    $('#scroll-left').click(function(e) {
```
OmgImAlexis: `on('click`

static/js/common/init.js
```diff
@@ -24,24 +24,63 @@ MEDUSA.common.init = function() {
...
+        e.preventDefault();
+        $('div.horizontal-scroll').animate({
+            scrollLeft: '-=153'
+        }, 1000, 'easeOutQuad');
+    });
+
+    $('#scroll-right').click(function(e) {
```
OmgImAlexis: `on('click`

static/js/common/init.js
```diff
@@ -24,24 +24,63 @@ MEDUSA.common.init = function() {
         return offset;
     }
 
+    let initHorizontalScroll = function() {
```
OmgImAlexis: This should be `const`.

OmgImAlexis: What exactly is this function doing? I have a feeling there's a way better way todo this.

p0psicles: It's to check if we need to show the left and right buttons. It tries to do that using the scrollWidth, clientWidth values. I'm running it on init and on resize

Issue #3006, opened 2017-08-12
Post Processor adds extra subtitle language extension


doopsuiker:
**Branch/Commit: Master
**OS: Windows 10
**What you did: After a download finishes, I let run a script that extracts subtitles as release.name.eng.srt
**What happened: After post processing, the subtitle file is renamed as release.name.eng.eng.srt. I don't let Medusa rename the episodes
**What you expected: The subtitle file should remain as release.name.eng.srt

fernandog:
@doopsuiker 
without logs there is nothing that can be done about it

doopsuiker:
@fernandog 
Here are the relevant logs:
`
2017-08-15 18:38:35 INFO     Thread-27 :: [49e9fc3] Successfully processed.
2017-08-15 18:38:35 INFO     Thread-27 :: [49e9fc3] Processing succeeded for C:\Users\***\Downloads\Series\rick.and.morty.s03e05.1080p.web-dl.mkv
2017-08-15 18:38:32 DEBUG    Thread-27 :: [49e9fc3] Transaction with 1 queries executed
2017-08-15 18:38:32 DEBUG    Thread-27 :: [49e9fc3] 275274: Setter sets location to E:\Series\English\Rick and Morty\Season 03\rick.and.morty.s03e05.1080p.web-dl.mkv
2017-08-15 18:38:32 DEBUG    Thread-27 :: [49e9fc3] Transaction with 1 queries executed
2017-08-15 18:38:32 DEBUG    Thread-27 :: [49e9fc3] Moving file from C:\Users\***\Downloads\Series\rick.and.morty.s03e05.1080p.web-dl.eng.srt to E:\Series\English\Rick and Morty\Season 03\rick.and.morty.s03e05.1080p.web-dl.eng.eng.srt 
2017-08-15 18:38:32 DEBUG    Thread-27 :: [49e9fc3] Moving file from C:\Users\***\Downloads\Series\rick.and.morty.s03e05.1080p.web-dl.mkv to E:\Series\English\Rick and Morty\Season 03\rick.and.morty.s03e05.1080p.web-dl.mkv 
2017-08-15 18:38:32 DEBUG    Thread-27 :: [49e9fc3] Found the following associated files for C:\Users\***\Downloads\Series\rick.and.morty.s03e05.1080p.web-dl.mkv: set([u'C:\\Users\\***\\Downloads\\Series\\rick.and.morty.s03e05.1080p.web-dl.eng.srt'])
[...]

PR #3022, opened 2017-08-16
Add tests to rename_associated_file and fix 2 bugs. #Fixes 3006

medariox:

- Fixes bug reported here: https://github.com/pymedusa/Medusa/issues/3006
- Fixes bug with subtitles and absolute paths
- Lint post_processor.py

### DIFF
--- a/medusa/failed_processor.py
+++ b/medusa/failed_processor.py
@@ -44,8 +44,8 @@ class FailedProcessor(object):
             parsed = NameParser().parse(releaseName)
         except (InvalidNameException, InvalidShowException):
             self.log(logger.WARNING, u'Not enough information to parse release name into a valid show. '
-                      u'Consider adding scene exceptions or improve naming for: {release}'.format
-                      (release=releaseName))
+                     u'Consider adding scene exceptions or improve naming for: {release}'.format
+                     (release=releaseName))
             raise FailedPostProcessingFailedException()
 
         self.log(u'Parsed info: {result}'.format(result=parsed), logger.DEBUG)
@@ -74,4 +74,5 @@ class FailedProcessor(object):
 
     @property
     def output(self):
+        """Return the concatenated log messages."""
         return '\n'.join(self._output)
--- a/tests/test_rename_associated_file.py
+++ b/tests/test_rename_associated_file.py
@@ -83,9 +83,9 @@ import pytest
     },
     {  # p12: MKV with new basename
         'new_path': 'media/shows/riko or marty/season 3/',
-        'new_basename': 'Riko or Morty S03E05 Episode Name',
+        'new_basename': 'Riko or Marty S03E05 Episode Name',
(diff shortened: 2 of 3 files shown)


fernandog:
thanks!

medusa/failed_processor.py
```diff
@@ -44,8 +44,8 @@ def process(self):
             parsed = NameParser().parse(releaseName)
         except (InvalidNameException, InvalidShowException):
             self.log(logger.WARNING, u'Not enough information to parse release name into a valid show. '
-                      u'Consider adding scene exceptions or improve naming for: {release}'.format
-                      (release=releaseName))
+                     u'Consider adding scene exceptions or improve naming for: {release}'.format
+                     (release=releaseName))
```
fernandog: I would improve this saying something like this: 
 `Consider adding scene exceptions (alternative name or add name without year)`
@medariox @duramato what do you think?

p0psicles: Hmm don't know. `add show name without year` is also rather vague

tests/test_rename_associated_file.py
```diff
@@ -0,0 +1,113 @@
...
+
+import pytest
+
+
+@pytest.mark.parametrize('p', [
+    {  # p0: Subtitle with language. No subtitles dir set.
+        'new_path': 'media/shows/great show/season 1/',
+        'new_basename': 'Great Show - S01E04 - Best Episode',
+        'filepath': 'downloads/tv/great.show.s01e04.720p.hdtv.x264-group.en.srt',
+        'expected': 'media/shows/great show/season 1/Great Show - S01E04 - Best Episode.en.srt'
```
fernandog: @medariox it should only rename if `app.RENAME_EPISODES` is True right ? Otherwise subtitle will be renamed and the video file won't. Agree?

But... the pt-br must be converted to pt-BR even if rename is not enabled.
So how to deal with that?

medariox: This is already being done somewhere (didnt have the time to check where exactly). Basically if new_basename gets passed as None or an empty string. If video isn't renamed, the associated files aren't renamed either. The info can still be renamed and the language code for subs normalized tho.

fernandog: Ok

medusa/failed_processor.py
duramato: `Consider adding scene exceptions (https://github.com/pymedusa/Medusa/wiki/Scene-exceptions-and-numbering)`

Of course the wiki link should be a shorten url

p0psicles: Yeah agree

fernandog: @medariox can you add this?

medariox: Imo what we currently have is not too bad. Why remove the suggestion to improve the naming? It's a possible way to handle it. Maybe we could just add the shortened link in brackets after "Consider adding scene exceptions".

PR #3026, opened 2017-08-17
refactor transmission authentication and implement Retry.

p0psicles:

### DIFF
--- a/medusa/clients/torrent/transmission_client.py
+++ b/medusa/clients/torrent/transmission_client.py
@@ -54,7 +54,7 @@ class TransmissionBasicAuth(auth.HTTPBasicAuth):
             request.url,
             headers=request.headers,
             data=data,
-            verify=request.verify,
+            verify=False,
             auth=auth.HTTPBasicAuth(self.username, self.password)
         )
 
--- a/medusa/clients/torrent/transmission_client.py
+++ b/medusa/clients/torrent/transmission_client.py
@@ -18,9 +18,11 @@ from medusa.helpers import (
     is_info_hash_processed,
 )
 from medusa.logger.adapters.style import BraceAdapter
+from medusa.session.core import MedusaSession
 
 from requests.adapters import HTTPAdapter
 from requests.compat import urljoin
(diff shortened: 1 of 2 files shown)


p0psicles:
@fernandog I believe this works (in an ugly way). Next is to look for a way to replace the current BasicAuth hook, with a custom hook, that also does the storing of the token.

fernandog:
Thanks. Will try. Should we add RetryError in MedusaSession exception handler (when max retries reached)

p0psicles:
No definitely not

p0psicles:
@labrys i've implemted a custom auth hook, subclassed from HTTPBasicAuth. We're still testing this, but i'd appreciate a review from you at some point, because you know more about this stuff then I do.

p0psicles:
As from review,
- [ ] need to change all the token references with session_id, as technically it's not a token that we're storing.
- [ ] Review all the docstrings, as there is allot of copy/pasting in there

medusa/clients/torrent/transmission_client.py
```diff
@@ -29,6 +32,52 @@
...
+            if r.status_code == 409 and r.headers.get('x-transmission-session-id'):
+                self.token = r.headers.get('x-transmission-session-id')
+                self.token_expires = time() + 3600
+
+            request.headers.update({'x-transmission-session-id': self.token})
+
+    def __call__(self, request):
+        self.authenticate(request)
+        request.headers.update({'x-transmission-session-id': self.token})
+        return super(TransmissionBasicAuth, self).__call__(request)
```
labrys: If the session id is meant to contain the authentication information, then calling `super().__call__` is incorrect because it would resend the username and password for every request.

p0psicles: It would set the `Authorization` header on each request, which I think is correct.

p0psicles:
@fernandog should be ready to start some testing in prod environment.

fernandog:
worked nice. thanks. nice job!

OmgImAlexis:
Closing and reopening to trigger CI.

PR #3030, opened 2017-08-18
Fix the queries to show the cached season packs

p0psicles:
Season packs can be stored in two ways..
1. With season 1 and episodes "empty" for ex: '||'
2. With season 1 and episodes filled with all the episodes of the season. Like '|1|2|3|..'.

no. 2 was not yet visualized. But is with this change.

~~Second "feature" thats added to this pr, is adjust the season search strings for the provider Nyaa.
As only little anime season packs, are released with season x. I've adjusted it to search for season pack keywords.~~


medusa/providers/torrent/rss/nyaa.py
```diff
@@ -143,5 +143,21 @@ def parse(self, data, mode):
 
         return items
 
+    def _get_season_search_strings(self, episode):
+        search_string = {
+            'Season': []
+        }
+        season_search_keywords = ('batch', 'complete', 'season')
```
p0psicles: I know it's a little offtopic regarding the PR. But I had to make this change before, I could fix the manual search cache query.

Maybe for season packs, we should go with using these keywords. Then we could store them like how we currently do for the 'PROPER' keywords.

@duramato thoughts?

duramato: Currently non anime doesn't need that as we don't support downloading `complete` releases, which are the complete show. Scene anime also follows normal "show's" numbering.

This could be usefull for foreign or non "standard" naming, like for example instead of `Season 1` releases with `Series 1` that exist.
So something like yoou propose would be usefull still for everything.

p0psicles: Then I suggest we add an attribute to the providers named `season_pack_keywords` (or something else). And if that attribute exists, we use the season_search_string creation, as shown in this PR?

p0psicles: Continuing this in: https://github.com/pymedusa/Medusa/pull/3369.
I'll limit this PR to the db query changes.

medusa/search/manual.py
```diff
@@ -223,16 +222,30 @@ def get_provider_cache_results(indexer, show_all_results=None, perform_search=No
...
+                # If were not looking for all results, meaning don't do the filter on season + ep, add sql
+                if not int(show_all_results):
+                    # If it's an episode search, pass season and episode.
+                    common_sql += " AND season = ? AND episodes LIKE ? "
+                    add_params += [season, "%|{0}|%".format(episode)]
+
+            else:
+                # If were not looking for all results, meaning don't do the filter on season + ep, add sql
+                if not int(show_all_results):
+                    list_of_episodes = '{0}{1}'.format(' episodes LIKE ', ' AND episodes LIKE '.join(
```
p0psicles: This is the meat of this PR.
It essentially makes sure we query for a season pack when a season is passed and no episodes are passed. Which is the most often occurring situation. Or when all episodes for the season are passed. Like: |1|2|3|4|5|...etc|.

medusa/providers/generic_provider.py
```diff
@@ -645,9 +645,9 @@ def _get_season_search_strings(self, episode):
             episode_string = show_name + self.search_separator
 
             if episode.series.air_by_date or episode.series.sports:
-                episode_string += str(episode.airdate).split('-')[0]
+                search_string['Season'].append(episode_string + str(episode.airdate).split('-')[0])
```
p0psicles: This I needed fix. Got broken few commits back. With the season search templates.

PR #3032, opened 2017-08-18
Skip propers for wrong season/result

fernandog:
cur_proper.show.episodes is the episode we are searching for.
cur_proper.parse_result is the possible proper parsed result

```
FINDPROPERS :: [Speedcd] :: [4c2ace8] Search string: Show S07E06 PROPER
FINDPROPERS :: [4c2ace8] Skipping proper for wrong season/episode: Show.S06E02.PROPER.HDTV.x264-BATV
FINDPROPERS :: [4c2ace8] Skipping proper for wrong season/episode: Show.S05E04.PROPER.720p.HDTV.x264-0SEC
```

### DIFF
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -159,6 +159,14 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
                 log.info('Skipping non-proper: {name}', {'name': cur_proper.name})
                 continue
 
+            if not cur_proper.show.episodes.get(cur_proper.parse_result.season_number) or \
+                    any([ep for ep in cur_proper.parse_result.episode_numbers
+                         if not cur_proper.show.episodes[cur_proper.parse_result.season_number].get(ep)]):
+                log.info('Skipping proper for wrong season/episode: {name}', {'name': cur_proper.name})
+                if cur_proper.name not in processed_propers_names:
+                    self.processed_propers.append({'name': cur_proper.name, 'date': cur_proper.date})
+                continue
+
             log.debug('Proper tags for {proper}: {tags}', {
                 'proper': cur_proper.name,
                 'tags': cur_proper.parse_result.proper_tags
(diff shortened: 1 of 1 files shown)


medusa/search/proper.py
```diff
@@ -159,6 +159,14 @@ def _get_proper_results(self):  # pylint: disable=too-many-locals, too-many-bran
                 log.info('Skipping non-proper: {name}', {'name': cur_proper.name})
                 continue
 
+            if not cur_proper.show.episodes.get(cur_proper.parse_result.season_number) or \
+                    any([ep for ep in cur_proper.parse_result.episode_numbers
+                         if not cur_proper.show.episodes[cur_proper.parse_result.season_number].get(ep)]):
+                log.info('Skipping proper for wrong season/episode: {name}', {'name': cur_proper.name})
+                if cur_proper.name not in processed_propers_names:
+                    self.processed_propers.append({'name': cur_proper.name, 'date': cur_proper.date})
```
medariox: I don't think you can do this here. What if it finds a proper for another episode that is legitimately needed (eg. within the proper range)? It shouldn't be ignored for every subsequent proper search.

fernandog: yeah, you are right. But somehow this code worked and also didn't worked. Hours later is snatched again. No idea what's wrong

fernandog: @medariox maybe we shouldn't add this to self.processed_propers so it would work as you said. What do you think?

p0psicles: You can't just skip the proper result here? Isn't that what we already did?

fernandog: What medariox said it in that loop im looking for S01E02 proper. So if is a season/episode mismatch like S01E01, it will be skipped. But in the next item, we are searching for S01E01, it will never get that proper because its on the "already processed propers"

p0psicles: aah I understand, and yes I agree. We should ignore it, but not throw it away.

fernandog:
Just need a review in the code as I did a while ago. Maybe it can be improved.

p0psicles:
If you upmerge you can merge

fernandog:
@p0psicles upmerge? didn't understood

p0psicles:
Merge develop in it.

fernandog:
@p0psicles 
need approve again

PR #3034, opened 2017-08-20
Make the homepage full width

OmgImAlexis:


p0psicles:
I choose for the 10 col width, as I thought it looked nice. So that was a personal preference. If i'm really the only one who thinks so, I don't mind in changing that.

If you do change this, you'll also need to look into the div.row.wide classes, which should be removed, together with the css. You can check those in the massUpdate and log pages.

OmgImAlexis:
The main reason for switching it to 12 is that all other pages now use the full width apart from the homepage depending on the width. Seems more universal now.

For now let's just merge this and I'll look at the other classes when I get a chance.

p0psicles:
The full width is currently  really more an exception then anything else. Just two or three pages.

duramato:
Screenshoots of this change?

OmgImAlexis:
Under 992px we already use the full width layout.

Before

After

duramato:
I feel that it had more whitespace to show more of background?

OmgImAlexis:
@duramato we either need to center it and reduce the size to 10 on all sizes or keep it at all width.

duramato:
Just center it then, why reduce the width too? 
It seems to me that it already reduces the width when needed, just needs to have the width uniform on both sides (that's what you mean by centering right?), mobile:

image

No need to reduce it for every display sizes, my desktop:

image

p0psicles:
The show list page, looks like it's not centered, because it's making use of a table. the table has to much info in it. And that's why it's trying to grow beyond the 10 columns width on the right. If you have a higher resolution monitor, it will not use that much space.

To get back on topic of the PR. This comes down to personal preference. If one changes this to full width, and merges the PR, a next can create a PR and change it back to col-lg-10.

So I see only two options here. 1. keep as it is. or 2. make is configurable.
Do keep in mind that for no. 2, it will be allot more work then just change the layout.

We would at least need to move the layout row, from the layout.mako to each individual page. (which might be a good thing, as then we could get rid of the row.wide hack.

fernandog:
@p0psicles @OmgImAlexis this worked for me

p0psicles:
U sure? Tried wide view and looked at massEdit or logs?

fernandog:
@p0psicles mine look ok. How it looks for you?

p0psicles:
@fernandog, didn't notice you also removed the .wide class. But yes then it should look good. So as is, you would have the choice to have it wide or narrow **for all pages**.

We could go with that for now. Or address this: "_Then we still can use your variable. But we could force some pages to be wide (when narrow is selected), like the logs. And we could remove the wide class._"

p0psicles:
Okay displayShow is really a bitch. For we kinda implemented a hack, where we'd close the row/column set in the default layout, and inserted new rows/columns, with the absolute positioned summaryBackGround in between. Apparently someone discovered that absolute positioned (full width) elements don't like to be in a bootstrapped grid.

I've cleaned up the displayShow page a little, but i'm stuck on the full width background. I'll add the commit. While I know it's not finished. If you want to start fresh, just revert the commit.

To display the issue:
image

fernandog:
non-wide:

image

p0psicles:
Okay while where hacking. Check again with latest commit.

p0psicles:
Again, if we're moving the content-row from the layout to the individual pages, we don't need this hack.

fernandog:
Thanks. Worked

p0psicles:
@OmgImAlexis what do you want to do with this PR?

medariox:
I think the wide layout looks great. What's holding this back?

p0psicles:
What's holding it back is that some like the narrow layout.

medariox:
Wasn't it made optional with [this](https://github.com/pymedusa/Medusa/pull/3034/commits/4783ad6be1af4dfe54edef0d6e61fa8387630b04) commit?

p0psicles:
Not as I remember

p0psicles:
Okay, if someone can give this some tests in full width and small with the dark and light css, then this should be okay now.

DisplayShow can now also be used full width as narrow.

medariox:
_Some issues I've noticed so far:_

Weird box always present on top:
img

Horizontal scrollbar always present even with enough space and high resolution (1080p in my case):
img2

p0psicles:
The weird box on all pages, is resolved.
The horizontal scrollbars, do you see them with the wide and narrow layout?

medariox:
@p0psicles 
It probably had something to do with the weird box, because it is fixed as well now. All good, thanks! 👍

fernandog:
@p0psicles need to change /views/manage_subtitleMissedPP.mako too

fernandog:
@p0psicles nevermind, its ok

medusa/__main__.py
```diff
@@ -922,6 +922,7 @@ def initialize(self, console_logging=True):
             app.SELECTED_ROOT = check_setting_int(app.CFG, 'GUI', 'selected_root', -1)
             app.BACKLOG_PERIOD = check_setting_str(app.CFG, 'GUI', 'backlog_period', 'all')
             app.BACKLOG_STATUS = check_setting_str(app.CFG, 'GUI', 'backlog_status', 'all')
+            app.LAYOUT_WIDE = bool(check_setting_int(app.CFG, 'GUI', 'layout_wide', 0))
```
medariox: `check_setting_bool` instead?

p0psicles: jip!

PR #3037, opened 2017-08-21
Fix exception thrown when no series added and viewlog is opened.

p0psicles:
* Renamed the #showID to #series-id, as 1. i'd rather use hyphened. And 2. the id showID is already used on pages. And i'd like to prevent this from being reused.

### DIFF
--- a/views/home.mako
+++ b/views/home.mako
@@ -12,9 +12,13 @@
 <meta data-var="max_download_count" data-content="${max_download_count}">
 </%block>
 <%block name="content">
-<% random_show = choice(app.showList) if app.showList else None %>
-<input type="hidden" id="showID" value="${getattr(random_show, 'indexerid', '')}" />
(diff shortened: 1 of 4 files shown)


fernandog:
@p0psicles AFAIK using '<%!' it caches variables so background will always be the same

views/home.mako
```diff
@@ -12,9 +12,13 @@
 <meta data-var="max_download_count" data-content="${max_download_count}">
 </%block>
 <%block name="content">
-<% random_show = choice(app.showList) if app.showList else None %>
-<input type="hidden" id="showID" value="${getattr(random_show, 'indexerid', '')}" />
-<input type="hidden" id="series_slug" value="${getattr(random_show, 'slug', '')}" />
+<%!
```
OmgImAlexis: @fernandog is right this should have been `<%`.

PR #3048, opened 2017-08-22
Fix "Error loading IMDb info: local variable 'countries' referenced b…

duramato:
…efore assignment"

### DIFF
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -8,6 +8,7 @@ from __future__ import unicode_literals
 import logging
 from collections import OrderedDict
 from datetime import datetime, timedelta
+
 from dateutil import parser
 
 from medusa.app import TMDB_API_KEY
@@ -16,6 +17,7 @@ from medusa.indexers.indexer_exceptions import IndexerError, IndexerException, I
 from medusa.logger.adapters.style import BraceAdapter
 
 from requests.exceptions import RequestException
+
 import tmdbsimple as tmdb
 
 log = BraceAdapter(logging.getLogger(__name__))
@@ -175,7 +177,7 @@ class Tmdb(BaseIndexer):
     def search(self, series):
         """Search TMDB (themoviedb.org) for the series name.
 
-        :param series: the query for the series name
+        :param series: The query for the series name
         :return: An ordered dict with the show searched for. In the format of OrderedDict{"series": [list of shows]}
         """
         series = series.encode('utf-8')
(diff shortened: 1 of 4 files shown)


fernandog:
@p0psicles returns string not tuple.

medusa/indexers/tmdb/tmdb.py
```diff
@@ -190,25 +192,25 @@ def search(self, series):
...
+        :return: A list with the show's country codes
         """
         show_info = self._get_show_by_id(tmdb_id)['series']
 
-        return show_info.get('origin_country', '')
+        if show_info and show_info.get('origin_country'):
+            return show_info['origin_country'].split('|')
 
     def _get_show_by_id(self, tmdb_id, request_language='en'):  # pylint: disable=unused-argument
-        """Retrieve tmdb show information by tmdb id, or if no tmdb id provided by passed external id.
```
medariox: @p0psicles 
I removed this because the ```tmdb_id``` isn't even optional and it was just a bit confusing.

medusa/subtitles.py
```diff
@@ -211,17 +215,18 @@ def from_ietf_code(code, unknown='und'):
...
     :return: the country name
     :rtype: str
     """
-    country_name = babelfish.Country(country).name if country and country in babelfish.COUNTRIES else ''
-
-    return country_name
+    try:
+        return Country(code.upper()).name
+    except ValueError:
+        return
```
fernandog: @medariox why check ValueError if we already have supported values in babelfish.COUNTRIES?

medariox: @fernandog 
It should be faster since it only looks up the value once, instead of twice (if in babelfish.COUNTRIES + Country(code)), it looks cleaner (at least to me) and it's a more pythonic approach.

medariox: Are we still talking about the code above?

p0psicles: I agree that using the exception handling looks cleaner. Tnx.

fernandog:
@medariox need updated `to_json` method. It only has `countries` not `country_code`

medariox:
@fernandog 
That's always been the case, why adding the countries now? ```imdb_countries``` are actually the country codes as a list (just like IMDb uses them) and as far as I can see the whole properties actually try to simulate what IMDb does. I don't think we should add another property if it isn't needed, it can be done at any point in the future if required.

fernandog:
@medariox you didn't understood. the `to_json` returns the JSON representation of the object. And it's missing the country_code. Everything is in there except the country_code

medariox:
@fernandog Added.

Issue #3012, opened 2017-08-15
[ FEATURE ] Snatched Removes File Name in GUI


bnevets27:
When an episode gets snatched to upgrade the quality, the previous version file name of the episode gets removed from the GUI even though the old file still exists. 

Expected behavior is that the filename and size of the file should stay in the GUI until it gets replaced with the snatched file.

Attached is a screenshot which will hopefully make more sense. Medusa shows 21 of 22 files but the filenames aren't shown for the snatched/upgraded episodes.

http://i.imgur.com/S9vRK4c.jpg

fernandog:
It's by design as the current status is SNATCHED and this status doesn't have size or filename - even its on disk.

p0psicles:
@bnevets27 i agree with you that this is not how you would expect this to work. But the underlying quality system is not easily modified. So don't expect us to change this anytime soon. But it will be fixed eventually.

labrys:
Closing this as it has been added to the  #282 **Master Feature List** to track all feature requests.  Discussion specific to this feature can continue in this thread.

PR #3051, opened 2017-08-23
Show file size and location for SNATCHED status

fernandog:
I don't agree because SNATCHED means no size and no location
But as @p0psicles said it should show I'm creating this so other can feedback

Related: https://github.com/pymedusa/Medusa/issues/3012


p0psicles:
It should show that the existing file if different from the snatched release. As long as there is not room for confusion I think it's a great addition.

fernandog:
@duramato comments?

duramato:
I have mixed feelings about this, on one way it makes sense to show the info of the **existing** but on the other hand is as the file wasn't there anymore, it's about to be replaced, thats why it's marked as snatched.

p0psicles:
This PR needs work. We should clearly show that the current existing file belongs to a previously downloaded episode. So there shouldn't be any room for confusion. So in other words, this pr also requires changes to the UI.

p0psicles:
ok I have a proposal. What if we mark the pill with a blue border, when it's snatching an already downloaded episode? Like this:
image

medariox:
This seems to be the best solution for now, but honestly I'm not a huge fan of blue border... kinda reminds me of Windows XP 😨

p0psicles:
Yeah I've used a turquoise blue in the last commit. But I think that's not that important. Can be pink or  purple, I don't care. But it should be clearly visible compared to all the color qualities we have.

fernandog:
Just add a "Existing file:" before the file label when status is snatched. Dont change pill

labrys:
Personally I think if the existing file information (such as size) is shown and the status is snatched its obvious that the file information there is for the existing file.

p0psicles:
Personally I like the small visual indication. As not everyone has size or file visible. But everyone has the status / quality pills visible. Especially on mobile.

fernandog:
Ok. Just giving my 2c as the issue is in the file label and not in the status pill

labrys:
I think the issue here is the same issue as with the database.  We are trying to use one indicator to show two separate statuses.  We really need to start splitting existing statuses from future (aka snatched) statuses.

p0psicles:
I'm going to implement a nice border color, and merge it.
If any objection, please state them.

fernandog:
My argument against is: the issue is in the file label and not in the status pill. Users will wonder what is the border.
as I said before I would add "Existing file:" before the file label.

NicoLeOca:
not all of us are in front of Medusa [put the figure you feel ok with] times a day.
I support @labrys and @fernandog comments, you can't use one indicator for several status
People will get confused quickly

labrys:
Personally I feel it's clear that if a file exists but the status is snatched then the file that exists will be replaced at some point.  What I feel is a bigger issue is that the quality of the existing file is not listed and gets overwritten in the database.  Also there is no way of listing multiple qualities snatched (for example when using manual search).

p0psicles:
What do we do? Just add a few fields in the db?

labrys:
Well the quick fix would be to split the `status` field into two parts, `existing_status` for files in the library and `pending_status` for the status of a snatch.  Ideally though, the status would be separated into its own table so you could have a one-to-many relationship between a single episode and multiple statuses.

p0psicles:
Okay, shouldn't be that much work, if we can create a table for it, and abstract it for the rest of the app.

p0psicles:
For now i'll remove my UI changes. And merge it as basic as can be.

medariox:
Too much changed to make this work, sorry.

PR #3057, opened 2017-08-24
Use post in stead of get, for massUpdate and massEdit.

p0psicles:

### DIFF
--- a/static/js/mass-update.js
+++ b/static/js/mass-update.js
@@ -12,7 +12,14 @@ $(document).ready(function() {
             return;
         }
 
-        window.location.href = $('base').attr('href') + 'manage/massEdit?toEdit=' + editArr.join('|');
+        const submitForm = $(
+            '<form method=\'post\' action=\'' + $('base').attr('href') + 'manage/massEdit\'>' +
+            '<input type=\'hidden\' name=\'toEdit\' value=\'' + editArr.join('|') + '\'/>' +
+            '</form>'
+        );
+        submitForm.appendTo('body');
+
+        submitForm.submit();
     });
 
     $('.submitMassUpdate').on('click', function() {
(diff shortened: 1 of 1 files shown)


fernandog:
Does this fix the `Cannot add all existing shows / mass update due to URL limitation` issue ?

p0psicles:
It does fix the mass update error. But it doesn't fix the add existing show error. I'll look at that separate.

static/js/mass-update.js
```diff
@@ -95,19 +102,18 @@ $(document).ready(function() {
...
-            toDelete: deleteArr.join('|'),
-            toRemove: removeArr.join('|'),
-            toMetadata: metadataArr.join('|')
+        if (updateArr.length + refreshArr.length + renameArr.length + subtitleArr.length + deleteArr.length + removeArr.length + metadataArr.length === 0) {
+            return false;
+        }
+        const url = $('base').attr('href') + 'manage/massUpdate';
+        const params = 'toUpdate=' + updateArr.join('|') + '&toRefresh=' + refreshArr.join('|') + '&toRename=' + renameArr.join('|') + '&toSubtitle=' + subtitleArr.join('|') + '&toDelete=' + deleteArr.join('|') + '&toRemove=' + removeArr.join('|') + '&toMetadata=' + metadataArr.join('|');
+        $.post(url, params, function() {
+            location.reload(true);
```
OmgImAlexis: Is it the best idea to refresh since we may end up sending the request twice because of the form?

p0psicles: I don't think you can send the post twice. The reload will simply reload the page.

labrys: With a nonce you could still refresh and delay the response until action is completed.  A nonce would also prevent a doubled action.

p0psicles: Let me stop you guys right there. I'm not interested in creating all kind of fancy features in our "old" frontend (like utilizing the websockets for this). I don't want to spend my time on that. If you think that the current PR doesn't hold enough value as is, then i'd say remove it. As I got lot's of things to work on (like bootstrapping the config pages or finishing super's binsearch provider).

Keep in mind, the massUpdate/Edit page in practice is used once by user, and then the user moves on in his/her use flow. Also I still don't see how we can send the request twice **by accident**.

Issue #2205, opened 2017-02-18
Make show selector auto size


fernandog:
May help avoid the two rows of menu and make div of buttons more to the left

![image](https://cloud.githubusercontent.com/assets/2620870/23093192/2bcdccc8-f5bb-11e6-8d09-96978ddea8af.png)

fernandog:
@p0psicles In the past I tried fixing this but the upper|down buttons are a mess to align and etc.  I think we can remove then. what do you think?

PR #3059, opened 2017-08-24
Show selector with auto width in displayShow

fernandog:
Implements https://github.com/pymedusa/Medusa/issues/2205

Before:
image

After:
image


p0psicles:
Screenees?

fernandog:
@p0psicles updated PR body

views/partials/showSelector.mako
```diff
@@ -1,8 +1,8 @@
-<div id="showSelector" class="hidden-print col-md-4 col-sm-12 col-xs-12 pull-md-left" style="margin-left:-21px">
+<div id="showSelector" class="hidden-print col-md-2 col-sm-12 col-xs-12 pull-md-left" style="margin-left:-21px">
     <div class="form-inline">
             <div>
                 <div class="select-show-group pull-left top-5 bottom-5">
-                    <select id="select-show" class="form-control input-sm-custom" style="height:25px;padding:1px">
+                    <select id="select-show" class="form-control input-sm-custom" style="width:auto;height:25px;padding:1px">
```
p0psicles: Can you move this to style.css while your at it?

p0psicles:
You tested on all screen sizes?

fernandog:
@p0psicles I resized browser window and looks good while doing it. Also tested on mobile

static/css/style.css
```diff
@@ -1116,6 +1116,12 @@ div#container-display-show {
     width: 300px;
 }
 
+#show-selector {
+    width: auto;
+    height:25px;
+    padding:1px;
+}
+
```
fernandog: @p0psicles didn't have the same effect

image

sharkykh: Probably getting overridden by bootstrap.
Try using `!important`, even though I read that it's unrecommended.

sharkykh: Correction:
You used `#show-selector` instead of `.show-selector`.

fernandog: @sharkykh that didn't work too.  Can you try? Feel free to commit

p0psicles: you added the id `show-selector`. that doesn't exist. You should modify the existing `select#select-show`. then it works fine.

p0psicles:
Why would you want to get rid of the 100% width? As currently it's filling out the container (div). So when your viewing on smaller sizes, it wraps to a new line, as how you can see it. You can make the select smaller, but that won't have any effect on the wrapping.

sharkykh:
@p0psicles You can also do `width: 100%;` within a `@media max-width: [insert width]px` (or something like that`

p0psicles:
@fernandog check it out.
I removed the bootstrap columns. And made it wrap to next line, using width: auto.
image

I made some other changes to make sure the show-selector looks good on narrow screen widths.

PR #3060, opened 2017-08-24
Warn about scene exception without year

fernandog:
@medariox

`SEARCHQUEUE-MANUAL-311945 :: [Rarbg] :: [4c2ace8] Consider adding 'Bull' in scene exceptions for show 'Bull (2016)'`


p0psicles:
I think we should have a more visual context based warning system. Where we know the user won't miss it. Like utilizing the alert modals in the top of the page. And make sure a user can "acknowledge" them. I think that, that would boost the value of this PR.

In this case we could show such an alert, in the manualSearch page.

Issue #3061, opened 2017-08-24
[ FEATURE ] Add release name to email notification


doopsuiker:
Hi,

Is it possible to add the full release name of an episode to the email notification?
In this way, we would exactly know what we downloaded instead of only "1080p WEB-DL"

fernandog:
@doopsuiker if you want to try, just stop Medusa and change that file like this:
https://github.com/pymedusa/Medusa/pull/3062/files

doopsuiker:
@fernandog 
I meant adding an extra line with the full release name like this:
Release name: Game.of.Thrones.S07E06.Beyond.the.Wall.1080p.AMZN.WEB-DL.DDP5.1.H.264-GoT.mkv

fernandog:
so you want both the "1080p WEB-DL" and the full release name ? Cause this change will show full release name instead of "1080p WEB-DL"

doopsuiker:
@fernandog 
I want an extra line with the full release name.
Your change only changes this:

Episode: Suits - 7x07 - Full Disclosure - 1080p WEB-DL
 instead of
Episode: 7x07 - Full Disclosure

I want an extra line like this:
Release name: Show.S07E06.1080p.AMZN.WEB-DL.DDP5.1.H.264-RG.mkv

labrys:
Closing this as it has been added to the  #282 **Master Feature List** to track all feature requests.  Discussion specific to this feature can continue in this thread.

Issue #168, opened 2016-03-18
[ FEATURE ] Different notification for episode being replaced with higher quality


OmgImAlexis:
At the moment we send out the same notification for new episodes being added and for episodes being replaced with a higher quality. It'd be nice to have the notification state that the episode is being replaced and/or even a separate notification toggle for the two different events.

OmgImAlexis:
This would also include things like when a version gets nuked and a proper gets snatched/downloaded as currently they send the same notification as the base file downloading.

@doucheymcdoucherson

fernandog:
https://github.com/pymedusa/SickRage/pull/734

p0psicles:
@fernandog can this be closed?

fernandog:
Higher quality not implemented.
Need some rewrite in notifiers.

labrys:
Added to master feature request list - discussion for feature will continue here even though issue is closed.

PR #3062, opened 2017-08-24
Improve notifiers by passing a dict and allow to send more information

fernandog:
Implements: https://github.com/pymedusa/Medusa/issues/3061


fernandog:
@OmgImAlexis with last commit this is fully implemented:

https://github.com/pymedusa/Medusa/issues/168

fernandog:
screenshot_20170825-120506

fernandog:
quick looking other notifiers seems their are not following any pattern to the message. 

Like, for NMA, the title is "common.NOTIFY_SNATCH" and the message its the (don't laugh) message.
For other notifiers the common.NOTIFY_SNATCH is in the body/message

So I won't touch them for now as I don't how to add this new info without knowning how they would appear in the notifier 

ping @p0psicles

fernandog:
Maybe pass as *args so we can keep current behavior and we can study each notifier at time

NicoLeOca:
HI @fernandog 
I use Pushover on IOS and I would be happy to test your improvement

fernandog:
I haven't touch that yet. Still waiting some discussion about possible pattern to notifiers text/msg

fernandog:
If no one want this, i will close it. as I said I would like some inputs about making default messages for all notifiers

NicoLeOca:
@fernandog 
Could you please tell us what we can talk about?
What possibilites do we have? which fields are available?

duramato:
The format should be like this:

image

If there is a "title" field like in that image, that's where the "Started Download", "Download Finished", etc. should go and in the body what you have in your example. Outherwise if the notifier just has a "body" it should be: "Started Download: Show - 1x01 -blah -1080p WEB-DL" for example

NicoLeOca:
@fernandog 
would you mind having a look at pushover notifier please?
I am really interested in this improvement

OmgImAlexis:
@fernandog what's the go with this?

p0psicles:
I think we should close this and create a Notification class. Gives more flexibility

PR #3069, opened 2017-08-28
Insensitive comparison in scene exceptions check

fernandog:
@duramato


fernandog:
@p0psicles can we add this to release too?

p0psicles:
at some point it's done. You understand that right? Else we can just as well stop doing release branches.

PR #3091, opened 2017-09-11
Fix not sending the correct xem url

duramato:

### DIFF
--- a/medusa/scene_exceptions.py
+++ b/medusa/scene_exceptions.py
@@ -38,6 +38,7 @@ logger = logging.getLogger(__name__)
 exceptions_cache = defaultdict(lambda: defaultdict(set))
 exceptionLock = threading.Lock()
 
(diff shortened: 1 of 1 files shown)


medusa/scene_exceptions.py
```diff
@@ -38,6 +38,7 @@
 exceptions_cache = defaultdict(lambda: defaultdict(set))
 exceptionLock = threading.Lock()
 
+VALID_XEM_ORIGINS = {'anidb', 'tvdb', }
```
medariox: Was this meant to be a tuple instead?

medariox: ^ @duramato

duramato: Not my doing, @labrys

labrys:
@medariox nope a set

medusa/scene_exceptions.py
labrys: Nope it was meant to be a set

PR #3094, opened 2017-09-13
Adding HeBits torrent provider

smoresmores:


medusa/providers/torrent/html/hebits.py
```diff
@@ -0,0 +1,199 @@
...
+                return items
+
+            # Skip column headers
+            for row in torrent_rows:
+                try:
+                    heb_eng_title = row.find('div', class_='bTitle').find(href=re.compile("details\.php")).find('b').get_text(strip=True)
+                    index = heb_eng_title.index('/')
+                    title = heb_eng_title[index+1:]
+
+                    download_url = urljoin(self.url, row.find('div', class_='bTitle').find(href=re.compile("download\.php"))['href'])
```
medariox: Please use urljoin() after the ```if not all([title, download_url]):``` check further below.

medusa/providers/torrent/html/hebits.py
```diff
@@ -0,0 +1,199 @@
...
+                    torrent_size = row.find('div', class_='bSize').get_text(strip=True)
+                    size = convert_size(torrent_size[5:], sep='') or -1
+
+                    item = {
+                        'title': title,
+                        'link': download_url,
+                        'size': size,
+                        'seeders': seeders,
+                        'leechers': leechers,
+                        'pubdate': None,
```
medariox: Provider has no publishing date?

medusa/providers/torrent/html/hebits.py
```diff
@@ -0,0 +1,199 @@
...
+            'password': self.password,
+        }
+
+        response = self.session.post(self.urls['login'], data=login_params)
+
+        if not response or not response.text:
+            log.warning('Unable to connect to provider')
+            return False
+
+        if response.text == 'OK':
```
medariox: Are you sure this is what you want? That expects the response.text to be exactly ```OK``` -  if you want to check if the response contains ```OK``` you would do it like this:
```
if 'OK' in response.text:
    return True
```
Same about the rest below.

smoresmores: Yes, this is the exact response. Tested and it's working.

medusa/providers/torrent/html/hebits.py
```diff
@@ -0,0 +1,202 @@
...
+            return True
+        elif response.text == 'Banned':
+            log.warning('User {0} is banned from HeBits', self.userename)
```
medariox: Small typo ```self.userename```.

medusa/providers/torrent/html/hebits.py
```diff
@@ -0,0 +1,202 @@
...
+        :param data: The raw response from a search
+        :param mode: The current mode used to search, e.g. RSS
+
+        :return: A list of items found
+        """
+        items = []
+
+        with BS4Parser(data, 'html5lib') as html:
+            torrent_table = html.find('div', class_='browse')
+            torrent_rows = torrent_table('div', class_=re.compile("^line")) if torrent_table else []
```
medariox: Please use single quotes everywhere (same for below).

medusa/providers/torrent/html/hebits.py
```diff
@@ -0,0 +1,202 @@
...
+
+            # Continue only if at least one release is found
+            if len(torrent_rows) < 2:
+                log.debug('Data returned from provider does not contain any torrents')
+                return items
+
+            for row in torrent_rows:
+                try:
+                    heb_eng_title = row.find('div', class_='bTitle').find(href=re.compile("details\.php")).find('b').get_text(strip=True)
+                    title = heb_eng_title.split('/')[1].strip()
```
medariox: Do you really need to strip() again here (the line above has already strip=True).

smoresmores: This one is necessary. The first isn't so I deleted the strip=True.

Issue #3140, opened 2017-09-30
Sizes display N/A in manual search for all NZB providers


labrys:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:** #dc8382b
**OS:** Windows
**What you did:**  Manual search an episode
**What happened:**  No sizes listed for any results.  All come up N/A
**What you expected:**  Sizes to be displayed
**Logs:**
```
N/A
```
![image](https://user-images.githubusercontent.com/1867464/31046001-b30d5952-a5be-11e7-85d9-a148dc2930f2.png)

medariox:
Which provider are we talking about here?

labrys:
All providers for me (all nzb ofc).

PR #3143, opened 2017-10-01
Fix newznab and torznab attributes not being parsed

labrys:
`newznab` and `torznab` namespaced tags are not parsed properly by BeautifulSoup 4 v4.6.0
Using a regex to find tags works until the bug is fixed.
https://bugs.launchpad.net/beautifulsoup/+bug/1720605


labrys:
Fixes #3140

PR #3151, opened 2017-10-07
don't use absolute episode numbering to season/ep transformation, whe…

p0psicles:
…n scene_season is detected.

This can happen for anime titles which map to a season scene name.

Potential fix for #3152 

Wil fix searches for releases:
* `[HorribleSubs].Kekkai.Sensen.&.Beyond.-.01.[1080p].mkv`
* `Shokugeki.no.Soma.S3.-.01.[1080p].mkv` (food wars)

### DIFF
--- a/medusa/name_parser/parser.py
+++ b/medusa/name_parser/parser.py
@@ -167,6 +167,7 @@ class NameParser(object):
             log.debug('Scene numbering enabled series {name} is anime',
                       {'name': result.show.name})
             scene_season = scene_exceptions.get_scene_exception_by_name(result.series_name)[1]
+
             for absolute_episode in result.ab_episode_numbers:
                 a = absolute_episode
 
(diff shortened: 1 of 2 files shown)


p0psicles:
This will fix anime titles not being recognized when they are not yet added to xem.

The culprit of the issue can be found here: https://github.com/pymedusa/Medusa/blob/feature/experimental-anime-parsing/medusa/scene_numbering.py#L161

And a number of bugs located in one really bad written function.
1. The first query used for anime parsing is this:
`SELECT absolute_number FROM scene_numbering WHERE indexer = ? and indexer_id = ? and scene_absolute_number = ? and scene_season = ?`

That does a query on scene_season + scene_absolute_number. This query relates to the custom absolute numbering in the shows displayShow UI. Like here:
image

When saving the scene_absolute number, it will only store the scene_absolute episode number. But not with the scene_season.

But when searching for a show that has a season scene name, it will also have the scene_season availabl.e As for the example: '[HorribleSubs].Kekkai.Sensen.&.Beyond.-.01.[1080p].mkv' translating to 2.

Unfortunately for anime scene_season is not stored in the db table scene_numbering, and thus is cannot find it.
image.

PR #3156, opened 2017-10-12
Fix thumbnails not being created. Fixes #3155

### LINKED ISSUE #3155: Episode thumbnails are not being created
**Branch/Commit:** 3f23fd2

**OS:** Debian Stretch (4.9.0-3-amd64 #1 SMP Debian 4.9.30-2+deb9u3 (2017-08-06) x86_64)

**What you did:** Added new episodes. Tried "Force Full Update".

**What happened:** Does not retrieve episode thumbnails. Log reports: "No thumb is available for this episode, not creating a thumb"

**What you expected:** The episode thumbnails to be created. They do exists on tvdb as can be seen at: https://www.thetvdb.com/?tab=season&seriesid=327109&seasonid=708372&lid=7

**Logs:**

> 2017-10-11 22:19:52 DEBUG    SHOWQUEUE-REFRESH :: [3f23fd2] 327109: Formatting pattern: %SN - S%0SE%0E - %EN -> Mr. Mercedes - S01E07 - 007 - Willow Lake
> 2017-10-11 22:19:52 DEBUG    SHOWQUEUE-REFRESH :: [3f23fd2] Metadata provider KODI 12+ creating episode thumbnail for Mr. Mercedes - S01E07 - 007 - Willow Lake
> 2017-10-11 22:19:52 DEBUG    SHOWQUEUE-REFRESH :: [3f23fd2] **No thumb is available for this episode, not creating a thumb**

More at: https://pastebin.com/mJf2uDtU

medariox:

This is a quick fix for TVDB v2 API currently not supporting thumbnails (or I missed something). The old method used helpers.validate_show(), but I don't know why and it can possibly be removed completely.

### DIFF
--- a/medusa/metadata/generic.py
+++ b/medusa/metadata/generic.py
@@ -15,7 +15,9 @@ from medusa.logger.adapters.style import BraceAdapter
 from medusa.metadata import helpers as metadata_helpers
 
 from requests.exceptions import RequestException
+
 from six import iterkeys
+
 import tmdbsimple as tmdb
 
 try:
@@ -158,6 +160,7 @@ class GenericMetadata(object):
     def get_episode_thumb_path(ep_obj):
         """
         Returns the path where the episode thumbnail should be stored.
+
         ep_obj: a Episode instance for which to create the thumbnail
         """
         if os.path.isfile(ep_obj.location):
@@ -182,7 +185,6 @@ class GenericMetadata(object):
         season: a season number to be used for the path. Note that season 0
                 means specials.
         """
-
         # Our specials thumbnail is, well, special
         if season == 0:
             season_poster_filename = u'season-specials'
@@ -200,7 +202,6 @@ class GenericMetadata(object):
         season: a season number to be used for the path. Note that season 0
                 means specials.
         """
-
         # Our specials thumbnail is, well, special
         if season == 0:
             season_banner_filename = u'season-specials'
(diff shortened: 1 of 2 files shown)


komerdoor:
I just tested it using this feature branch. If a episode thumbnail is available on tvdb and it does not exist on disk it downloads it. After removing the Kodi texture.db I let kodi regenerate all thumbnails from the -thumb.jpg files, but it does not regenerate the ones dowloaded using the code from this branch. The older -thumb.jpg files work.

New one that does not work in Kodi:
> Mr.Mercedes.S01E09.720p.HDTV-Medusa-thumb.jpg: JPEG image data, Exif standard: [TIFF image data, big-endian, direntries=12, height=405, bps=0, PhotometricIntepretation=RGB, orientation=upper-left, width=720], progressive, precision 8, 400x225, frames 3

Old one (created by SickRage) that does work in Kodi:
> Mr.Mercedes.S01E01.720p.HDTV-SickRage-thumb.jpg: JPEG image data, JFIF standard 1.01, resolution (DPI), density 72x72, segment length 16, baseline, precision 8, 400x225, frames 3

p0psicles:
It's a while ago so I forgot. But there was a reason why it wasn't implemented as before. Have a strong feeling it had something to do with api v2. And not wanting to eliminate any ties to old api in case that got decommissioned.

medariox:
@komerdoor 
No idea what could be causing that.

@p0psicles 
The code I replaced was something like 4 years old. The alternative seems to be not having thumbnails at all. We can also wait until TVDB adds support in the v2 API, but why not have something that at least works until then?

komerdoor:
It works now in Kodi for an episode that just got added. It now had the same thumbnail as on tvdb. The other ones that did not work were already added before the fix. It seems that even after removing texture.db and restarting, Kodi does not re-read the -thumb files.

medusa/metadata/generic.py
```diff
@@ -368,30 +369,22 @@ def create_season_all_banner(self, show_obj):
             return self.save_season_all_banner(show_obj)
         return False
 
-    def _get_episode_thumb_url(self, ep_obj):
+    def _get_episode_thumb_url(self, series, episode):
```
p0psicles: Do we really need to pass the series? I believe the series is already available by doing ep.series?

medariox: Possibly, I will have another look.

medariox: We need information from the indexer series object, we can't get that from the episode object currently. I called it ```indexer_series``` now to make it more obvious.

medariox:
Turns out the API v2 actually provides the thumbnail. It's just a bit inefficient as we have to query the single episodes, but that's how the API is designed. Should we use the API nonetheless?

labrys:
I would say only if it was done asynchronously.  Otherwise the overhead wouldn't be worth it.

p0psicles:
I think I didn't implemented it back then, because I made the assumption that it was going to be used on every indexing or adding of a new show. But of course that's not true. Adding a new show doesn't necessarily have any episodes on disc, and therefor does not call the episode thumb or nfo api.

PR #3191, opened 2017-10-28
Fix imdb info

duramato:

### DIFF
--- a/medusa/tv/series.py
+++ b/medusa/tv/series.py
@@ -1575,21 +1575,19 @@ class Series(TV):
         # Set retrieved IMDb ID as imdb_id for externals
         self.externals['imdb_id'] = self.imdb_id
 
-        self.imdb_info = {
+        self.imdb_info.update({
             'imdb_id': imdb_obj.imdb_id,
             'title': imdb_obj.title,
             'year': imdb_obj.year,
             'akas': '',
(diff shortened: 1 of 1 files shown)


p0psicles:
Imdb_info is already populated at that point?

duramato:
`imdb_info` always exists as an empty dictionary, and as the countries are added before this changed code https://github.com/pymedusa/Medusa/blob/fix/imdb-info/medusa/tv/series.py#L1566-L1567 before this change ìmdb_info` would get overwritten by this

medariox:
The empty string probably needs to stay for the DB (didn't check).

medariox:
Did you make sure that the DB is going to be fine without countries and country_codes set? None isn't the same as an empty string, let's avoid to create inconstancies in the DB.

p0psicles:
I'll make sure.

p0psicles:
I checked. But, those keys are not available if there are no country codes right?
If there not available, there not written to the db. In that case for sqlite there is not much of a difference between not writing to the field, or writing an empty string.

I could use this as a workaround to be sure:
```
self.imdb_info['countries'] = self.imdb_info.get('countries', '')
self.imdb_info['country_codes'] = self.imdb_info.get('country_codes', '')
```

PR #3217, opened 2017-11-02
Fixed custom newznab providers not saving the categories after restart.

### LINKED ISSUE #3215: Storage of Custom Newznab provider media categories
[Using latest docker image of master from linuxserver/medusa]

Editing Custom Newznab providers pulls their categories properly and "update categories" works as expected in the UI to store the correct categories for each provider.

This persists until restart, at which point these are lost. While running and after shut down, I have grep'ped for a particular sample category and cannot find that string anywhere in any config, so it seems it's not being written properly into a config for the custom Newznab provider itself.

p0psicles:

In the end it was a stupid typo.
Do realize that with this fix the cats are saved as a string. While we have done allot to store lists really as lists in the config.

If we want to store this as a list, it'l take more work.

Will fix #3215

### DIFF
--- a/medusa/providers/nzb/newznab.py
+++ b/medusa/providers/nzb/newznab.py
@@ -45,7 +45,7 @@ class NewznabProvider(NZBProvider):
     Tested with: newznab, nzedb, spotweb, torznab
     """
 
-    def __init__(self, name, url='', api_key='0', cat_ids=('5030', '5040'), default=False, search_mode='eponly',
+    def __init__(self, name, url='', api_key='0', cat_ids=None, default=False, search_mode='eponly',
                  search_fallback=False, enable_daily=True, enable_backlog=False, enable_manualsearch=False):
         """Initialize the class."""
         super(NewznabProvider, self).__init__(name)
@@ -65,7 +65,7 @@ class NewznabProvider(NZBProvider):
         self.needs_auth = self.api_key != '0'
         self.public = not self.needs_auth
 
-        self.cat_ids = list(cat_ids) if cat_ids else ['5030', '5040']
+        self.cat_ids = cat_ids or ['5030', '5040']
 
         self.torznab = False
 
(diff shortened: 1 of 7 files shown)


medariox:
Seeing the effort that was put into converting everything to lists, I think it would be a pity to not have this setting stored as list as well. I don't see this bug as too urgent. If you want to make it a list, take your time and go for it IMO.

p0psicles:
Still doing some tests. I'll ask the others when everything seems okay on my installations.
Config changes are always tricky.

medusa/version_checker.py
```diff
@@ -31,9 +31,10 @@
 
 from medusa.logger.adapters.style import BraceAdapter
 
-from . import app, db, helpers, notifiers, ui
-from .github_client import get_github_repo
-from .session.core import MedusaSession
+from medusa import app, db, notifiers, ui
```
p0psicles: These changes got in because of some experiments. But they also don't hurt.

medusa/providers/nzb/newznab.py
```diff
@@ -44,7 +45,7 @@ class NewznabProvider(NZBProvider):
     Tested with: newznab, nzedb, spotweb, torznab
     """
 
-    def __init__(self, name, url='', api_key='0', cat_ids='5030,5040', default=False, search_mode='eponly',
+    def __init__(self, name, url='', api_key='0', cat_ids=('5030', '5040'), default=False, search_mode='eponly',
```
medariox: I would make `cat_ids` optional at this point since they are being assigned in the constructor anyway.

medariox: I would make cat_ids optional at this point since they are being assigned in the constructor anyway.

medariox: I would make `cat_ids` optional (without default value) at this point since they are being assigned in the constructor anyway.

p0psicles: True

medariox:
Saving the categories causes all custom newznab providers to be removed for me...
To reproduce, just save some categories for a custom newznab provider and save.

PR #3239, opened 2017-11-07
Use the Medusa version in the user agent

duramato:


p0psicles:
Let's move out the version from the USER_AGENT string. And put it into it's dedicated constant.
It can stay in common for now. But then it can be used to import from another module.

Issue #3248, opened 2017-11-09
Getting subtitles for Archived episodes --> Too many requests


Olivier6767:
Branch: master
Commit: b16e0e5eb4ded727eebd2ab2cd9683ab597662f5
Version: v0.1.17

I get constant errors from several subtitle providers: Too many requests.
After turning on debug, I found out that Medusa is regularly trying to download subtitles for shows and episodes marked as Archived. Since I have many of them, it generates too many requests to the subtitle providers and eventually it exceeds their request quota.

**More information:**
I have many shows where I had downloaded the first seasons in local language (french). I did this by adding <french> in the required words field in the Advanced season settings.
Once I had watched one season, I marked all the episodes for this season as Archived.
For the latest seasons, since the French version is usually not available upon release, I remove the required keyword <french> and I enable Search subtitles in the Main season settings.

What I found out by reviewing the debug logs, is that Medusa still **wants to download all missing subtitles for old seasons marked as Archived**. This represents a huge list of subtitles.
This exceeds the subtitle provider's quota and generates daily errors for 'Too Many Requests'.
Therefore no subtitles are downloaded for newly fetched episodes.

Is this a normal behavior?
Is there a setting that would avoid Medusa trying to download old subtitles (for episodes marked as Archived)?

p0psicles:
It's definitely not expected behavior. Will put up a fix for it shortly.

p0psicles:
@fernandog in our sql query for episodes that should search for subtitles, i found this: `"AND (e.status LIKE '%4' OR e.status LIKE '%6') "`.
It looks like where looking explicitly for episodes that have a status of `archived`. You have any idea why?

fernandog:
ARCHIVED means It wont snatch any other quality but If show has subtitle enabled and EP is missing subs It will download it.  So ARCHIVED is not related with subs

p0psicles:
okay, but is the right? I mean, when you archive a show, your mostly done with it right?

fernandog:
Also It dont pick ALL missing Episodes in a each run. It limits.

Addic7ed active blocks subliminal. It can be after 1 request or 10. Please try opensubtiles

fernandog:
You are done with snatches. Not subs. Changing this wont make Addic7ed works.

p0psicles:
[...]

PR #3266, opened 2017-11-13
Don't search subtitles for archived episodes

fernandog:
Archived status is generally used to mark a episode as downloaded and watched, so it doesn't need subtitles

@duramato all good?

image

Related to: https://github.com/pymedusa/Medusa/issues/3248


medariox:
There is [some more](https://github.com/pymedusa/Medusa/search?utf8=%E2%9C%93&q=status+LIKE+%256&type=).

PR #3275, opened 2017-11-16
add real patch to series

OmgImAlexis:
Since I couldn't find anything in our js using this API path I think we should be fine merging this.


medusa/server/api/v2/series.py
```diff
@@ -96,18 +102,28 @@ def patch(self, series_slug, path_param=None):
...
-                elif value is False:
-                    series.unpause()
-                else:
-                    return self._bad_request('Invalid request body: pause')
-                done[key] = value
-
-        return self._ok(done)
+        accepted = {}
+        ignored = {}
+        patches = {
```
OmgImAlexis: @ratoaq2 I know config uses self.patches but we need access to the series object which isn't available in the class scope.

p0psicles: Ok if i'm understanding this correct, then what your doing here is getting the series object, and changing an attribute if it needs to be patched.

But the config uses this command, to store the new `in-mem` values into the db: `app.instance.save_config()`
You don't have something similar for the series object? Or i'm missing something. So these changes will not be persistent.

OmgImAlexis: The patches is a list of patchable fields, this is literally copied and pasted from the other apiv2 end points for patching.

OmgImAlexis: The only difference is config uses `self.patches` where as this uses a local variable called patches.

p0psicles:
I need to reverse engineer the apiv2 to figure out how to use the patch method.
Don't we have this documented somewhere with some real examples?

p0psicles:
Some notes.
For passing booleans, the api only accepts passing it as a `"scene": true` and not `"scene": "true"`.
Do we want to limit it to that?

p0psicles:
@medariox you know why it's still hanging on code beat?

medariox:
@p0psicles Try to rebase.

OmgImAlexis:
@p0psicles 
> For passing booleans, the api only accepts passing it as a "scene": true and not "scene": "true".
Do we want to limit it to that?

We do this as the API accepts JSON so the booleans should be booleans not strings. I'd rather keep it that way if possible.

PR #3280, opened 2017-11-18
Use latest Subliminal develop branch

fernandog:


p0psicles:
The code you removed is now included in subliminal?

fernandog:
Yes

fernandog:
with this version of subliminal with have:
- Alternative titles when search for subs (it uses our scene exceptions)
- No more traceback/exceptions when service is unavailable. It will log as unavailable instead
- Tons of fixes to LTV
- Fixed unicode search for podnapisi and use new url
- Faster search for podnapisi

Issue #3252, opened 2017-11-09
Feature Request: Delete file when "new file is same size"


phraymond:
I'd like to have an option so that if/when this happens, I can tell Medusa to delete the new file rather than ignore it:

Old size: 2.09 GB
New size: 2.09 GB
New file is the same size.

OmgImAlexis:
Why are you downloading episodes again if you already have them?

Seems kinda pointless to allow the user to download a whole file just to delete it.

labrys:
@OmgImAlexis some possible use-cases
* manually queue multiple episodes to get the first available
* search snatches a file multiple times
* file is downloaded as a backlog but the parser determines its a different quality than its naming would suggest

phraymond:
Exactly, @labrys, thank you.

labrys:
@phraymond Feature is in progress.  It is now configurable, but need to still add the settings to the config and the UI.

phraymond:
I saw the note that it had been added to code. Is there a place where I can enable it? Thanks!

p0psicles:
Not finished. It's still in progress. As soon as #3285 has been merged, you can test it in develop.

labrys:
Sorry between holidays and illness I havent had a chance to finish polishing it off.

phraymond:
No worries. Health and family take priority.

rafaello104:
Any word on where I can enable this? Or if it's available yet? 
I am having to do this manually and it could really help me out .

Thanks in advance 🦄

medariox:
@rafaello104 
The PR was never completed, so the feature is yet to be added. You can see an overview of all requested/completed feature requests here: https://github.com/pymedusa/Medusa/issues/282

flick1999:
Lots of updates lately. Perhaps this one can make the list? :-)

p0psicles:
a number of devs left the project. So we have to do with the resources that are available. There is still a lot of other work to be finished first.

PR #3285, opened 2017-11-21
Configurable post-processor quality replacement

labrys:
This is the first step to making the post processor quality replacement configurable.  Right now its hardcoded as "strict".  This means it does not replace qualities except for replacing an allowed with a preferred.  Current default behaviour is to replace with higher qualities if both are allowed or both are preferred.  When finished it will address #3252.

TODO:
- [X] Pass configuration to method instead of hardcoded
- [X] Add preconfigured profiles such as "Strict", "Permissive", "Upgrade", "Downgrade" etc
- [x] Add setting to config file
- [ ] Add setting to UI
- [ ] Add setting to manual post processing to override configured setting


p0psicles:
@labrys as your already working on the post-processor.
I think it would be really valuable to create a post-processing thread. And track each postprocessing process using a UUID. Then external applications can track the pp status, using the id. And our UI will not lock up when a postprocess action is started.

labrys:
Open it in a new PR. I don't want notifications for this.

Issue #3295, opened 2017-11-23
Post-Processed seeding torrents location don't work with remote torrent client


darkbroodzed:
**Branch/Commit:** Any
**OS:** Any
**What you did:**  set Post-Processed seeding torrents location 
**What happened:** error in log
**What you expected:** moving torrent to new location after post process
**Logs:**
```
PASTE LOGS HERE
```
2017-11-23 15:30:19 WARNING  POSTPROCESSOR :: [b16e0e5] Not possible to move torrent after post-processing because seed location is invalid

its fail because of wrong logic in file process_tv.py

def move_torrent(info_hash, release_names):
        """Move torrent to a given seeding folder after PP."""
        if not os.path.isdir(app.TORRENT_SEED_LOCATION):
            logger.log('Not possible to move torrent after post-processing because seed location is invalid',
                       logger.WARNING)
return False

always "false" for any non local path, but torrent client can be on different device

lilp69:
Check if your user have right to write in both location.

darkbroodzed:
yep i have it
pi@raspberrypi:/mnt/dlink_a2/Transmission $ ls -la
total 40
drwxrwxrwx  8  501  501  4096 Nov 22 17:38 .
drwxrwxrwx 16 root root  4096 Nov 23 14:58 ..
drwxrwxrwx 72  501  501 12288 Nov 23 14:23 complete
drwxrwxrwx  2 1002  501  4096 Nov 19 12:19 couchpotato
drwxrwxrwx  2  501  501  4096 Nov 23 15:22 incomplete
drwxrwxrwx  5 pi   pi    4096 Nov 23 15:22 medusa
drwxrwxrwx  4 pi   pi    4096 Nov 23 00:30 seed
drwxrwxrwx  2  501  501  4096 Oct  2  2014 torrent

fernandog:
@darkbroodzed not sure if transmission allows move to non local folder. Did you try using the Transmission UI to do that? I think it will complain because path is not mapped

darkbroodzed:
guys i showed you code where logic fail.
In my installation i comment code
if not os.path.isdir(app.TORRENT_SEED_LOCATION):
logger.log('Not possible to move torrent after post-processing because seed location is invalid',
logger.WARNING)
return False
restart and all start work as expected.

Ok. Mb my English isn't perfect so i try explain it again:
I have sever with Medusa (Machine number one)
[...]

PR #3298, opened 2017-11-23
Fix moving torrent feature not working with remote client

fernandog:
I can't check `isdir` for a remote location
If folder is not valid Transmission will return an error

Related to https://github.com/pymedusa/Medusa/issues/3295


p0psicles:
So what was the goal of this check?

fernandog:
@p0psicles make sure user selected a valid folder but as you can read the conversation in the issue we can skip it

PR #3322, opened 2017-11-26
Improve Newpct search, add new tests

medariox:


medusa/providers/torrent/html/newpct.py
```diff
@@ -179,6 +179,21 @@ def parse(self, data, mode):
...
+            search_matches = NewpctProvider.anime_search_regex.match(search_string)
+            name = search_matches.group(1)
+            chapter = '1{0}'.format(search_matches.group(2))
+        else:
+            search_matches = NewpctProvider.search_regex.match(search_string)
+            name = search_matches.group(1)
+            chapter = '{0}{1}'.format(search_matches.group(2), search_matches.group(3))
+
+        norm_name = re.sub(r'\W', '-', name)
+        title = norm_name.replace('--', '-').strip('-').lower()
```
snamds: This norm_name replace and strip will make the parse fail. Both should be removed.

I think you put them because of the code I showed you bu in that code there where a list of address to try. Both versions, with and without replace, where tried.
I've been trying to find a show that needs the replace but i can't find it. It's better for you to remove both the replace and the strip.
Sorry for the misleading.

See the examples here:
"1992 (Mille Novecento Novantadue)"
"http://www.newpct.com/series-hd/1992--mille-novecento-novantadue-/2075"
Notice the "-" after novantadue

"Blade - La Serie"
"http://www.newpct.com/series/blade---la-serie/875"

"Erase Una Vez... El Espacio"
"http://www.newpct.com/series/erase-una-vez----el-espacio/2533"

medariox: These are not the urls we query in my code.
What we need is that e.g. MacGyver (2016) becomes macgyver-2016.
`http://www.newpct.com/descargar-serie/macgyver-2016/capitulo-22/hdtv/`
The replace and split are both needed there.

snamds: Ok , the big part of series don't need the replace/strip but macGyver does.
So _parse_title has to return a list. If the title is different with and without replace/strips, both titles should be returned and added to the "search_urls"of the search function

medariox: From what I could see it's quite the contrary, most series require the replace and strip. Do you have more examples? The only one I could find is this `http://www.newpct.com/descargar-serie/1992--mille-novecento-novantadue-/capitulo-110/hdtv-720p-ac3-5-1/` and I'm not going to double the search attempts for one show.

snamds: check this link for example
http://www.newpct.com/series/letter/b
Every single show with parenthesis has double "--" and ends with "-"

snamds: Ah ok I see what you mean.
The address for the show is this:
http://www.newpct.com/series/bron--el-puente-/1721
but the address for the episodes is this:
http://www.newpct.com/descargar-serie/bron-el-puente/capitulo-39/hdtv/

Ok. It's ok the way you are doing it. Sorry

medariox: Exactly! Yeah, the site is terrible 😞

snamds: One last thing (you are going to dislike me)
I think the replace should be done with "re.sub(r'-+', '-'"
The "Blade - La Serie" or "Erase Una Vez... El Espacio" have more than two "-" to replace

medariox: What matters is still the "download page" (don't get confused with the "show page"). Don't see any issues with that one:
`http://www.newpct.com/descargar-serie/erase-una-vez-el-espacio/capitulo-11/dvdrip/` (single -, no replace needed)
"Blade" doesn't seem to have any download page at all.

No hate at all! I appreciate all feedback I can get to get this provider working as good as possible.

snamds: Yeah I know is empty, it's just an example, but if someone use the scene exception "Blade - La Serie" because this is the name he sees in the newpct site, the algorithm will fail. _parse_title will "blade - la serie" --> "blade---la-serie" --> "blade--la-serie" because of the replace. With "re.sub(r'-+', '-'" would do "blade - la serie" --> "blade---la-serie" --> "blade-la-serie"

medariox: Do you have a real example where it matters? Otherwise, I'd like to avoid one more regex.

egibar:
I see a problem with newpct or newpct1.com, for example if you try to search for vikings in spanish, the is no torrents, you have to search as vikingos. I think that the solution can be easy.

medariox:
You mean easier than adding "vikingos" as scene exception for that show? We are always open to suggestions and code improvements.

egibar:
I thought that if you have the metadata info that already names the show as Vikingos, only with changing the string that passes the indexer and put as vikingos it could be done instead of adding as scene exception, but I am not sure.

egibar:
Also I think is better newpct1.com than newpct.com

PR #3339, opened 2017-11-29
Integrate Dockerfile for easier builds

bobbysteel:
To enable easier Docker builds, integrate the Dockerfile and helper scripts.

The current linuxserver/medusa image is great, but builds aren't triggered as it only polls their linuxserver/docker-medusa repo which is rarely updated. This will enable auto rebuilding upon commits to master.


Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# install app
+COPY . /app/medusa/
+
+# copy local files
+COPY docker/root/ /
+
+# ports and volumes
+EXPOSE 8081
+VOLUME /config /downloads /tv
```
OmgImAlexis: Would you mind adding a volume mount for anime since some people do like to keep them separate.

OmgImAlexis:
We may want to move `docker` to `.docker` so we keep the root dir clean.

bobbysteel:
@OmgImAlexis done!

bobbysteel:
I'm open to that - anyone else have a view? I chose the docker/ folder
structure to mirror huginn which seems to use it successfully
https://github.com/huginn/huginn. But I'm not bothered either way really.

bobbysteel:
Done now @OmgImAlexis

medariox:
Can you rebase, please?

labrys:
@bobbysteel can you change the `MAINTAINER` to a URL or email address?  Additionally, the `MAINTAINER` instruction is deprecated in favor of `LABEL maintainer`, see the [docs](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated).

bobbysteel:
Sure - I can assign to myself or link a url to issues here on github - what
would you suggest?

PR #3340, opened 2017-11-29
Fix Yggtorrent

### LINKED ISSUE #3334: Search system of Yggtorrent has changed
Hi, like i said, the search system has changed, and medusa can't search in Yggtorrent anymore.

```
2017-11-28 18:45:00 ERROR    SEARCHQUEUE-FORCED-257655 :: [Yggtorrent] :: [b96b3ee] Failed parsing provider.
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/torrent/html/yggtorrent.py", line 141, in parse
    pubdate = self.parse_pubdate(pubdate_raw, human_time=True)
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/generic_provider.py", line 557, in parse_pubdate
    return datetime.now(tz.tzlocal()) - timedelta(seconds=seconds)
TypeError: unsupported type for timedelta seconds component: NoneType
```
regards

labrys:
Fixes #3334


labrys:
This is more of a hot-fix, need to check the actual provider results to see if structure has actually changed.

adaur:
https://github.com/pymedusa/Medusa/blob/master/medusa/providers/torrent/html/yggtorrent.py#L48

There is no
`'heure': 'hour',`

Might be unrelated, but who knows.

medariox:
I'd like to avoid this hot-fix and go to the root of this issue. There is surely a reason why the parsing is failing. @adaur That translation missing is likely to be the issue. Any other translations missing?

adaur:
It seems good. However, a failing pubdate should probably raise a warning but not make the whole search fail

p0psicles:
It shouldn't make the search fail anymore.

PR #3394, opened 2017-12-03
Added reflinking as option for the post processor

Kriskras99:

This adds reflink support for CoW filesystems, currently only BTRFS, upstream is working on adding support for Apple's APFS and I'm looking into Windows' ReFS.

### DIFF
--- a/medusa/helpers/__init__.py
+++ b/medusa/helpers/__init__.py
@@ -14,7 +14,6 @@ import os
 import platform
 import random
 import re
-import reflink
 import shutil
 import socket
 import ssl
@@ -60,6 +59,10 @@ from six.moves import http_client
 log = BraceAdapter(logging.getLogger(__name__))
 log.logger.addHandler(logging.NullHandler())
 
+try:
+    import reflink
+except ImportError:
+    reflink = None;
 
 try:
     from urllib.parse import splittype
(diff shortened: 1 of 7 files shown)


medusa/helpers/__init__.py
```diff
@@ -420,6 +421,66 @@ def move_and_symlink_file(src_file, dest_file):
...
+def reflink_file(src_file, dest_file):
+    """Copy a file from source to destination with a reference link.
+
+    :param src_file: Source file
+    :type src_file: str
+    :param dest_file: Destination file
+    :type dest_file: str
+    """
+    try:
+        reflink.reflink(str(src_file), str(dest_file))
```
medariox: Why the str() here?

requirements.txt
```diff
@@ -20,6 +20,7 @@ profilehooks==1.9.0
 PyGithub==1.35
 PyNMA==1.0
 python-dateutil==2.6.1
+reflink=0.1.4
```
medariox: Double `==` instead of single `=`. You also need to include the reflink package in the `ext` folder, as we currently still bundle external packages.

fernandog:
I would add a note quick explaining this setting

Kriskras99:
@fernandog Were would you suggest adding this note?
@medariox What would be the best way to bundle `cffi`?

medariox:
@Kriskras99 
I'm afraid we can't add packages that have C dependencies until we have a proper setup script. 😞

fernandog:
@Kriskras99 I would add here: https://github.com/pymedusa/Medusa/pull/3394/files#diff-78821c2496564ec7b9c01846a747a2caL79

@medariox what do you think?

medariox:
@fernandog 
Sure, but the main issue is that we currently can't add the `cffi` dependency.

labrys:
@medariox it could be added as an optional feature that you have to manually install the package and it uses it only if available, similar to mediainfo.

PR #3417, opened 2017-12-06
Transmission. Error: argument of type 'NoneType' is not iterable

fernandog:
```
File Medusa\medusa\session\hooks.py", line 31, in log_url
    if 'multipart/form-data' not in request.headers.get('content-type') \
```

### DIFF
--- a/medusa/session/hooks.py
+++ b/medusa/session/hooks.py
@@ -27,9 +27,14 @@ def log_url(response, **kwargs):
     log.debug('User-Agent: {}'.format(request.headers['User-Agent']))
 
     if request.method.upper() == 'POST':
-        body = request.body \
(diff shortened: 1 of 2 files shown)


medusa/session/hooks.py
```diff
@@ -28,8 +28,8 @@ def log_url(response, **kwargs):
 
     if request.method.upper() == 'POST':
         body = request.body \
-            if 'multipart/form-data' not in request.headers.get('content-type') \
-            else request.body[1:99].replace('\n', ' ') + '...'
+            if 'multipart/form-data' not in request.headers.get('content-type', '') \
+            else request.body[1:99].replace('\n', ' ') + '...' if request.body else None
```
p0psicles: better use `''` instead of None, as where decoding a few lines down. And None.decode('utf-8') will cause an exception.

PR #3429, opened 2017-12-07
Add missing mappings for YggTorrent, fixes #3427

### LINKED ISSUE #3427: [APP SUBMITTED]: No translation mapping available for value: minutes
### INFO
        **Python Version**: `2.7.9 (default, Jun 29 2016, 13:08:31) [GCC 4.9.2]
        **Operating System**: `Linux-3.16.0-4-amd64-x86_64-with-debian-8.9`
        **Locale**: `UTF-8`
        **Database**: `44.8`
        **Commit**: pymedusa/Medusa@1a8dc66c7ec9231be0d3444710e476437c9ae1c8
        **Link to Log**: https://gist.github.com/ba8bb91c8f92d3cd18d74ce092a0b5f6
        ### ERROR
        <pre>
2017-12-07 07:55:51 ERROR    SEARCHQUEUE-FORCED-321285 :: [Yggtorrent] :: [1a8dc66] No translation mapping available for value: minutes
</pre>
        ---

adaur:

I thought they were not needed since they are not to be translated but apparently they are.

Once again, I think this should be a soft warning, not a hard error preventing the snatch...

Also could we do a hotfix for this one? Otherwise I guess we'll see at least 10 tickets about this until next release.

### DIFF
--- a/medusa/providers/torrent/html/yggtorrent.py
+++ b/medusa/providers/torrent/html/yggtorrent.py
@@ -45,6 +45,8 @@ class YggtorrentProvider(TorrentProvider):
 
         # Miscellaneous Options
         self.translation = {
+            'seconde': 'second',
+            'secondes': 'seconds',
             'minute': 'minute',
             'minutes': 'minutes',
             'heure': 'hour',
@@ -53,6 +55,7 @@ class YggtorrentProvider(TorrentProvider):
             'jours': 'days',
             'mois': 'month',
             'an': 'year',
+            'ans': 'years',
             'années': 'years'
         }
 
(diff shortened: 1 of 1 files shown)


p0psicles:
Yeah the question now is.. Should we hotfix this?

adaur:
Maybe wait a couple of days to see if there are any other blatant errors that (also?) need to be hotfixed

medariox:
@adaur 
The logging of error or exception isn't what is preventing the snatch, it's rather the [continue](https://github.com/pymedusa/Medusa/blob/72089e6c07f8ed0e0aadd1082edf50647d3cd6e7/medusa/providers/torrent/html/yggtorrent.py#L150) which can safely be removed IMO - after all pubdate is just a nice-to-have info and shouldn't prevent download.

fernandog:
yes, remove the continue. other providers pubdate default is None. If available then set

p0psicles:
I want to add that the prio to this is not that high. As the chance you will miss a release is very low. At least for releases that are skipped with a pubdate of 'minutes', it will most certainly come back in next daily search.

adaur:
Thanks for your input. It's done.

@p0psicles good point.

fernandog:
is there anything else to do here?

fernandog:
Approved

PR #3430, opened 2017-12-07
Add network information while adding new show

fernandog:
image

### DIFF
--- a/medusa/server/web/home/add_shows.py
+++ b/medusa/server/web/home/add_shows.py
@@ -101,7 +101,7 @@ class HomeAddShows(Home):
         for i, shows in iteritems(results):
             final_results.extend({(indexerApi(i).name, i, indexerApi(i).config['show_url'], int(show['id']),
                                    show['seriesname'].encode('utf-8'), show['firstaired'] or 'N/A',
-                                   show['network'] or 'N/A') for show in shows})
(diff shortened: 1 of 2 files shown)


p0psicles:
What does it look like using tmdb or tvmaze?

fernandog:
Doesn't matter which indexer

fernandog:
we use the common field called `network`

image

fernandog:
btw should we order by date? Seems there is no order in the result list

medusa/server/web/home/add_shows.py
```diff
@@ -100,7 +100,8 @@ def searchIndexersForShowName(search_term, lang=None, indexer=None):
 
         for i, shows in iteritems(results):
             final_results.extend({(indexerApi(i).name, i, indexerApi(i).config['show_url'], int(show['id']),
-                                   show['seriesname'].encode('utf-8'), show['firstaired']) for show in shows})
+                                   show['seriesname'].encode('utf-8'), show['firstaired'] or 'N/A',
+                                   show['network'] or 'N/A') for show in shows})
```
p0psicles: Are there any networks which should be encoded?

fernandog: I don't know. Same applies for show names right? No issues about that

fernandog:
@p0psicles anything else?

PR #3450, opened 2017-12-11
Example of how to use marsmallow on the Series class.

p0psicles:


p0psicles:
@medariox pull in the last commit and try to start medusa.
You'll hit an exception. That's caused by our 'complex' routes.

OmgImAlexis:
@p0psicles what're we doing with this PR?

p0psicles:
Yeah good question.
It would be a really great addition if we could get around the bug with the complex routes.

For now I think we should keep this stored here.
We also still need to implement the rest of the routes. So until then, this won't have a really high prio.

PR #3478, opened 2017-12-16
Feature/generic

labrys:
Clean up generic provider.

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -185,7 +185,14 @@ class GenericProvider(object):
                         search_result.date = datetime.today()
                         search_result.show = show_obj
 
-        return results
+        return self.remove_duplicate_urls(results)
+
+    def remove_duplicate_urls(self, items):
+        items_list_without_dups = []
+        for item in items:
+            if item['link'] not in [_['link'] for _ in items_list_without_dups]:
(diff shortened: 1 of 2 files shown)


medusa/providers/generic_provider.py
```diff
@@ -228,31 +228,40 @@ def find_search_results(self, show, episodes, search_mode, forced_search=False,
...
-            for item in items_list:
-                items[self.get_quality(item, anime=show.is_anime)].append(item)
-
-            # temporarily remove the list of items with unknown quality
-            unknown_items = items.pop(Quality.UNKNOWN, [])
-
-            # make a generator to sort the remaining items by descending quality
-            items_list = (items[quality] for quality in sorted(items, reverse=True))
+        # Remove duplicate items using 'link' as primary key
+        unique_items = {
```
p0psicles: @labrys does this mess with the order, as it being converted to a dict?

labrys: I wasn't overly concerned about order since we are sorting by quality later.  However if you would prefer it to stay in original order that can easily modified to:
```python
unique_items = OrderedDict(
    (item['link'], item)
    for item in items_list
).values()
```

p0psicles: Yes please, then I can use it to fix the issue with daily search that I caused.

Issue #3436, opened 2017-12-08
Custom providers can't have double pipes (||) in the name


ward0:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:**
**OS:**
**What you did:**
**What happened:**
**What you expected:**
**Logs:**
```
Any custom 'Configure Custom
Newznab Providers' and for the variable 'provider name' set || in the name and you will have a error. change column to something else?
```

medariox:
I'm not sure if I understood this correctly. You set as provider name `II`?

ward0:
example provider name: '1337x || Jackett'
a provider name that works: '1337x(space)Jackett'

I use '|| Jackett' in the name to get more information for myself.

medariox:
How is `||` any better than `|` or anything else really? I've tagged it as a bug now, but this is something that we can't really change in the foreseeable future. That's how SR used to handle the providers separation in the config and we still want to support users coming from SR and older versions for now.

ward0:
@medariox just wanted to let you guys know that its a small bug. I just changed the || to something else indeed.

labrys:
I think this, at least for the foreseeable future is a won't fix.  We could maybe put a note though in the UI saying `||` is not valid.

ward0:
@labrys I was more informing you guys... A note in the UI would be handy indeed.

PR #3491, opened 2017-12-18
Don't allow save custom provider if bad name

fernandog:
Fix https://github.com/pymedusa/Medusa/issues/3436

any suggestions?


p0psicles:
Yeah,  I think for 1. the character is singular, as in '|' and not '||'. And 2. it should be handled in the frontend. Or at least that's my preference.

fernandog:
@p0psicles if you try, it will show a JS alert showing user can't save it. Please try

Afaik the issue is the double pipes, and not the single pip

p0psicles:
No havent' tested, but I believe you. But imo the request should never hit the backend, as it can be handled with a field validation in the frontend.

fernandog:
There is already a function that validates the provider, inclusing if a name was set. I'm using it.
If you think all this validations should be done in JS please close it

image

medariox:
Please hold this back until the Jackett config is done.

fernandog:
@medariox isn't the issue here how we handle with old configs as you said in the issue? how is this related to Jackett?

medariox:
It is because I've already changed these functions locally and it is going to be messy otherwise.

fernandog:
@medariox changed branch to `add-torznab-providers`. no conflicts

fernandog:
@medariox may I merge into your branch?

medariox:
@fernandog
If it isn't a priority I'd like to finish the PR first.

fernandog:
@pymedusa/developers rebased

fernandog:
Thanks

PR #3495, opened 2017-12-18
Implement basic cache of 1 day for guessit geet_expected_titles funct…

p0psicles:
…ion results.

### DIFF
--- a/medusa/name_parser/guessit_parser.py
+++ b/medusa/name_parser/guessit_parser.py
@@ -13,6 +13,7 @@ from medusa.name_parser.rules import default_api
 
 
 EXPECTED_TITLES_EXPIRATION_TIME = timedelta(days=1).total_seconds()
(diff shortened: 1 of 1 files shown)


p0psicles:
After a discussion with rato I came to the conclusion, that there is no way of implementing an efficient caching for this.

Previously rato had implemented caching through a dogpile decorator, using the app.showLists to create a 'hash' / key for dogpile. Using this mechanic it would create a new cache every time the showLists changes or scene exception modified. This was the _key_generator() function.

Implementing the _key_generator() function is unavoidable. But to be able to generate the key, you would need to loop through the shows anyway. So when trying to prevent the 3 or 4 loops in get_expected_titles, it will cost me 1 or 2 in the key_generator. Saving me maybe 2 loops in the end.

So allot of complexity for little gain.

PR #3497, opened 2017-12-19
Get speed.cd login URL instead of looping a list of URLs

fernandog:
They keep changing login URL (post)

@duramato

### DIFF
--- a/medusa/providers/torrent/html/speedcd.py
+++ b/medusa/providers/torrent/html/speedcd.py
@@ -5,7 +5,6 @@
 from __future__ import unicode_literals
 
 import logging
-import re
 import traceback
 
 from medusa import tv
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/html/speedcd.py
```diff
@@ -183,20 +181,34 @@ def login(self):
...
         }
 
-        for login_url in self.urls['login']:
-            response = self.session.post(login_url, data=login_params)
-            if not response or not response.text:
-                log.debug('Unable to connect to provider using login URL: {url}',
-                          {'url': login_url})
-                continue
+        if not (self.urls['login_post'] and self.login_url()):
+            log.debug('Unable get login URL')
```
p0psicles: Unable to get login url

medusa/providers/torrent/html/speedcd.py
```diff
@@ -183,20 +181,34 @@ def login(self):
...
-            if re.search('Incorrect username or Password. Please try again.', response.text):
-                log.warning('Invalid username or password. Check your settings')
-                return False
-            return True
+        response = self.session.post(self.urls['login_post'], data=login_params)
+        if not response or not response.text:
+            log.debug('Unable to connect to provider using login URL: {url}',
+                      {'url': self.urls['login_post']})
+            return False
+        if re.search('Incorrect username or Password. Please try again.', response.text):
```
p0psicles: You need a regex for this? Can't just do 'sring' in response.text ?

fernandog: No ideia. I just removed the for loop. need to check history

PR #3500, opened 2017-12-19
Feature/broken union

fernandog:
First it did: 
 `if not self.urls['login_post'] and not self.login_url():` 
then I "improved" to:
`if not (self.urls['login_post'] and not self.login_url()):` 

but second one is not the same as `self.login_url` is not called.
Had to revert. Why is that?
@p0psicles

### DIFF
--- a/medusa/providers/torrent/html/speedcd.py
+++ b/medusa/providers/torrent/html/speedcd.py
@@ -180,7 +180,7 @@ class SpeedCDProvider(TorrentProvider):
             'password': self.password,
         }
 
-        if not (self.urls['login_post'] and self.login_url()):
+        if not self.urls['login_post'] and not self.login_url():
             log.debug('Unable to get login URL')
             return False
 


medariox:
Just remove the second negation.

fernandog:
@medariox code as in the PR works. What I don't get is why:

not (x and b)  is not the same as not a and not b

The self.login_url returns True when it manage to get the login url so I must check if not self.login_url() in the second part

medariox:
With the brackets, you first evaluate what is in the brackets.
With the first code, you only get `True` if both statements are `False`. With the second code, you can get `True` in 3 different situations:
`not (True and False)` is `True`
`not (False and True)` is `True`
`not (False and False)` is `True`
It's not nearly the same thing.

fernandog:
Ok. So all good. My goal is to check when both are False.

labrys:
@fernandog If you want to read more about it, it is called `De Morgan's Laws`.

If you want to check that two statements are false you can either do `not A and not B` or you can do `not (A or B)`.  In python you can also use `not any([A, B, C, ....])` if you want to make sure that every statement is false.

PR #3502, opened 2017-12-20
Replace cached images in forced show update

fernandog:
POC

We could also create a button in displayShow  that calls `fill_cache` that will only replace images.


fernandog:
image

fernandog:
BTW this way I did is blocking UI as its not a schedule task. Need to change it

PR #3503, opened 2017-12-20
Remove all images for a show, using the ImageClass().

p0psicles:

### DIFF
--- a/medusa/server/web/home/handler.py
+++ b/medusa/server/web/home/handler.py
@@ -1793,6 +1793,7 @@ class Home(WebRoot):
 
         # force the update
         try:
+            show_obj.remove_images()
             app.show_queue_scheduler.action.updateShow(show_obj)
         except CantUpdateShowException as e:
             ui.notifications.error('Unable to update this show.', ex(e))
--- a/medusa/image_cache.py
+++ b/medusa/image_cache.py
@@ -18,12 +18,17 @@
 
 from __future__ import unicode_literals
 
+import logging
 import os.path
 import warnings
 
-from . import app, logger
-from .helper.exceptions import ShowDirectoryNotFoundException
-from .metadata.generic import GenericMetadata
+from medusa import app
+from medusa.logger.adapters.style import BraceAdapter
+from medusa.helper.exceptions import ShowDirectoryNotFoundException
+from medusa.metadata.generic import GenericMetadata
+
+log = BraceAdapter(logging.getLogger(__name__))
+log.logger.addHandler(logging.NullHandler)
 
 BANNER = 1
 POSTER = 2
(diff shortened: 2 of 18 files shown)


medusa/server/web/home/handler.py
```diff
@@ -1793,6 +1793,7 @@ def updateShow(self, show=None):
 
         # force the update
         try:
+            show_obj.remove_images()
```
p0psicles: See one line change.

p0psicles: Point is, that this is an object oriented approach. As the logic to delete the files in now in the ImageClass, where it should be. To make it more clean, we could even replace some of logic for cleaning up the images after a show delete, using this method.

fernandog:
Ok but as I said it won't work if user only one want to update images and not rescan ALL files.

My other PR has an action in mass update that only update images and doesn't rescan all files from disk

p0psicles:
What's wrong with rescanning? I thought that was the solution, we agreed on? Just do a force full update right?
And if you want to add the massUpdate part. No prob. just use the ImageClass the remove and repopulate the images.

fernandog:
@p0psicles sure but I also want to add an option to only update images. Rescanning a 10 season show with  tons of episodes may be wrong if user want only new images. Understood my pov?

p0psicles:
You can use the Image Class for that. Fine if you add it in this Pr.

fernandog:
@p0psicles ok but I still don't know how to NOT lock the ui when calling a unscheduled task

labrys:
Accidentally pushed to wrong branch.  Reverted after.

p0psicles:
@fernandog just port your changes for the massEdit to this pr.
And the massEdit is a fast action. It's not needed to run in a thread. Threads are used for heavy duty.

Only difference now, compared to your pr. Is that you have to use the command from this PR to remove the images. And after that, you recreate them.

fernandog:
@p0psicles I cherry-picked and changed code BUT its reading ALL files from disk.

I did for only one show with 100 episodes and I got 100 logs like this:
` Setter sets location to`

as you can see in the code:
```python
    @location.setter
    def location(self, value):
        log.debug('{id}: Setter sets location to {location}',
                  {'id': self.series.indexerid, 'location': value})
        self._location = value
        self.file_size = os.path.getsize(value) if value and self.is_location_valid(value) else 0
```

If calls os.path to get size of all files and check if file exists.
It does all that just to update images.
I don't agree with this actions

p0psicles:
I tested this (also the massupdate) and besides one small error, ALl LGTM.

fernandog:
@p0psicles did you read my message about it reading all files from disk when doing this image update?

p0psicles:
Yeah commented on it on slack. But I can't reproduce it. But I didn't have any episodes in my shows. So maybe that's why?

PR #3504, opened 2017-12-20
Migrate all the db's to have the shows id included, making the db model multi-indexer safe

### LINKED ISSUE #3205: tvdb Scene exceptions used for tmdb shows
The tvdb Scene exceptions are being used for tmdb shows, becouse of id colisions.

https://github.com/pymedusa/medusa.github.io/blob/master/scene_exceptions/scene_exceptions_tvdb.json#L269-L274
My show has an id in TMDb of `73178` which matches `Two Guys a Girl and a Pizza Place` on tvdb, but it should only use exceptions for their respective indexer
image

p0psicles:
In a next iteration we can rename the columns to all match `indexer_id` + `series_id`.

- [x] Add the indexer_id columns to main.db, failed.db and cache.db. Provide the indexer_id fields with a value of `-1`
- [x] Post migration cleanup all the indexer_id fields in the tables. Scanning the show library, and replacing the `-1` with the correct indexer_id if available.

fixes #3205 

Here's the new setup. everything with a `+` is going to be added.
*main.db*
```
table blacklist
show_id
+ indexer_id

table history
showid
+ indexer_id

table imdb_info
indexer_id
+ imdb_info_id (new autonumber PK)
+ indexer

table whitelist
show_id
+ indexer_id
```

*failed.db*
```
table history
showid
+ indexer_id
```

*cache.db*
```
(each) table providers
indexerid
+ indexer

table scene_names
indexer_id
+ indexer
```

### New coding guidelines
* For the indexer id, use the naming `indexer_id` (previously `indexer` was used)
* For a series id, use the naming `series_id` (previously `indexerid`, `indexer_id` was used)
* Whenever referring to a series object, name the object `series_obj`.
* Routes that pass on a series, need to use the following format: `home/route?indexername=[indexer name]&seriesid=series id`. For example: `home/route?indexername=tvdb&seriesid=12345`
* The locate a series from the app.showList, use the new function show.Show.find_by_id(app.showList, indexer_id, series_id). `show.Show.find()` has been deprecated with an deprecation warning.

*js*
* Ajax requests that need to specify a series, will need to pass the indexer_name and series_id.
series-id and indexer-name can be stored in html, using the following inputs:
```
<input type="hidden" id="series-id" value="${show.series_id}" />
<input type="hidden" id="indexer-name" value="${show.indexer_name}" />
<input type="hidden" id="series-slug" value="${show.slug}" />
```
This has already been implemented on displayShow and editShow.

In javascript, a selector can be used to used the get the values.
Like:
```
var indexerName = $('#indexer-name').val();
var seriesId = $('#series-id').val();
```

### Known issues
These will be fixed in this PR.
- [x] Backlog search
- [x] Forced search
- [x] Episode statistics
- [x] History data migration
- [x] Cache images with indexer in path
- [x] Migrate / refresh images.
- [x] Edit show and add a scene exception
- [x] Manage episode status bug
- [x] History mako error, using compact modus
- [x] Show refresh get's xem mappings for every episode
- [x] Fix manage episode status 
- [x] Fix manage subtitles
- [x] Fix bug in the name_cache
- [x] Snach/Download history not visible in snatchSelection
- [x] Fix testRename
- [x] Fix adding shows using recommended lists (tested with trakt and anidb).
- [x] Fix backlog overview page, after I broke it,.. again.
- [x] Fix xem_refresh when having shows with same id.
- [x] Fix apiv1 when retrieving images.
- [x] Fix addshow, temporary show poster (thumb) causes exception.
- [x] Fix Access to SearchResult.show.

These issues will remain. And will probably not be resolved in this PR.
* Complex queries that join / inner join and correlate the tv_shows and tv_episodes table. We need a new unique identifier to join these tables.
For ex:
** medusa/server/web/home/handler.py:177
** medusa/show/coming_episodes.py:87


p0psicles:
TODO:
./views/history.mako
./views/manage_subtitleMissed.mako
./views/manage_subtitleMissedPP.mako
./views/partials/header.mako
./views/partials/seasonEpisode.mako
./views/testRename.mako

p0psicles:
@labrys please check the latest commit: https://github.com/pymedusa/Medusa/pull/3504/commits/7a31bf97e94cd65843c1e953b17967e48f7f7dd8
I've made the changes to add the indexer_name.
Also want to compliment you on the image_cache changes. Was really no work, to adapt it now.

Still need to add the transition of existing images of course.

medusa/black_and_white_list.py
```diff
@@ -33,22 +33,22 @@
...
 
-    def __init__(self, show):
+    def __init__(self, series):
         """Init method.
 
         :param show_id:
         :type show_id: int
         """
-        if not show.indexerid:
+        if not series.indexerid:
```
p0psicles: @medariox @fernandog you might not have known this. But whenever passing a series object, i've used the variable name `series_obj` everywhere. Only when accessing the series as an attribute of for example a SearchResult class, i used `series`.

medusa/name_parser/parser.py
```diff
@@ -36,12 +36,12 @@
 class NameParser(object):
     """Responsible to parse release names."""
 
-    def __init__(self, show=None, try_indexers=False, naming_pattern=False, parse_method=None,
+    def __init__(self, series=None, try_indexers=False, naming_pattern=False, parse_method=None,
```
p0psicles: Same here, should be series_obj. If renamed consistently with the other renames.

medusa/metadata/generic.py
```diff
@@ -390,6 +390,9 @@ def _get_episode_thumb_url(self, indexer_series, episode):
             thumb_url = getattr(indexer_episode, 'filename', None)
             if thumb_url:
                 return thumb_url
+            else:
+                log.debug(u'Tried to get the episodes thumbnail image for episode {0!r}, but the indexer {1!r} '
```
p0psicles: When the thumb_url is missing, you wil get a debug now.

medusa/metadata/generic.py
```diff
@@ -844,8 +847,10 @@ def _get_show_data(self, series_obj):
         series_id = series_obj.series_id
 
         try:
-            if not (series_obj.indexer_api and all([series_obj.indexer_api.config[u'banners_enabled'],
-                                                    series_obj.indexer_api.config[u'actors_enabled']])):
+            # If an episode object is passed, we always want to create a new indexer_api object.
+            if ep_obj or not (series_obj.indexer_api and all(
```
p0psicles: This will make sure we get a new indexer_api object, when an ep_obj is passed.

medusa/metadata/kodi_12plus.py
```diff
@@ -214,7 +214,7 @@ def _ep_data(self, ep_obj):
         """
         eps_to_write = [ep_obj] + ep_obj.related_episodes
 
-        series_obj = self._get_show_data(ep_obj.series)
+        series_obj = self._get_show_data(ep_obj.series, ep_obj)
```
p0psicles: kodi12 is the only one that creates episode thumbnails, but in the future, we should pass the ep_obj if we want to get episode related show data.

medusa/post_processor.py
```diff
@@ -1255,7 +1272,7 @@ def process(self):
         # generate nfo/tbn
         try:
             # Reset the indexer_api for this series, as we don't want to use 'old' cached (to the series's indexer_api attribute) content.
-            ep_obj.series.indexer_api = None
+            self.clear_indexer_api(ep_obj)
```
p0psicles: Reset the cached indexer_api for the series. But only do it once. Meaning we get the series data from the indexer once during a post process action.

p0psicles:
@fernandog you renamed to series in the show_parser. I think you should add a property `show` with deprecated warning. As I think you forgot more renames.

medusa/failed_processor.py
```diff
@@ -62,7 +62,7 @@ def process(self):
 
         if segment:
             self.log(logger.DEBUG, u'Adding this release to failed queue: {release}'.format(release=releaseName))
-            cur_failed_queue_item = FailedQueueItem(parsed_result.series, segment)
+            cur_failed_queue_item = FailedQueueItem(parse_result.series, segment)
```
fernandog: was about to commit that. thanks. pycharm was updating

medariox: No problem.

p0psicles:
@fernandog could you please add the `@property` `show` to the name_parser class. I suspect that there are still references to show. And by adding it as a backup property, we're safe.
Think about some debug log that will only hit under specific circumstances.

fernandog:
I renamed all. Checked twice.

p0psicles:
Todo for myself. Include some exception handling for the image migration. To prevent users of not being able to start up.

p0psicles:
Broke the websockets. Need to find a way to handle the config loaded event twice.

PR #3511, opened 2017-12-21
Classless ImageCache

labrys:
Rewrite image caching and media classes.


p0psicles:
LGTM. And i'll be merging this also in my PR. I'm going to add the indexer_name as an additional layer to the directory structure. So if there is anything wrong, i'll notice soon enough.

Issue #3489, opened 2017-12-18
Subtitle files not being post-processed


johnvick:
Medusa Info: | Branch: masterCommit: 1a8dc66c7ec9231be0d3444710e476437c9ae1c8Version: v0.1.20Database: 44.8
-- | --
Python Version: | 2.7.9 (default, Aug 13 2016, 17:56:53) [GCC 4.9.2]
SSL Version: | OpenSSL 1.0.1t 3 May 2016
OS: | Linux-4.9.61+-armv7l-with-debian-8.10
Locale: | en_GB.UTF-8

I have found that when I download an episode that comes with a separate subtitle file that the subtitle file (.srt) is not moved to the destination directory. If I change its name to match the video file (if need be) the same happens.

There is not much in the log:

```
2017-12-18 18:34:03 INFO     Thread-60 :: [1a8dc66] Processing /mnt/usb_2/completed/Gomorra S03 E09 - 1080p x264 - ITA (ENG SUBS) [BRSHNKV]/Gomorra S03 E09 - x264 .mp4
2017-12-18 18:34:03 INFO     Thread-60 :: [1a8dc66] New file: /mnt/usb_2/completed/Gomorra S03 E09 - 1080p x264 - ITA (ENG SUBS) [BRSHNKV]/Gomorra S03 E09 - x264 .mp4
2017-12-18 18:34:03 INFO     Thread-60 :: [1a8dc66] New size: 452.44 MB
```
I don't remember this happening with SickRage and my setting are the same as they were.
Thanks.

medariox:
Do you have subtitles enabled for that show?

Please always post debug logs.

johnvick:
I didn't have as the show always comes with them so I didn't think Medusa needed to search for them. I'll try again with this setting enabled. Next episode due out soon and I've enabled debug logs and will report back. Thanks for your help.

johnvick:
Just tried this again with the box ticked to search for subtitles. The download contained a separate SRT file which was not processed. I renamed it to  match the video file and tried post-processing again, after deleting the episode from the destination folder and marking it as wanted in Medusa, but same result, subtitle file not copied. Is Medusa meant to look for the subtitle file in the download directory first and then look elsewhere (subtitle sites) if not found? Does the name of the subtitle file have to exactly match the name of the episode file?

This is the relevant log:

2017-12-19 16:27:19 DEBUG    Thread-17 :: [1a8dc66] Processing path: /mnt/usb_2/completed
[...]

PR #3518, opened 2017-12-25
Fix PP renaming subtitle to .srt.srt

fernandog:
Fixes https://github.com/pymedusa/Medusa/issues/3489
@medariox when calling from_ietf_code with `srt`, it returns an alpha3 code with `srt` in Language object but no country information 

IMO if no country, so the Language object is not a valid language, so we should not use it. 
Let me know what do you think

This issue was introduced recently as moving a SRT with no language wasn't a issue since I first started to used subtitles in Medusa/SR

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -367,10 +367,11 @@ class PostProcessor(object):
             new_extension = 'nfo-orig'
 
         elif is_subtitle(filepath):
-            sub_code = filepath.rsplit('.', 2)
+            split_path = filepath.rsplit('.', 2)
             # len != 3 means we have a subtitle without language
-            if len(sub_code) == 3:
-                code = sub_code[1].lower().replace('_', '-')
+            if len(split_path) == 3:
+                sub_code = split_path[1]
+                code = sub_code.lower().replace('_', '-')
                 if from_code(code, unknown='') or from_ietf_code(code, unknown=''):
                     # TODO remove this hardcoded language
                     if code == 'pt-br':
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -367,14 +367,16 @@ class PostProcessor(object):
             new_extension = 'nfo-orig'
 
         elif is_subtitle(filepath):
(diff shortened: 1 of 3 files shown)


medariox:
Isn't this more a babelfish issue then? `srt` isn't a valid alpha3 code, right? Why is it returning a result in the first place?

fernandog:
Well, it's returning a Language object to be consistent but they set None as country

medariox:
How is that consistent? From the code and documentation, it should raise `ValueError('%r is not a valid language' % language)`: https://github.com/Diaoul/babelfish/blob/master/babelfish/language.py#L75-L82

fernandog:
srt is a valid Language btw
https://github.com/Diaoul/babelfish/blob/master/babelfish/data/iso-639-3.tab#L6047

called here: https://github.com/Diaoul/babelfish/blob/master/babelfish/language.py#L26-L30

if you have a better fix close this then

medusa/post_processor.py
```diff
@@ -367,14 +367,16 @@ def rename_associated_file(new_path, new_basename, filepath):
...
-                    code = 'pt-BR'
-                new_extension = code + '.' + extension
-                extension = sub_code + '.' + extension
+            sub_code = filepath.rsplit('.', 2)
+            # len != 3 means we have a subtitle without language
+            if len(sub_code) == 3:
+                code = sub_code[1].lower().replace('_', '-')
+                if from_code(code, unknown='') or from_ietf_code(code, unknown=''):
+                    # TODO remove this hardcoded language
+                    if code == 'pt-br':
```
fernandog: pt-br > 3 ?

medariox: Nope, that's something different. Try to `filepath.rsplit('.', 2)` and you will understand what's going on.

fernandog: we also support `en` , `pt`, and `it` and language codes

medariox: That should be fine, see the tests.

PR #3541, opened 2017-12-27
replace boobysteel docker hub link with pymedusa's one

OmgImAlexis:
@bobbysteel you may want to leave your build up incase others are using it.


bobbysteel:
Cheers will do. Will also add a note pointing to the official shortly

Issue #3462, opened 2017-12-13
TypeError: can't compare offset-naive and offset-aware datetimes


pucelano7:
**OS:** DSM 6 (NAS Synology)
**What happened:** Daily Search is FALSE...when I restart MEDUSA the value of Daily Search change to TRUE but the second time it has to search change to FALSE again.
**Logs:**
```
2017-12-13 20:57:04 ERROR    DAILYSEARCHER :: [1a8dc66] Exception generated: astimezone() cannot be applied to a naive datetime
Traceback (most recent call last):
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/scheduler.py", line 90, in run
    self.action.run(self.force)
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/search/daily.py", line 89, in run
    end_time = show_air_time.astimezone(app_timezone) + timedelta(minutes=try_int(show.runtime, 60))
ValueError: astimezone() cannot be applied to a naive datetime
2017-12-13 20:47:24 ERROR    Thread-16 :: [1a8dc66] Exception generated: can't compare offset-naive and offset-aware datetimes
Traceback (most recent call last):
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/server/web/core/base.py", line 263, in async_call
    result = function(**kwargs)
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/server/web/core/base.py", line 351, in schedule
    results = ComingEpisodes.get_coming_episodes(ComingEpisodes.categories, app.COMING_EPS_SORT, False)
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/show/coming_episodes.py", line 121, in get_coming_episodes
    results.sort(ComingEpisodes.sorts[sort])
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/show/coming_episodes.py", line 40, in <lambda>
    'date': (lambda a, b: cmp(a['localtime'], b['localtime'])),
TypeError: can't compare offset-naive and offset-aware datetimes
```
Parameters of search:
![captura1](https://user-images.githubusercontent.com/24883267/33968338-bcc2359c-e067-11e7-8435-78f3262eba54.PNG)

Here you can see the FALSE value in Daily Search
![captura](https://user-images.githubusercontent.com/24883267/33968383-e34731a4-e067-11e7-975b-923720516e06.PNG)

[...]

PR #3543, opened 2017-12-28
Remove broad exception for news. Remove duplicate log

fernandog:
~~TODO: Need to update local file every release~~

~~About news, we show a link for the news when we failed to get it from github~~


p0psicles:
Do we really want to keep this in the repo? Shouldn't we just cache in db?

fernandog:
@p0psicles used to be in local file. We moved to the repo because the file is always being updated right so user could get file soon instead of waiting for new releases. Righ @duramato ?

p0psicles:
But then using the cache db would be preferred imo. Then you have best of both worlds.

fernandog:
It's already in the cache DB. We get the file and import to network_timezones table

p0psicles:
Ow. Then what's the issue? Why should we keep a file next to it?

fernandog:
@p0psicles the issue is failing to connect to cdn to get the updates. And bigger issue is in new installs when cache.db doesn't exist

fernandog:
@p0psicles Did you read the issue?

fernandog:
It's also a fix for this:
https://github.com/pymedusa/Medusa/issues/3462

labrys:
Posted my opinion on slack.

medusa/version_checker.py
```diff
@@ -306,15 +306,12 @@ def check_for_new_news(self, force=False):
...
-            news = self.session.get(app.NEWS_URL).text
-        except Exception:
-            log.warning(u'check_for_new_news: Could not load news from repo.')
-            news = ''
-
-        if not news:
-            return ''
+        log.debug(u'Checking GitHub for latest news.')
+        response = self.session.get(app.NEWS_URL)
+        news = response.text
```
p0psicles: If response is None, youl get an exception on the response.text

fernandog: Will change thanks

fernandog:
@p0psicles sorry, I did another commit and it dismissed your approve. I done now

Issue #3567, opened 2018-01-02
Medusa fails to start with entware-ng


stefaang:
**Branch/Commit:** master
**OS:** unix with entware-ng on x64 platform
It's a commonly used 3rd party package manager for headless platforms such as routers and NAS boxes. e.g. synology, qnap ...
**What you did:** 
```
opkg install busybox git git-http python python-pyopenssl
git clone git://github.com/pymedusa/Medusa
cd Medusa
python SickBeard.py --help
```
**What happened:**
Failed to start
**What you expected:**
The help should be shown
**Logs:**
```
python SickBeard.py --help
Traceback (most recent call last):
  File "SickBeard.py", line 4, in <module>
    from medusa.__main__ import main
  File "/opt/share/medusa/Medusa/medusa/__main__.py", line 67, in <module>
    from . import (
  File "/opt/share/medusa/Medusa/medusa/providers/__init__.py", line 10, in <module>
    from medusa.providers.nzb import (
  File "/opt/share/medusa/Medusa/medusa/providers/nzb/anizb.py", line 14, in <module>
    from medusa.providers.nzb.nzb_provider import NZBProvider
  File "/opt/share/medusa/Medusa/medusa/providers/nzb/nzb_provider.py", line 8, in <module>
    from medusa.providers.generic_provider import GenericProvider
  File "/opt/share/medusa/Medusa/medusa/providers/generic_provider.py", line 18, in <module>
    from medusa import (
  File "/opt/share/medusa/Medusa/medusa/config.py", line 30, in <module>
    from medusa.version_checker import CheckVersion
  File "/opt/share/medusa/Medusa/medusa/version_checker.py", line 31, in <module>
    from medusa.github_client import get_github_repo
  File "/opt/share/medusa/Medusa/medusa/github_client.py", line 6, in <module>
    import github
  File "/opt/share/medusa/Medusa/ext/github/__init__.py", line 37, in <module>
    from MainClass import Github, GithubIntegration
  File "/opt/share/medusa/Medusa/ext/github/MainClass.py", line 34, in <module>
    import jwt
  File "/opt/share/medusa/Medusa/ext/jwt/__init__.py", line 19, in <module>
    from .api_jwt import (
  File "/opt/share/medusa/Medusa/ext/jwt/api_jwt.py", line 8, in <module>
    from .api_jws import PyJWS
  File "/opt/share/medusa/Medusa/ext/jwt/api_jws.py", line 7, in <module>
    from .algorithms import (
[...]

PR #3573, opened 2018-01-05
Only apply UTF-8 monkey patch when necessary

stefaang:

Fixes https://github.com/pymedusa/Medusa/issues/3567

### DIFF
--- a/medusa/init/filesystem.py
+++ b/medusa/init/filesystem.py
@@ -82,7 +82,7 @@ def _varkwargs(**kwargs):
 
 def make_closure(f, handle_arg=None, handle_output=None):
     """Apply an input handler and output handler to a function.
-    
+
     Used to ensure UTF-8 encoding at input and output.
     """
     return patch_output(patch_input(f, handle_arg), handle_output)
@@ -90,7 +90,7 @@ def make_closure(f, handle_arg=None, handle_output=None):
 
 def patch_input(f, handle_arg=None):
     """Patch all args and kwargs of function f.
-    
+
     If handle_arg is None, just return the original function.
     """
     def patched_input(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


medariox:
Thanks for the contribution!

We still need to monkey patch to support [unicode input for pyOpenSSL](https://github.com/stefaang/Medusa/blob/6e903f348a082d5243f9fb8bb110a05b4f8258a2/medusa/init/filesystem.py#L118), so this solution won't cut it.

Don't you think that there is a problem with the Python implementation in Entware-ng?

stefaang:
The problem was resolved by skipping the `lambda x: x` input wrapper, introduced by Medusa. It was failing on closed file object pointers being forwarded as arguments. 
I couldn't identify any root cause in the openwrt python build scripts yet.

I'd argue that monkey patching python core modules is a bit more dirty ... 
I'll refactor my fix to include the output wrapper (thanks for spotting that by the way)

medariox:
I didn't mean to say that your solution is dirty in any way, quite the contrary actually. It just needs to handle what I pointed above as well, that's all. We are grateful for any help and contribution.

stefaang:
Sorry, I couldn't squash the commits due to issues with line endings.
Anyway, glad to help you guys out!

medariox:
Thanks! The code looks really good. Can you [fix the flake8 issues](https://travis-ci.org/pymedusa/Medusa/builds/326382056#L926-L929) when you get the chance?

medariox:
Argh, need to remove the whitespaces from the blank lines, see [travis issue](https://travis-ci.org/pymedusa/Medusa/builds/326413659#L926-L928).

labrys:
Also as an aside, the monkey patching fails for python 3.

Issue #3580, opened 2018-01-07
[Feature] Stop searching for preferred qualities after initial search period


moryoav:
There's an issue in Medusa with the current implementation of "qualities". 

As it is set up now, a user can have a list of "Allowed" qualities for a show (for example: 720p, 1080p, etc) and "Preferred" (for example: 4k qualities). This is a typical scenario I am sure many users with 4k TVs started using, what with a very limited amount of shows that have 4k qualities.

The rationale is for a show to download in HD quality but if there's a 4k quality then the user would prefer that over the HD. If HD was downloaded first, the 4k version should download later. The problem is that there is no definition for time expiration for searching for "Preferred" quality. A show can end, an episode can be watched a long time ago, and it'll still try to search and download a Preferred version for that show. Logically, I would say that if within a certain amount of days you can't find the Preferred quality, no point to continue looking. 

My feature request: Under Settings->Search Settings->Episode Search, under "proper search days" (which is a similar concept to this feature request, just with "Proper" versions), add an option for "Preferred search days". If that number of days passed since downloading the Allowed version, stop looking.

labrys:
That's a logical request.  You can get this type of behaviour through manually changing episodes to an Archived status when you no longer want them to search for a preferred quality.  A question for your feature request though, would you still want a 4k search result downloaded if it was found during a daily search?

moryoav:
It would make sense for episodes to automatically archive themselves after a certain amount of days and so they won't be searched at all during a normal daily search.

labrys:
Just to make sure you understand what a daily search is vs a backlog search.  Daily search is a constant ongoing process that checks your provider for any recently added releases and downloads anything that's wanted.  Its a very efficient way of getting missing items.  Backlog search on the other hand does a search for a single missing item at a time on each provider and is significantly less efficient but is more likely to find results.  Archived status will not replace existing files with a preferred quality even if found later during a daily search, but an auto-archiving feature would likely be easier to implement than just preventing backlog searching.

moryoav:
[...]

PR #3583, opened 2018-01-07
Don't search preferred if preferred expiration days is reached

fernandog:
POC
Need to decide if min=7 days and which value is default.

Usefull for users that doesn't delete episodes after watch it

As requested in: https://github.com/pymedusa/Medusa/issues/3580

image


fernandog:
@p0psicles no interest on this one? If not, close it.

p0psicles:
Hi long time. Hop around discord if you want a chat.

PR #3606, opened 2018-01-14
Fix tmdb image typecasting for rating and ratingcount

labrys:
TMDB `rating` and `ratingcount` are `str` instead of `float` and `int` respectively.

### DIFF
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -304,6 +304,7 @@ class Tmdb(BaseIndexer):
         """
         key_mapping = {'file_path': 'bannerpath', 'vote_count': 'ratingcount', 'vote_average': 'rating', 'id': 'id'}
         image_sizes = {'fanart': 'backdrop_sizes', 'poster': 'poster_sizes'}
+        typecasts = {'rating': float, 'ratingcount': int}
 
         log.debug('Getting show banners for {0}', sid)
         _images = {}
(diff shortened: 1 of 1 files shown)


medusa/indexers/tmdb/tmdb.py
```diff
@@ -304,6 +304,7 @@ def _parse_images(self, sid):
         """
         key_mapping = {'file_path': 'bannerpath', 'vote_count': 'ratingcount', 'vote_average': 'rating', 'id': 'id'}
         image_sizes = {'fanart': 'backdrop_sizes', 'poster': 'poster_sizes'}
+        typecasts = {'rating': float, 'ratingcount': int}
```
p0psicles: They should be stored as str, as storing them as float or int will result an error when trying to create the xml's for the nfo on some metadata providers.

labrys: Then the nfo creation should be fixed too, but as all the other providers use float/int for those values I don't see why it should affect nfo creation.

p0psicles: I'll go through the Metadata providers. Was already did some in the Imdb Pr.

p0psicles: I made it more consistent over de different providers.
Also fixed an issue with the show rating in the kodi metadata.
All should now also use text_type instead of str.

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -98,7 +101,7 @@ def _object_to_dict(self, tvdb_response, key_mapping=None, list_separator='|'):
                             continue
 
                         if isinstance(value, list):
-                            if list_separator and all(isinstance(x, (str, unicode)) for x in value):
+                            if list_separator and all(isinstance(x, (string_types)) for x in value):
```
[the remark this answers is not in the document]
labrys: Also all the `isinstance(x, int)` should be `isinstance(x, six.integer_types)` as Python 2 `long` is not addressed.

Issue #3613, opened 2018-01-16
TypeError: remove() takes exactly 1 argument (2 given)


scott451:
hello,

i use Medusa on DSM 6.1.4. Today I updated Git on the DSM to 1.15.1 I also update Python to 2.7.14. Following this update Medusa refused to start. I have tried everything including a re-install to no avail.

following is the error message i see when i try to start. sorry to not have more info and thanks for your help.

- scott

```
root@Valinor:/var/packages/sickbeard-custom# /var/packages/sickbeard-custom/scripts/start-stop-status start
Starting SickBeard Custom ...
/usr/local/ffmpeg/bin/ffprobe: /lib/libstdc++.so.6: no version information available (required by /usr/local/ffmpeg/lib/libx265.so.87)
/usr/local/ffmpeg/bin/ffprobe: /lib/libstdc++.so.6: no version information available (required by /usr/local/ffmpeg/lib/libx265.so.87)
Traceback (most recent call last):
  File "/usr/local/sickbeard-custom/var/SickBeard/SickBeard.py", line 4, in <module>
    from medusa.__main__ import main
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/__main__.py", line 67, in <module>
    from . import (
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/__init__.py", line 10, in <module>
    from medusa.providers.nzb import (
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/nzb/anizb.py", line 14, in <module>
    from medusa.providers.nzb.nzb_provider import NZBProvider
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/nzb/nzb_provider.py", line 8, in <module>
    from medusa.providers.generic_provider import GenericProvider
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/generic_provider.py", line 18, in <module>
    from medusa import (
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/config.py", line 30, in <module>
    from medusa.version_checker import CheckVersion
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/version_checker.py", line 31, in <module>
    from medusa.github_client import get_github_repo
  File "/volume1/@appstore/sickbeard-custom/var/SickBeard/medusa/github_client.py", line 6, in <module>
    import github
[...]

Issue #3618, opened 2018-01-17
Medusa can't start anymore in Synology


mikeo999:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:**
**OS:** Synology DSM 6.1.4.15217 Update 5
**What you did: Update where available for DSM also on for Pyhton to 2.7.14-17 version.
**What happened: The next day a saw that Medusa was stopped. Trying to start it failed with an error **failed to run the package service**
Medusa was running fine before this.

Then i try to start it manually with Putty
`/var/packages/sickbeard-custom/scripts/start-stop-status start`

And get this error.

**Logs:**
```

```
Traceback (most recent call last):
  File "/usr/local/sickbeard-custom/var/SickBeard/SickBeard.py", line 4, in <module>
    from medusa.__main__ import main
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/__main__.py", line 67, in <module>
    from . import (
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/__init__.py", line 10, in <module>
    from medusa.providers.nzb import (
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/nzb/anizb.py", line 14, in <module>
    from medusa.providers.nzb.nzb_provider import NZBProvider
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/nzb/nzb_provider.py", line 8, in <module>
    from medusa.providers.generic_provider import GenericProvider
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/providers/generic_provider.py", line 18, in <module>
    from medusa import (
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/config.py", line 30, in <module>
    from medusa.version_checker import CheckVersion
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/version_checker.py", line 31, in <module>
    from medusa.github_client import get_github_repo
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/medusa/github_client.py", line 6, in <module>
    import github
  File "/volume2/@appstore/sickbeard-custom/var/SickBeard/ext/github/__init__.py", line 37, in <module>
    from MainClass import Github, GithubIntegration
[...]

PR #3619, opened 2018-01-17
Import tempfile early on to ensure it is monkey patched

stefaang:
- [ x ] PR is based on the DEVELOP branch
- [ x ] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [ x ] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

Fixes openwrt / entware / synology issues with monkeypatching.

Deeper explanation: after printing the `f, args, type(args[0]), args[0].__type__` in the `patch_output` wrapper. 
```
<built-in function unlink> (u'/tmp/JhfhpO',) <type 'unicode'> <type 'unicode'>
<built-in function unlink> (<closed file '<fdopen>', mode 'w+b' at 0x244f600>, u'/tmp/tmpwipQTM') <type 'instance'> tempfile._TemporaryFileWrapper
<built-in function unlink> (<closed file '<fdopen>', mode 'w+b' at 0x2455850>, u'/tmp/tmpMSDo2T') <type 'instance'> tempfile._TemporaryFileWrapper
<built-in function unlink> (<closed file '<fdopen>', mode 'w+b' at 0x6a0f3c0>, u'/tmp/tmpW40zBz') <type 'instance'> tempfile._TemporaryFileWrapper
```
I suspected garbage collection issues with the cleanup with tempfiles. When I tried to drop these args, I saw that simply importing the tempfile module was sufficient (no need to drop them even).

Fixes https://github.com/pymedusa/Medusa/issues/3613, https://github.com/pymedusa/Medusa/issues/3618 and its first appearance in https://github.com/pymedusa/Medusa/issues/3567.

### DIFF
--- a/medusa/init/filesystem.py
+++ b/medusa/init/filesystem.py
@@ -8,6 +8,8 @@ import shutil
 import sys
 import tarfile
 
+# unused import but needs to be monkeypatched here
+import tempfile
 import certifi
 
 from six import binary_type, text_type
--- a/medusa/init/filesystem.py
+++ b/medusa/init/filesystem.py
@@ -8,8 +8,7 @@ import shutil
 import sys
 import tarfile
 
-# unused import but needs to be monkeypatched here
-import tempfile
+import tempfile # noqa # pylint: disable=unused-import
 import certifi
 
 from six import binary_type, text_type


medusa/init/filesystem.py
```diff
@@ -8,6 +8,8 @@
 import sys
 import tarfile
 
+# unused import but needs to be monkeypatched here
+import tempfile
```
p0psicles: # pylint: disable=unused-import

fernandog: I don't think Travis/flake8 read that

p0psicles: Well allot of IDE's do. So imo it doesn't do harm. And it prevents someone from removing it in the future, who think.. "hey an unused import". Putting the comment inline with pylint: makes it less user error prone.

stefaang: fixed for both

PR #3629, opened 2018-01-19
Added eslint-config xo and eslint, to be used with ide's that support…

p0psicles:
… .eslintrc.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+static/js/lib/**
+static/js/*.min.js
+static/js/vender.js
+static/js/api.js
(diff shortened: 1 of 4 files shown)


fernandog:
Does It work for pycharm?

p0psicles:
For reference, my pycharm config (ms visual code also works)
image

OmgImAlexis:
The `eslint-config` you added looks like the new one which is going to cause nothing but issues since it adds new rules that we can't use yet like requiring `const` and `let` instead of `var`. This should have been held off and the older version should have been used.

PR #3636, opened 2018-01-20
Add PrivateHD provider

Kriskras99:


medusa/providers/torrent/html/privatehd.py
```diff
@@ -0,0 +1,185 @@
...
+            'remember': 1,
+            'submit': 'Login',
+        }
+
+        response = self.session.post(self.urls['login'], data=login_params)
+        if not response or not response.text:
+            log.warning('Unable to connect to provider')
+            return False
+
+        if re.search('These credentials do not match our records.', response.text):
```
p0psicles: No need to use a regex for this. You can just do: `'These credentials do not match our records.' in response.text`

medusa/providers/torrent/html/privatehd.py
```diff
@@ -0,0 +1,184 @@
+# coding=utf-8
+
+"""Provider code for HDSpace."""
```
medariox: PrivateHD

medusa/providers/torrent/html/privatehd.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        results = []
+        if not self.login():
+            return results
+
+        for mode in search_strings:
+            log.debug('Search mode: {0}', mode)
+
+            for search_string in search_strings[mode]:
+                print("Search string:" + search_string)
```
medariox: Remove the print please

medusa/providers/torrent/html/privatehd.py
```diff
@@ -0,0 +1,184 @@
...
+            log.warning('Unable to connect to provider')
+            return False
+
+        if 'These credentials do not match our records.' in response.text:
+            log.warning('Invalid username or password. Check your settings')
+            return False
+
+        return True
+
+    def _check_auth(self):
```
medariox: This is not used?

medusa/providers/torrent/html/privatehd.py
```diff
@@ -0,0 +1,184 @@
...
+                        'pubdate': pubdate,
+                    }
+                    if mode != 'RSS':
+                        pass
+                    log.debug('Found result: {0} with {1} seeders and {2} leechers',
+                              title, seeders, leechers)
+
+                    items.append(item)
+                except (AttributeError, TypeError, KeyError, ValueError, IndexError):
+                    log.error('Failed parsing provider. Traceback: {0!r}',
```
medariox: Use log.exception here instead

medariox: log.exception('Failed parsing provider.')

medusa/providers/torrent/html/privatehd.py
```diff
@@ -0,0 +1,184 @@
...
+        self.password = None
+
+        # URLs
+        self.url = 'https://privatehd.to'
+        self.urls = {
+            'login': urljoin(self.url, 'auth/login'),
+            'search': urljoin(self.url, 'torrents'),
+        }
+
+        # Proper Strings
```
medariox: No proper string needed?

medusa/providers/torrent/html/privatehd.py
Kriskras99: It seems that the majority of providers use log.error, or is log.exception the new way?

medusa/providers/torrent/html/privatehd.py
Kriskras99: `_check_auth` is used by the `GenericProvider` class in the [`find_search_results` function](https://github.com/pymedusa/Medusa/blob/bd0832551d836713481dd0a93f4118372bb5d963/medusa/providers/generic_provider.py#L207), the provider will work without it because the [default implementation](https://github.com/pymedusa/Medusa/blob/bd0832551d836713481dd0a93f4118372bb5d963/medusa/providers/generic_provider.py#L549) always returns `True`.

medariox: Yep, that's because most of the time it's not really needed / redundant. If you want to add it, can you please raise AUtheException like here? https://github.com/pymedusa/Medusa/blob/1a8dc66c7ec9231be0d3444710e476437c9ae1c8/medusa/providers/torrent/html/gftracker.py#L220-L221

medusa/providers/torrent/html/privatehd.py
medariox: Yes, we change it recently.

Kriskras99:
For some reason there were already two icons for PrivateHD, but they were a bit outdated so I added a new one and removed the old ones.

Kriskras99:
@p0psicles @medariox I made the requested changes, is it good now?

PR #3714, opened 2018-02-05
Feature/increment cache invalidation

p0psicles:

~~@OmgImAlexis do you know if an alternative solution? A way to invalidate the cache, without needing to change the url?~~

~~To be clear, i'm all for a better alternative.~~

I misread and thought we were always sending 200 from server. But apparently we already have a fine working ETag. So i'll use this PR to revert my hard caching.

### DIFF
--- a/medusa/server/api/v2/series_asset.py
+++ b/medusa/server/api/v2/series_asset.py
@@ -33,6 +33,6 @@ class SeriesAssetHandler(BaseRequestHandler):
         if not asset:
             return self._not_found('Asset not found')
 
-        self.set_header('Cache-control', 'public, max-age=86400')
+        self.set_header('Cache-control', 'no-cache, must-revalidate')
 
         self._ok(stream=asset.media, content_type=asset.media_type)
--- a/medusa/server/api/v2/series_asset.py
+++ b/medusa/server/api/v2/series_asset.py
@@ -33,6 +33,4 @@ class SeriesAssetHandler(BaseRequestHandler):
         if not asset:
             return self._not_found('Asset not found')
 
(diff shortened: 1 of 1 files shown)


medusa/image_cache.py
```diff
@@ -163,6 +163,8 @@ def replace_images(series_obj):
     """
     remove_images(series_obj)
     fill_cache(series_obj)
+    # Update the increment, to bypass the client side caching.
+    app.INC += 1
```
p0psicles: Maybe this feels a bit awkward. But we and hide this in a simple decorator. Then a function that should up the incremental, would look like:
```
@invalidate_browser_cache
def replace_images(series_obj):
    """
    Replace cached images for a series based on image type.

    :param series_obj: Series object
    """
    remove_images(series_obj)
    fill_cache(series_obj)
```

themes-default/slim/static/js/core.js
```diff
@@ -36,7 +36,7 @@ const UTIL = {
         $('[asset]').each(function() {
             const asset = $(this).attr('asset');
             const series = $(this).attr('series');
-            const path = apiRoot + 'series/' + series + '/asset/' + asset + '?api_key=' + apiKey;
+            const path = apiRoot + `series/${series}/asset/${asset}?api_key=${apiKey}&${MEDUSA.config.pid}${MEDUSA.config.inc}`;
```
p0psicles: This will essentially place the PID+INC value after every image request.
If for example the PID = 1234, we start with the INC value 1. The value will be 12341.
After refreshing an image, the incremental value will be bumped to 2. And the value 12342 is added, making it essentially refresh the image.

PR #3765, opened 2018-02-13
remove old vue files

OmgImAlexis:
@p0psicles please check this through but from memory nothing relies on these directories so there isn't much point in keeping it.

If we do still have anything pointing at the vue directories let me know and I'll fix it up.


p0psicles:
I think you'll have to double check the gulp build scripts, as we did copy some of the files over to dist.

PR #3768, opened 2018-02-13
Recommended series: Use dogpile for caching of the imdb, trakt and anidb shows

### LINKED ISSUE #3749: Add Recommended -> IMDB gives error
Hi,

adding recommended shows from IMDB doesn't work. there are no errors or else to be seen in logs. 
the gui shows an error relating to u'ratingCount' ?

error

Branch: master
Commit: ccbc0ce9a2d3977a71bde03e29be503464ebe3cd
Version: v0.2.0
Database: 44.8

p0psicles:
* Fixed imdb recommended list, for series without the possibility to rate.

All shows from recommended lists, are cached now on disk.
Cache expires after 1 week.

Also fixes #3749 

### DIFF
--- a/medusa/show/recommendations/recommended.py
+++ b/medusa/show/recommendations/recommended.py
@@ -28,6 +28,7 @@ from medusa.cache import recommended_series_cache
 from medusa.indexers.utils import indexer_id_to_name
 from medusa.session.core import MedusaSession
 from simpleanidb import Anidb
+from six import binary_type, iteritems
 
 
 session = MedusaSession()
@@ -166,3 +167,23 @@ def cached_get_imdb_series_details(imdb_id):
     Use dogpile cache to return a cached id if available.
     """
     return imdb_api.get_title(imdb_id)
+
+
+def create_key_from_series(namespace, fn, **kw):
+    """Generate a key limiting the amount of dictionaries keys that are allowed to be used."""
+
+    def generate_key(*arg):
+        """
+        Generate the key.
+        The dogpiled items should have the following format:
+            `dict(keys={'key1': 'value_for_key1', 'key2': 'value_for_key2}, item=to_be_cached_item)`.
+        Following this standard we can cache every object, using this key_generator.
+        """
+        keys = arg[1]['keys']
+        return b'{namespace}_{arguments}'.format(
+            namespace=namespace, arguments=b'_'.join(
+                binary_type(v) for v in keys
+            )
+        )
+
+    return generate_key
--- a/medusa/show/recommendations/recommended.py
+++ b/medusa/show/recommendations/recommended.py
@@ -173,16 +173,16 @@ def cached_get_imdb_series_details(imdb_id):
 def create_key_from_series(namespace, fn, **kw):
     """Generate a key limiting the amount of dictionaries keys that are allowed to be used."""
 
-    def generate_key(*arg):
+    def generate_key(*arg, **kwargs):
         """
         Generate the key.
         The dogpiled items should have the following format:
             `dict(keys={'key1': 'value_for_key1', 'key2': 'value_for_key2}, item=to_be_cached_item)`.
         Following this standard we can cache every object, using this key_generator.
         """
-        keys = arg[1]['keys']
+        storage_keys = kwargs.pop('storage_keys')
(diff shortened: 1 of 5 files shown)


medusa/show/recommendations/trakt.py
```diff
@@ -36,9 +36,9 @@ def __init__(self):
         self.cache_subfolder = __name__.split('.')[-1] if '.' in __name__ else __name__
         self.recommender = "Trakt Popular"
         self.default_img_src = 'trakt-default.png'
-        self.anidb = Anidb(cache_dir=app.CACHE_DIR)
         self.tvdb_api_v2 = indexerApi(INDEXER_TVDBV2).indexer()
 
+    @recommended_series_cache.cache_on_arguments()
```
ratoaq2: The argument here is a show object. You need to provide an "object to key" function.

p0psicles: Maybe your right. But it seems to work fine. Maybe it pickling the object?

ratoaq2: It works because it's the same object instance. If the cache is persistent and you restart Medusa, the show object will be a different instance

p0psicles: No it still used the correct cached key

p0psicles: Give it a try, you'll see

p0psicles: This is a key that has been generated:
```
medusa.show.recommendations.imdb:_create_recommended_show|{u'rating': 8.8, u'votes': 50663, u'name': u'This Is Us', u'imdb_url': u'http://www.imdb.com/title/tt5555260/', u'outline': u'A heartwarming story of the struggles of the most unique set of triplets, and the wonderful parents who raised them.', u'image_url_large': u'https://images-na.ssl-images-amazon.com/images/M/MV5BOWU3M2I0YWItNWIzNC00MTBmLTkzMDgtMTdmOGMzOTI2YmVjXkEyXkFqcGdeQXVyNjMxNzcwOTI@._V1_.jpg', u'image_url': u'https://images-na.ssl-images-amazon.com/images/M/MV5BOWU3M2I0YWItNWIzNC00MTBmLTkzMDgtMTdmOGMzOTI2YmVjXkEyXkFqcGdeQXVyNjMxNzcwOTI@.__SY600_AL_.jpg', u'year': 2016, u'imdb_tt': u'tt5555260', u'image_path': u'images/imdb_popular/MV5BOWU3M2I0YWItNWIzNC00MTBmLTkzMDgtMTdmOGMzOTI2YmVjXkEyXkFqcGdeQXVyNjMxNzcwOTI@._V1_.jpg'}
```

It could still be better to calculate the key our selves. But that would be to not make it refresh, every time the rating or number of votes changes?

p0psicles: I see the confusion now. I named it `show_obj` while in fact it's a dict. That's why dogpile can handle it. It would be really different if a real object was passed.

p0psicles: I'm getting the hang of it. This dogpile hocus pocus is really awesome. I might even just use it to create an anidb indexer. As we need heavy caching for that.

p0psicles:
@medariox i standardized it a little. But to make sure it uses the key we want. I had to envelop the show data together with the key we want to use.

I also implement a function that will give us all cached recommended shows. So we can do something `get_all_recommended_series_from_cache(['imdb', 'anidb', 'trakt'])` and that will return all recommended show objects.

medusa/show/recommendations/recommended.py
```diff
@@ -131,3 +138,89 @@ def check_if_anime(self, anidb, tvdbid):
...
+        """
+        keys = arg[1]['keys']
+        return b'{arguments}'.format(
+            arguments=b'_'.join(binary_type(v) for v in keys)
+        )
+
+    return generate_key
+
+
+def update_recommended_series_cache_index(indexer, new_index):
```
p0psicles: @ratoaq2 i've created this function to store an index of all the keys that i've added.
do you know if there is already a native dogpile solution for this? I want to do something like return all shows with this namespace for example.

PR #3769, opened 2018-02-14
update xo

OmgImAlexis:
@p0psicles I don't think we should update legacy since we have effectively frozen it.


OmgImAlexis:
@p0psicles maybe we should look at moving the themes out of the main repo like I suggested and use npm to handle downloading them? It'd fix the issue of us needing 3 copies of each mako file.

p0psicles:
We have to throw that in the group and discuss the impact. Especially on the develop life cycle.

PR #3774, opened 2018-02-14
Added the ability to use an external CA Bundle for self-signed cert

abeloin:

This PR allow Medusa's requests.session to connect to services(e.g. rTorrent) that are using self-signed certificate without having to disable the SSL verification if REQUESTS_CA_BUNDLE or CURL_CA_BUNDLE is set, replicating requests override behaviour.

Tested in Linux. Should also work in Windows and MacOS


p0psicles:
I'm okay with adding this feature, we only don't want to start relying on environment variables.
Not every user is as tech savvy as we are. And until now we've always added these options as configuration options.

I'd suggest you add the option to the general -> advanced settings.

abeloin:
OK, I've made the changes. Let me know if you see anything else.

jfnz:
Should this feature also allow the ability to use deep-SSL inspection with a custom CA cert for connections out to things like GitHub and IMDB for update checks/information retrieval?

I note that at this stage it doesn't appear although Medusa will reference the OS certificate store (Windows, in this case) for trusted root CAs. This is causing SSL certificate errors on my deployment due to deep-SSL inspection running on the firewall between my Medusa server and the Internet.

PR #3781, opened 2018-02-18
FIX: shows with semi-colons in name return no results on Torrent9 (e.…

darkweaver87:
…g. NICS: New Orleans)


medariox:
@p0psicles 
It's because Torrent9 replaces all non-words with `-` in their search queries. I'll push another PR that uses regex substitution later.

PR #3805, opened 2018-02-24
Fix torrent_checker scheduler not working properly

abeloin:

This PR fix an issue with torrent_checker not starting/stopping properly after enabling it in the config and at startup.

Also added remove_ratio_reached() and remove_torrent() in generic.py as they should be a requirement for torrent_checker.py

Replaced the log debug in torrent_checker.py to logger.exception as I think remove_ratio_reached() should handle it's own exception and anything else should constitute an error.

### DIFF
--- a/medusa/__main__.py
+++ b/medusa/__main__.py
@@ -1327,7 +1327,7 @@ class Application(object):
                 app.trakt_checker_scheduler.silent = True
             app.trakt_checker_scheduler.start()
 
-            if app.USE_TORRENTS and app.REMOVE_FROM_CLIENT:
+            if app.USE_TORRENTS and app.REMOVE_FROM_CLIENT and app.TORRENT_METHOD != 'blackhole':
                 app.torrent_checker_scheduler.enable = True
             app.torrent_checker_scheduler.silent = False
             app.torrent_checker_scheduler.start()
--- a/medusa/config.py
+++ b/medusa/config.py
@@ -440,6 +440,32 @@ def change_PROCESS_AUTOMATICALLY(process_automatically):
         app.auto_post_processor_scheduler.silent = True
 
 
+def change_remove_from_client(new_state):
+    """
+    Enable/disable TorrentChecker thread
+    TODO: Make this return true/false on success/failure
+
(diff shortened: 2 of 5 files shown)


medusa/__main__.py
```diff
@@ -1327,7 +1327,7 @@ def start_threads():
                 app.trakt_checker_scheduler.silent = True
             app.trakt_checker_scheduler.start()
 
-            if app.USE_TORRENTS and app.REMOVE_FROM_CLIENT:
+            if app.USE_TORRENTS and app.REMOVE_FROM_CLIENT and not app.TORRENT_METHOD == 'blackhole':
```
medariox: Instead of `not app.TORRENT_METHOD == 'blackhole'` I'd use `app.TORRENT_METHOD != 'blackhole'`.

medusa/server/web/config/search.py
```diff
@@ -80,6 +80,11 @@ def saveSearch(self, use_nzbs=None, use_torrents=None, nzb_dir=None, sab_usernam
         app.TORRENT_METHOD = torrent_method
         app.USENET_RETENTION = try_int(usenet_retention, 500)
 
+        if not app.TORRENT_METHOD == 'blackhole' and app.TORRENT_METHOD in ('transmission', 'deluge', 'deluged'):
```
medariox: Same as above.

medusa/config.py
```diff
@@ -440,6 +440,32 @@ def change_PROCESS_AUTOMATICALLY(process_automatically):
         app.auto_post_processor_scheduler.silent = True
 
 
+def change_REMOVE_FROM_CLIENT(remove_from_client):
```
medariox: Please keep `change_REMOVE_FROM_CLIENT` all lowercase. I know the other functions are also like this, but it's not like it should be.

I'd also rename the parameter `remove_from_client` to `new_state` or something more explicit.

medariox:
Thanks to you!

This looks ready to me, but I don't use the feature at all.
@fernandog any feedback?

fernandog:
What ever abeloin does I'm ok

PR #3815, opened 2018-02-27
Feature/fix tabs

OmgImAlexis:
This removes the need for the `document.write` hack we used for the tabs by replacing it with the `full_url` passed via page templates.


p0psicles:
Just to make sure, it's reverse proxy safe?

OmgImAlexis:
It "should be" but I haven't tested as I don't have anything setup right now.

PR #3816, opened 2018-02-28
Fix systemd startup

tkimball83:

Summary: Make sure the medusa service is started after the network becomes available.

The current (default) systemd startup options will fire up medusa before the network becomes available.  This is fine if web_host is is set to 0.0.0.0, but if its set to the address of eno1/eth0 (eg, 192.168.0.254) then medusa will immediately die with the following error.

MAIN :: [ccbc0ce] Starting Medusa [master] using u'/opt/medusa/config.ini'
TORNADO :: [ccbc0ce] Starting Medusa on http://192.168.0.254:8081/
TORNADO :: [ccbc0ce] Could not start the web server on port 8081, already in use!

The error above is actually incorrect, which needs to be fixed as well.  The port 8081 is available, but the address of 192.168.0.254 is not as medusa tried to start before network.target.

This fix should ensure that medusa starts up correctly regardless of which interface/address you assign web_host to.

### DIFF
--- a/runscripts/init.systemd
+++ b/runscripts/init.systemd
@@ -44,6 +44,7 @@
 
 [Unit]
 Description=Medusa Daemon
+After=network.target
 
 [Service]
 User=medusa


p0psicles:
closed/opened to trigger travis build

PR #3818, opened 2018-02-28
move quality-chooser to vue

OmgImAlexis:
@p0psicles I still need to fix the forms on the pages that use this but it should be fine to test.

~**MAKE SURE NOT TO SAVE SHOWS WITH THIS IT WILL 100% KILL THE SHOW QUALITY**~

Edit: This should be working on both the edit and new show page. Please let me know if there's any issues.

Closes https://github.com/pymedusa/Medusa/issues/3909


p0psicles:
Looks nice. But because you replace the mako, it has to be implemented in a big bang right? Can't we create new mako vue import files. And them implement the components one page at a time?

OmgImAlexis:
I'm good with that. I still need to fix the forms but I could rename this and copy the old file back in?

Which page would we use first? Edit or new show?

OmgImAlexis:
I think the only issue I found was on the edit page it would show blank instead of the quality you have it set as so I need to fix that.

themes-default/slim/views/inc_qualityChooser.mako
```diff
@@ -1,5 +1,54 @@
+<script type="text/x-template" id="quality-chooser-template">
```
p0psicles: This file needs to be in themes-default/slim/views/vue-components/qualityChooser.mako

OmgImAlexis: Did you want me to replace this file with the old content and just move the new code to a new file?

p0psicles:
@OmgImAlexis as discussed we should remove all direct python calls / mako vars. And pass those along through props. This will result in a clean set of components, that we can move to .Vue files in the end.

themes-default/slim/views/vue-components/qualityChooser.mako
```diff
@@ -0,0 +1,199 @@
...
+                <h5>But it will stop searching when one of these is downloaded:  <label id="preferredExplanation">{{preferredExplanation.join(', ')}}</label></h5>
+                </template>
+            </div>
+            <div v-f="seriesSlug">
+                <h5 class="red-text" id="backloggedEpisodes" v-html="backloggedEpisodes"></h5>
+            </div>
+            <div id="archive" v-if="archive">
+                <h5>
+                    <b>
+                        Archive downloaded episodes that are not currently in <a target="_blank" href="manage/backlogOverview/"><font color="blue"><u>backlog</u>.</font></a>
```
p0psicles: This is not showing anything when changing to a lower quality. I suspect you wanted to subsittute the `.`

themes-default/slim/views/editShow.mako
```diff
@@ -18,6 +18,20 @@
 % if show.is_anime:
     <script type="text/javascript" src="js/blackwhite.js?${sbPID}"></script>
 % endif
+<script src="js/lib/vue.js"></script>
+<script src="js/lib/vue-async-computed@3.3.0.js"></script>
+<%include file="/vue-components/inc_qualityChooser.mako"/>
```
p0psicles: Still including the old file.

themes-default/slim/views/vue-components/qualityChooser.mako
```diff
@@ -0,0 +1,199 @@
...
+
+%>
+<script>
+Vue.component('quality-chooser', {
+    name: 'quality-chooser',
+    template: '#quality-chooser-template',
+    data() {
+        return {
+            // Python convertions
+            allowedQualities: ${convert(allowed_qualities)},
```
p0psicles: Can we change this so it's using props/events and not directly the python vars?

OmgImAlexis: For now let's leave this here since otherwise each file that uses this will need that python code which will just duplicate it.

p0psicles: In this case I don't fully agree. But it's up to you. And what works best for you. So i'm fine with it.
Besides this, you think it's ready to merge? Needs some more reviewing? When this is ready I can merge it into my editShow pr.

p0psicles: Also it'l need a rebase.

OmgImAlexis: I think after a rebase it'll still need a little bit of testing as I believe it still has the issue of getting the wrong quality when you first open the edit show page.

p0psicles: OK I'll merge it into my branch. Let's test there.

sharkykh:
- [ ] Remove missed `<script type="text/javascript" src="js/quality-chooser.js?${sbPID}"></script>` in `editShow.mako`
- [ ] Remove `quality-chooser.js.map` files
- [ ] Verify Backlog warning / Archive episode works now
- [ ] Fix massEdit quality chooser:
     - [ ] Update component to support receiving quality as a property.
     - [ ] Update component to support the `<Keep> (value="keep")` option, and show it for `massEdit` only.
     - [ ] Make sure mass-editing quality actually saves the correct quality.

p0psicles:
@OmgImAlexis @sharkykh i'm a bit confused. There aren't any child -> parent event's defined nor methods to patch the quality settings for a show directly? How should we update selected quality settings?

sharkykh:
@p0psicles If I understand you question,
Then I think this is what is needed for `massEdit`, since you have to "pass" the quality to the component in `<quality-chooser quality="[quality]"></quality-chooser>`, so it displays the correct one.

p0psicles:
Yeah have been talking with @medariox about it. And imo we should split quality into allowed and preferred. So we can cleanly pass them as properties but also store them independently.

Else we would have so always store the combined quality. But then you keep glaring to combine them in frontend.

OmgImAlexis:
> Fix addShows_addExistingShow.mako - remove JS import, add component import.

@sharkykh can you check my last commit fixed this?

p0psicles:
Also emits an event when one of the qualities (allowed/preferred) changes.
Can be used in the parent like:
`<quality-chooser @update="saveQualities"></quality-chooser>`

p0psicles:
What else do we need to do for this?

sharkykh:
With the latest commit:
-  Mass Edit quality chooser is working.
   `< Keep >` option will show up when editing two or more shows and at least one quality setting is different from the others.
- Props added to the component:
  `overallQuality` [Number] - combined quality (allowed+preferred) to set in the quality chooser on load.
  Defaults to the converted Python var: `show.quality` if a show object is detected, `app.QUALITY_DEFAULT` otherwise.
  `keep` [Boolean] - This is used to mark the component for massEdit - for the added feature of the `Keep` option for the preset selector. Defaults to `false`.
- Change in events:
  The qualities are now `$emit()`-ing individually, as:
  `update:quality:allowed` for allowed qualities
  `update:quality:preferred` for preferred qualities

**Notes:**
- Recalculating the backlogged episodes after archiving **is not working.** This is a small visual bug.
  Marked it as `@FIXME` for the sake of merging faster.

PR #3852, opened 2018-03-11
More changes for Python 3 compatibility

medariox:
- [x PR is based on the DEVELOP branch

@duramato


labrys:
@medariox Up to you but instead of making the sql queries bytes, you should fix the db and make them all unicode.  It makes for fewer issues all around.

medariox:
@labrys You're absolutely right. That will be done after we are comfortable with the current changes.

labrys:
Just an FYI, I've already done it in my fork and it's working well.  I've been running py3 for about 3 weeks now with no issues.  I recreated my database due to other changes I've made, so I would test migration before you implemented it.

PR #3881, opened 2018-03-20
Fixing issue of no matches showing when only one result available (hebits)

smoresmores:

When only one result appeared in the search it was not parsing it.
Tiny fix.

### DIFF
--- a/medusa/providers/torrent/html/hebits.py
+++ b/medusa/providers/torrent/html/hebits.py
@@ -117,7 +117,7 @@ class HeBitsProvider(TorrentProvider):
             torrent_rows = torrent_table('div', class_=re.compile('^line')) if torrent_table else []
 
             # Continue only if at least one release is found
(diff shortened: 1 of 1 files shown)


p0psicles:
We use mostly < 2 when the table header is also included in de results.

smoresmores:
Yes, which in this case the header is not included. As I am using regex to find all the results that start with "line" and this doesn't include the header. That's why it didn't work when only one result was found, because the size of the list is one and it worked as there are no results at all.
Now this should work fine though.

myakove:
Works for me with this fix.

medariox:
@smoresmores 
Can you rebase the PR please?

p0psicles:
Meaning can you merge develop into this branch pls.

smoresmores:
Sorry for the delay.
Hope it's fine now.

PR #3884, opened 2018-03-20
add vue to all pages and convert all links to app-link

OmgImAlexis:
This allows us to start on all the pages as we now have access to global components like app-link and asset.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
         'pep8-naming',
         'pytest',
         'pytest-cov',
-        'pytest-flake8',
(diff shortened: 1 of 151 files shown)


p0psicles:
Just double checking. I know it shouldn't be the case. But do we have any styles  on the anchors?

OmgImAlexis:
It shouldn't matter as this renders anchor links just as if we had them on the page normally. The <app-link> won't show on the HTML once vue starts.

p0psicles:
Don't know where those flake warning are coming from. But i'll fix it later for you.

Issue #3888, opened 2018-03-21
History page not showing up anymore (develop)


medariox:
Since this commit: https://github.com/pymedusa/Medusa/pull/3884

Also the pages `/news`, `/changes` and `/IRC`:
![img](https://snapr.pw/i/9a353ac965.png)

OmgImAlexis:
Yep I'll fix that.

medariox:
Fixed

PR #3889, opened 2018-03-21
fix missing vue pages

OmgImAlexis:
Closes https://github.com/pymedusa/Medusa/issues/3888

### DIFF
--- a/themes-default/slim/static/js/core.js
+++ b/themes-default/slim/static/js/core.js
@@ -190,25 +190,6 @@ function wsCheckNotifications() {
 // Listen for the config loaded event.
 window.addEventListener('build', e => {
     if (e.detail === 'medusa config loaded') {
-        /**
-         * Search for anchors with the attribute indexer-to-name and translate the indexer id to a name using the helper
-         * function MEDUSA.config.indexers.indexerIdToName().
-         *
-         * The anchor is rebuild using the indexer name.
-         */
-        $('[data-indexer-to-name]').each((index, target) => {
-            const indexerId = $(target).attr('data-indexer-to-name');
-            const indexerName = MEDUSA.config.indexers.indexerIdToName(indexerId);
-
-            const re = /indexer-to-name/gi;
-
-            $.each(target.attributes, (index, attr) => {
-                if (attr.name !== 'data-indexer-to-name' && target[attr.name]) {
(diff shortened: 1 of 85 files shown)


sharkykh:
Can you check the `/errorlogs/` route? Seems broken to me as well

sharkykh:
same for `/addRecommended/`

sharkykh:
Also found this is broken (at least those two lines):
https://github.com/pymedusa/Medusa/pull/3884/files#diff-c05bb2c59f6b274b8c5134313046b898R51

Error is:
```
SyntaxException: (SyntaxError) EOL while scanning string literal (<unknown>, line 1) (u'\'realpage + \'/\'}?traktList=${traktList}#tabs-1">Ma') in file 'C:/Medusa/themes/light/templates/addShows_trendingShows.mako' at line: 51 char: 75
```

sharkykh:
I have a show with a double quotation mark in its name - `תאג"ד`, since the last PR it broke some stuff:

Shows list:
before

after


Show page:
before
chrome_2018-03-22_01-49-56
after
chrome_2018-03-22_01-50-17

sharkykh:
Not sure if related to my last comment - recent shows links are broken for me:
I get the link - `https://localhost:8081/home/displayShow?indexername=indexer-to-name&seriesid=316913`.
This seems to be unintended?
https://github.com/pymedusa/Medusa/blob/1d62c9064c260d003cc825f069b980c85c5837ce/themes/light/templates/partials/header.mako#L36

OmgImAlexis:
@sharkykh that should be everything fixed. Can you please review.

sharkykh:
Still seeing an error in `addShows_recommended.mako` (lines 51-52) and the character escape issue in `partials/showheader.mako`.
Apart from that and the recent shows, LGTM.

sharkykh:
By the way, it seems like the `indexer-to-name` stuff is injected using JS.
But it's late here, so I might be wrong.

OmgImAlexis:
@sharkykh recent should be fine now.

@p0psicles I've removed your `data-indexer-to-name` function and moved the relevant code to the `app-link` so you can now just pass a `indexer-id` to `app-link` and it'll replace the placeholder with the correct value.

I also added placeholder as a prop so you can change the placeholder as it defaults to `indexer-to-name`.

PR #3896, opened 2018-03-22
Fix Git token re-encryption

### LINKED ISSUE #3895: `git_token` is encrypted multiple times and becomes unusable
**Branch/Commit:** 527c2326aefb788a5359de0c06843dd8f68d4fcd
**OS:** Windows 7

`git_token` config entry gets re-encrypted and loses access to the raw, unencrypted value.

**Steps to reproduce:**
1. Enable the *Encrypt passwords* option in General -> Advanced
2. Put in a fresh Github Personal Access Token and save the General config.
3. Check `config.ini`, original token is now encrypted once - save that encrypted value
4. Restart Medusa
5. Go to General config and save the config.
6. Check `config.ini`, original token is now encrypted *twice* - notice that the length is a little longer than the first one.
7. **The token is now unusable.** At this point you will start getting the following warning:
    > Invalid Github credentials. Please check your Github credentials in Medusa settings.

Repeat steps 4 to 6. The token just keeps getting re-encrypted and becomes unusable.

image

sharkykh:
This is one way to solve the issue (tested, working).
Though, it may be a bit confusing to see other encrypted config entries without the `encrypted=True`.

In my mind - it's either this fix or changing the entry name to `git_token_password`, which looks weird.
(Note: Adding `token` as a keyword for encryption like `password` would cause a bigger change)

Your thoughts?

Fixes #3895

### DIFF
--- a/medusa/__main__.py
+++ b/medusa/__main__.py
@@ -437,7 +437,7 @@ class Application(object):
             app.GIT_AUTH_TYPE = check_setting_int(app.CFG, 'General', 'git_auth_type', 0)
             app.GIT_USERNAME = check_setting_str(app.CFG, 'General', 'git_username', '')
             app.GIT_PASSWORD = check_setting_str(app.CFG, 'General', 'git_password', '', censor_log='low')
-            app.GIT_TOKEN = check_setting_str(app.CFG, 'General', 'git_token', '', censor_log='low')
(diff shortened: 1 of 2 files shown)


medariox:
I think this is a good solution and the best we can do for now.

PR #3904, opened 2018-03-24
add 1 hour cache to series assets

OmgImAlexis:
I've noticed we're missing the `cache-control` header for our series assets. So far I've seen quite a good improvement in loading times for the homepage.


OmgImAlexis:
Instead let's wait till tornado is updated.

Ref: https://github.com/tornadoweb/tornado/pull/2270

PR #3918, opened 2018-03-26
Template fixes

sharkykh:
* `snatchSelection.mako`:
   Search results table has 12 columns (I'm assuming this is from when you added the Published column)
   This just looked weird (bottom of the table)

* `genericMessage.mako` was missing the Vue enabling script. This made navigation not work.

* `inc_qualityChooser.mako`
  `</br>` is not a valid html tag.

* `file-browser.mako`
  Fix 2 unmatched tags: `div` & `app-link`


p0psicles:
From what I remember. It would not respond to submit when selecting a folder.

PR #3920, opened 2018-03-26
Fix sorting by published date on manual search results table

### LINKED ISSUE #3917: Sorting published date in manual search does not work
Medusa Info: | Branch: master                 Commit: d6eb72d4e1019e5d1286ae6ac1d7e4b318229ebe                 Version: v0.2.1                 Database: 44.9
-- | --
Python Version: | 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:25:58) [MSC v.1500 64 bit (AMD64)]
SSL Version: | OpenSSL 1.0.2k  26 Jan 2017
OS: | Windows-10-10.0.14393
Locale: | nl_NL.cp1252

Sorting published date in manual search does not work like expected, it does work with Added date for instance.
image

Example for added date
image

sharkykh:
Fixes #3917 

Changed this to be like the `Added` date.


medusa/search/manual.py
```diff
@@ -300,8 +300,8 @@ def get_provider_cache_results(series_obj, show_all_results=None, perform_search
             i['pretty_size'] = pretty_file_size(i['size']) if i['size'] > -1 else 'N/A'
             i['seeders'] = i['seeders'] if i['seeders'] >= 0 else '-'
             i['leechers'] = i['leechers'] if i['leechers'] >= 0 else '-'
-            i['pubdate'] = sbdatetime.convert_to_setting(parser.parse(i['pubdate'])).strftime(
-                app.DATE_PRESET + ' ' + app.TIME_PRESET) if i['pubdate'] else '-'
+            i['pubdate'] = parser.parse(i['pubdate']).strftime(
+                '%Y-%m-%d %H:%M:%S') if i['pubdate'] else '-'
```
fernandog: not using the `convert_to_setting`, it won't convert the datetime to user local timezone.

`added` is already in user local timezone, so don't need to be converted

sharkykh: https://github.com/pymedusa/Medusa/blob/be284322b5b0ed8cf8624a3827f3210501573335/medusa/sbdatetime.py#L110-L117

That bit of code means if you select "network" as your timezone in config, you get 'published' times in an incorrect timezone. That's not the correct solution in my opinion.

sharkykh:
Should be correct now.

fernandog:
I can't test. Anyone?

medusa/search/manual.py
```diff
@@ -300,8 +300,8 @@ def get_provider_cache_results(series_obj, show_all_results=None, perform_search
             i['pretty_size'] = pretty_file_size(i['size']) if i['size'] > -1 else 'N/A'
             i['seeders'] = i['seeders'] if i['seeders'] >= 0 else '-'
             i['leechers'] = i['leechers'] if i['leechers'] >= 0 else '-'
-            i['pubdate'] = sbdatetime.convert_to_setting(parser.parse(i['pubdate'])).strftime(
-                app.DATE_PRESET + ' ' + app.TIME_PRESET) if i['pubdate'] else '-'
+            i['pubdate'] = parser.parse(i['pubdate']).astimezone(app_timezone).strftime(
+                '%Y-%m-%d %H:%M:%S') if i['pubdate'] else '-'
```
medariox: The code looks good to me, but why did you remove the time and date preset? It was meant to show the pubdate as set in the config.

sharkykh: image

None of the date formats matched.
If I change one, I'll want to change the others... But what is ideal? Should I apply the user's preference for "timeago" to all datetimes here?

fernandog: I still think the issue is in the table sort parsing, in the javascript fle

sharkykh: Also, Published is the only column that is unsortable (of the datetime columns), probably because we need to have a timestamp for each one in the code - so that tableSorter can easily sort them.

fernandog: @OmgImAlexis which file does the tablesorting parsing for date fields in the manual search?

sharkykh: https://github.com/pymedusa/Medusa/blob/be284322b5b0ed8cf8624a3827f3210501573335/themes-default/slim/static/js/home/snatch-selection.js#L75-L97

fernandog: I can't find this code for manual search table sorter:

https://github.com/pymedusa/Medusa/blob/master/themes/light/assets/js/home/index.js#L103-L126

sharkykh: I'm working on it again now. I also noticed if you try to sort by quality it's not sorting correctly. Fixing tableSorter config should fix this as well.

fernandog:
@sharkykh nice! This used to work. I don't know how much code was deleted to stop working

sharkykh:
You guys are doing a big overhaul of the frontend, it's bound to happen to everything.
Also I'm pretty sure tableSorter is *one of the oldest* features of the UI...

medariox:
@fernandog Can you approve if you think the PR is good now?

fernandog:
I didn't test also JS is not something I understand. If you said is good then ok

PR #3921, opened 2018-03-26
[Feature] Show floating notification badge in small views

sharkykh:
Just a small QoL thing...
Show a floating notification badge (for errors / warnings / news) on the navbar toggle button in when using small views such as mobile.

chrome_2018-03-26_20-47-10

### DIFF
--- a/medusa/server/web/core/base.py
+++ b/medusa/server/web/core/base.py
@@ -111,6 +111,7 @@ class PageTemplate(MakoTemplate):
             'newsBadge': '',
             'toolsBadge': '',
             'toolsBadgeClass': '',
+            'floatingMenuBadge': '',
             'base_url': base_url + app.WEB_ROOT + '/',
             'realpage': '',
             'full_url': base_url + rh.request.uri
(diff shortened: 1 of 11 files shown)


medusa/server/web/core/base.py
```diff
@@ -111,6 +111,7 @@ def __init__(self, rh, filename):
             'newsBadge': '',
             'toolsBadge': '',
             'toolsBadgeClass': '',
+            'floatingMenuBadge': '',
```
OmgImAlexis: No new python vars please. This should be implemented client side.

themes-default/legacy/static/css/style.css
```diff
@@ -2843,6 +2843,20 @@ select,
     padding: 0;
 }
 
+.floating-badge {
```
OmgImAlexis: No updating legacy please.

medusa/server/web/core/base.py
sharkykh: Do you mean 'reading' the badge info from the tools badge and recreating it?

sharkykh: I should be more specific - most of the info needed to create a badge with the correct color is only server side (apart from the unread news count which is global (`app.UNREAD_NEWS`).

OmgImAlexis: I mean there should be no changes to the python code for any new ui changes. Everything should now be done client side. The badges and HTML that're currently handled server side like the ones above will all be removed very soon.

PR #3929, opened 2018-03-28
Newpct restructre

snamds:
Compared to previous version:
- Avoid second web fecth for every daily content result
- Newpct packs of episodes work in manual searches
- Season search work
- Independent of newpct url episode naming


medariox:
Sorry, but I don't think we can accept a provider with that amount of customization. You said that the download wouldn't require an additional request, but isn't storing `details_url` as `link` exactly the same? I think it makes more sense to replace Newpct with an alternative Spanish provider. Anyway - thanks for the effort you put in the PR.

snamds:
My bad. I didn't realize the current medusa code only did two requests in the rss mode. I thought another request was done for every result listed.

I think newpct (tvsinpagar, torrentlocura, torrentrapid, descargas2020, all the same) is the way to go. It's  by far the web with more content and the one that updates faster. There are another pages but they appear and get closed. And the ones that resist use newpct releases. If you go to mejortorrent and download a torrent, you'll see it's a newpct file. If you go to elitetorrent.biz, you will find a newpct file.

The current medusa code stores an attempt to guess the torrent url content. This is not going to work anymore. They have added a kind of timestamp in every torrent url. This PR stores the link to the details url page. This details page is only downloaded if the item is requested to be downloaded by medusa.

Is there anything I can do? What are you not comfortable with? Just trying to help

snamds:
For those interested in newpct releases, this code has been moved to jackett.

PR #3961, opened 2018-04-05
Fix Torrent9 torrent provider

medariox:


medusa/providers/generic_provider.py
```diff
@@ -141,7 +141,7 @@ def download_result(self, result):
             verify = False if self.public else None
 
             if download_file(url, filename, session=self.session, headers=self.headers,
-                             hooks={'response': self.get_url_hook}, verify=verify):
```
medariox: @p0psicles 
I've removed this because it was preventing the Cloudflare hook to work. Since the hook is attached to the session, I don't think it is required at all. Possibly something that was needed before the current hooks implementation?

PR #4005, opened 2018-04-12
Missing torrent download path in WebUI command to uTorrent (per issue #3963); Add optional dynamic/per-show label for uTorrent; fix WebUI for seeding time set as "-1" (disabled)

rafi-d:
1. Change/fix for  uTorrent "snatches" to be directed to the exact download destination sub-folder location, based on that destination defined in Medusa. If there was no destination defined, we'll use the show NAME as destination sub-folder (this is commented in the source)

2. Add an option to use per-show label, by using "%N" notation in the search->utorrent->label(s) edit-boxes. This can be "mixed" with other text like "tv-Medusa_%N" . Description text is also added to the GUI for the three themes .

Thanks

PR #4006, opened 2018-04-12
Fix uTorrent "pause" WebUI command (search option)

rafi-d:
Change  to "stop" WebAPI command, since "pause" is not recommended (removed from uTorrent toolbar), and harmful to the torrents' swarm.  Basically - it has similar effect - stops the added torrent from proceeding  with the download, as intended.

Thanks

Rafi 

PS: I was unable to change to develop branch... Please help.


OmgImAlexis:
Duplicate of https://github.com/pymedusa/Medusa/pull/4005

Issue #2449, opened 2017-03-22
Subtitles issue with db


duramato:
Happen while searching for subtitle from the Post-Processor

```
2017-03-21 19:28:07 WARNING  Thread-18 :: [79ebec8] Failed to refine video
Traceback (most recent call last):
  File "/home/pi/Medusa/lib/subliminal/core.py", line 543, in refine
    refiner_manager[refiner].plugin(video, **kwargs)
  File "/home/pi/Medusa/lib/subliminal/refiners/tvdb.py", line 267, in refine
    results = search_series(video.series.lower())
  File "/home/pi/Medusa/lib/dogpile/cache/region.py", line 1220, in decorate
    should_cache_fn)
  File "/home/pi/Medusa/lib/dogpile/cache/region.py", line 825, in get_or_create
    async_creator) as value:
  File "/home/pi/Medusa/lib/dogpile/lock.py", line 154, in __enter__
    return self._enter()
  File "/home/pi/Medusa/lib/dogpile/lock.py", line 87, in _enter
    value = value_fn()
  File "/home/pi/Medusa/lib/dogpile/cache/region.py", line 780, in get_value
    value = self.backend.get(key)
  File "/home/pi/Medusa/lib/dogpile/cache/backends/file.py", line 220, in get
    with self._dbm_file(False) as dbm:
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/home/pi/Medusa/lib/dogpile/cache/backends/file.py", line 215, in _dbm_file
    "w" if write else "r")
  File "/usr/lib/python2.7/anydbm.py", line 79, in open
    raise error, "need 'c' or 'n' flag to open new db"
error: need 'c' or 'n' flag to open new db
```

Full log: https://gist.github.com/duramato/be8bbe41f5a01db74032d737b6cabae5

This happen after i enabled **"Erase subtitles cache on next boot"**
@ratoaq2 @fernandog 
Dont know if it's Medusa's issue or an upstream one

ratoaq2:
Did you reboot?

duramato:
Yes, this happen after the reboot

ratoaq2:
Shutdown, remove cached files and start

fernandog:
here works fine:

2017-03-24 08:07:49 INFO     EVENT-QUEUE :: [b4a8a54] Saving all shows to the database
[...]

Issue #4009, opened 2018-04-13
[APP SUBMITTED]: error: need 'c' or 'n' flag to open new db


tomsema:
### INFO
**Python Version**: `2.7.12 (default, Dec  4 2017, 14:50:18) [GCC 5.4.0 20160609]`
**Operating System**: `Linux-4.4.0-119-generic-x86_64-with-Ubuntu-16.04-xenial`
**Locale**: `UTF-8`
**Branch**: [master](../tree/master)
**Database**: `44.9`
**Commit**: pymedusa/Medusa@d6eb72d4e1019e5d1286ae6ac1d7e4b318229ebe
**Link to Log**: https://gist.github.com/e14066318bbb5ec4eb0e7f58fa88b69c
### ERROR
<pre>
2018-04-13 11:03:10 ERROR    Thread-29 :: [d6eb72d] Exception generated: need 'c' or 'n' flag to open new db
Traceback (most recent call last):
  File "/opt/medusa/<a href="../blob/d6eb72d4e1019e5d1286ae6ac1d7e4b318229ebe/medusa/server/web/core/base.py#L283">medusa/server/web/core/base.py</a>", line 283, in async_call
    result = function(**kwargs)
  File "/opt/medusa/<a href="../blob/d6eb72d4e1019e5d1286ae6ac1d7e4b318229ebe/medusa/server/web/home/handler.py#L2287">medusa/server/web/home/handler.py</a>", line 2287, in manual_search_subtitles
    found_subtitles = subtitles.list_subtitles(tv_episode=ep_obj, video_path=video_path)
  File "/opt/medusa/<a href="../blob/d6eb72d4e1019e5d1286ae6ac1d7e4b318229ebe/medusa/subtitles.py#L331">medusa/subtitles.py</a>", line 331, in list_subtitles
    cache.set(subtitle_key.format(id=subtitle.id).encode('utf-8'), subtitle)
  File "/opt/medusa/<a href="../blob/d6eb72d4e1019e5d1286ae6ac1d7e4b318229ebe/ext/dogpile/cache/region.py#L984">ext/dogpile/cache/region.py</a>", line 984, in set
    self.backend.set(key, self._value(value))
  File "/opt/medusa/<a href="../blob/d6eb72d4e1019e5d1286ae6ac1d7e4b318229ebe/ext/dogpile/cache/backends/file.py#L237">ext/dogpile/cache/backends/file.py</a>", line 237, in set
    with self._dbm_file(True) as dbm:
[...]

Issue #4008, opened 2018-04-13
Manual Subtitles Search Error


tomsema:
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit: master
**OS:Linux-4.4.0-119-generic-x86_64-with-Ubuntu-16.04-xenial
**SSL Version: OpenSSL 1.0.2g
**Python Version: 2.7.12

**What you did: Manual Subtitles Search. 
**What happened: Subtitles was not found and received error
**What you expected: Receive subtitles for episode
**Logs:**
```
PASTE LOGS HERE
2018-04-13 11:03:10 ERROR    Thread-29 :: [d6eb72d] Exception generated: need 'c' or 'n' flag to open new db
Traceback (most recent call last):
  File "/opt/medusa/medusa/server/web/core/base.py", line 283, in async_call
    result = function(**kwargs)
  File "/opt/medusa/medusa/server/web/home/handler.py", line 2287, in manual_search_subtitles
    found_subtitles = subtitles.list_subtitles(tv_episode=ep_obj, video_path=video_path)
  File "/opt/medusa/medusa/subtitles.py", line 331, in list_subtitles
    cache.set(subtitle_key.format(id=subtitle.id).encode('utf-8'), subtitle)
  File "/opt/medusa/ext/dogpile/cache/region.py", line 984, in set
    self.backend.set(key, self._value(value))
  File "/opt/medusa/ext/dogpile/cache/backends/file.py", line 237, in set
    with self._dbm_file(True) as dbm:
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/opt/medusa/ext/dogpile/cache/backends/file.py", line 215, in _dbm_file
    "w" if write else "r")
  File "/usr/lib/python2.7/anydbm.py", line 79, in open
    raise error, "need 'c' or 'n' flag to open new db"
error: need 'c' or 'n' flag to open new db

medariox:
Thanks for the bug report. Fixed with: https://github.com/pymedusa/Medusa/pull/4011

PR #4011, opened 2018-04-13
Configure subtitle cache after "erase subtitle cache" action

fernandog:

### DIFF
--- a/medusa/__main__.py
+++ b/medusa/__main__.py
@@ -1088,7 +1088,7 @@ class Application(object):
                 except OSError as e:
                     logger.warning(u'Unable to remove subtitles cache files. Error: {error}', error=e)
                 # Disable flag to erase cache
-                app.SUBTITLES_ERASE_CACHE = 0
+                app.SUBTITLES_ERASE_CACHE = False
 
             # Check if we need to perform a restore of the cache folder
             Application.restore_cache_folder(app.CACHE_DIR)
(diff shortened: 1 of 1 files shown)


medusa/__main__.py
```diff
@@ -1094,6 +1090,10 @@ def initialize(self, console_logging=True):
                 # Disable flag to erase cache
                 app.SUBTITLES_ERASE_CACHE = 0
```
sharkykh: Should be `app.SUBTITLES_ERASE_CACHE = False`

medusa/__main__.py
```diff
@@ -1094,6 +1090,10 @@ def initialize(self, console_logging=True):
...
 
+            # Check if we need to perform a restore of the cache folder
+            Application.restore_cache_folder(app.CACHE_DIR)
```
sharkykh: Keep these two lines where they we're. This is the restore feature. You only need the next line here IMO.

sharkykh:
Fixes #2449, #4008, #4009

medusa/__main__.py
fernandog: If we restore before we could erase if the erase setting is enabled. Right?

sharkykh: Yes. But, I mean.. in that case, what do you think the user wants to do?Restore the old cache or remove it?

sharkykh: Oh, I see what you mean now...
I guess you're right :+1:

PR #4024, opened 2018-04-14
Fix "Add Existing Show" page

### LINKED ISSUE #3909: Adding existing shows
Branch: develop
Commit: fc7ca7f2ed12fa25935fb4024afc47da078793ef
Version: v0.1.22
Database: 44.9

When trying to add existing shows not Displaying folders within these directories and cant add any additional directories. None of the buttons work.
image

No errors in logs

sharkykh:
### AS SEEN ON SLACK... :laughing: 

Fixes the "Add Existing Show" page, just so it works right now, until it can be fully converted/moved to Vue.
Left some comments in for future reference, like the `rootDirText` value not actually updating / triggering the `change` event.
**Edit:** `root-dirs` needs to be converted to Vue.js before this can be fixed.

Fixes #3909 

**Idea...**
Fix `/addShows/existingShows/`'s "Customize Options" to have a direct effect on the added shows.
Currently, when you don't check "Prompt me to set settings for each show", you have to save the options as default for it to have an effect on the added shows.
If you don't save, it uses the current default options.


themes/light/templates/addShows_addExistingShow.mako
```diff
@@ -12,6 +12,8 @@ const startVue = () => {
         el: '#vue-wrap',
         data() {
             return {
+                header: '${header}',
+                title: '${title}',
```
sharkykh: This might not work as expected.
https://github.com/zzzeek/mako/blob/a582879d89ac6e37b9daa8b3c820ebedddad2aea/mako/runtime.py#L215-L234
That will raise a name error if you try and cast `str( )` on it (which is what `${   }` basically does)
I think this'll work:
```
header: '${header or ''}',
title: '${title or ''}',
```

sharkykh: Obviously it will raise the exception only when header/title are actually undefined.

sharkykh: TL;DR

After checking this,
1. My solution doesn't work if it's the variable is undefined (that's because [`strict_undefined=True`](http://docs.makotemplates.org/en/latest/runtime.html#context-variables)).
2. `header` and `title` are set to the value `FixME` by default when constructing `PageTemplate`. They should always be available in the template context, **and if they aren't - that's a bug to be fixed,** and it will throw a Mako error, like so:
```
2018-04-15 15:30:21 DEBUG    Thread_13 :: [6d54662] File C:\Medusa\medusa\server\web\core\base.py, line 147, in render
2018-04-15 15:30:21 DEBUG    Thread_13 :: [6d54662] File C:\Medusa\ext\mako\template.py, line 471, in render_unicode
2018-04-15 15:30:21 DEBUG    Thread_13 :: [6d54662] File C:\Medusa\ext\mako\runtime.py, line 838, in _render
2018-04-15 15:30:21 DEBUG    Thread_13 :: [6d54662] File C:\Medusa\ext\mako\runtime.py, line 873, in _render_context
2018-04-15 15:30:21 DEBUG    Thread_13 :: [6d54662] File C:\Medusa\ext\mako\runtime.py, line 899, in _exec_template
2018-04-15 15:30:21 DEBUG    Thread_13 :: [6d54662] File C:/Medusa/themes/light/templates/layouts/main.mako, line 128, in render_body
2018-04-15 15:30:21 DEBUG    Thread_13 :: [6d54662] File C:/Medusa/themes/light/templates/addShows_addExistingShow.mako, line 5, in render_scripts
2018-04-15 15:30:21 INFO     Thread_13 :: [6d54662] NameError: 'heade' is not defined
```

OmgImAlexis: @sharkykh yep, all the main vars in mako templates are set before we load anything in. Most using `FixMe` so we can easily find unset vars.

themes-default/slim/views/addShows_addExistingShow.mako
```diff
@@ -14,8 +14,87 @@ const startVue = () => {
         },
         data() {
             return {
```
OmgImAlexis: Please add `metaInfo: { title: 'TITLE_HERE' }` replacing `TITLE_HERE` with the correct title.

sharkykh: Exists on lines 12-14.
(I rebased on current develop)

sharkykh:
This is working right now, apart from the need to refresh the page after adding/editing/deleting a root dir.
This is good enough for now, because to fix that we need to convert root-dirs to Vue.

PR #4057, opened 2018-04-20
Always show "Keep" option for quality chooser on mass edit page

sharkykh:
As suggested on Slack.

For the current sceduled release.
This is basically restoring a functionality that's available in current master, and I implemented it in a wrong way when I converted the page to Vue.

**Always** show "Keep" option for quality chooser on **mass edit page**
If the qualities of the edited shows are the same, it will still select the correct quality.
But, you get the option to just select "Keep".
So if you changed a bunch of options including the quality, then changed your mind before submitting the form, you don't have to remember what quality was selected, or refresh the page.


sharkykh:
This is meant for the current sceduled release.
This is basically restoring a functionality that's available in current master, and I implemented it in a wrong way when I converted the page to Vue.

PR #4059, opened 2018-04-20
Vueify Root Dirs

sharkykh:
Move Root Dirs into a Vue component.
**New feature:** Adding/Editing a root dir now verifies that the root dir doesn't already exist.
**Changes:** 
`/api/v2/internal/existingSeries` now takes the root dir index (or comma separated indices).
Meaning it's location in the list (not including the first element - `app.ROOT_DIRS[1:]`).
Given:
```py
app.ROOT_DIRS = [
    0,
    "C:\TV",
    "C:\Anime"
]
```
For: `GET /api/v2/internal/existingSeries?rootDirs=0`
Response will be the data from `C:\TV`.

For: : `GET /api/v2/internal/existingSeries?rootDirs=0,1`
Response will be the data from `C:\TV` and `C:\Anime`.

~To be merged after #4112~


themes-default/slim/views/vue-components/root-dirs.mako
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            this.saveRootDirs();
+        },
+        setDefault() {
+            if (this.selectedDir === this.defaultDir) return;
+            this.defaultDir = this.selectedDir;
+            this.saveRootDirs();
+        },
+        saveRootDirs() {
```
sharkykh: @OmgImAlexis Should this update `MEDUSA.config.rootDirs` too?

themes-default/slim/views/vue-components/root-dirs.mako
```diff
@@ -0,0 +1,146 @@
...
+        text-align: center;
+    }
+</style>
+<script type="text/x-template" id="root-dirs-template">
+    <div id="root-dirs-wrapper">
+        <div class="root-dirs-selectbox">
+            ## @TODO: remove id attribute (use Vue $refs to access component data)
+            <select v-model="selectedDir" name="rootDir" id="rootDirs" size="6">
+                <option v-for="(curDir, index) in rootDirs" :value.number="index">{{markDefault(index) + curDir}}</option>
+            </select>
```
sharkykh: This setup is now causing me issues:
- If I use `v-model="selectedDir"` and set `:value.number="index"`, I need to add an attribute for the raw root dir path - `:data-path="curDir"`, but that means forms that use [`VueSubmitForm`](https://github.com/pymedusa/Medusa/blob/9f4e60a05106bbd6e27573c778c31f299f1568c3/themes-default/slim/static/js/vue-submit-form.js) take the value, which is the index, and not the actual path.
- If I use `v-model="selectedDir"` and set `:value="curDir"` I can't update the selection via Vue?

Am I missing something big here?

OmgImAlexis: Have you tried without value being set? `v-model` updates value so just by having the option set as selected it should fine.

themes-default/slim/views/vue-components/root-dirs.mako
OmgImAlexis: For now let's keep changes local, thinking of removing `MEDUSA` and replacing it a puex/vuex store.

themes-default/slim/views/vue-components/root-dirs.mako
sharkykh: That's exactly my point. `v-model` only works with one value. I don't remember if I tried not setting a value **with that `v-for` in place. I assume it's going to use the first variable as the value, which is the path.
That means I gotta change everything because then I can't set selected/default via the option index.

I assume I have to use something new for that - is it this?
```
computed: {
    selectedDir: {
        get () {
        },
        set (newValue) {
        }
    }
}
```

OmgImAlexis: Oh likely. You should be able that without an issue.

sharkykh:
This is now complete IMO. I want to #4112 to be merged first.

sharkykh:
To-do:
- Ultimately I don't want to send full root-dir paths to the API. Thinking about using the root dir index in the list, and send those.
- Fix conflicts & run `gulp`.

sharkykh:
I feel like this is ready for review.
I've been using this branch for a while now and it looks stable.

themes-default/slim/views/addShows_addExistingShow.mako
```diff
@@ -91,22 +70,37 @@ const startVue = () => {
...
+                        return indices;
+                    }, []);
+                if (indices.length === 0) {
                     this.dirList = [];
                     this.isLoading = false;
                     return;
                 }
-                const params = { 'root-dir': JSON.stringify(this.selectedRootDirs) };
+
+                const params = { 'root-dirs': indices.join(',') };
```
OmgImAlexis: query parameters should be camelCase

themes-default/slim/views/addShows_addExistingShow.mako
```diff
@@ -127,29 +121,21 @@ const startVue = () => {
                         .trigger("updateAll");
```
OmgImAlexis: Use single quotes.

themes-default/slim/views/addShows_addExistingShow.mako
```diff
@@ -184,7 +170,7 @@ const startVue = () => {
                         <li><app-link href="addShows/existingShows/#tabs-2">Customize Options</app-link></li>
                     </ul>
                     <div id="tabs-1" class="existingtabs">
-                        <%include file="/inc_rootDirs.mako"/>
+                        <root-dirs @update:root-dirs-value="rootDirsUpdated"></root-dirs>
```
OmgImAlexis: You should look into https://vuejs.org/v2/guide/components-custom-events.html#Binding-Native-Events-to-Components since this can be treated like an input field.

themes-default/slim/views/addShows_addExistingShow.mako
```diff
@@ -197,15 +183,15 @@ const startVue = () => {
                 <hr>
                 <p><b>Displaying folders within these directories which aren't already added to Medusa:</b></p>
                 <ul id="rootDirStaticList">
-                    <li v-for="rootDir in rootDirs" class="ui-state-default ui-corner-all" @click="toggleRootDir(rootDir)">
-                        <input type="checkbox" class="rootDirCheck" v-model.sync="selectedRootDirs" :value="rootDir" style="cursor: pointer;">
-                        <label><b style="cursor: pointer;">{{rootDir}}</b></label>
+                    <li v-for="(rootDir, idx) in rootDirs" class="ui-state-default ui-corner-all" @click="rootDirs[idx].selected = !rootDirs[idx].selected">
```
OmgImAlexis: Please don't use abbreviations. This should be using `index`.

themes-default/slim/views/addShows_addExistingShow.mako
sharkykh: I think I searched the code and there wasn't any existing endpoint that had a multi-word parameter name. I'll change it.

themes-default/slim/views/addShows_addExistingShow.mako
sharkykh: I'm only watching for changes in value, so it doesn't refresh the directory list if you change the selection in the root-dirs component - only if you change the default root-dir, or add/edit/delete a root dir.
As far as I could tell, there is no native event for that (you can't reliably detect when an option was edited, added or removed from a select?

sharkykh: I can see how this can simplify the root-dirs component if I bind a `change` event handler directly on the `<select>` element. I can do that, and looks like it is working.

However I still need a watcher for the "root-dirs-value", which is basically this: `0|C:\TV|D:\TV`

OmgImAlexis: So we'd need to provide two seperate native elements, a select and a input box?

sharkykh: An input box? Why?

sharkykh: As I said on Slack,
I'm not planning on keeping the `#rootDirText` input element.
It will be removed when `addShowOptions` is converted to Vue, so I don't see a reason to add a native event watcher on it.

But I did add `v-bind="$attrs" v-on="$listeners"`. Not sure if we're going to use it.

medariox:
What's the status here?

sharkykh:
@medariox 
I think @OmgImAlexis is good with my JS code, so maybe it just needs a review of the Python code :thinking:

medusa/server/api/v2/internal.py
```diff
@@ -53,24 +51,27 @@ def get(self, resource, path_param=None):
...
-            default_index = int(app.ROOT_DIRS[0])
-            default_root_dir = root_dirs[default_index]
-            root_dirs.remove(default_root_dir)
-            root_dirs.insert(0, default_root_dir)
-        except (IndexError, ValueError):
-            pass
+        root_dirs_indices = self.get_argument('rootDirs', '')
+
+        if root_dirs_indices:
+            root_dirs_indices = set(root_dirs_indices.split(','))
```
medariox: Not a biggie, but I think this can go just after line 66? One `if` less 🎉 
The rest of the python code LGTM.

sharkykh: Yes, looks like it to me too. Good catch! Thanks.
I will fix it as soon as I can.

PR #4070, opened 2018-04-22
Move config to route

OmgImAlexis:
I've moved config to a template and added a handler for routes.

This way we can have the templates setup and ready now instead of waiting till everything is done.

> To add a page to the router just create the component and then push to the window.routes array. Make sure to add the `template` tag and remove the `el` tag.

```js
window.routes.push({
    path: '/config',
    component
});
```


p0psicles:
Okay. But help & info isn't currently showing any info.
image

Is that expected?

PR #4096, opened 2018-04-23
Cache: Fix incorrect return value

### LINKED ISSUE #4092: [APP SUBMITTED]: TypeError: object of type 'bool' has no len()
### INFO
**Python Version**: `2.7.14 (default, Feb 23 2018, 14:28:08) [GCC 4.9.3 20150311 (prerelease)]`
**Operating System**: `Linux-4.4.15+-x86_64-with-glibc2.2.5`
**Locale**: `None`
**Database**: `44.9`
**Commit**: pymedusa/Medusa@77b20916ab577b82112ebc624f816054788c63f2
**Link to Log**: https://gist.github.com/a9e8675deaed951b2f10e04f8d5baca0
### ERROR
<pre>
2018-04-23 21:42:52 ERROR    SEARCHQUEUE-DAILY-SEARCH :: [RARBG2] :: [77b2091] DailySearchQueueItem Exception, error: object of type 'bool' has no len()
  File "/volume1/@appstore/sickbeard-custom/var/<a href="../blob/77b20916ab577b82112ebc624f816054788c63f2/SickBeard/medusa/search/queue.py#L263">SickBeard/medusa/search/queue.py</a>", line 263, in run
    found_results = search_for_needed_episodes(force=self.force)
  File "/volume1/@appstore/sickbeard-custom/var/<a href="../blob/77b20916ab577b82112ebc624f816054788c63f2/SickBeard/medusa/search/core.py#L463">SickBeard/medusa/search/core.py</a>", line 463, in search_for_needed_episodes
    cur_provider.cache.update_cache()
  File "/volume1/@appstore/sickbeard-custom/var/<a href="../blob/77b20916ab577b82112ebc624f816054788c63f2/SickBeard/medusa/tv/cache.py#L239">SickBeard/medusa/tv/cache.py</a>", line 239, in update_cache
    cache_db_con.mass_action(results)
  File "/volume1/@appstore/sickbeard-custom/var/<a href="../blob/77b20916ab577b82112ebc624f816054788c63f2/SickBeard/medusa/db.py#L203">SickBeard/medusa/db.py</a>", line 203, in mass_action
    querylist = [i for i in querylist if i is not None and len(i)]
TypeError: object of type 'bool' has no len()
</pre>
---

sharkykh:
The altered line used to not exist, as can be seen [here](https://github.com/pymedusa/Medusa/commit/18056a265f866fc0352761cccd1777e8287985ae#diff-04bdc0be0e7d1bf798758336637007b6).

As I explained [in this comment](https://github.com/pymedusa/Medusa/issues/4092#issuecomment-383708438), that function should return `None`.
Python methods that exit without a `return <VALUE>` statement, or with just a `return`, return `None`.
The commit above changed that to `False`.

This should fix #4092.

### DIFF
--- a/medusa/db.py
+++ b/medusa/db.py
@@ -218,7 +218,7 @@ class DBConnection(object):
                                 logger.log(qu[0] + " with args " + str(qu[1]), logger.DEBUG)
                             sql_results.append(self._execute(qu[0], qu[1], fetchall=fetchall))
                     self.connection.commit()
-                    logger.log(u"Transaction with " + str(len(querylist)) + u" queries executed", logger.DEBUG)
+                    logger.log(u"Transaction with " + str(len(sql_results)) + u" queries executed", logger.DEBUG)
 
                     # finished
                     break
--- a/medusa/db.py
+++ b/medusa/db.py
@@ -198,9 +198,8 @@ class DBConnection(object):
         :param fetchall: Boolean, when using a select query force returning all results
         :return: list of results
         """
-        querylist = querylist or []
-        # remove None types
-        querylist = [i for i in querylist if i is not None and len(i)]
+        # Remove Falsey types
+        querylist = (q for q in querylist or [] if q)
 
         sql_results = []
         attempt = 0
--- a/medusa/tv/cache.py
+++ b/medusa/tv/cache.py
@@ -310,7 +310,7 @@ class Cache(object):
         else:
             log.debug('The data returned from the {0} feed is incomplete,'
                       ' this result is unusable', self.provider.name)
-        return False
+        return None
 
     @property
     def updated(self):


labrys:
@sharkykh

imo you should change [L203](https://github.com/pymedusa/Medusa/blob/master/medusa/db.py#L203) from `querylist = [i for i in querylist if i is not None and len(i)]` to:
```python
querylist = [
    item for item in querylist
    if item and len(item)
]
```
This handles both None and Falsey items such as bool or empty values.  The `and len(item)` is likely unnecessary as well.

Finally you can probably improve performance by making it a generator leaving the final version:
```python
querylist = (
    item for item in querylist
    if item
)
```

sharkykh:
@labrys Thank you for the input. I agree.
I decided to keep the change from `False` to `None` as well, because it makes more sense if the `Cache._parse_item` method returns one of two types - either a string or a `None`, rather than three (`bool`), as is the case with `Cache.add_cache_entry()`.

labrys:
Since you went with a generator don't forget to change [L222](https://github.com/pymedusa/Medusa/blob/master/medusa/db.py#L222) as generators don't have a length so you can't do `len(gen)`.  I would probably opt for `len(sql_results)`.

sharkykh:
I'm gonna stick to the simple stuff for now...

labrys:
Good choice :)

PR #4100, opened 2018-04-24
app-link: Fix undefined url with empty anonRedirect

### LINKED ISSUE #3359: Empty "Anonymous redirect" results in broken links in add show
Branch: master                
Commit: b96b3ee10760ffa03a15a207841a1a2bc65798dd                 
Version: v0.1.19                 
Database: 44.8
Python Version: | 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]
SSL Version: | OpenSSL 1.0.2j  26 Sep 2016
OS: | Windows-10-10.0.14393
Locale: | nl_NL.cp1252

When no link is used in Anonymous redirect this will result in a broken link in the add show page all other pages seem to work fine.
image

image
The links point to:
- undefinedhttp://thetvdb.com/?tab=series&id=321239&lid=7
- undefinedhttp://www.tvmaze.com/shows/16579
- undefinedhttps://www.themoviedb.org/tv/69478

sharkykh:
Tested, works for me.
Fixes #3359 

@OmgImAlexis If you have any more fixes queued up for `app-link` maybe we should combine them into one PR.

### DIFF
--- a/themes-default/slim/views/vue-components/app-link.mako
+++ b/themes-default/slim/views/vue-components/app-link.mako
@@ -44,7 +44,7 @@ Vue.component('app-link', {
             const isAnonymised = url => {
                 return url.startsWith(MEDUSA.config.anonRedirect);
             };
-            const anonymise = url => MEDUSA.config.anonRedirect ? MEDUSA.config.anonRedirect + url : null;
+            const anonymise = url => MEDUSA.config.anonRedirect ? MEDUSA.config.anonRedirect + url : url;
             if (!to) {
                 if (href) {
                     // @TODO: Remove this once we move to vue only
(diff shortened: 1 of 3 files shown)


OmgImAlexis:
The one problem I have with this is it could leak users if they somehow are missing the field? Maybe we need a disable option?

sharkykh:
@OmgImAlexis If they're missing the field, they're probably missing the config, If they're missing the config, what good would a disable option do? :thinking:

OmgImAlexis:
If they disable it using a checkbox we know they don't want to use a anon redirect otherwise they could be missing the field because they pressed backspace, etc.

p0psicles:
This can be merged?

PR #4120, opened 2018-04-26
Fix /config for web root users

sharkykh:
@OmgImAlexis  According to https://router.vuejs.org/en/api/options.html#base we should use a relative path and not the full URL for the router `base`.
Tested and working for both an empty `WEB_ROOT` and a non-empty one.
Notice that I added `/` that will always be there.

@medariox @p0psicles

### DIFF
--- a/themes-default/slim/views/layouts/main.mako
+++ b/themes-default/slim/views/layouts/main.mako
@@ -134,7 +134,7 @@
             if (!window.app) {
                 console.info('Loading Vue with router since window.app is missing.');
                 const router = new vueRouter({
-                    base: '${app.WEB_ROOT}/',
+                    base: document.body.getAttribute('api-root').replace('api/v2/', ''),
                     mode: 'history',
                     routes
                 });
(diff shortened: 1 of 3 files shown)


themes-default/slim/views/layouts/main.mako
```diff
@@ -134,7 +134,7 @@
             if (!window.app) {
                 console.info('Loading Vue with router since window.app is missing.');
                 const router = new vueRouter({
-                    base: document.getElementsByTagName('base')[0].getAttribute('href'),
+                    base: '${app.WEB_ROOT}/',
```
OmgImAlexis: Shouldn't adding python... Let's just fix this in js.
```
Try using `window.baseUrl.replace('/api/v2', '')`
```

OmgImAlexis: That may work, but that's not what I was going for.baseUrl is the relative baseUrl to the domain.

PR #4129, opened 2018-04-28
convert timeago to vue

OmgImAlexis:
This should replace the need for the other timeago plugin we have but I can't remove it directly at the moment as it's in the `vender.js`.


sharkykh:
What is the new `/js/lib/timeago.js`?
I would like to have some sort list for the standalone JS libs and their versions...

OmgImAlexis:
@sharkykh that's https://raw.githubusercontent.com/hustcc/timeago.js/master/dist/timeago.min.js

https://github.com/hustcc/timeago.js

sharkykh:
Right now on my front page sorting is broken. Should be an easy fix.
image

themes-default/slim/views/partials/schedule/list.mako
```diff
@@ -46,11 +46,11 @@
         <tr class="${show_div}">
             <td align="center" nowrap="nowrap" class="triggerhighlight">
                 <% airDate = sbdatetime.sbdatetime.convert_to_setting(cur_result['localtime']) %>
-                <time datetime="${airDate.isoformat('T')}" class="date">${sbdatetime.sbdatetime.sbfdatetime(airDate)}</time>
+                <time datetime="${airDate.isoformat('T')}" title="${airDate.isoformat('T')}" class="date">{{'${airDate.isoformat('T')}' | formatDate}}</time>
             </td>
             <td align="center" nowrap="nowrap" class="triggerhighlight">
                 <% ends = sbdatetime.sbdatetime.convert_to_setting(cur_ep_enddate) %>
-                <time datetime="${ends.isoformat('T')}" class="date">${sbdatetime.sbdatetime.sbfdatetime(ends)}</time>
+                <time datetime="${airDate.isoformat('T')}" title="${airDate.isoformat('T')}" class="date">{{'${airDate.isoformat('T')}' | formatDate}}</time>
```
sharkykh: You changed `ends` to `airDate`, it's not using the correct date-time ATM.

themes-default/slim/views/home.mako
```diff
@@ -21,6 +21,100 @@ const startVue = () => {
             return {
                 header: 'Show List'
             };
+        },
+        mounted() {
+            $('#showListTableSeries:has(tbody tr), #showListTableAnime:has(tbody tr)').tablesorter({
```
OmgImAlexis: @sharkykh I tried using `nextTick` here to prevent the issue you're talking about but I could get get it working without `nextTick` so I don't know what todo now.

themes-default/slim/views/layouts/main.mako
```diff
@@ -129,6 +130,14 @@
         <%include file="/vue-components/quality-chooser.mako"/>
         <%include file="/vue-components/language-select.mako"/>
         <script>window.routes = [];</script>
+        <script>
+            Vue.filter('formatDate', str => {
+                if (MEDUSA.config.fuzzyDating) {
+                    return timeago().format(str);
+                }
+                return (new Date(str)).toLocaleDateString();
```
sharkykh: I guess this is fine for now, but this needs to use the user preference.

themes-default/slim/views/history.mako
```diff
@@ -220,9 +219,8 @@ const startVue = () => {
                 % for hItem in compactResults:
                     <tr>
                         <td align="center" class="triggerhighlight">
-                            <% airDate = sbdatetime.sbfdatetime(datetime.strptime(str(hItem.actions[0].date), History.date_format), show_seconds=True) %>
                             <% isoDate = datetime.strptime(str(hItem.actions[0].date), History.date_format).isoformat('T') %>
-                            <time datetime="${isoDate}" class="date">${airDate}</time>
+                            <time datetime="${isoDate}" title="${isoDate}" class="date">{{'${isoDate}' | formatDate}}</time>
```
sharkykh: Tooltips should show date-times in a nice format IMO.

themes-default/slim/views/home.mako
sharkykh: `textExtraction` seems outdated for it anyway. See below.

themes-default/slim/views/home.mako
```diff
@@ -21,6 +21,100 @@ const startVue = () => {
...
+                debug: false,
+                sortList: [[7, 1], [2, 0]],
+                textExtraction: (function() {
+                    return {
+                        0: node => $(node).find('span').attr('timestamp'),
+                        1: node => $(node).find('span').attr('timestamp'),
```
sharkykh: I got it working by changing these two to:
```js
                        0: node => $(node).find('span').attr('timestamp'),
                        1: node => $(node).find('span').attr('timestamp'),
```

themes-default/slim/views/history.mako
OmgImAlexis: The old timeago did exactly what I've got here so I didn't change it over.

themes-default/slim/views/home.mako
OmgImAlexis: Oh... I forgot I had thought about using a span with timestamp but then went with the old format.

themes-default/slim/views/history.mako
sharkykh: As we discussed, should be using a string representation of the date-time.

themes-default/slim/views/layouts/main.mako
OmgImAlexis: What user preference I didn't know we had a format for dates, if we do is it available in the API?

OmgImAlexis: Oh is that timepreset? Something like this? `"%I:%M %p"`

PR #4056, opened 2018-04-20
Feature/utorrent dest folder

p0psicles:
- [ ] PR is based on the DEVELOP branch
- [ ] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [ ] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

Rebased this locally from #4005 and pushed as new branch.
@rafi-d i'll try to give you push access to this branch.

PR #4133, opened 2018-04-28
Develop

rafi-d:
Improving efficiency of search-results analysis by skipping redundant checks:
Once "best result" is equal to "current result" (best_result = cur_result), there is no point in doing any more "research" on two identical strings....


rafi-d:
A side note: 
The current logic seems to not handle the use cases where "preferred_words" and "undesired words" are *both* present.  When two releases do have preferred words, it should give priority to the one that has NO undesired words.

rafi-d:
Hi,
I've added a new file: core_min_size_filter.py file, just for review (as requested by you couple of weeks ago). It is unrelated to this specific PR. Feel fre to move it to another branch if you like.

This includes extra minimal code for a proposed new feature - for the user to be able to specify a min file size for his downloads. 

The goal was to:
- Simplify/automate as much as possible:
* Five only one new UI setting - the minimum files size for a 45 min episode @1080p
* the logic will normalize  the relevant sizes for other qualities and compressions
* the user file-size setting can be global  only or also per series
- This will not effect any of the current filters it set as disable/0 (default)
- UI was not changed. I im plemented  a hard-coded size, for testing - 1.2G

I am attaching a draft-UI required, and a test log . I am not knowledgeable enough to make those changes. 

I believe this can be very useful to avoid random small files creeping in.  Max size filter can be added too.

Thanks

Rafi

min-file-size_feature
[Min-file-size_feature_test-log.txt](https://github.com/pymedusa/Medusa/files/1958097/Min-file-size_feature_test-log.txt)

p0psicles:
In not against a feature like this. But it has to be allot more refined and flexible then this for it to be considered mergeable.

Like having a central config for defining qualities icw file sizes.

But the biggest problem is our overal quality configuration. Its dated and needs a massive rewrite before it can be extended. There are already multiple issues with it, which are hard to be addressed. 

Adding more complexity to it, will exponentially increase issues related to it. That why where hesitant to add anything like this.

p0psicles:
If your really determined to imorove the quality system, where more then happy to send you a slack invite. But then be prepared to invest allot of time. As your touching upon one of the most complex parts of the code.

rafi-d:
Hi,
There are two separate things here:

1. In core.py - just some  fixes to if statement. I believe these should just be fixed/merged

2. The proposed new (and minimal) file-size filter - which is for your review/discretion. I have been using it here for a couple of weeks  now, and it looks helpful. Also "harmless to he existing quality filters, either good or bad... :) 

Sorry, I will not be able to invest a lot of time in overall re-writes, but am using/aware of what others have (like sonarr), and this is my suggestion for implementing something similar,  but with with minimal effort, that can be useful to many users.

So - up to you. I will not be offended either way... ;) 

Rafi

PS: My other PR #4056   is  pending your review. I hope you can find the time to merge it.

OmgImAlexis:
If we're going to add the file size limits, etc. this then it really should be added on the quality itself as that is where it'll matter.

Saying I only want a 1GB file on every episode doesn't really mean much and if anything we have more users wanting a hard top limit not a bottom limit.

OmgImAlexis:
Just putting this out there but I'm not approving anything that auto-normalises the sizes based on a single input. There are way too many variables and if we're going to do this it needs to be done right the first time not just by guessing.

rafi-d:
PS: 
If you want to try it, I can easily add a max size limit too, plus, two temporary inputs as pewdefined/reserved  words: %MINSIZE=yyy and %MAXSIZE=yyy in the "Ignore words" edit box.

rafi-d:
This is what sonarr has. I don't see a point in setting 20 numbers when one needs only one or two max per his own show... 
sonarr_sizes

OmgImAlexis:
@rafi-d we don't expect users to change all the settings we have right now and yet we still have them. The setting per quality will allow this idea to actually work.

If you want to add something based on Sonarr's quality system then I'd be all for that.

OmgImAlexis:
If you want to send me an email at xo@wvvw.me I'll add you to our Slack and we can discuss. It'll be easier than using issues.

> In this case, why not start like I suggested (simple, just two min-max settings, BUT also per-show), see the feedback, and decide then? 

The main issue I have with this is people tend to fade away and not stick around to maintain features, we also have a lot of other code that'll need to be drastically changed to handle all of this. I'm sure there are loads of edge cases that you have missed that we'll be hunting for the year or two if it is meged. This is why I want to full discuss anything that changes the quality or show setting.

I'm sure @p0psicles is in the same mindset too.

The idea I liked and most users have wanted is where you can have different sizes per quality profile as well as adding your own quality profiles. I think 3. would be able to replace what you have now; As in we could ship with default that the user could change.

I can't see 5. being added at all since it's not as simple as you're saying. What if a show is releases as a double but listed as a single? Does that mean it'll skip it now since it's too large and too long? Also what about 10 bit vs 8 bit HEVC? I've been up to a 5GB difference between files that're 8 and 10 bit. Same goes for `5.1` vs `2.1` again on a HEVC I've been 500MB+ changes on some because of the audio.

I think we should start off just moving out current qualities to using the size per quality and then later on we can add other restrictions to create what Sonarr calls "profiles".

rafi-d:
(sent  by email)
I am aware of audio,10vs8 bit, double episodes (which sonarr solved by auto-doubling the size of SxE1-s) . I believe such variations cannot be resolved automatically based on names only, and mediainfo is not available (and too late…). So, as long as the user has control over the size-numbers and the preferred “words” – he can adjust as best he can per his own use-case/show-selection.

Once he will do that for his  selected qualities/compression etc (as today, no change), the code can help by normalizing for the rest (if enabled), but I really do not expected to it  to happen (like using other qualities for that show). No one likes that… 

My idea is to keep it simple, size-per one quality per show – the one selected, and not to fear optional auto-factoring. Yes, you can give 10x2 global numbers as defaults (2 per current quality), tho I think two are enough to start with with a single quality. I believe an option to also automatically re-factor when one is being changed can be useful too.

rafi-d:
Hi,
Just updated here with my new design/logic (all done in core.py for now to make it easy to review/track).  
This is running fine here for some time now. It IS following your restrictions of:
- Wait for vue, no GUI changes...
- Every user can also define now per-show filter (not only a GLOBAL/general one) , which will limit exactly to the file-sizes he is used to download. 

I suggest to consider merging it (with some logging reductions)  so to get users' feedback, and enhance it when vue will come (in a year or so?)... Why wait if this can be of help to others.

Smart-file-size-filter
======================

A. Requirements
---------------
1. Backwods compatieble: if no limits-/keywords are being used - ignore this filter
2. Minimal GUI usage: Since migration to vue is on-going and expected to be so for a long time.
3. Future enhance-able to a more user-controlled - per-quality size-limits

B. Concept
----------
1. Needs minimal user input to start with- can use only two keywords to define the min/max limits (globally, @1080p/45min/x264)). 
   Will be  auto-scaled for all qualities, and all runtimes
2. On the other hand - can be fully user definable. Users can define their exact size requirement(s) at the show(s) level.
   The user will reference here the exact size he is customed to download for that show.
3. Code changes - kept to minimum - one file (for now).

C. Design/Logic
---------------
1. Use %MINSIZE= %MAXSIZE=  "Ignored" words to specify the size limits
2. Sizes are in MBytes (example: %MINSIZE=1500 is 1.5GB)
3. Add a log of the result(s) for user reference 
4. [optional: color code the sizes in manual search. Not implemented]
5. Logic:

 During search/pick_best_result():
                                        - try find  min/maxsize keyword in current show's Ignore Words list.
                                         -> if found - use those size-limits as is for the found results
                                         -> if not found - try find the min/maxsize keywords in the global Ignore Words list
                                           -> if found - scale up/down per quality, runtime and compression
                                           -> if not found - no size limit is used
6. New class/Data Structures:
class ExtendedQuality(Quality)
-> QualityScreenPixels = NumDict - screen sizes used for scaling
-> QualityFileSizes = NumDict - [min,max] touple per quality - for future use with GUI

D. Documentation/guide
-----------------------
1. Instructions in news/changelog (optional link to wiki page)
2. Recommended: couple of texts additions in Ignored-words settings screens.

wallopthecat:
This looks great and is something I miss from Sonarr

PR #4153, opened 2018-05-01
Only retrieve release groups from anidb for anime shows.

p0psicles:

### DIFF
--- a/medusa/tv/series.py
+++ b/medusa/tv/series.py
@@ -2095,9 +2095,11 @@ class Series(TV):
         data['config']['defaultEpisodeStatus'] = self.default_ep_status_name
         data['config']['aliases'] = self.aliases
         data['config']['release'] = NonEmptyDict()
-        data['config']['release']['blacklist'] = bw_list.blacklist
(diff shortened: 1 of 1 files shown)


medusa/tv/series.py
```diff
@@ -2095,9 +2095,11 @@ def to_json(self, detailed=True):
         data['config']['defaultEpisodeStatus'] = self.default_ep_status_name
         data['config']['aliases'] = self.aliases
         data['config']['release'] = NonEmptyDict()
-        data['config']['release']['blacklist'] = bw_list.blacklist
-        data['config']['release']['whitelist'] = bw_list.whitelist
-        data['config']['release']['allgroups'] = get_release_groups_for_anime(self.name)
+        # These are for now considered anime-only options, as they query anidb for available release groups.
+        if self.is_anime:
```
OmgImAlexis: Won't this make it so it's missing the key blacklist instead of just setting it to be an empty array?

medariox: Wouldn't that represent the current situation? Only anime shows have bl/wl.

PR #4157, opened 2018-05-03
Feature/fix setting status back to wanted

### LINKED ISSUE #4156: When setting status back to 'WANTED' quality of the existing file get's lost
### Before submitting your issue:

Enable debug logging in Medusa settings, reproduce the error (be sure to disable after the bug is fixed)

**Branch/Commit:**
**OS:**
**What you did:**
1. Downloaded an episode, but it was a bad release.
2. Changed it's status to wanted
3. Started a backlog

**What happened:**
All results were ignored because the existing quality was 'UNKNOWN'.
**What you expected:**
That it would replace the existing file.
**Logs:**
```
2018-05-03 06:14:55 DEBUG    SEARCHQUEUE-BACKLOG-284131 :: [NZBGeek] :: [781bfd2] 284131: 'The Seven Deadly Sins' S02E08 status is: 'SNATCHED (UNKNOWN)'. Ignoring result with quality '720p HDTV'. Reason: Existing quality is UNKNOWN. Ignoring new quality (edited)
```

Imo there are three issues here:
1. When you reset the status of an ep to wanted, you overwrite the status to 3. With that, you lose the combined status info. And thus lose the original quality.
2. When it can't detect an existing quality, it won't replace. I still don't understand the reasoning behind this.
3. It's not using `knowit` to detect the quality. Where in my case, this would also have resolved the issue.

p0psicles:
### ~Currently it will break the quality system. As we have a limited set of combinable qualities / statuses. And with this change, it technically enables you to set every combination.~

Fix #4156


medusa/tv/series.py
```diff
@@ -1377,6 +1377,14 @@ def make_ep_from_file(self, filepath):
                     continue
 
             else:
+                # If the episodes status is set to archive, leave it alone.
```
p0psicles: It's still a hypothesis, but I have a suspicion that is the reason that sometime the episode status switches from archived back to a downloaded status. And which will result it to be available for "upgrades".

I think it has something todo, with when something on the filesystem changes. Like maybe the file itself. And then it can be evaluated by the should_refresh method, which can also change it's status.

Shortcutting it, maybe not fix the the root cause. But I think it wil realy prevent this from happening again.

@fernandog @medariox can you think of situations where we don't want to shortcut this when the status is ARCHIVED?

p0psicles:
We now have these weird checks, where we have quality overview's which consist of a status + every possible quality.
https://github.com/pymedusa/Medusa/blob/feature/fix-setting-status-back-to-wanted/medusa/search/core.py#L223

like:
`if curEpObj.status not in Quality.DOWNLOADED:`
And then in medusa.common, Quality.DOWNLOADED is build up like this:
`Quality.DOWNLOADED = [Quality.composite_status(DOWNLOADED, x) for x in Quality.qualityStrings if x is not None]`

We could also just compare the status, instead of doing the `in`. Like:
`if curEpObj.composite_status not in common.DOWNLOADED`

Any reason not to change this?

p0psicles:
This will now introduce a new status in displayShow.
image

You can now have a status "WANTED" + a quality. The quality is for the existing quality before the episode was changed to WANTED.

p0psicles:
This happened last night. But I honestly do not know where this comes from.
Where did it get the key `3276803` from?

https://github.com/pymedusa/Medusa/blob/feature/fix-setting-status-back-to-wanted/medusa/search/daily.py#L99

p0psicles:
Scratch this PR. It can't cope with some qualities having a false-ish value like -1 for UNKNOWN. We'll have to split them anyway.

medusa/tv/episode.py
```diff
@@ -1970,3 +2018,66 @@ def airdate_modify_stamp(self):
...
+                {'name': self.series.name, 'filepath': filepath}
+            )
+
+            new_quality = Quality.name_quality(filepath, self.series.is_anime)
+
+            if old_status in (SNATCHED, SNATCHED_PROPER, SNATCHED_BEST) or (
+                    old_status == DOWNLOADED and old_location):
+                new_status = DOWNLOADED
+            else:
+                new_status = ARCHIVED
```
sharkykh: This means everything else will be tagged as `Archived` when refreshing a show,
meaning episodes with `Wanted` or `Failed` status and a file path would get tagged as `Archived`.
That is something to consider.
@medariox and I tried to wrap our heads around it, but ultimately I think it's just a decision that needs to be made...

medusa/tv/series.py
```diff
@@ -1323,8 +1271,8 @@ def make_ep_from_file(self, filepath):
         :return:
         :rtype: Episode
         """
-        if not os.path.isfile(filepath):
-            log.info(u"{indexer_id}: That isn't even a real file dude... {filepath}",
+        if not (os.path.isfile(filepath) and helpers.is_media_file(filepath)):
```
sharkykh: Don't know if the `is_media_file` check is really needed, 
since `Series.load_episodes_from_dir` only lists media files?
https://github.com/pymedusa/Medusa/blob/abdd3b6d2008a23c4069629ed68561572184289c/medusa/tv/series.py#L969

medusa/tv/episode.py
```diff
@@ -403,6 +408,49 @@ def air_date(self):
...
+
+    @property
+    def splitted_status(self):
+        """Return the existing status removing the quality from it."""
+        return Quality.split_composite_status(self._status)
+
+    @property
+    def splitted_status_status(self):
+        """Return the status from the status/quality composite."""
+        return self.splitted_status.status
```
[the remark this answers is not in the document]
p0psicles: As this is a rather safe move for now. The bigger refactor will happen when we split quality and status in the db.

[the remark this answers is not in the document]
p0psicles: Bit we can change the names. But id rather not refactor it more then ehat weve currently have

tests/test_update_status.py
```diff
@@ -78,36 +78,53 @@ def create(filepath, status, size):
...
+    },
+    {  # p17: Previous status was Wanted
+        'status': Quality.composite_status(WANTED, Quality.FULLHDTV),
+        'location': 'Show.S01E17.1080p.HDTV.X264-GROUP.mkv',
+        'filepath': 'Show.S01E17.720p.HDTV.X264-GROUP.mkv',
+        'new_size': 38,
+        'expected': Quality.composite_status(ARCHIVED, Quality.HDTV)
+    },
 ])
 def test_should_refresh(p, create_episode, create_file):
```
sharkykh: `test_update_status`

tests/test_update_status.py
```diff
@@ -0,0 +1,143 @@
...
+    },
+    {  # p17: Previous status was Wanted
+        'status': Quality.composite_status(WANTED, Quality.FULLHDTV),
+        'location': 'Show.S01E17.1080p.HDTV.X264-GROUP.mkv',
+        'filepath': 'Show.S01E17.720p.HDTV.X264-GROUP.mkv',
+        'new_size': 38,
+        'expected': Quality.composite_status(ARCHIVED, Quality.HDTV)
+    },
+])
+def test_should_refresh(p, create_episode, create_file):
```
sharkykh: `test_update_status`

medusa/tv/series.py
medariox: You are right, but since want to keep separating the logic, I thought we could keep it. It is coming from the old `should_refresh_file`: https://github.com/pymedusa/Medusa/pull/4157/files#diff-01e0cabaabc8a78fb76ee292b696e98aL1282

medusa/tv/episode.py
medariox: What should be added is: this situation can only happen if the file was `manually` changed. Just triggering a refresh of the show won't change its status. Also now the status goes from WANTED to ARCHIVED only if an episode was already previously present. Related tests: https://github.com/pymedusa/Medusa/pull/4157/files#diff-fe5305567785e70fe5a0a06240cbcb5fR24

medusa/common.py
```diff
@@ -867,27 +878,27 @@ def __missing__(self, key):
         :raise KeyError: if the key is invalid and can't be determined from Quality
         """
         # convert key to number
-        key = self.numeric(key)  # raises KeyError if it can't
-        if key in self.qualities:  # if key isn't found check in qualities
-            current = Quality.split_composite_status(key)
+        key = int(key)  # raises KeyError if it can't
```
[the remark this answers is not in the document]
medariox: This wasn't meant to mimic the old `NumDict` in any way. The comments were old and misleading. I have updated the StatusStrings code accordingly.

p0psicles:
Whats the status?

PR #4154, opened 2018-05-01
Remove the plus button for adding new items to the list.

p0psicles:
Items are now saved onChange event.

- [ ] PR is based on the DEVELOP branch
- [ ] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [ ] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

PR #4163, opened 2018-05-05
Update editShow UI (select-list, releasegroups, link back to displayShow)

sharkykh:
- Fix javascript error on non-anime shows
- Update `select-list` to make it clear that you have to click the plus button, and make it more intuitive.
  Supersedes #4154 

- Add link back to `displayShow`. The method of getting that link is temporary. When we're using vue-router we should be able to fix that.
image
image


p0psicles:
Yes lets go with this. Ill close the other pr.

themes-default/slim/views/editShow.mako
```diff
@@ -146,6 +146,11 @@ const startVue = () => {
                 const preferred = this.series.config.qualities.preferred.reduce(reducer, 0);
 
                 return allowed | preferred << 16
+            },
+            displayShowUrl() {
+                // @TODO: Change the URL generation to use `this.series`. Currently not possible because
+                // the values are not available at the time of app-link component creation.
+                return window.location.pathname.replace('editShow', 'displayShow') + window.location.search;
```
sharkykh: I don't like it, but this is the best way to **avoid** having to add a watch to the `app-link` component (like I did with `quality-chooser`, `file-browser`, `select-list`, etc.)

PR #4191, opened 2018-05-12
Fix show link obstructed by poster link (Fixes #4060)

### LINKED ISSUE #4060: Show title link is obstructed by the poster link
**Branch/Commit:** develop / 47cfea0b73d078a0865c297c392a8f879463302c
The Poster and 'show/hide specials' links are being tagged as false-links in the `app-link` component.
These issues are caused by the combination of Vue and [core.js#L36-L50](https://github.com/pymedusa/Medusa/blob/47cfea0b73d078a0865c297c392a8f879463302c/themes-default/slim/static/js/core.js#L36-L50).
- The link on the show name is partly 'covered' by the poster link (about the width of the poster), making it unclickable. Maybe it would be better to add an option to the `asset` component that will add an asset app-link wrapper to the asset?
  When this is fixed, the banner is missing a similar asset link, if that wasn't intended.
- The show/hide specials link is getting moved to a new line because of `false-link` css

sharkykh:
- Fix show title link being obstructed by poster link - Fixes #4060 [1]
- Add link to the banner too
- Fix Show/Hide specials link being tagged as a false link - don't use `app-link` - Fixes #4060 [2]
- Fix Season Jump offset:
  Needed to increase offset from target due to changes in the navbar
  (two rows instead of one for a total of ~90px)
- Move the Season Jump select down a bit.
- Move the first season table down a bit:
  The header's background was overlapping with the show header. It just looked wrong.


themes/dark/templates/displayShow.mako
```diff
@@ -35,7 +35,7 @@ const startVue = () => {
 <%include file="/partials/showheader.mako"/>
 
 <div class="row">
-    <div class="col-md-12 horizontal-scroll">
+    <div class="col-md-12 horizontal-scroll" style="top: 12px">
```
p0psicles: Cant use a class for this one?

sharkykh: I don't see a reason to make a class for this little thing which is used once...

PR #4193, opened 2018-05-13
Remove shutil_custom

medariox:

I see no reason to keep using it until we have a confirmed bug - otherwise, this can only do more harm than good.


labrys:
@medariox This is meant to get around (as much as possible) the following restrictions of using shutil to copy files:

>On POSIX platforms, this means that file owner and group are lost as well as ACLs. On Mac OS, the resource fork and other metadata are not used. This means that resources will be lost and file type and creator codes will not be correct. On Windows, file owners, ACLs and alternate data streams are not copied.

medariox:
@labrys You mean the custom shutil? In that case, why isn't shutil.copy2() used instead?

labrys:
Sorry missed a line in the quote.. that message starts:

> Warning Even the higher-level file copying functions (shutil.copy(), shutil.copy2()) can’t copy all file metadata.

labrys:
Also the existing shutil custom is poorly implemented due to possible race conditions.

labrys:
A safer version of copyfile_custom would be the following:

```python
import os
import stat

from shutil import Error, _samefile
try:
   from shutil import SpecialFileError
except ImportError:
   from shutil import Error as SpecialFileError

try:
    O_BINARY = os.O_BINARY  # Windows
except AttributeError:
    O_BINARY = 0


READ_FLAGS = os.O_RDONLY | O_BINARY
WRITE_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_TRUNC | O_BINARY
BUFFER_SIZE = 128*1024


def copyfile_custom(src, dst):
    """Copy data from src to dst."""

    def special_file(fn):
        try:
            st = os.stat(fn)
        except OSError:  # File most likely does not exist
            pass
        else:
            if stat.S_ISFIFO(st.st_mode):
                raise SpecialFileError("`%s` is a named pipe" % fn)

    with os.open(src, READ_FLAGS) as fin, os.open(dst, WRITE_FLAGS) as fout:
        if _samefile(src, dst):
            raise Error("`%s` and `%s` are the same file" % (src, dst))

        special_file(src)
        special_file(dst)

        while True:
            data = fin.read(BUFFER_SIZE)
            if data == '':
                break
            fout.write(data)
```

This streamlines the imports, defines the constants a single time only, uses context managers to ensure files get closed properly, and moves the same file and special file checks as close as possible to the actual writing (i.e. after the file operation has begun) to minimize race conditions.

medariox:
Thanks @labrys, this looks like a good improvement. I'm still unsure about the advantages of this over `copy2` and if there are really any after all.

labrys:
I also have my doubts that this accomplished anything with the metadata.  To really copy the metadata you would need to also use shutil.copymode and shutil.copystat and possibly other custom methods such as those in pywin32 to copy as much as possible.  As for using shutil.copy2 it doesn't have a way to set buffer size which can make it slow (in theory, I have not tested it).

Issue #4141, opened 2018-04-29
General Update Kodi button not working


Rouzax:
Medusa Info: | Branch: master                 Commit: 77b20916ab577b82112ebc624f816054788c63f2                 Version: v0.2.2                 Database: 44.9
-- | --
Python Version: | 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:25:58) [MSC v.1500 64 bit (AMD64)]
SSL Version: | OpenSSL 1.0.2k  26 Jan 2017
OS: | Windows-10-10.0.14393
Locale: | nl_NL.cp1252

The general `Update Kodi` menu item does not work any more.
Link goes to: `https://xxx.yyy/home/updateKODI/`
![image](https://user-images.githubusercontent.com/4103090/39408794-191728fe-4bdc-11e8-947c-fb8fe22ed239.png)

```
Error
Invalid show ID 
```

Naffles:
Is this the same as the error i get when i try to update Emby for an individual show? Error is: 

WARNING  Thread_18 :: [77b2091] EMBY: Provider unknown

There is no corresponding Emby log at all so nothing is being sent to the emby server.

sharkykh:
@Naffles No, that's a different issue.
It seems like EMBY only supports using TVDB ID to identify a show.
We'll have to do something else to fix that.

Naffles:
That's interesting... I use TMDB as default anyway (option set in settings) ... so shouldn't it be sending that through?

Naffles:
Sorry misread. TMDB vs TVDB... Is this going to be hard to fix? I'm surprised the issue hasn't come up previously - does no-one use Medusa/Emby is combo?

sharkykh:
As I said, **EMBY** doesn't seem to support updating the TV library using indexers other than TVDB.
I'm working on a workaround for that.

Naffles:
Can confirm that this is now working... however...

When updating a specific show there is no INFO level log that it has been done, only DEBUG. I initially thought it still wasn't working as there was nothing in the standard log - is it possible to log an additional INFO level notification that it has occurred? 

If not no biggie - like i said it looks to be working now... although it is also a bit strange that there is nothing after the for colons (i.e. no post data or HTTP response).

MEDUSA:
2018-05-21 17:42:34 DEBUG    Thread_16 :: [212cd1c] EMBY: HTTP response: 
[...]

PR #4199, opened 2018-05-14
Update Emby Notifier

sharkykh:
- Add support for indexers other than TVDB by using the show's external ids.
  (Emby's `/Library/Series/Updated` API only supports TVDB IDs) - Should fix https://github.com/pymedusa/Medusa/issues/4141#issuecomment-388614665
- Use `MedusaSession` instead of `urllib`
- Fix flake8 docstring warnings
- Remove `u` prefix from strings.

### DIFF
--- a/medusa/notifiers/emby.py
+++ b/medusa/notifiers/emby.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 
+"""Emby notifier module."""
 from __future__ import unicode_literals
 
 import json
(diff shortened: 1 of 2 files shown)


sharkykh:
@Naffles
I would appreciate if you could test this.
Both with the "Test" function on the config page, and by sending a specific show update.

Naffles:
Legend! I'm Aussie so walking to work atm. Will test when home tonight. Cheers

Naffles:
Can confirm working now working :) Thanks legends!

PR #4221, opened 2018-05-19
Split composite status into status and quality, add Unknown as valid quality

p0psicles:
This PR contains major changes to our quality system.
1. It splits the composite qualities into status (`Downloaded`, `Skipped`, etc.) and quality (`HDTV`, `720p WEB-DL`, etc.). This allows us to remove a lot of superfluous code and will make a better quality system possible in the future.
2. Existing qualities are being shifted by one spot. This means that we can use `Unknown` as a valid quality (the lowest available), thus eliminating a bunch of annoying hacks and allowing it to be set even as Preferred (e.g. for `x265` content)

It goes without saying that these are massive changes that should be tested thoroughly over a minimum period of time. It would be great if anyone could give this a spin in the next couple of days. The upgrade process is automated and backup files for `main.db`, `failed.db` and `cache.db` will be created.

---


p0psicles:
Pff this will require allot of testing.

p0psicles:
@medariox with the latest commits, it's now running somewhat stable. But the history table in failed.db should also have the qualities shifted. not?

medusa/common.py
```diff
@@ -507,16 +508,16 @@ def scene_quality_from_name(name, quality):
...
+        if quality == Quality.SDDVD:
             rel_type = ' BDRip'
             if re.search(r'br(-| |\.)?(rip|mux)', name):
                 rel_type = ' BRRip'
             elif re.search(r'dvd(-| |\.)?(rip|mux)', name):
                 rel_type = ' DVDRip'
 
         # If any WEB type
-        if quality in (32, 64, 1024, 8192):
+        if quality in (Quality.HDWEBDL, Quality.FULLHDWEBDL, Quality.UHD_4K_WEBDL,
```
p0psicles: Shouldnt we create groups for these?

medariox: I don't see it as necessary.

sharkykh: My opinion is we should, **but only if** it's needed in more than one or two places.

OmgImAlexis:
Just adding a note here. @sharkykh if we don't already we'll need to look at adding an endpoint that gives us the current quality, status, etc. settings.

I know we can hardcode for now but once we get to adding custom qualities and such it'll be worth it.

sharkykh:
@OmgImAlexis 
> an endpoint that gives us the current quality, status, etc. settings.

... of what?
A specific show? (`GET /api/v2/series/:slug` should have everything you need)
Global defaults? (will do when converting  `inc_addShowOptions.mako` to Vue)

Or do you mean the values from the `medusa/common.py` file?
An endpoint that just returns possible values like status IDs and names, quality IDs and names , etc.?

medusa/show/history.py
```diff
@@ -110,18 +110,18 @@ def _get_actions(action):
...
         return result or []
 
-    action_fields = ('action', 'provider', 'resource', 'date', 'proper_tags', 'manually_searched')
+    action_fields = ('action', 'quality', 'provider', 'resource', 'date', 'proper_tags', 'manually_searched')
     # A specific action from history
     Action = namedtuple('Action', action_fields)
     Action.width = len(action_fields)
 
-    index_fields = ('indexer_id', 'show_id', 'season', 'episode', 'quality')
+    index_fields = ('indexer_id', 'show_id', 'season', 'episode')
```
sharkykh: As per the discussion on Slack, I think this is part of what affects the compact layout.

I'm against this change being made right now. I disagree with waiting for the Vue conversion to change the way it is displayed with this branch.
Please keep it as it is now on `develop`.

medusa/history.py
```diff
@@ -104,10 +99,10 @@ def log_download(ep_obj, filename, new_ep_quality, release_group=None, version=-
 
     action = ep_obj.status
 
-    _logHistoryItem(action, ep_obj, quality, filename, provider, version, size=size)
+    _log_history_item(action, ep_obj, filename, provider, version, size=size)
 
 
-def logSubtitle(ep_obj, status, subtitle_result):
+def logSubtitle(ep_obj, subtitle_result):
```
sharkykh: Might as well fix the name to snake_case.

medusa/subtitles.py
```diff
@@ -493,7 +492,7 @@ def save_subs(tv_episode, video, found_subtitles, video_path=None):
         if app.SUBTITLES_HISTORY:
             logger.debug(u'Logging to history downloaded subtitle from provider %s and language %s',
                          subtitle.provider_name, subtitle.language.opensubtitles)
-            history.logSubtitle(tv_episode, status, subtitle)
+            history.logSubtitle(tv_episode, subtitle)
```
sharkykh: This is the only use case of `history.logSubtitle`, might as well fix the name to snake_case.

medusa/subtitles.py
```diff
@@ -993,7 +992,7 @@ def dhm(td):
                 "WHERE "
                 "s.subtitles = 1 "
                 "AND s.paused = 0 "
-                "AND e.status LIKE '%4' "
+                "AND e.status = '4' "
```
sharkykh: Please use `DOWNLOADED` (update import)

medusa/trakt_checker.py
```diff
@@ -223,17 +223,17 @@ def remove_episode_trakt_collection(self, filter_show=None):
...
             params = []
             main_db_con = db.DBConnection()
-            selection_status = ['?' for _ in Quality.DOWNLOADED + Quality.ARCHIVED]
+            status = [DOWNLOADED, ARCHIVED]
             sql_selection = b'SELECT s.indexer, s.startyear, s.indexer_id, s.show_name,' \
                             b'e.season, e.episode, e.status ' \
                             b'FROM tv_episodes AS e, tv_shows AS s WHERE e.indexer = s.indexer AND ' \
                             b's.indexer_id = e.showid and e.location = "" ' \
-                            b'AND e.status in ({0})'.format(','.join(selection_status))
+                            b'AND e.status in ({0})'.format(','.join(['?']*len(status)))
```
sharkykh: I think the coding style is `['?'] * len(statuses)`

medusa/trakt_checker.py
```diff
@@ -223,17 +223,17 @@ def remove_episode_trakt_collection(self, filter_show=None):
 
             params = []
             main_db_con = db.DBConnection()
-            selection_status = ['?' for _ in Quality.DOWNLOADED + Quality.ARCHIVED]
+            status = [DOWNLOADED, ARCHIVED]
```
sharkykh: The name `statuses` is more ideal IMO.

medusa/trakt_checker.py
```diff
@@ -274,13 +274,13 @@ def add_episode_trakt_collection(self):
         if app.TRAKT_SYNC and app.USE_TRAKT:
 
             main_db_con = db.DBConnection()
-            selection_status = ['?' for _ in Quality.DOWNLOADED + Quality.ARCHIVED]
+            status = [DOWNLOADED, ARCHIVED]
```
sharkykh: Same as above.

medusa/trakt_checker.py
```diff
@@ -274,13 +274,13 @@ def add_episode_trakt_collection(self):
...
             sql_selection = b'SELECT s.indexer, s.startyear, s.indexer_id, s.show_name, e.season, e.episode ' \
                             b'FROM tv_episodes AS e, tv_shows AS s ' \
                             b'WHERE e.indexer = s.indexer AND s.indexer_id = e.showid ' \
-                            b"AND e.status in ({0}) AND e.location <> ''".format(','.join(selection_status))
+                            b"AND e.status in ({0}) AND e.location <> ''".format(','.join(['?']*len(status)))
```
sharkykh: Same as above.

medusa/trakt_checker.py
```diff
@@ -337,12 +337,11 @@ def remove_episode_watchlist(self):
         if app.TRAKT_SYNC_WATCHLIST and app.USE_TRAKT:
 
             main_db_con = db.DBConnection()
-            status = Quality.DOWNLOADED + Quality.ARCHIVED
-            selection_status = [b'?' for _ in status]
+            status = [DOWNLOADED, ARCHIVED]
```
sharkykh: Same as above.

medusa/trakt_checker.py
```diff
@@ -337,12 +337,11 @@ def remove_episode_watchlist(self):
...
             sql_selection = b'SELECT s.indexer, s.startyear, e.showid, s.show_name, e.season, e.episode ' \
                             b'FROM tv_episodes AS e, tv_shows AS s ' \
                             b'WHERE e.indexer = s.indexer ' \
-                            b'AND s.indexer_id = e.showid AND e.status in ({0})'.format(b','.join(selection_status))
+                            b'AND s.indexer_id = e.showid AND e.status in ({0})'.format(b','.join([b'?']*len(status)))
```
sharkykh: Same as above.

medusa/trakt_checker.py
```diff
@@ -382,12 +381,11 @@ def add_episode_watchlist(self):
         if app.TRAKT_SYNC_WATCHLIST and app.USE_TRAKT:
 
             main_db_con = db.DBConnection()
-            status = Quality.SNATCHED + Quality.SNATCHED_BEST + Quality.SNATCHED_PROPER + [WANTED]
-            selection_status = [b'?' for _ in status]
+            status = [SNATCHED, SNATCHED_BEST, SNATCHED_PROPER, WANTED]
```
sharkykh: Same as above.

medusa/trakt_checker.py
```diff
@@ -382,12 +381,11 @@ def add_episode_watchlist(self):
...
             sql_selection = b'SELECT s.indexer, s.startyear, e.showid, s.show_name, e.season, e.episode ' \
                             b'FROM tv_episodes AS e, tv_shows AS s ' \
                             b'WHERE e.indexer = s.indexer AND s.indexer_id = e.showid AND s.paused = 0 ' \
-                            b'AND e.status in ({0})'.format(b','.join(selection_status))
+                            b'AND e.status in ({0})'.format(b','.join([b'?']*len(status)))
```
sharkykh: Same as above.

medusa/tv/series.py
```diff
@@ -1752,11 +1757,10 @@ def refresh_dir(self):
                     with cur_ep.lock:
                         # if it used to have a file associated with it and it doesn't anymore then
                         # set it to app.EP_DEFAULT_DELETED_STATUS
-                        if cur_ep.location and cur_ep.status in Quality.DOWNLOADED:
+                        if cur_ep.location and cur_ep.status in DOWNLOADED:
```
sharkykh: Since `Quality.DOWNLOADED` was a list, I think this changes the logic. So use:
```py
if cur_ep.location and cur_ep.status == DOWNLOADED:
# or
if cur_ep.location and cur_ep.status in (DOWNLOADED, ):
# or
if cur_ep.location and cur_ep.status in [DOWNLOADED]:
```

medusa/tv/series.py
```diff
@@ -2188,11 +2192,13 @@ def want_episode(self, season, episode, quality, forced_search=False,
         )
         return should_replace
 
-    def get_overview(self, ep_status, backlog_mode=False, manually_searched=False):
+    def get_overview(self, ep_status, ep_quality, backlog_mode=False, manually_searched=False):
         """Get the Overview status from the Episode status.
 
         :param ep_status: an Episode status
         :type ep_status: int
+        :param ep_quality: an Episode status
```
sharkykh: *quality

Also: A few lines below we do:
`ep_status = try_int(ep_status) or UNSET`
Should there be one for quality too?
`ep_quality = try_int(ep_quality) or Quality.NA`

medusa/common.py
```diff
@@ -666,15 +617,10 @@ def should_replace(ep_status, old_quality, new_quality, allowed_qualities, prefe
         :param search_type: The search type, that started this method
         :return: True if the old quality should be replaced with new quality.
         """
-        if ep_status and ep_status not in Quality.DOWNLOADED + Quality.SNATCHED + Quality.SNATCHED_PROPER:
+        if ep_status and ep_status not in (DOWNLOADED, SNATCHED, SNATCHED_PROPER):
             if not force:
                 return False, 'Episode status is not DOWNLOADED|SNATCHED|SNATCHED PROPER. Ignoring new quality'
```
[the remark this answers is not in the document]
medariox: What do you mean? This looks correct to me.

medusa/databases/main_db.py
```diff
@@ -257,16 +256,29 @@ def fix_subtitles_codes(self):
...
+        Check for a status bigger than 12, and translate to a status + quality, as these are old composite statuses.
 
+        This can be removed when all code that creates composite statuses has been migrated.
+        Until then this can be used to keep the DB sane.
+        """
+        log.info(u'Convert composite statuses in tv_episodes to status + quality.')
+        sql_results = self.connection.select("SELECT status FROM tv_episodes where status > 12 GROUP BY status;")
+        for status in sql_results:
+            log.info(u'Split composite status into status and quality for %s', status[b'status'])
+            split = common.Quality.split_composite_status(status[b'status'])
```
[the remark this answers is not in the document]
medariox: I believe this excludes "simple" statuses from the SQL query, as up to 12 they are not composite (status + quality): https://github.com/pymedusa/Medusa/blob/master/medusa/common.py#L104-L115. The old quality values can't be "simple" in the old status column IIRC, they can only be composite and that would make the > 12.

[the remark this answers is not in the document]
sharkykh: Let's say I have `status = 3204` (which is **Downloaded 720p WEB-DL** on `develop`)
This will extract it, and set the values in database to:
```py
status = 4  # Downloaded
quality = 32  # 1080p HDTV - on this branch
```

What I'm saying is, the value should be shifted `val >> 1`, just like in the migration code at the bottom of this file.
Needs to be:
```py
status = 4  # Downloaded
quality = 64 # 720p WEB-DL - on this branch
```

----

And adding to that, see what the next "sanity check" function does:
```py
    def update_status_unknown(self):
        """Changes any `UNKNOWN` quality to 1."""
        log.info(u'Update status UNKONWN from tv_episodes')
        self.connection.action("UPDATE tv_episodes SET quality = 1 WHERE quality = 65536;")
```

If you didn't shift the quality values, you won't have `quality = 65536` to begin with.

themes-default/slim/views/displayShow.mako
```diff
@@ -280,15 +280,18 @@ const startVue = () => {
...
+                        %>
+                        % if cur_quality != Quality.NA:
                             <td class="col-status triggerhighlight">${statusStrings[cur_status]} ${renderQualityPill(cur_quality)}</td>
                         % else:
                             <td class="col-status triggerhighlight">${statusStrings[cur_status]}</td>
                         % endif
                     <td class="col-search triggerhighlight">
                         % if int(epResult["season"]) != 0:
-                            % if (int(epResult["status"]) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST + Quality.DOWNLOADED ) and app.USE_FAILED_DOWNLOADS:
+                            % if app.USE_FAILED_DOWNLOADS and (int(epResult["status"]) in (SNATCHED, SNATCHED_PROPER, SNATCHED_BEST, DOWNLOADED)):
```
sharkykh: The extra parentheses can be removed.

themes-default/slim/views/displayShow.mako
```diff
@@ -297,7 +300,7 @@ const startVue = () => {
                         % else:
                             <app-link class="epManualSearch" id="${str(show.indexer)}x${str(show.series_id)}x${str(epResult["season"])}x${str(epResult["episode"])}" name="${str(show.indexer)}x${str(show.series_id)}x${str(epResult["season"])}x${str(epResult["episode"])}" href="home/snatchSelection?indexername=${show.indexer_name}&seriesid=${show.series_id}&amp;season=${epResult["season"]}&amp;episode=${epResult["episode"]}"><img data-ep-manual-search src="images/manualsearch.png" width="16" height="16" alt="search" title="Manual Search" /></app-link>
                         % endif
-                        % if int(epResult["status"]) not in Quality.SNATCHED + Quality.SNATCHED_PROPER and app.USE_SUBTITLES and show.subtitles and epResult["location"]:
+                        % if app.USE_SUBTITLES and show.subtitles and epResult["location"] and (int(epResult["status"]) not in (SNATCHED, SNATCHED_PROPER, SNATCHED_BEST)):
```
sharkykh: The extra parentheses can be removed.

themes-default/slim/views/manage_backlogOverview.mako
```diff
@@ -144,13 +144,14 @@ const startVue = () => {
                 </tr>
                 % for cur_result in showSQLResults[(cur_show.indexer, cur_show.series_id)]:
                     <%
-                        old_status, old_quality = Quality.split_composite_status(cur_result['status'])
-                        archived_status = Quality.composite_status(ARCHIVED, old_quality)
+                        old_status = cur_result['status']
+                        old_quality = cur_result['quality']
+                        archived_status = ARCHIVED
```
sharkykh: We can remove it. The only place this is used in is line 175:
https://github.com/pymedusa/Medusa/blob/942eb3095ace01860979f1f6caea4eae1928c925/themes-default/slim/views/manage_backlogOverview.mako#L175

So we can just substitute it with `ARCHIVED`.

themes-default/slim/views/manage_episodeStatuses.mako
```diff
@@ -42,28 +42,28 @@ Manage episodes with status <select name="whichStatus" class="form-control form-
...
     # Do not allow setting to bare downloaded or archived!
-    statusList.remove(common.DOWNLOADED)
-    statusList.remove(common.ARCHIVED)
+    statusList.remove(DOWNLOADED)
+    statusList.remove(ARCHIVED)
     if int(whichStatus) in statusList:
         statusList.remove(int(whichStatus))
-    if int(whichStatus) in [common.SNATCHED, common.SNATCHED_PROPER, common.SNATCHED_BEST] + common.Quality.ARCHIVED + common.Quality.DOWNLOADED and app.USE_FAILED_DOWNLOADS:
-        statusList.append(common.FAILED)
+    if app.USE_FAILED_DOWNLOADS and (int(whichStatus) in (SNATCHED, SNATCHED_PROPER, SNATCHED_BEST, ARCHIVED, DOWNLOADED)):
```
sharkykh: The extra parentheses can be removed.

medusa/common.py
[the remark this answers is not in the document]
medariox: Oh, you are talking about the log message. Got it now :)

medusa/post_processor.py
```diff
@@ -896,8 +890,6 @@ def _should_process(current_quality, new_quality, allowed, preferred):
         :param preferred: Qualities that are preferred
         :return: Tuple with Boolean if the quality should be processed and String with reason if should process or not
         """
-        if current_quality is common.Quality.NONE:
-            return False, 'There is no current quality. Skipping as we can only replace existing qualities'
```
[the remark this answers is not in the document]
medariox: I've asked myself why we shouldn't be replacing `N/A` and I couldn't find a valid reason.

medusa/databases/failed_db.py
```diff
@@ -111,3 +112,87 @@ def execute(self):
...
+
+        sql_results = self.connection.select('SELECT status FROM history GROUP BY status;')
+        for result in sql_results:
+            split = Quality.split_composite_status(result[b'status'])
+            self.connection.action('UPDATE history SET status = ?, quality = ? WHERE status = ?;',
+                                   [split.status, split.quality, result[b'status']])
+
+    def inc_major_version(self):
+        major_version, minor_version = self.connection.version
+        major_version += 1
```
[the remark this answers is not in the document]
medariox: I disagree. If the minor version doesn't exist it shouldn't be added, as that would be misleading.

medusa/tv/series.py
medariox: I've removed the try_int and also the or. If we pass something that isn't valid we need to find the cause of it, try_int is generally a mistake.

medusa/show/history.py
p0psicles: Afaik the changes in history are all mako related. Changes in python are for separating status and quality

medusa/trakt_checker.py
p0psicles: That's what it says right? Do you mean the binary string conversion? That's just there to make sure the end string is also binary.

sharkykh: No, sorry. And I didn't even notice I dropped the `b` prefix.
@medariox it wasn't intentional.
I meant the spaces around the asterisk.

medusa/databases/failed_db.py
```diff
@@ -111,3 +112,87 @@ def execute(self):
...
+        """
+        Shift all qualities << 1.
+
+        This makes it possible to set UNKNOWN as 1, making it the lowest quality.
+        """
+        log.info('Shift qualities in history one place to the left.')
+        sql_results = self.connection.select("SELECT quality FROM history GROUP BY quality ORDER BY quality DESC;")
+        for result in sql_results:
+            quality = result[b'quality']
+            new_quality = quality << 1
```
sharkykh: I just woke up, but... You're supposed to shift right by one bit, not left. Right? :thinking: 
This applies to other locations too.

medariox: No, how do you come to that conclusion now?

medusa/databases/main_db.py
```diff
@@ -706,3 +727,163 @@ def create_series_dict():
...
+        for status in sql_results:
+            split = common.Quality.split_composite_status(status[b'action'])
+            self.connection.action("UPDATE history SET action = ? WHERE action = ?;",
+                                   [split.status, status[b'action']])
+
+        self.inc_minor_version()
+
+
+class ShiftQualities(AddSeparatedStatusQualityFields):
+    """Shift all qualities one place to the left."""
```
sharkykh: Same for this class - shoulf be shifting right, not left?
I hope I'm wrong though.

medusa/databases/main_db.py
```diff
@@ -37,13 +37,15 @@ def check(self):
         self.fix_invalid_airdates()
         #  self.fix_subtitles_codes()
         self.fix_show_nfo_lang()
-        self.convert_archived_to_compound()
+        # self.convert_archived_to_compound()
```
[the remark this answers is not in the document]
p0psicles: I wasnt sure at the time if we should keep it for the long term. But it can be removed.

medusa/show/history.py
sharkykh: Just based on the values here, I'm assuming it must have an effect on that.
This is exactly what defines a row in compact layout. Now that you removed `quality`, it's joining everything under one row.

I'm not 100% sure, but it looks like that to me, purely based on the Python code.

p0psicles: Ill check

medusa/databases/main_db.py
```diff
@@ -850,6 +844,14 @@ def shift_tv_qualities(self):
         for result in sql_results:
             quality = result[b'quality']
             new_quality = quality << 1
+
+            # UNKNOWN quality value is 65536 (1 << 16) instead of 32768 (1 << 15) after the shift
+            # Qualities in the tv_shows table have the combined values of allowed and preferred qualities.
+            # Preferred quality couldn't contain UNKNOWN
+            if new_quality & 65536 > 0:  # If contains UNKNOWN allowed quality
+                new_quality -= 65536  # Remove it
+                new_quality |= common.Quality.UNKNOWN  # Then re-add it using the correct value
```
sharkykh: Maybe this should be `new_quality |= 1`, in case we're changing the value again.

medariox: I think it's good as it is now.

medusa/databases/failed_db.py
```diff
@@ -24,8 +24,8 @@ def execute(self):
         queries = [
             ('CREATE TABLE failed (release TEXT, size NUMERIC, provider TEXT);',),
             ('CREATE TABLE history (date NUMERIC, size NUMERIC, release TEXT, provider TEXT,'
-             ' old_status NUMERIC DEFAULT 0, showid NUMERIC DEFAULT -1, season NUMERIC DEFAULT -1,'
-             ' episode NUMERIC DEFAULT -1);',),
+             ' status NUMERIC DEFAULT 0, quality NUMERIC DEFAULT 0, showid NUMERIC DEFAULT -1',
```
medariox: Should `status` default to `-1` instead since that's the default status? https://github.com/pymedusa/Medusa/blob/feature/add-db-status-quality-fields/medusa/common.py#L103

sharkykh: Yes. You're right.

sharkykh:
[Closed by mistake]

Oops, I think I broke something.
I just put this branch on my main instance of Medusa and my tv_episodes table is messed up.

sharkykh:
@medariox @p0psicles I don't believe that was intended:
image

We should still have the options to choose a downloaded & archived quality.

sharkykh:
I tried to fix the history "bug" myself but hit a snag.
FWIW I found that because the `action` database column was a composite status, `Action` could be defined by it, and `Index` was defined by the `quality` column.
But because of the composite statuses split, now `Action` has `quality` (because it has to), and `Index` doesn't have it, so when the items are compacted it's compacting all of the history items of the same indexer, show, season and episode, ignoring the quality of the action.

p0psicles:
Both History and episode status management where broken for me.

p0psicles:
I dont know. I feel like we should really do more testing. The last two bugs i noticed where really obvious. And i couldnt test it sooner. As my db is a mess.

medariox:
That's why we should merge it to develop ASAP, to have some more people actually test it.

p0psicles:
Good point

PR #4236, opened 2018-05-21
Update root yarn.lock

sharkykh:


.travis.yml
```diff
@@ -10,8 +10,8 @@ install:
   - pip install dredd_hooks
   - pip install PyYAML
   - pip install six
-  - nvm install v8.9.0
-  - nvm use v8.9.0
+  - nvm install node
```
OmgImAlexis: We should be pinning our version still.

OmgImAlexis: @sharkykh we can use 10, etc. just make sure to pin the version so it's always the same. Libs keep changing their allowed version to disallow new versions, etc. so that's why I was against just using the latest.

PR #4237, opened 2018-05-21
Remove NotifyMyAndroid

sharkykh:
## Merge [Squash] on/after May 24th
image


sharkykh:
With this PR I also worked on the notifiers' icons image sprite: `32x_sprite_colored_notifiers.png`
I added a PSD file to `themes-default/slim/resources/32x_sprite_colored_notifiers.psd`. as I found it easier to work on a PSD file rather than the PNG file.
I think `themes-default/slim/resources` is a good place for these kind of files.

I moved every icon to be in a box of exactly 32px by 32px. This makes adding/removing icons easier, and in the CSS - each icon's position is exactly `-32px` from the previous one.
i
i

PR #4244, opened 2018-05-23
Update guessit to 3.0.0 (+subliminal update)

sharkykh:
* Migration Docs: https://github.com/guessit-io/guessit/blob/3.0.0/docs/migration2to3.rst
* ~Waiting on #4221.~
* In order to update `guessit`, `subliminal` has to be updated as well.
  ~[I started a branch here](https://github.com/pymedusa/subliminal/commits/guessit-3), but hit a snag with `WEBRip` and such.~ pymedusa/subliminal#1
* *Tip:* Review without the first commit (the vendored guessit update)


labrys:
Careful there's a regression in Guessit for titles with periods in it (e.g. Marvel's Agents of S.H.E.I.L.D.) where it drops some of the periods (e.g. `S H.E.I.L.D.`).  Might not be a problem with expected titles, but should be checked.

sharkykh:
@labrys You sure? I get the correct title on both `guessit` and `tests/report_guessit.py`.
Can you give me an example filename for it?

labrys:
[SSCE](http://sscce.org/)
```python
import guessit
guess = guessit.guessit("Marvel's Agents of S.H.I.E.L.D.")
print(guess['title'])
# Marvel's Agents of S H.I.E.L.D.
#                     ^
```
See guessit-io/guessit#546

ratoaq2:
It also happens in 2.1.4... Upgrading to 3.0 shouldn't be a problem

medusa/name_parser/rules/__init__.py
```diff
@@ -16,6 +16,9 @@
 from medusa.name_parser.rules.rules import rules
 
 
+# @FIXME: Workaround
+default_api.configure({})
```
sharkykh: I don't know if that's right.

medusa/name_parser/rules/rules.py
```diff
@@ -35,7 +35,8 @@
 
 from guessit.rules.common.comparators import marker_sorted
 from guessit.rules.common.formatters import cleanup
-from guessit.rules.properties.release_group import clean_groupname
+# @FIXME: `clean_groupname` is no longer available...
+# from guessit.rules.properties.release_group import clean_groupname
```
sharkykh: It was a public function, now it's an inner function inside the `release_group` function.
`clean_groupname` is used in `ReleaseGroupPostProcessor`.

sharkykh: It was a public function, now it's an inner function inside the `release_group` function.
`clean_groupname` is used in `ReleaseGroupPostProcessor`.

medusa/name_parser/rules/rules.py
```diff
@@ -35,8 +35,8 @@
 
 from guessit.rules.common.comparators import marker_sorted
 from guessit.rules.common.formatters import cleanup
-from guessit.rules.properties import website
-from guessit.rules.properties.release_group import clean_groupname
+# @FIXME: `clean_groupname` is no longer available...
+# from guessit.rules.properties.release_group import clean_groupname
```
sharkykh: It was a public function, now it's an inner function inside the `release_group` function.
`clean_groupname` is used in `ReleaseGroupPostProcessor`.

sharkykh:
@OmgImAlexis 0.2.7 is awfully optimistic, considering I don't know how to proceed with this PR.

medusa/name_parser/rules/rules.py
```diff
@@ -914,6 +915,73 @@ def when(self, matches, context):
...
+            "source": "Web",
+            "other": "Rip",
+            "audio_codec": "AAC",
+            "audio_channels": "2.0",
+            "video_codec": "H.264",
+            "release_group": "monkee",
+            "type": "episode"
+        }
+
+    without the rule:
```
p0psicles: With?

sharkykh:
Need to update `medusa/subtitle_providers/itasa.py`. I have a commit, I'll push it when I can.

PR #4251, opened 2018-05-25
Fix #4246 - No search results in APIv1 sb.searchindexers

### LINKED ISSUE #4246: [API v1] sb.searchindexers does not return anything
Hi guys,

I received a bug report for my Android app (https://github.com/MGaetan89/ShowsRage/issues/149) related to the search feature in the API v1.
When using this entry point, Medusa (version 0.2.4) does not return anything, while the web UI returns some shows for the same query.
Here is a sample request that returns nothing for me from the API (in the API builder), and a lot of shows on the web UI:
`/api/v1/{api_key}/?cmd=sb.searchindexers&name=how`

Could you have a look to see what is going on?
I know that you have an API v2 now, and I am planning to include it at some point. But until then, I will need v1 to work as expected ;)

sharkykh:
Fixes #4246 - No search results from API v1 `sb.searchindexers`
Also fixes #1895 - `tuple index out of range` (when searching tvmaze, tmdb or all indexers)

:heavy_check_mark: Tested.

### DIFF
--- a/medusa/server/api/v1/core.py
+++ b/medusa/server/api/v1/core.py
@@ -26,7 +26,6 @@ import json
 import logging
 import os
 import time
-from builtins import str
 from collections import OrderedDict
 from datetime import date, datetime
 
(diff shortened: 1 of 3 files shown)


medusa/server/api/v1/core.py
```diff
@@ -64,7 +64,7 @@
 
 from requests.compat import unquote_plus
 
-from six import binary_type, iteritems, itervalues, string_types, text_type, viewitems
+from six import binary_type as bytes, iteritems, itervalues, string_types, text_type as str, viewitems
```
labrys: This is a really bad practice.

sharkykh:
The build is failing because of a `LookupError` exception raised by `imdbpie` while adding a show via API v2. That is fixed in #4253.

sharkykh:
Succeeded by #4260.

PR #4262, opened 2018-05-28
Add YouTube Red network icon

sharkykh:


sharkykh:
(For the record, I prefer `-`)
If we want to do that, this is the code that needs to be changed: https://github.com/pymedusa/Medusa/blob/develop/medusa/tv/series.py#L351-L354

I was going to say that means we need to make sure all the templates use the series asset API endpoint, but it seems like that is already the case.

sharkykh:
@OmgImAlexis 
We can merge this PR now.

Here's another branch for removing the spaces from the file names.
https://github.com/pymedusa/Medusa/compare/develop...feature/network-logos-use-hyphen

PR #4267, opened 2018-05-28
Vueify/new show

sharkykh:
Depends on :heavy_check_mark: ~#4059~ + :heavy_check_mark: ~#4283~ 

----

~There are some commits that don't belong in this branch. I will clean it when the time comes :)~

Showing off the new UI :wink: (gif is a bit outdated though)
<details>
<summary>GIF</summary>

</details>


themes-default/slim/views/addShows_newShow.mako
```diff
@@ -215,19 +207,67 @@ const startVue = () => {
...
+                    formData = new FormData(this.$refs.addShowForm);
+                }
+
+                this.otherShows.forEach(nextShow => formData.append('other_shows', nextShow));
+
+                const response = await apiRoute.post('addShows/addNewShow', formData);
+                const { data } = response;
+                const { result, message, redirect, params } = data;
+
+                if (message.length !== 0) {
```
OmgImAlexis: Since this isn't an array use `if (message) {}`. Same goes for the `redirect.length` below.

OmgImAlexis:
@sharkykh is this going to be ready for `0.2.6`?

sharkykh:
@OmgImAlexis It depends how far I'm willing to go with this.
It's already pretty good, I just have some ideas I can most certainly do in another PR.

medusa/server/api/v2/internal.py
```diff
@@ -136,3 +138,92 @@ def resource_existing_series(self):
...
+        if indexer_id > 0 and indexer_id not in enabled_indexers:
+            return self._bad_request('Invalid indexer id.')
+
+        if not language or language == 'null':
+            language = app.INDEXER_DEFAULT_LANGUAGE
+
+        search_terms = [query]
+
+        # If search term ends with what looks like a year, enclose it in ()
+        matches = re.match(r'^(.+ |)([12][0-9]{3})$', query)
```
OmgImAlexis: Can probably add this as another PR but we also need to do `vs` and `vs.` as I found with `Red vs. Blue`.

themes-default/slim/views/addShows_newShow.mako
```diff
@@ -1,14 +1,36 @@
...
 %>
 <%block name="scripts">
 <script type="text/javascript" src="js/add-show-options.js?${sbPID}"></script>
 <script type="text/javascript" src="js/blackwhite.js?${sbPID}"></script>
-<script src="js/lib/frisbee@2.0.4.min.js"></script>
-<script src="js/lib/vue-frisbee.min.js"></script>
-<script src="js/vue-submit-form.js"></script>
+## <script src="js/lib/frisbee@2.0.4.min.js"></script>
+## <script src="js/lib/vue-frisbee.min.js"></script>
+## <script src="js/vue-submit-form.js"></script>
```
OmgImAlexis: These should be removed before merging.

themes-default/slim/views/addShows_newShow.mako
```diff
@@ -94,17 +523,12 @@ const startVue = () => {
                         <%include file="/inc_addShowOptions.mako"/>
                     </div>
                 </fieldset>
-                % for curNextDir in other_shows:
-                <input type="hidden" name="other_shows" value="${curNextDir}" />
-                % endfor
-                <input type="hidden" name="skipShow" v-model="skipShow" />
             </form>
-            <br>
+            <br />
```
OmgImAlexis: The other style is preferred in HTML5.

themes-default/slim/views/addShows_newShow.mako
OmgImAlexis: They're used in other files right? I was mainly talking about the python comment.

sharkykh: Ah well, no, they're unused.
`frisbee` (`vue-frisbee`) is only used in `vue-submit-form`, and `vue-submit-form` *was* only used in `newShow`.

OmgImAlexis: If they're unused remove the libs as we can always add them back.

PR #4271, opened 2018-05-29
Feature/hotswap themes

sharkykh:
Currently you have to restart Medusa for the theme to update..
I agree it's a bit of a hack, but still.
This can wait for another release.


sharkykh:
Let's do this?

PR #4288, opened 2018-05-30
Add vue component name-pattern.mako

p0psicles:
Will be used in postprocessing.
In future also for creating custom search strings.


p0psicles:
This PR is finished imo. It's working fine as for the naming part of the vueify-postprocessing PR.
But when merged I can continue on utilizing the naming templates component for the search template PR.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+            namingPattern: {
+                type: String,
+                default: ''
+            },
+            /**
+             * An array of available preset naming patterns.
+            */
+            namingPresets: {
+                type: Array,
+                default: () => []
```
OmgImAlexis: No need for the function, just use `default: []`.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+            multiEpStyle: {
+                type: Number,
+                default: null
+            },
+            /**
+             * Availale multi ep style
+             */
+            multiEpStyles: {
+                type: Array,
+                default: () => []
```
OmgImAlexis: No need for the function, just use `default: []`.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+            namingPresets: {
+                type: Array,
+                default: () => []
+            },
+            /**
+             * The selected multi ep style
+             */
+            multiEpStyle: {
+                type: Number,
+                default: null
```
OmgImAlexis: Remove default since null isn't a Number type.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                    pattern: pattern,
+                    anime_type: animeType
+                }
+
+                if (selectedMultiEpStyle) {
+                    params.multi = selectedMultiEpStyle;
+                }
+
+                try {
+                    const result = await apiRoute.get('config/postProcessing/testNaming', {params: params});
```
OmgImAlexis: If you switch this to `return apiRoute.get('config/postProcessing/testNaming', {params: params}).then(res => res.data);` then you wouldn't need the `async`.

OmgImAlexis: Also since this is async you should return something yet if this throws the catch is just going to log.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                    return result.data;
+                } catch (e) {
+                    console.warning(e);
+                }
+            },
+            async updatePatternSamples() {
+                // If it's a custom pattern, we need to get the custom pattern from this.customName
+                const pattern = (this.isCustom) ? this.customName : this.pattern;
```
OmgImAlexis: Remove the `()`.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                } catch (e) {
+                    console.warning(e);
+                }
+            },
+            async updatePatternSamples() {
+                // If it's a custom pattern, we need to get the custom pattern from this.customName
+                const pattern = (this.isCustom) ? this.customName : this.pattern; 
+                
+                // Update single
+                this.namingExample = await this.testNaming(pattern, false, this.animeType) + '.ext';
```
OmgImAlexis: If this fails then the namingExample will be set to undefined or worse the promise itself.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                
+                // Update multi if needed
+                if (this.isMulti) {
+                    this.namingExampleMulti = await this.testNaming(pattern, this.selectedMultiEpStyle, this.animeType) + '.ext';
+                    this.checkNaming(pattern, this.selectedMultiEpStyle, this.animeType);
+                }
+            },
+            update(newValue) {
+                this.$emit('change', {
+                    pattern: (this.isCustom) ? this.customName : this.pattern,
```
OmgImAlexis: No need for `()`.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                    type: this.type,
+                    multiEpStyle: this.selectedMultiEpStyle,
+                    custom: this.isCustom,
+                    enabled: this.isEnabled,
+                    animeNamingType: this.animeType
+                });
+            },
+            async checkNaming(pattern, selectedMultiEpStyle, animeType) {
```
OmgImAlexis: Again with the async, there are places above that don't use await so they'll likely remain as a unresolved promise until you do something with them.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                const vm = this;
```
OmgImAlexis: Instead of using `vm` use deconstruction to get a const or an arrow function to stay in scope.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                const params = {
+                    pattern: pattern,
```
OmgImAlexis: Replace with `pattern,`.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                }
+
+                if (selectedMultiEpStyle) {
+                    params.multi = selectedMultiEpStyle;
+                }
+
+                try {
+                    const result = await apiRoute.get('config/postProcessing/isNamingValid', {params: params});
+                    if (result.data === 'invalid') {
+                        $(vm.$el).find('#naming_pattern').qtip('option', {
```
OmgImAlexis: This would be nicer like so.

OmgImAlexis: You could move the first two lines further up as well.

themes-default/slim/views/vue-components/name-pattern.mako
```diff
@@ -0,0 +1,491 @@
...
+                get: function() {
+                    return (this.isCustom) ? 'Custom...' : this.pattern;
+                },
+                set: function(value) {
+                    this.pattern = value;
+                }
+            }
+        },
+        mounted() {
+            this.pattern = this.namingPattern;
```
OmgImAlexis: Why aren't all of these set in data? The mounted shouldn't be setting default values.

themes-default/slim/views/vue-components/name-pattern.mako
p0psicles: What should I return here then?

themes-default/slim/views/vue-components/name-pattern.mako
p0psicles: I'm no handling the error in the testNaming function.
If something goes wrong, it will return an empty string for now.

themes-default/slim/views/vue-components/name-pattern.mako
p0psicles: Can you provide an example?

themes-default/slim/views/vue-components/name-pattern.mako
p0psicles: I'm setting the values for the local data using the properties. That's what you should do in mounted right? data() i am using to set the defaults.

themes-default/slim/views/vue-components/name-pattern.mako
p0psicles: Never mind I see what you mean. It's the same as what you mentioned below.:
```
                const { $el } = this;
                const el = $($el);
```

themes-default/slim/views/vue-components/name-pattern.mako
OmgImAlexis: All other fields using arrays are working fine. Does this get set something other than an array somewhere in the component?

themes-default/slim/views/vue-components/name-pattern.mako
OmgImAlexis: Why not set the default on the prop and then just pass `pattern` as a prop instead of `namingPattern` ?

themes-default/slim/views/vue-components/name-pattern.mako
p0psicles: No. For example multiEpStyles is only assigned to this.availableMultiEpStyles on line 407

themes-default/slim/views/vue-components/name-pattern.mako
p0psicles: I can't directly use the properties, as then I would change them by reference. And Vue.js doesn't like that

themes-default/slim/views/vue-components/name-pattern.mako
OmgImAlexis: @p0psicles marked as resolved. All good.

PR #4292, opened 2018-06-01
Replace spaces with hyphens in network logo names

sharkykh:
@medariox Sorry our discussion doesn't show up here
https://github.com/pymedusa/Medusa/commit/c1530abc91af99030c6128ffd3f33c34f50ef05e#comments


medusa/tv/series.py
```diff
@@ -351,7 +351,7 @@ def is_sports(self):
     @property
     def network_logo_name(self):
         """Get the network logo name."""
-        return self.network.replace(u'\u00C9', 'e').replace(u'\u00E9', 'e').lower()
+        return self.network.replace(u'\u00C9', 'e').replace(u'\u00E9', 'e').replace(' ', '-').lower()
```
medariox: My point is: we should remove it as it makes no sense, because: 1) it replaces only one unicode char 2) it tries to replace it twice. 😕

sharkykh: About 2: It's not the same character twice.
[`\u00C9`](https://m.unicode-table.com/en/00C9/) [`\u00E9`](https://m.unicode-table.com/en/00e9/)

If we do I'd have to find the network names that originally had it and re-add the correct char instead of `e`.

medariox: True that, I didn't see the difference 😃 Let's not complicate this further at this point.

PR #4298, opened 2018-06-02
Fix adding 2+ shows (that need searching) using Add Existing Show

sharkykh:
Fixes the bug where after adding the first show you were redirected to the home page instead of the next show.

Make `addShows/addNewShow` return a redirection url instead of redirecting
Backported from #4267 and tweaked
Tested.

@OmgImAlexis Apologies.


medusa/server/web/home/add_shows.py
```diff
@@ -417,14 +432,16 @@ def addNewShow(self, whichSeries=None, indexer_lang=None, rootDir=None, defaultS
...
-
-            # peel off the next one
-            next_show_dir = other_shows[0]
-            rest_of_show_dirs = other_shows[1:]
+                return json_redirect('/home/')
 
             # go to add the next show
-            return self.newShow(next_show_dir, rest_of_show_dirs)
+            return json_redirect(
+                '/addShows/newShow/',
```
OmgImAlexis: Won't this also need the other query params like if the user wants auto settings or settings picked manually for each, etc?

sharkykh: No, because the form on `addShow/existingShows` doesn't POST to this endpoint (`addShow/addNewShow`), but rather to [`addShow/addExistingShows`](https://github.com/pymedusa/Medusa/blob/a4448f97e0cd4df3d9c3717eeec211d30e19e934/medusa/server/web/home/add_shows.py#L548) (notice the no `add` at the start).
This ^ endpoint handles the `promptForSettings` stuff and then redirects us to the `addShow/newShow` **page** with the correct parameters. That's why the first show that needs searching always works.

It's only `addShow/newShow` that POSTs to `addShow/addNewShow` which needs this patch, at least for right now.

PR #4299, opened 2018-06-03
Fix snatchSelection refreshing page every 5 seconds

sharkykh:
Refresh the page only when the search is finished,
and use workarounds until we convert to Vue.


medariox:
Like I said on Slack I'm against this change, as a search can take some time to finish and it is important to provide some kind of visual feedback before it finishes.

PR #4313, opened 2018-06-07
Feature/always pp manual searched files

p0psicles:

This PR will make sure that snatches done through a manual search, are always considered "priority" download. Meaning that although they might have already been downloaded in the past. A manual search result will always be post-processed.

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -812,61 +812,60 @@ class PostProcessor(object):
                  season, episode] + snatched_statuses
             )
 
-            if tv_episodes_result:
+            if tv_episodes_result and tv_episodes_result[0][b'quality'] == quality:
                 # Check if the snatch is a manual snatch
                 if tv_episodes_result[0][b'manually_searched'] == 1:
                     self.manually_searched = True
 
-                if tv_episodes_result[0][b'quality'] == quality:
-                    # Second: get the quality of the last snatched epsiode
-                    # and compare it to the quality we are post-processing
-                    history_result = main_db_con.select(
-                        'SELECT quality, info_hash '
+                # Second: get the quality of the last snatched epsiode
+                # and compare it to the quality we are post-processing
+                history_result = main_db_con.select(
+                    'SELECT quality, info_hash '
+                    'FROM history '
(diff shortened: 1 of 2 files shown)


p0psicles:
@medariox there are comparable checks in process_tv.py and post_processor.py. It seems i need to make changes to both?

p0psicles:
OK, so process_tv.py is the manual postprocessor which scans a folder. Then this looks like it could be a lot more DRY.

medusa/process_tv.py
```diff
@@ -489,6 +489,29 @@ def unrar(self, path, rar_files, force=False):
...
+
+        if all(
+            [history_result,
+             history_result[0]['manually_searched'] == '1',
+             history_result[0]['action'] in (SNATCHED, SNATCHED_BEST, SNATCHED_PROPER)]
+        ):
+            self.log('Last snatch for this video was manual searched: {0}'.format(
+                video_file
+            ), logger.DEBUG)
+            return True
```
medariox: The problem with that is that the `video_file` name doesn't always match the torrent/nzb name. So it's not 100% reliable.

medusa/post_processor.py
```diff
@@ -1040,7 +1041,7 @@ def process(self):
                    (common.Quality.qualityStrings[new_ep_quality]), logger.DEBUG)
 
         # see if this is a priority download (is it snatched, in history, PROPER, or BEST)
-        priority_download = self._is_priority(old_ep_quality, new_ep_quality)
+        priority_download = self._is_priority(old_ep_quality, new_ep_quality) or self.manually_searched
```
medariox: This will cause the file to be processed over and over if it's not moved away from the pp folder.

p0psicles: Ok so your suggesting that we need to keep track of an additional flag. Which is enabled when it has been processed once?

medariox: Sorry, I was actually wrong, it should work just fine in `post_processor` now. The issue was rather in `process_tv`. Should be good now.

medariox:
@p0psicles Can you please let me know if it works for you this way? I've tested with a few releases and it seems to work well so far.

PR #4332, opened 2018-06-11
add puex store and vuex websockets

OmgImAlexis:
These are all libs apart from the `store.js`.


themes/dark/assets/js/lib/vuex.js
```diff
@@ -0,0 +1,938 @@
+/**
```
sharkykh: Looks like this was committed by mistake? None of the other themes have this file, and it's not imported in `main.mako`.

themes-default/slim/static/js/store.js
```diff
@@ -0,0 +1,195 @@
...
+            });
+        }
+    },
+    plugins: []
+});
+
+Vue.use(VueNativeSock, 'ws://localhost:8081/x/ws/ui', {
+    store: window.store,
+    format: 'json'
+});
```
sharkykh: Do you know if doing this makes it try to connect to the websocket?
Do you need to add `connectManually: true,` to disable that?

themes-default/slim/static/js/store.js
```diff
@@ -0,0 +1,195 @@
...
+                    release: null,
+                    aliases: null,
+                    subtitlesEnabled: null,
+                    anime: null
+                }
+            }
+        }
+    },
+    mutations: {
+        [LOGIN_PENDING]() {},
```
sharkykh: What sorcery is this? :scream: 
Can you explain why it's in brackets?

OmgImAlexis: const x = 'a';[x]

OmgImAlexis: const x = 'a';
const y = {};[a] =

OmgImAlexis:
@p0psicles any ideas if we pass anything but JSON?

Just noticed this section with `{ format: json }` enabled.
https://github.com/nathantsoi/vue-native-websocket#with-format-json-enabled

sharkykh:
@OmgImAlexis 
We **currently** don't pass anything but JSON through the websocket.
And frankly I don't think we will ever pass anything else.

sharkykh:
Just a small note: It currently opens two WebSocket connections.
Can we move the notification handling to this file right now?

OmgImAlexis:
@sharkykh yes that was my plan after merging this.

PR #4339, opened 2018-06-11
editShow: Replace checkboxes with toggle-button

p0psicles:


themes-default/slim/views/editShow.mako
```diff
@@ -50,7 +52,8 @@ const startVue = () => {
                     {text: 'Ignored', value: 'Ignored'}
                 ],
                 seriesLoaded: false,
-                saving: false
+                saving: false,
+                enableButton: true
```
sharkykh: Unused?

themes-default/slim/views/editShow.mako
```diff
@@ -225,14 +228,16 @@ const startVue = () => {
                         <div class="form-group">
                             <label for="subtitles" class="col-sm-2 control-label">Subtitles</label>
                             <div class="col-sm-10 content">
-                                <input type="checkbox" id="subtitles" name="subtitles" v-model="series.config.subtitlesEnabled"/> search for subtitles
+                                <toggle-button width="45" heigth="22" id="subtitles" name="subtitles" v-model="series.config.subtitlesEnabled" :sync="true"></toggle-button>
```
sharkykh: 1. Boolean properties can be: `<toggle-button ... sync></toggle-button>`
3. There's a typo in the height property (not `heigth`)
2. I wonder if we can somehow "subclass" the component to set our preferred defaults, so we wouldn't have to keep adding `width="45" height="22"` everywhere.
@OmgImAlexis Mixins maybe? I'm not sure I understand the docs ATM.

Looks like these apply to all the buttons.

OmgImAlexis: The width and height should be set in the component with props then just set the prop to have a default of these values.

OmgImAlexis: Ah, I see it's a third-party component, we could add a mixin I think.

p0psicles: ok will add the component

p0psicles: Tried it but couldnt get it to work. I suggest keep it like is.

themes-default/slim/views/editShow.mako
p0psicles: Ill remove this and then merge

themes-default/slim/views/editShow.mako
sharkykh: @p0psicles I was able to figure it out :)

PR #4346, opened 2018-06-12
convert series to show

OmgImAlexis:
@pymedusa/administrators if you could all try this as I changed almost every file.


OmgImAlexis:
@p0psicles I'm going to try and just convert the js over and I'll sort the python later.

PR #4357, opened 2018-06-12
move notifications to vue

OmgImAlexis:


themes-default/slim/static/js/store.js
```diff
@@ -171,10 +175,34 @@ const store = new Puex({
...
+                state.socket.messages.push(message);
+            }
         },
         // Mutations for websocket reconnect methods
         [SOCKET_RECONNECT](state, count) {
+            let error = '';
+            error += 'Error connecting to websocket. Please check your network connection. ';
+            error += 'If you are using a reverse proxy, please take a look at our wiki for config examples.';
+
+            displayNotification('notice', error);
```
sharkykh: This should be in `SOCKET_RECONNECT_ERROR` instead.

themes-default/slim/static/js/store.js
```diff
@@ -171,10 +175,34 @@ const store = new Puex({
...
         [SOCKET_ONMESSAGE](state, message) {
+            const { data, event } = message;
+            const { body, hash, type, title } = data;
+
+            // Set the current message
             state.socket.message = message;
+
+            // Show the notification to the user
+            if (event === 'notification') {
+                displayNotification(type, title, body, hash);
```
sharkykh: Something is weird. Using the following action, I can't seem to get more than 3 notifications to show.
Maybe it's unrelated.
`for (let i=0; i < 10; i++) { window.store.dispatch('testNotifications') }`
```js
        testNotifications(store) {
            const rand = Math.floor((Math.random() * 10000)).toString();
            store.commit(SOCKET_ONMESSAGE, {
                event: 'notification',
                data: {
                    type: `error`,
                    title: `test-${rand}`,
                    body: `test-${rand}<br><i class="test-class">hello <b>world</b></i><ul><li>item 1</li><li>item 2</li></ul>`,
                    hash: `notification-${rand}`
                }
            });
        }
```

OmgImAlexis: Are all the commits showing in the vuex dev panel?

sharkykh: Yeah, I think it's unrelated. Even with `new Notification("message")` it's a bit buggy...

PR #4359, opened 2018-06-13
Handle exceptions on download method test web endpoints

sharkykh:


OmgImAlexis:
@sharkykh we should look at moving the tests to the API.

sharkykh:
Yeah I know, in time...

PR #4361, opened 2018-06-13
move more of schedule to vue

OmgImAlexis:
I'll work on the layout templates themselves in another PR.

This does currently require us to still refresh to update the page as the sub templates are using Python still.

@sharkykh any idea what we should do because of the above?


themes-default/slim/views/schedule.mako
```diff
@@ -19,14 +21,34 @@ const startVue = () => {
...
+            {
+                layout: {
+                    get() {
+                        const { config } = this;
+                        return config.layout.schedule;
+                    },
+                    set(layout) {
+                        const { $store } = this;
+                        const page = 'schedule';
+                        $store.dispatch('setLayout', { page, layout });
```
OmgImAlexis: Please only use this for updating the layout. Just provide it with the page and layout you want to set. It'll `PATCH` to the API and then update the config in the store.

themes-default/slim/views/schedule.mako
```diff
@@ -19,14 +21,34 @@ const startVue = () => {
...
+                    }
+                }
+            }
+        ),
         mounted() {
-            if ($.isMeta({ layout: 'schedule' }, ['list'])) {
+            const { $store, $route, config, layout } = this;
+            $store.dispatch('getConfig');
+            $store.dispatch('getShows');
```
OmgImAlexis: These will be moved to the main vue instance when all pages are done.

themes-default/slim/views/schedule.mako
```diff
@@ -58,7 +80,7 @@ const startVue = () => {
                 $.ajaxEpSearch();
             }
 
-            if ($.isMeta({ layout: 'schedule' }, ['banner', 'poster'])) {
+            if (layout in ['banner', 'poster']) {
```
sharkykh: Did you mean this?
```js
if (['banner', 'poster'].includes(layout)) {
```

themes-default/slim/static/js/store.js
```diff
@@ -244,11 +244,35 @@ const store = new Puex({
...
+        },
+        setLayout(store, { page, layout }) {
+            const { dispatch } = store;
+
+            return api.patch('config/main', {
+                layout: {
+                    [page]: layout
+                }
+            }).then(setTimeout(() => {
+                dispatch('getConfig');
```
sharkykh: 1. Why is a timeout needed here?
2. Since we currently refresh the page I suggest commenting out this line.

themes-default/slim/views/schedule.mako
```diff
@@ -19,14 +21,34 @@ const startVue = () => {
...
+                }
+            }
+        ),
         mounted() {
-            if ($.isMeta({ layout: 'schedule' }, ['list'])) {
+            const { $store, $route, config, layout } = this;
+            $store.dispatch('getConfig');
+            $store.dispatch('getShows');
+
+            if (layout == 'list') {
```
sharkykh: Triple equals here?

themes-default/slim/views/schedule.mako
sharkykh: I just checked and the syntax you used works on objects, but not on arrays.

themes-default/slim/static/js/store.js
```diff
@@ -233,17 +233,17 @@ const store = new Puex({
     // No actions should write to the store
     // Please use store.commit to fire off a mutation that'll update the store
     actions: {
-        getConfig(store) {
+        getConfig() {
             return api.get('/config/main').then(res => {
                 store.commit(ADD_CONFIG, res.data);
```
sharkykh: Did you mean to replace these stores with context as well?

OmgImAlexis: Nope, I'm using the global `store`. Just happens context also contains a small section of the store object. I think puex exports `dispatch` and `commit` only.

themes/light/templates/schedule.mako
```diff
@@ -38,70 +38,74 @@ const startVue = () => {
...
-            if (layout === 'list') {
-                const sortCodes = {
-                    date: 0,
-                    show: 2,
-                    network: 5
-                };
-                const sort = config.comingEpsSort;
-                const sortList = (sort in sortCodes) ? [[sortCodes[sort], 0]] : [[0, 0]];
+            const unwatch = this.$watch('layout', () => {
+                unwatch();
```
OmgImAlexis: Shouldn't this be outside of the function itself? 😕

sharkykh: I wasn't sure if it works this way. My point is to unwatch as soon as I can because I don't want the code to run more than once.

OmgImAlexis: Oh it should, just checked the docs and $this.watch returns the function needed to unregister it.

sharkykh:
> @sharkykh any idea what we should do because of the above?

@OmgImAlexis 
IMO, since we can't load Mako templates on demand apart from actually loading the page,
and we can't use jQuery to load the DOM because Vue won't parse it,
I don't see a way until they're converted to Vue with vue-router.

themes-default/slim/static/js/store.js
sharkykh: @OmgImAlexis 
I know this is already merged but working on something else I noticed the `.then()` doesn't wait for the API call to return.
Don't know exactly what the issue is, but try looking on the network tab (enable `preserve log` if needed), on the Waterfall section of the network tab, you can see that the request issued by `.then()` doesn't wait for the previous one to finish.

PR #4363, opened 2018-06-13
Add new values to the config API endpoint

sharkykh:
Commit messages say it all.


OmgImAlexis:
Does this allow `/api/v2/config` to return all three sections?

OmgImAlexis:
You'll likely need to update the dredd api description file.

sharkykh:
`GET /api/v2/config` - You get everything (`main`, `qualities`, `statuses`) - the object will be split into the sections.
`GET /api/v2/config/main` - Just one section as the main object.
`GET /api/v2/config/qualities?append=statuses,main` - Get `qualities`, `statuses` and `main` with the same schema as the first call.

OmgImAlexis:
Is that last one really needed? The endpoints should be independent. If you need multiple either get /config or each of them and merge on the client.

OmgImAlexis:
@sharkykh 

`GET /api/v2/config` is returning an array with a single object that contains the configs. Can you please change this to returning just the object there's no need for the outer array.

sharkykh:
This is @ratoaq2 's initial pagination code, I haven't touched that. So not really sure.

OmgImAlexis:
Since I can't ever see a reason to paginate the config I've removed that, how this?

```diff
diff --git a/medusa/server/api/v2/config.py b/medusa/server/api/v2/config.py
index 6aa217beb..e34a2e583 100644
--- a/medusa/server/api/v2/config.py
+++ b/medusa/server/api/v2/config.py
@@ -98,7 +98,7 @@ class ConfigHandler(BaseRequestHandler):
             for section in config_sections:
                 config_data[section] = getattr(DataGenerator, 'data_' + section)()

-            return self._paginate([config_data])
+            return self._ok(data=config_data)

         config_data = getattr(DataGenerator, 'data_' + identifier)()

```

sharkykh:
Yep, that's correct.
I pushed this change and also made sure it still works.

OmgImAlexis:
You may need to change the `store.js` as I think it's still getting the first array item.

themes-default/slim/static/js/store.js
```diff
@@ -233,9 +240,16 @@ const store = new Puex({
     // No actions should write to the store
     // Please use store.commit to fire off a mutation that'll update the store
     actions: {
-        getConfig() {
-            return api.get('/config/main').then(res => {
-                store.commit(ADD_CONFIG, res.data);
+        getConfig(context, section) {
+            return api.get('/config/' + (section || '')).then(res => {
+                if (!section) {
```
sharkykh: That should be `if (section) {` if I'm not mistaken.

PR #4367, opened 2018-06-13
Develop

FireBladeNooT:
Hi,
Could you allow choosing or add using scene name for anime search?
Also it could be great to configure anime and choose if we want absolute number or not (S00E00).

For example:
Anime: My Hero Academia
Scene name: Boku no hero academia
Torrent looking for:  Boku.no.Hero.Academia.S03E10.VOSTFR.720p.WEBRiP.x264-GODSPACE (mp4)

So in this example, the most efficient would be check "Anime" for nyaa search, not check "Absolute Numbering" and check "Scene Name for search".

To resume options:
- Anime (Remove absolute property)
- Absolute (New option to choose absolute number)
- Scene Name (New option to choose scene name)
- All type of search (it is like actual search + search "scene name S00E00")

Or it could be options for each kind of search:
- Anime
- Normal numbering
- Absolute numbering
- Normal name
- Scene Name
The main difficult in this proposal will be to managed all combinations possible and mainly combination with nothing checked.

What do you think about it?

Sincerely


p0psicles:
Why did you create a pull request.

FireBladeNooT:
@p0psicles It's a mistake, I would make an issue. sorry

PR #4372, opened 2018-06-15
fix assets not loading in some home layouts

### LINKED ISSUE #4216: Poster image not being displayed
Hi,

I'm running the master branch of Medusa and I've noticed that I have two shows where the posters aren't showing up for the show.  It's just the generic Medusa image.  Yet if I go into the show info page and click on the poster, the poster image will then be displayed full screen.  So the system has the poster, it's just not displaying the small poster on the various pages.

I've tried deleting the show and adding it back, but that doesn't fix the issue.  Has anyone else noticed this problem before?

Thanks

OmgImAlexis:
Should close #4216

### DIFF
--- a/themes-default/slim/views/partials/home/banner.mako
+++ b/themes-default/slim/views/partials/home/banner.mako
@@ -152,7 +152,9 @@
                     </td>
                     <td align="center">
                     % if cur_show.network:
-                        <span title="${cur_show.network}" class="hidden-print"><img id="network" class="show-network-image" width="54" height="27" src="images/network/nonetwork.png" lazy="on" series="${cur_show.slug}" asset="network" alt="${cur_show.network}" title="${cur_show.network}" /></span>
+                        <span title="${cur_show.network}" class="hidden-print">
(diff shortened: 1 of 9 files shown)


themes-default/slim/views/partials/home/banner.mako
```diff
@@ -152,7 +152,9 @@
                     </td>
                     <td align="center">
                     % if cur_show.network:
-                        <span title="${cur_show.network}" class="hidden-print"><img id="network" class="show-network-image" width="54" height="27" src="images/network/nonetwork.png" lazy="on" series="${cur_show.slug}" asset="network" alt="${cur_show.network}" title="${cur_show.network}" /></span>
+                        <span title="${cur_show.network}" class="hidden-print">
+                            <asset default="images/network/nonetwork.png" series-slug="${cur_show.slug}" type="network" cls="show-network-image" :link="false" width="54" height="27" alt="${cur_show.network}" title="${cur_show.network}"></asset>
```
sharkykh: 1. No need for `:link="false"` it seems, since it defaults to false. But maybe it's good for readability?
2. I remember having an issue when trying to move the network images to use the asset component.
I think it was the missing error handling when a network logo isn't available.

OmgImAlexis: 1. Yeah let's leave it for readability.

PR #4374, opened 2018-06-15
moved config to http-vue-loader

OmgImAlexis:
@p0psicles

### DIFF
--- a/themes-default/slim/static/js/lib/http-vue-loader.min.js
+++ b/themes-default/slim/static/js/lib/http-vue-loader.min.js
@@ -1,0 +1,470 @@
+(function umd(root,factory){
+	if(typeof module==='object' && typeof exports === 'object' )
+		module.exports=factory()
+	else if(typeof define==='function' && define.amd)
+		define([],factory)
+	else
+		root.httpVueLoader=factory()
+})(this,function factory() {
+	'use strict';
+
+	var scopeIndex = 0;
+
+	StyleContext.prototype = {
+
+		withBase: function(callback) {
+
+			var tmpBaseElt;
+			if ( this.component.baseURI ) {
+
+				// firefox and chrome need the <base> to be set while inserting or modifying <style> in a document.
+				tmpBaseElt = document.createElement('base');
+				tmpBaseElt.href = this.component.baseURI;
+
+				var headElt = this.component.getHead();
+				headElt.insertBefore(tmpBaseElt, headElt.firstChild);
+			}
+
+			callback.call(this);
+
+			if ( tmpBaseElt )
+				this.component.getHead().removeChild(tmpBaseElt);
+		},
+
+		scopeStyles: function(styleElt, scopeName) {
+
+			function process() {
+
+				var sheet = styleElt.sheet;
+				var rules = sheet.cssRules;
+
+				for ( var i = 0; i < rules.length; ++i ) {
+
+					var rule = rules[i];
+					if ( rule.type !== 1 )
+						continue;
+
+					var scopedSelectors = [];
+
+					rule.selectorText.split(/\s*,\s*/).forEach(function(sel) {
+
+						scopedSelectors.push(scopeName+' '+sel);
+						var segments = sel.match(/([^ :]+)(.+)?/);
+						scopedSelectors.push(segments[1] + scopeName + (segments[2]||''));
+					});
+
+					var scopedRule = scopedSelectors.join(',') + rule.cssText.substr(rule.selectorText.length);
+					sheet.deleteRule(i);
+					sheet.insertRule(scopedRule, i);
+				}
+			}
+
+			try {
+				// firefox may fail sheet.cssRules with InvalidAccessError
+				process();
+			} catch (ex) {
+
+				if ( ex instanceof DOMException && ex.code === DOMException.INVALID_ACCESS_ERR ) {
+
+					styleElt.sheet.disabled = true;
+					styleElt.addEventListener('load', function onStyleLoaded() {
+
+						styleElt.removeEventListener('load', onStyleLoaded);
+
+						// firefox need this timeout otherwise we have to use document.importNode(style, true)
+						setTimeout(function() {
+
+							process();
+							styleElt.sheet.disabled = false;
+						});
+					});
+					return;
+				}
+
+				throw ex;
+			}
+		},
+
+		compile: function() {
+
+			var hasTemplate = this.template !== null;
+
+			var scoped = this.elt.hasAttribute('scoped');
+
+			if ( scoped ) {
+
+				// no template, no scopable style needed
+				if ( !hasTemplate )
+					return;
+
+				// firefox does not tolerate this attribute
+				this.elt.removeAttribute('scoped');
+			}
+
+			this.withBase(function() {
+
+				this.component.getHead().appendChild(this.elt);
+			});
+
+			if ( scoped )
+				this.scopeStyles(this.elt, '['+this.component.getScopeId()+']');
+
+			return Promise.resolve();
+		},
+
+		getContent: function() {
+
+			return this.elt.textContent;
+		},
+
+		setContent: function(content) {
+
+			this.withBase(function() {
+
+				this.elt.textContent = content;
+			});
+		}
+	};
+
+	function StyleContext(component, elt) {
+
+		this.component = component;
+		this.elt = elt;
+	}
+
+
+	ScriptContext.prototype = {
+
+		getContent: function() {
+
+			return this.elt.textContent;
+		},
+
+		setContent: function(content) {
+
+			this.elt.textContent = content;
+		},
+
+		compile: function(module) {
+
+			var childModuleRequire = function(childURL) {
+
+				return httpVueLoader.require(resolveURL(this.component.baseURI, childURL));
+			}.bind(this);
+
+			var childLoader = function(childURL, childName) {
+
+				return httpVueLoader(resolveURL(this.component.baseURI, childURL), childName);
+			}.bind(this);
+
+			try {
+				Function('exports', 'require', 'httpVueLoader', 'module', this.getContent()).call(this.module.exports, this.module.exports, childModuleRequire, childLoader, this.module);
+			} catch(ex) {
+
+				if ( !('lineNumber' in ex) ) {
+
+					return Promise.reject(ex);
+				}
+				var vueFileData = responseText.replace(/\r?\n/g, '\n');
+				var lineNumber = vueFileData.substr(0, vueFileData.indexOf(script)).split('\n').length + ex.lineNumber - 1;
+				throw new (ex.constructor)(ex.message, url, lineNumber);
+			}
+
+			return Promise.resolve(this.module.exports);
+		}
+	};
+
+	function ScriptContext(component, elt) {
+
+		this.component = component;
+		this.elt = elt;
+		this.module = { exports:{} };
+	}
+
+
+	TemplateContext.prototype = {
+
+		getContent: function() {
+
+			return this.elt.innerHTML;
+		},
+
+		setContent: function(content) {
+
+			this.elt.innerHTML = content;
+		},
+
+		getRootElt: function() {
+
+			var tplElt = this.elt.content || this.elt;
+
+			if ( 'firstElementChild' in tplElt )
+				return tplElt.firstElementChild;
+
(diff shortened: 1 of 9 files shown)


sharkykh:
IMO, we have to do something about the fact it loads the config from the API twice.
image
One from `core.js` and the other from the store.

themes-default/slim/static/js/templates/config.vue
```diff
@@ -0,0 +1,58 @@
...
+<script>
+module.exports = {
+    name: 'config',
+    metaInfo: {
+        title: 'Help & Info'
+    },
+    computed: Object.assign(store.mapState(['config']), {
+    }),
+    mounted() {
+        store.dispatch('getConfig');
```
OmgImAlexis: This isn't needed if the parent Vue instance already has access to the config which it would once I move it. I just figured I'd start with moving the config page as is.

themes-default/slim/views/config.mako
```diff
@@ -1,110 +1,12 @@
...
-            pythonVersion: undefined,
-            release: undefined,
-            sourceUrl: undefined,
-            sslVersion: undefined,
-            webRoot: undefined,
-            wikiUrl: undefined
-        };
-    },
-    async mounted() {
-        const { data } = await api.get('config/main');
```
OmgImAlexis: If you notice this would already hit the config twice since it's got the api check here. I just moved it to a store dispatch instead of a raw api method.

OmgImAlexis:
Okay, this doesn't work if the component uses another component. I'm not sure if there's anyway to fix that so maybe we should skip this for now?

OmgImAlexis:
I've added the 404 route and added it to the global VueRouter. That's also been moved to it's own file.

---

Once a `.mako` file has been moved over completely to Vue the template can be switched to `index.mako` as it's setup to load as a SPA.

---

`Vue-meta` should be removed once all the pages are added to the router but it's needed still otherwise the title of each page doesn't get set.

labrys:
@pymedusa/support I don't know if anyone's addressed it but with so many things moving to javascript you should implement an issue submitter and in app logs for all the js code.

p0psicles:
As soon as we can ditch the mako files. We can start using webpack. And with that unit and end2end test our vue components.

Were not that far off. With allot of pahlges already transitioned to vue.

sharkykh:
I already have a local branch that automatically logs Vue errors to Python.
I just wasn't sure about my code.

themes-default/slim/static/js/templates/config.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+    <div id="config-content">
+        <table class="infoTable" cellspacing="1" border="0" cellpadding="0" width="100%">
+            <tr v-if="config.release">
```
sharkykh: My changes from #4477 were reverted here.

themes-default/slim/static/js/lib/http-vue-loader.min.js
```diff
@@ -0,0 +1,470 @@
+(function umd(root,factory){
```
sharkykh: This file is named `.min.js` yet it is not minimised.

medusa/server/web/config/handler.py
```diff
@@ -46,7 +46,7 @@ def index(self):
         """
         Render the Help & Info page
         """
-        t = PageTemplate(rh=self, filename='config.mako')
+        t = PageTemplate(rh=self, filename='index.mako')
 
         try:
```
sharkykh: This route can be simplified now.
Many of the values are already served from the API.

The route function can be:
```py
    def index(self):
        """
        Render the Help & Info page.

        [Converted to VueRouter]
        """
        t = PageTemplate(rh=self, filename='index.mako')
        return t.render(submenu=self.ConfigMenu(), topmenu='config')
```

themes-default/slim/static/js/templates/config.vue
```diff
@@ -0,0 +1,55 @@
...
+            <tr><td><i class="icon16-config-github"></i> Source:</td><td><app-link :href="config.sourceUrl">{{config.sourceUrl}}</app-link></td></tr>
+            <tr><td><i class="icon16-config-mirc"></i> IRC Chat:</td><td><app-link href="irc://irc.freenode.net/#pymedusa" rel="noreferrer"><i>#pymedusa</i> on <i>irc.freenode.net</i></app-link></td></tr>
+        </table>
+    </div>
+</template>
+
+<script>
+module.exports = {
+    name: 'config',
+    computed: Object.assign(store.mapState(['config']), {
```
sharkykh: I don't see why this and the `dispatch` below are needed here.
Why did you add the following mixin if not for that?
https://github.com/pymedusa/Medusa/blob/f535c3d89c28e97861ec5cf6cb36f8d725d21fb2/themes-default/slim/views/layouts/main.mako#L163-L177

medusa/server/web/config/handler.py
sharkykh: Oh yeah, and also remove the unused imports at the top of the file.

medusa/server/web/config/handler.py
```diff
@@ -2,16 +2,7 @@
 
 """Base handler for Config pages."""
 
-from __future__ import unicode_literals
```
sharkykh: `__future__` imports should usually stay. This one in particular.
Basically, they're the first steps of making Python 2 compatible with Python 3.
Sorry for not being more explicit...

themes-default/slim/static/js/templates/config.vue
```diff
@@ -0,0 +1,68 @@
...
+
+<script>
+const { store } = window;
+
+module.exports = {
+    name: 'config',
+    computed: store.mapState(['config']),
+    mounted() {
+        const { $store } = this;
+        $store.dispatch('getConfig');
```
sharkykh: You didn't respond to my review about that.
I don't understand why we still need to get the config if we enabled it globally with the mixin?

OmgImAlexis: I have to look into it but for some reason it's not getting the config set to anything but `null` like it is when we first load the store.

OmgImAlexis:
After trying to convert more templates locally I found that the jQuery tabs refuse to load if a Vue HTML element is found inside of it and if I move the component back to being loaded locally via the x-templates it works fine.

sharkykh:
@OmgImAlexis 
Have you tried calling the tabs initializations (`$(element).tabs()` IIRC) on the component `mounted()`?

themes-default/slim/static/js/templates/config.vue
sharkykh: I removed that code and it still works fine for me? :thinking:

OmgImAlexis:
Yes, I also tried commenting it out and running it after everything has loaded via the console and nothing happens.

themes-default/slim/static/js/templates/config.vue
OmgImAlexis: Who knows then.. shall I remove it?

themes-default/slim/static/js/templates/config.vue
```diff
@@ -0,0 +1,64 @@
...
+        </table>
+    </div>
+</template>
+
+<script>
+const { store } = window;
+
+module.exports = {
+    name: 'config',
+    computed: store.mapState(['config']),
```
sharkykh: I don't know if it affects performance, but this mapState is not needed.

themes-default/slim/static/js/templates/config.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+    <div id="config-content">
```
sharkykh: How do you feel about:
```vue
    <div v-if="!config.pythonVersion" class="vue-loading" id="config-content">
        <h3>Loading...</h3>
    </div>
    <div v-else id="config-content">
```

It gives it a cleaner look in the meantime, but a little more work when it's no longer needed.
I added `class="vue-loading"` so if we keep it the same on all `.vue` files it would be easy to remove them.

themes-default/slim/static/js/lib/http-vue-loader.js
```diff
@@ -0,0 +1,470 @@
+(function umd(root,factory){
```
sharkykh: I still prefer a minimised version of this file, unless you believe it would be easier to debug issues?

themes-default/slim/static/js/templates/config.vue
OmgImAlexis: I was thinking we could just add a loading component before the `<router-view>`?

```
<loading v-if="loading"></loading>
<router-view v-else></router-view>

Comment (sharkykh) [reply themes-default/slim/static/js/templates/config.vue]: How would you know when the values were loaded?

File: themes/dark/templates/config.mako [hunk 12]
```diff
@@ -1,122 +1,5 @@
 <%inherit file="/layouts/main.mako"/>
```
sharkykh: These files didn't get deleted on dark and light themes, need to manually remove

themes-default/slim/static/js/templates/config.vue
OmgImAlexis: I still want to keep it there as I'd like to disable the mixin for `.vue` files.

themes-default/slim/static/js/lib/http-vue-loader.js
OmgImAlexis: Yeah, I think for now it'll be easier.

themes-default/slim/static/js/templates/config.vue
OmgImAlexis: Well for example on this you'd set `loading: true` on data and then set a function to watch `config` when it updates set `loading: false`.

themes-default/slim/static/js/router.js
```diff
@@ -0,0 +1,47 @@
...
+            path: '/not-found',
+            name: 'not-found',
+            meta: {
+                title: '404',
+                header: '404 - page not found'
+            },
+            component: notFoundComponent
+        }, {
+            path: '*',
+            redirect: '/not-found'
```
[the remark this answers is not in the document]
OmgImAlexis: Not getting this issue on dev with and without webroot so I'm not sure what's causing it on your end.Do you maybe still have the service worker loaded?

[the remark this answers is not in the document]
sharkykh: I don't use a webroot. ServiceWorker is loaded but I don't it has anything to do with it.

Try navigating from say, `/home` to `/schedule` by clicking the Schedule nav link. You should get this issue.
Don't know why it only seems to happen on `/schedule` route.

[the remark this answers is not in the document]
OmgImAlexis: That 100% has something todo with it. Remove the service worker.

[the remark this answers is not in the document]
sharkykh: Oh, I just figured it out.
https://github.com/pymedusa/Medusa/blob/9eaf64841e85608af4ea1cff3a516237a9ef7bed/themes/light/templates/schedule.mako#L15
We need to remove this line.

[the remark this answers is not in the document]
OmgImAlexis: Oh.. no we just need to add the route for it. I'm moving the headers and titles over to the `router.js`. Just don't add a component property and it works fine.

For example.

```js
routes: Object.assign(routes || [], [{
    path: '/config',
    name: 'config',
    meta: {
        title: 'Help & Info',
        header: 'Medusa Configuration'
    },
    component: configComponent
}, {
    path: '/home',
    name: 'home',
    meta: {
        title: 'Home',
        header: 'Show list'
    }
}, {
    path: '/history',
    name: 'history',
    meta: {
        title: 'History',
        header: 'History'
    }
}, {
    path: '/not-found',
    name: 'not-found',
    meta: {
        title: '404',
        header: '404 - page not found'
    },
    component: notFoundComponent
}, {
    path: '*',
    redirect: '/not-found'
}])
```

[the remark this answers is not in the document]
OmgImAlexis: We then use this in the pages that aren't `.vue` since they don't use the global Vue.

```
header() {
    return this.$route.meta.header;
},
```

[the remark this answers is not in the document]
sharkykh: This is the only page that has the `router` as a property now.
I still think removing it is the correct option since it'll be a while until schedule is moved to the router.

PR #4376, opened 2018-06-15
fix asset 404

OmgImAlexis:
@sharkykh please check this. I just blindly added the code. /hides


themes-default/slim/views/vue-components/asset.mako
```diff
@@ -22,16 +22,17 @@ Vue.component('asset', {
...
     },
     computed: {
         src() {
-            const {seriesSlug, type, isVisible} = this;
+            const { error, seriesSlug, type, isVisible } = this;
             const apiRoot = document.getElementsByTagName('body')[0].getAttribute('api-root');
             const apiKey = document.getElementsByTagName('body')[0].getAttribute('api-key');
 
-            if (!isVisible || !seriesSlug || !type) {
+            if (error || !isVisible || !seriesSlug || !type) {
```
sharkykh: This may actually cause a infinite loop if the default image is throwing a 404 too:
404 -> `error = true` -> `src = this.default` -> 404 -> `error = true` -> `src = this.default` ...

sharkykh:
As I said on Slack, after testing this it seems I was wrong before.
The API `/series/:slug/asset/:type` endpoint returns the default asset image if it's not available.
That's assuming all the values passed to `<asset>` (like `series-slug` and `type`) are correct.

OmgImAlexis:
@sharkykh 
The endpoint shouldn't return the default image at all. That API is wrong then as it should 404 is the asset is missing.

OmgImAlexis:
Checkout [the issue](https://bugs.chromium.org/p/chromium/issues/detail?id=124534) that's open about Chrome.

sharkykh:
That's very unfortunate.
And the issue is listed as fixed, closed October 2014.
The fix isn't even really a fix, just an option to hide it.

sharkykh:
@OmgImAlexis 
Looks like FF also doesn't honor the `error` event handler override.

How do you feel about adding a query parameter to return the default asset if an asset is not available?

sharkykh:
@OmgImAlexis 
I can approve this PR,
but I think my [initial review about the infinite loop](https://github.com/pymedusa/Medusa/pull/4376#discussion_r195710278) still stands.

themes-default/slim/views/vue-components/asset.mako
OmgImAlexis: If `this.default` is an error then wouldn't that mean the server is down? Any ideas on how we can get around this?

OmgImAlexis: We could include a base64 image for this so we wouldn't need to default back to a URL based one?

sharkykh: > If this.default is an error then wouldn't that mean the server is down?

Either that, or the url in default was incorrect.
> Any ideas on how we can get around this?

Error count instead of a boolean?
@error="error++;"
if (error == 2)

sharkykh: I just verified, there's no issue with that.
Seems like Vue is handling this correctly - If the image from the API fails it falls back to `this.default` and if `this.default` fails, nothing happens, since the value of `error` didn't change from `true`.

sharkykh:
I just verified, there's no issue with that.
Seems like Vue is handling this correctly - If the image from the API fails it falls back to `this.default` and if `this.default` fails, nothing happens, since the value of `error` didn't change from `true`.

Issue #4369, opened 2018-06-14
[APP SUBMITTED]: 


007revad:
### INFO
**Python Version**: `2.7.14 (default, May 28 2018, 14:47:18) [GCC 4.7.3 20130102 (prerelease)]`
**Operating System**: `Linux-3.10.102-x86_64-with-glibc2.2.5`
**Locale**: `UTF-8`
**Branch**: [master](../tree/master)
**Database**: `44.11`
**Commit**: pymedusa/Medusa@c65ca2607fae74405dcceaaf8c02f3d29d0313ea
**Link to Log**: https://gist.github.com/630e44e3868f93cdf7fb75418c951727
### ERROR
<pre>
2018-06-14 15:13:45 ERROR    SHOWQUEUE-REFRESH :: [c65ca26] 256002: Error while refreshing show America's Book of Secrets. Error: (404)
Reason: HTTPError
404 Client Error: Not Found for url: https://tvdb2.plex.tv/series/256002/actors
Traceback (most recent call last):
  File "/volume2/@appstore/sickbeard-custom/var/<a href="../blob/c65ca2607fae74405dcceaaf8c02f3d29d0313ea/SickBeard/medusa/show_queue.py#L686">SickBeard/medusa/show_queue.py</a>", line 686, in run
    self.show.refresh_dir()
  File "/volume2/@appstore/sickbeard-custom/var/<a href="../blob/c65ca2607fae74405dcceaaf8c02f3d29d0313ea/SickBeard/medusa/tv/series.py#L1713">SickBeard/medusa/tv/series.py</a>", line 1713, in refresh_dir
    self.load_episodes_from_dir()
  File "/volume2/@appstore/sickbeard-custom/var/<a href="../blob/c65ca2607fae74405dcceaaf8c02f3d29d0313ea/SickBeard/medusa/tv/series.py#L986">SickBeard/medusa/tv/series.py</a>", line 986, in load_episodes_from_dir
    cur_episode = self.make_ep_from_file(os.path.join(self.location, media_file))
  File "/volume2/@appstore/sickbeard-custom/var/<a href="../blob/c65ca2607fae74405dcceaaf8c02f3d29d0313ea/SickBeard/medusa/tv/series.py#L1352">SickBeard/medusa/tv/series.py</a>", line 1352, in make_ep_from_file
    root_ep.create_meta_files()
[...]

PR #4377, opened 2018-06-15
Fix similar issues tests and empty issue titles

sharkykh:
- Fix find_similar_issues logging and tests
- Fix empty issue title and add test
  Fixes titles like #4369

### DIFF
--- a/medusa/issue_submitter.py
+++ b/medusa/issue_submitter.py
@@ -112,6 +112,9 @@ class IssueSubmitter(object):
         """Find similar issues in the GitHub repository."""
         results = dict()
         issues = github_repo.get_issues(state='all', since=datetime.now() - max_age)
+        log.debug('Searching for issues similar to:\n{titles}', {
+            'titles': '\n'.join([line.issue_title for line in loglines])
+        })
         for issue in issues:
             if hasattr(issue, 'pull_request') and issue.pull_request:
                 continue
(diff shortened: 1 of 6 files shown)


medusa/issue_submitter.py
```diff
@@ -121,19 +126,25 @@ def find_similar_issues(cls, github_repo, loglines, max_age=timedelta(days=180))
...
-                log.debug('Searching for issues similar to: {0}', log_title)
 
                 # Apply diff ratio overrides on first-matched basis, default = 0.9
                 diff_ratio = next((override[1] for override in cls.TITLE_DIFF_RATIO_OVERRIDES
                                    if override[0] in log_title.lower()), 0.9)
 
                 if cls.similar(log_title, issue_title, diff_ratio):
+                    log.debug(
+                        'Found issue #{number} ({issue})'
+                        ' to be similar ({ratio:.0%}}) to {log}',
```
medariox: `{ratio:.0%}}` one `}` too much.

medusa/issue_submitter.py
```diff
@@ -183,6 +194,7 @@ def result(message, level=logging.WARNING):
         except RateLimitExceededException:
             return result(self.RATE_LIMIT)
         except (GithubException, IOError) as error:
+            log.debug('Issue submitter failed with error: {!r}', error)
```
medariox: I know that it is optional, but I'd still go for `{0!r}`.

medusa/issue_submitter.py
sharkykh: Oops, good catch! Thanks!

medusa/logger/__init__.py
```diff
@@ -346,7 +346,10 @@ def key(self):
     @property
     def issue_title(self):
         """Return the expected issue title for this logline."""
-        result = self.traceback_lines[-1] if self.traceback_lines else self.message
+        if self.traceback_lines:
+            result = next((line for line in reversed(self.traceback_lines) if line.strip()), self.message)
```
medariox: Is the `if` statement really needed?

sharkykh: Yes. Traceback lines can have an empty line, causing issue titles like this:
https://github.com/pymedusa/Medusa/issues/4369

PR #4388, opened 2018-06-17
Improve shutil_custom. Should fix #4362

### LINKED ISSUE #4362: Regular bad file descriptor errors, OSX
```
the.bold.type.s02e02.720p.web.x264-tbs.mkv to /mnt/Resources/TV/The xxx xxx/Season.02/The.xxx.xxx.S02E02.xxx yyy zzz.720p WEB-DL.mkv
2018-06-13 03:38:44 ERROR    Thread_16 :: [c65ca26] OSError:
  File "/app/medusa/medusa/helpers/__init__.py", line 286, in copy_file
    shutil.copyfile(src_file, dest_file)
  File "/app/medusa/lib/shutil_custom/__init__.py", line 44, in copyfile_custom
    os.write(fout, x)
OSError: [Errno 9] Bad file descriptor
2018-06-13 03:38:44 DEBUG    Thread_16 :: [c65ca26] Setting permissions for /mnt/Resources/TV/The xxx xxx/Season.02/The.xxx.xxx.S02E02.xxx yyy zzz.720p WEB-DL.mkv to 438 as parent directory has 511
```
Seems to happen about once per day. Destination is fine (NFS mount to NAS) and file seems to copy just fine regardless.

OSX is up to date and running Dockerized.

medariox:

Big thanks to @labrys for the code!
I've only changed the imports: `SpecialFileError` was meant to be `SameFileError`
and removed the function overwrite in `process_tv` as that's not needed.

Should fix #4362

### DIFF
--- a/lib/shutil_custom/__init__.py
+++ b/lib/shutil_custom/__init__.py
@@ -26,10 +26,10 @@ def copyfile_custom(src, dst):
     if _samefile(src, dst):
         raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
 
-    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
-        special_file(src)
-        special_file(dst)
+    special_file(src)
+    special_file(dst)
 
+    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
         while True:
             buf = fsrc.read(BUFFER_SIZE)
             if not buf:
--- a/lib/shutil_custom/__init__.py
+++ b/lib/shutil_custom/__init__.py
@@ -28,7 +28,9 @@ def copyfile_custom(src, dst):
             if stat.S_ISFIFO(st.st_mode):
                 raise SpecialFileError("`%s` is a named pipe" % fn)
 
-    with open(src, READ_FLAGS) as fin, open(dst, WRITE_FLAGS) as fout:
+    fdin = os.open(src, READ_FLAGS)
+    fdout = os.open(dst, WRITE_FLAGS)
+    with os.fdopen(fdin) as fin, os.fdopen(fdout) as fout:
         if _samefile(src, dst):
             raise SameFileError("`%s` and `%s` are the same file" % (src, dst))
 
--- a/lib/shutil_custom/__init__.py
+++ b/lib/shutil_custom/__init__.py
@@ -1,5 +1,6 @@
 import os
 import stat
+from io import open
 from shutil import SpecialFileError, _samefile
 try:
     from shutil import SameFileError
--- a/lib/shutil_custom/__init__.py
+++ b/lib/shutil_custom/__init__.py
@@ -12,8 +12,6 @@ except AttributeError:
     O_BINARY = 0
 
 
-READ_FLAGS = os.O_RDONLY | O_BINARY
-WRITE_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_TRUNC | O_BINARY
 BUFFER_SIZE = 128*1024
 
 
@@ -28,9 +26,7 @@ def copyfile_custom(src, dst):
             if stat.S_ISFIFO(st.st_mode):
                 raise SpecialFileError("`%s` is a named pipe" % fn)
 
-    fdin = os.open(src, READ_FLAGS)
-    fdout = os.open(dst, WRITE_FLAGS)
-    with open(fdin) as fin, open(fdout) as fout:
+    with open(src, 'rb') as fin, open(dst, 'wb') as fout:
         if _samefile(src, dst):
             raise SameFileError("`%s` and `%s` are the same file" % (src, dst))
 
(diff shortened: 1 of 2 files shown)


lib/shutil_custom/__init__.py
```diff
@@ -28,7 +28,7 @@ def special_file(fn):
             if stat.S_ISFIFO(st.st_mode):
                 raise SpecialFileError("`%s` is a named pipe" % fn)
 
-    with os.open(src, READ_FLAGS) as fin, os.open(dst, WRITE_FLAGS) as fout:
+    with open(src, READ_FLAGS) as fin, open(dst, WRITE_FLAGS) as fout:
```
labrys: The builtin `open` does not support flags, thus the use of `os.open`.

medariox: Yep and `os.open` doesn't support context managers for python < 3.6(?). What shall we do?

labrys: You use `os.fdopen` as the context manager with the file descriptors returned by `os.open`

lib/shutil_custom/__init__.py
```diff
@@ -30,7 +30,7 @@ def special_file(fn):
 
     fdin = os.open(src, READ_FLAGS)
     fdout = os.open(dst, WRITE_FLAGS)
-    with os.fdopen(fdin) as fin, os.fdopen(fdout) as fout:
+    with open(fdin) as fin, open(fdout) as fout:
```
labrys: A binary write mode would still be required for the output: `open(fdout, 'wb')`

labrys: Let me do some reading on the default flags used with the builtin open.  If my assumption is correct the flags we are setting manually are the same as the defaults provided by the higher level open builtin.

medariox: Sure thing, just let me know. Thanks!

sharkykh: Afaik the built-in open is not an alias for io.open on Python < 3.

sharkykh: > Since reading/writing here in raw binary though it makes no difference using `open` vs `io.open`

Docs for the built-in `open` on Python 2.7 doesn't say it accepts file descriptor objects like `io.open` does.
So IMO you need to use `io.open` regardless.

labrys: So looks like all the flags can be removed and just use `open` in `rb` and `wb` modes.

sharkykh: In any case I suggest using `io.open` for compatibility between the Python versions.

labrys: @sharkykh you're right again, 2.7 open does not accept a fd.  I've all but abandoned python 2.7 since i converted my medusa to pure python 3.6

labrys: @sharkykh  Sorry didn't finish my though.  You're right that in 2.7 `open` does not use accept file descriptors but the point is since the modes `'rb'` and `'wb'` are equivalent to the flags we use there is no need to even bother with `os.open` and file descriptors.

medariox: Done

lib/shutil_custom/__init__.py
```diff
@@ -6,35 +6,31 @@
...
-            raise SameFileError("`%s` and `%s` are the same file" % (src, dst))
+            raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
 
         special_file(src)
         special_file(dst)
 
-        while True:
-            data = fin.read(BUFFER_SIZE)
-            if data == '':
+        while 1:
```
labrys: Style preference only, but I think `while True:` is clearer than `while 1:`

medariox: Changed 👌

lib/shutil_custom/__init__.py
```diff
@@ -1,52 +1,36 @@
...
-                except NameError:
-                    raise Error("`%s` is a named pipe" % fn)
+                raise SpecialFileError("{!r} is a named pipe".format(fn))
 
-    try:
-        # Windows
-        O_BINARY = os.O_BINARY
-    except:
-        O_BINARY = 0
+    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
```
sharkykh: For consistency's sake - my point still stands.
We should be using `io.open` for Python 3 compatibility, same arguments.
For the record, `python-future` does that too.

I doesn't matter much in this context, however we decided we are going to phase out `python-future`, and we should be replacing all built-in `open` with `io.open` to make sure it's compatible with both major versions.

lib/shutil_custom/__init__.py
```diff
@@ -1,52 +1,36 @@
...
+        if _samefile(src, dst):
+            raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
```
sharkykh: This check can/should be before you open the files.

lib/shutil_custom/__init__.py
medariox: Good point, changed.

PR #4401, opened 2018-06-17
Fix toggle-button.

p0psicles:
* Remove the medusa-button wrapper, as it wasn't supposed to be in there at all.


themes-default/slim/views/layouts/main.mako
```diff
@@ -145,6 +145,7 @@
                     routes
                 });
             }
+            Vue.use(window['vue-js-toggle-button'].default);
```
p0psicles: @OmgImAlexis should I import here?

sharkykh: If you ask me, the component is already imported on line 120, and I see no reason not to put this here.

OmgImAlexis: Yeah Vue is a global so you only need it once per page, if it's in the `main.mako` once it'll be on every single Vue instance on that page. No need to re-register it.

OmgImAlexis: @sharkykh if the imported script doesn't auto install then that line will still be needed.

PR #4410, opened 2018-06-19
Improve detection of quality from filename

medariox:

1) Renamed `scene_quality` to `quality_from_name` as it is more accurate.
2) The resolution is now enough to determine a quality (without source it defaults to HDTV)
3) Fixed detection of `mpeg` releases
   Added detection of Amazon releases
   Improved detection of Netflix releases
4) Added tests for `quality_from_name`

### DIFF
--- a/medusa/common.py
+++ b/medusa/common.py
@@ -265,15 +265,13 @@ class Quality(object):
         """
         Return the quality from an episode filename.
 
-        If no quality is achieved it will try scene_quality regex.
-
         :param name: to parse
-        :param anime: Boolean to indicate if the show we're resolving is Anime
+        :param anime: Boolean to indicate if the show we're resolving is anime
         :param extend: boolean to extend methods to try
         :return: Quality
         """
-        # Try Scene names first
-        quality = Quality.scene_quality(name, anime)
+        # Try getting the quality from the filename
+        quality = Quality.quality_from_name(name, anime)
         if quality != Quality.UNKNOWN:
             return quality
 
@@ -284,12 +282,12 @@ class Quality(object):
         return Quality.UNKNOWN
 
     @staticmethod
-    def scene_quality(name, anime=False):
+    def quality_from_name(name, anime=False):
         """
         Return the quality from the episode filename with the regex.
 
         :param name: Episode filename to analyse
-        :param anime: Boolean to indicate if the show we're resolving is Anime
+        :param anime: Boolean to indicate if the show we're resolving is anime
         :return: Quality
         """
         from medusa.tagger.episode import EpisodeTags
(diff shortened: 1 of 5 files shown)


medusa/common.py
```diff
@@ -310,67 +308,62 @@ def scene_quality(name, anime=False):
...
+                # WEB-DL
+                elif ep.web or any([ep.amazon, ep.itunes, ep.netflix]):
+                    result = Quality.UHD_8K_WEBDL if is_4320p else Quality.UHD_4K_WEBDL
+                # HDTV
+                else:
+                    result = Quality.UHD_8K_TV if is_4320p else Quality.UHD_4K_TV
 
         # Is it HD?
         elif ep.vres in [1080, 720]:
+            is_1080p = ep.vres == 1080
```
sharkykh: Same here, `is_1080`.

medusa/common.py
```diff
@@ -310,67 +308,62 @@ def scene_quality(name, anime=False):
...
             elif sd_options:
                 result = Quality.SDTV
 
             return Quality.UNKNOWN if result is None else result
 
         # Is it UHD?
-        if ep.vres in [2160, 4320] and ep.scan == 'p':
-            # BluRay
-            full_res = (ep.vres == 4320)
-            if ep.avc and ep.bluray:
```
sharkykh: Just to be clear, removing the `ep.avc` tag checks means any video codec is accepted.
```py
avc = re.compile(r'([xh].?26[45]|(?:he|a)vc)', re.IGNORECASE)
```
https://regex101.com/r/e3tGNV/1/

Could you clarify which cases this change applies to?

medusa/common.py
```diff
@@ -310,67 +308,62 @@ def scene_quality(name, anime=False):
...
+                # HDTV
+                else:
+                    result = Quality.FULLHDTV if is_1080p else Quality.HDTV
+            elif ep.scan == 'i' and ep.tv == 'hd' and (ep.mpeg or (ep.raw and ep.avc_non_free)):
                 result = Quality.RAWHDTV
         elif ep.hrws:
             result = Quality.HDTV
 
         # Is it SD?
-        elif ep.xvid or ep.avc:
```
sharkykh: Again here, you're basically removing codec checks.

medusa/common.py
duramato: If I'm not mistaken it's intended.

medariox: Previously the function was called `scene_quality` and in that case, it made sense to check for the codec, as it is required by scene standards to be present in the release name. My goal here was to make it more generic so that we have a quick way to determine the quality from the name. The reasoning behind it is: do we really need to know the codec to determine the quality from the name? I don't think we do and since most of the users are probably using non-scene namings (without codec or source, but only with resolution) it made sense to remove that check altogether.

medusa/common.py
```diff
@@ -310,67 +308,62 @@ def scene_quality(name, anime=False):
...
-            if ep.avc and ep.bluray:
-                result = Quality.UHD_4K_BLURAY if not full_res else Quality.UHD_8K_BLURAY
-            # WEB-DL
-            elif (ep.avc and ep.itunes) or ep.web:
-                result = Quality.UHD_4K_WEBDL if not full_res else Quality.UHD_8K_WEBDL
-            # HDTV
-            elif ep.avc and ep.tv == 'hd':
-                result = Quality.UHD_4K_TV if not full_res else Quality.UHD_8K_TV
+        if ep.vres in [2160, 4320]:
+            is_4320p = ep.vres == 4320
```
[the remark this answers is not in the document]
medariox: Good point! Will change it later.

medusa/common.py
sharkykh: Ah, very well, makes sense.

PR #4448, opened 2018-06-22
Fix TorrentDay not returning results with invalid JSON

medariox:

### DIFF
--- a/medusa/providers/torrent/json/torrentday.py
+++ b/medusa/providers/torrent/json/torrentday.py
@@ -4,6 +4,7 @@
 
 from __future__ import unicode_literals
 
+import json
 import logging
 import re
 
@@ -96,9 +97,9 @@ class TorrentDayProvider(TorrentProvider):
                     continue
 
                 try:
-                    jdata = response.json()
-                except ValueError:
-                    log.debug('No data returned from provider')
+                    jdata = json.loads(response.content.decode('unicode-escape'))
+                except ValueError as error:
+                    log.error("Couldn't deserialize JSON document. Error: {0!r}", error)
                     continue
 
                 results += self.parse(jdata, mode)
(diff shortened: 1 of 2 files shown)


medusa/providers/torrent/json/torrentday.py
```diff
@@ -96,9 +97,9 @@ def search(self, search_strings, age=0, ep_obj=None, **kwargs):
                     continue
 
                 try:
-                    jdata = response.json()
-                except ValueError:
-                    log.debug('No data returned from provider')
+                    jdata = json.loads(response.content.decode('unicode-escape'))
```
sharkykh: Think about `{"name": "Escaped chars test: \"torrent\". It\'s dangerous..."}`.
```python
>>> print ur'{"name": "Escaped chars test: \"torrent\". It\'s dangerous..."}'
{"name": "Escaped chars test: \"torrent\". It\'s dangerous..."}
>>> print ur'{"name": "Escaped chars test: \"torrent\". It\'s dangerous..."}'.decode('unicode-escape')
{"name": "Escaped chars test: "torrent". It's dangerous..."}
```

Maybe just replace all `r"\'"` (= `"\\'"`)  with `"'"`?

labrys: In my opinion that's error-prone.  A batteries included option: `pip install dirtyjson`
```python
import dirtyjson as json
json.loads(response.content)
```

labrys: @sharkykh You're likely not doing any `dump(s)` in provider code, so importing as json shouldn't be a problem, however per the zen of python "Explicit is better than implicit." so keeping it as dirtyjson is reasonable.

medariox: Done. Thanks for the good suggestions 👍

PR #4454, opened 2018-06-23
Patch race condition when changing schedule layout

sharkykh:
@OmgImAlexis 
This is where I found out [about this](https://github.com/pymedusa/Medusa/pull/4361#discussion_r197620353):
> I know this is already merged but working on something else I noticed the `.then()` doesn't wait for the API call to return.
Don't know exactly what the issue is, but try looking on the network tab (enable `preserve log` if needed), on the Waterfall section of the network tab, you can see that the request issued by `.then()` doesn't wait for the previous one to finish.


OmgImAlexis:
Is there anything we can do server side for this as this is still a race condition with the server returning the old value even after we PATCH as long as it's before it can update the `app.*` vars.

PR #4456, opened 2018-06-24
WebSocket: Message class and configUpdated event

sharkykh:


sharkykh:
I can easily add an ISO timestamp to each message, meaning when was it created, or when it was pushed.
It would usually be the same, so let's go with created.
Will that help with anything?

OmgImAlexis:
@sharkykh didn't we decide it wasn't needed?

sharkykh:
@OmgImAlexis We talked about the notification `hash`, which I said isn't required for other events.
With a timestamp we can potentially compare to make sure users don't get spammed with notifications from the server side, then there shouldn't be any need to save them in store?

OmgImAlexis:
I think for now let's leave it since we don't use a timestamp anywhere.

sharkykh:
Rebased. Needs testing again.

PR #4462, opened 2018-06-25
Vueify/misc changes

OmgImAlexis:
Check commit messages.


themes-default/slim/views/history.mako
```diff
@@ -105,22 +109,22 @@ const startVue = () => {
...
-                        <option value="10" ${'selected="selected"' if limit == 10 else ''}>10</option>
-                        <option value="25" ${'selected="selected"' if limit == 25 else ''}>25</option>
-                        <option value="50" ${'selected="selected"' if limit == 50 else ''}>50</option>
-                        <option value="100" ${'selected="selected"' if limit == 100 else ''}>100</option>
-                        <option value="250" ${'selected="selected"' if limit == 250 else ''}>250</option>
-                        <option value="500" ${'selected="selected"' if limit == 500 else ''}>500</option>
-                        <option value="750" ${'selected="selected"' if limit == 750 else ''}>750</option>
-                        <option value="1000" ${'selected="selected"' if limit == 1000 else ''}>1000</option>
-                        <option value="0"   ${'selected="selected"' if limit == 0   else ''}>All</option>
+                        <option value="10" :selected="limit === 10">10</option>
```
p0psicles: Cant this be a little cleaner using v-for?

OmgImAlexis: I forgot about `v-model`. 🙄

themes-default/slim/static/js/store.js
```diff
@@ -9,7 +9,7 @@ Vue.use(Puex);
 // There are no naming conventions so try and match
 // similarly to what we already use when adding new ones.
 const mutationTypes = {
-    LOGIN_PENDING: '🔒 Login Pending',
+    LOGIN_PENDING: '🔒 Logging in..',
```
sharkykh: Three dots or [&#8230; (ellipsis)](https://www.fileformat.info/info/unicode/char/2026/index.htm) please.

themes-default/slim/static/js/store.js
```diff
@@ -50,6 +50,16 @@ const {
 
 const store = new Puex({
     state: {
+        auth: {
+            // @TODO: Set this as false when we add in vue(jwt) login
```
sharkykh: What is this `@TODO` for? The value is already false.

themes-default/slim/static/js/store.js
```diff
@@ -50,6 +50,16 @@ const {
...
+            isAuthenticated: false,
+            user: {},
+            tokens: {
+                access: null,
+                refresh: null
+            },
+            err: null
```
sharkykh: I thought non-abbreviations were preferred :stuck_out_tongue:

themes-default/slim/static/js/store.js
```diff
@@ -338,6 +440,14 @@ const store = new Puex({
...
+            const { dispatch } = context;
+
+            if (section !== 'main') {
+                return;
+            }
+            return api.patch('config/' + section, config).then(setTimeout(() => dispatch('getConfig'), 500));
+        },
         getShow(context, { indexer, id }) {
             return api.get('/series/' + indexer + id).then(res => {
                 store.commit(ADD_SHOW, res.data);
```
sharkykh: Some actions are using `store.commit` and some `context.commit`.
Can we stick to one or the other?
I don't understand the difference, which means I'm bound to use the incorrect one.

themes-default/slim/views/config_search.mako
```diff
@@ -14,9 +14,12 @@ window.app = {};
 const startVue = () => {
     window.app = new Vue({
         el: '#vue-wrap',
+        store,
         metaInfo: {
             title: 'Config - Episode Search'
         },
+        computed: Object.assign(store.mapState([]), {
```
sharkykh: You pull the config but don't assign anything. Missed it?
Now that `config` is available from the global mixin,
maybe you don't need to pull the config here or map the state at all?

themes-default/slim/views/editShow.mako
```diff
@@ -16,7 +16,6 @@ const startVue = () => {
                 seriesSlug: $('#series-slug').attr('value'),
                 seriesId: $('#series-id').attr('value'),
                 indexerName: $('#indexer-name').attr('value'),
-                config: MEDUSA.config,
```
sharkykh: What? Why? It's being used.
https://github.com/pymedusa/Medusa/blob/b8207ce7ec20efa971bbb998933b00ca61c5e61b/themes-default/slim/views/editShow.mako#L142-L145

Unless you meant to replace it with the store config...

themes-default/slim/views/history.mako
```diff
@@ -22,17 +22,34 @@ const startVue = () => {
...
+                },
+                set(layout) {
+                    const { $store } = this;
+                    const page = 'history';
+                    $store.dispatch('setLayout', { page, layout });
+                }
+            }
+        }),
         mounted() {
+            const { layout, config } = this;
```
sharkykh: I didn't checkout the branch, but...
[We already had an issue with this on the schedule page.](https://github.com/pymedusa/Medusa/commit/7cd3d70a682ef13e9001c2ed2c530cf13856b0ef)

themes-default/slim/views/history.mako
```diff
@@ -104,23 +108,23 @@ const startVue = () => {
...
-                        <option value="50" ${'selected="selected"' if limit == 50 else ''}>50</option>
-                        <option value="100" ${'selected="selected"' if limit == 100 else ''}>100</option>
-                        <option value="250" ${'selected="selected"' if limit == 250 else ''}>250</option>
-                        <option value="500" ${'selected="selected"' if limit == 500 else ''}>500</option>
-                        <option value="750" ${'selected="selected"' if limit == 750 else ''}>750</option>
-                        <option value="1000" ${'selected="selected"' if limit == 1000 else ''}>1000</option>
-                        <option value="0"   ${'selected="selected"' if limit == 0   else ''}>All</option>
-                    </select>
+                <select v-model="limit" name="history_limit" id="history_limit" class="form-control form-control-inline input-sm">
+                    <option value="10">10</option>
```
sharkykh: I don't believe these would work, because `limit` is an integer but these values are parsed as strings.

themes-default/slim/views/manage_failedDownloads.mako
```diff
@@ -21,7 +21,8 @@ const startVue = () => {
         },
         data() {
             return {
-                header: 'Failed Downloads'
+                header: 'Failed Downloads',
+                limit: '${limit}'
```
sharkykh: Just note that here you do it differently than on `history.mako` (`limit: parseInt('${limit}', 10)`)

themes-default/slim/views/layouts/main.mako
```diff
@@ -133,6 +133,25 @@
...
+                        // We wait 1000ms to allow the mutations to show in vue dev-tools
+                        // Please see https://github.com/egoist/puex/issues/8
+                        setTimeout(() => {
+                            const { store } = window;
+                            store.dispatch('login');
+                            store.dispatch('getConfig');
+                        }, 1000);
+                    }
+                },
+                // Make auth and config accessable to all components
```
sharkykh: `accessible` /hides

themes-default/slim/static/js/core.js
```diff
@@ -33,27 +33,12 @@ const UTIL = {
         }
 
         const body = document.body;
-        $('[asset]').each(function() {
```
sharkykh: Be wary. I tried to do that and it broke all kinds of stuff related to assets when combined with `Vue`, `VueInViewport`, `isotope`, `tablesorter`, etc...

themes-default/slim/static/js/store.js
```diff
@@ -247,9 +320,21 @@ const store = new Puex({
...
         [LOGIN_PENDING]() {},
-        [LOGIN_SUCCESS]() {},
-        [LOGIN_FAILED]() {},
-        [LOGOUT]() {},
+        [LOGIN_SUCCESS](state, user) {
+            state.auth.user = user;
+            state.auth.isAuthenticated = true;
+            state.auth.err = null;
+        },
+        [LOGIN_FAILED](state, err) {
```
sharkykh: `err` is an object containing `{ err, credentials }` as can be seen on line 424:
https://github.com/pymedusa/Medusa/blob/b8207ce7ec20efa971bbb998933b00ca61c5e61b/themes-default/slim/static/js/store.js#L424

themes-default/slim/views/history.mako
OmgImAlexis: This uses `setLayout` which currently reloads the page so it should be fine.

themes-default/slim/views/history.mako
OmgImAlexis: It worked fine for all of the other number based selects so it should be fine here.

OmgImAlexis: Fixed.

themes-default/slim/static/js/core.js
OmgImAlexis: I can't find `asset=` anywhere in the themes directory so I don't think we have anything left that uses it even in js. 🙏 I could be wrong though.

themes-default/slim/views/history.mako
sharkykh: It's not the same issue. The issue was they weren't available **when the page loaded**.

themes-default/slim/views/config_search.mako
```diff
@@ -18,8 +18,6 @@ const startVue = () => {
         metaInfo: {
             title: 'Config - Episode Search'
         },
-        computed: Object.assign(store.mapState([]), {
-        }),
```
sharkykh: You're still pulling the config on this file: https://github.com/pymedusa/Medusa/blob/520807626e929bd657c38f5892908d95359ce609/themes-default/slim/views/config_search.mako#L206-L207

themes-default/slim/views/history.mako
OmgImAlexis: Since it's injected to the root instance it should be available in the store before this component is even loaded as mixins run their mounted, etc. before the component itself.

You should actually be able to revert the change you needed for schedule if you're using this branch.

themes-default/slim/views/config_search.mako
```diff
@@ -202,6 +203,9 @@ const startVue = () => {
             };
         },
         mounted() {
+            const { $store } = this;
+            $store.dispatch('getConfig');
```
sharkykh: I thought this wasn't needed anymore.

themes-default/slim/views/history.mako
sharkykh: I just checked and this *is* an issue (on history page, with subtitles enabled),
unless something else is wrong.

themes-default/slim/static/js/store.js
```diff
@@ -448,19 +450,20 @@ const store = new Puex({
             return api.patch('config/' + section, config).then(setTimeout(() => dispatch('getConfig'), 500));
         },
         getShow(context, { indexer, id }) {
+            const { commit } = context;
             return api.get('/series/' + indexer + id).then(res => {
-                store.commit(ADD_SHOW, res.data);
+                commit(ADD_SHOW, res.data);
```
sharkykh: Sorry if I caused it, but on schedule page this error pops up:
```
Uncaught (in promise) TypeError: Cannot read property 'mutations' of undefined
    at commit (puex.js:136)
    at shows.forEach.show (store.js:466)
    at Array.forEach (<anonymous>)
    at api.get.then.res (store.js:465)
```

OmgImAlexis: Try the latest commit I pushed. I think I fixed this.

themes-default/slim/views/history.mako
OmgImAlexis: If you add `console.log(config.subtitles.enabled)` under this what do you get?

sharkykh: I put the log below this line
And another log above https://github.com/pymedusa/Medusa/blob/990d701805c2ea959678c47093971afe47ce26f2/themes-default/slim/views/history.mako#L67

`null`. Both locations.

OmgImAlexis: What if you use `this.config` in both places?

themes-default/slim/static/js/store.js
sharkykh: Yes, looks good now, I'll sync these to the other themes.

By the way, schedule is now a bit outdated. I'll fix that.

themes-default/slim/views/history.mako
sharkykh: It's the same.
The issue is when mounted is called, the config has yet to be committed to the store.
Vue doesn't wait for the global mixin's `created()` to finish getting the config.

OmgImAlexis: What do you suggest we switch this to then?

sharkykh: Actually no, I keep getting different errors. But I already messed with some files so please hold.

sharkykh: Actually no, the one before `if (config.subtitles.enabled) { ` line
can't use `this` because it's jQuery, I believe.

themes-default/slim/static/js/core.js
sharkykh: I agree that `asset=` is no longer being used.

themes-default/slim/views/history.mako
sharkykh: @OmgImAlexis 
I don't know if there's anything we can do about it right now.
Only solution I have in mind is the one I used on `schedule.mako`, which works on this page too.

sharkykh:
@OmgImAlexis 
I'll update this comment with issues as I find them.

- **`/home/`**: Posters not showing up before you scroll. This is probably because isotope changes the DOM.

themes-default/slim/views/partials/home/poster.mako
```diff
@@ -155,7 +155,7 @@
                                     <td class="show-table">
                                     % if cur_show.network:
                                         <span title="${cur_show.network}">
-                                            <asset default="images/network/nonetwork.png" series-slug="${cur_show.slug}" type="network" cls="show-network-image" :link="false" alt="${cur_show.network}" title="${cur_show.network}"></asset>
+                                            <asset default="images/network/nonetwork.png" series-slug="${cur_show.slug}" lazy="false" type="network" cls="show-network-image" :link="false" alt="${cur_show.network}" title="${cur_show.network}"></asset>
```
p0psicles: Your setting it to the string 'false' not the boolean. Also its the default value. So you can just leave out right?

sharkykh:
You changed the network asset by mistake and as @p0psicles said you set `lazy` to a string and not the boolean.
But, after correcting those, it seems to fix both issues.

themes-default/slim/views/partials/home/poster.mako
sharkykh: Your're correct about the false not being a boolean, but the default value is set to true (as it should), so we can't leave it out.
Although it won't stay there forever.

OmgImAlexis:
@sharkykh I think I did the rebase correctly.

PR #4471, opened 2018-06-26
Fix "already added" link using incorrect indexer and show id

sharkykh:
It was using the indexer's name and show id instead of the values of the show added to the library.
Made the `displayShow` link invalid as it pointed to a show that didn't exist in the library.

### DIFF
--- a/medusa/server/api/v2/internal.py
+++ b/medusa/server/api/v2/internal.py
@@ -208,10 +208,11 @@ class InternalHandler(BaseRequestHandler):
         # Get all possible show ids
         all_show_ids = {}
         for show in app.showList:
-            for external in show.externals:
-                indexer_id = reverse_mappings.get(external)
-                show_id = show.externals[external]
-                all_show_ids.setdefault(indexer_id, []).append(show_id)
+            for ex_indexer_name, ex_show_id in iteritems(show.externals):
+                ex_indexer_id = reverse_mappings.get(ex_indexer_name)
+                if not ex_indexer_id:
+                    continue
+                all_show_ids[(ex_indexer_id, ex_show_id)] = (show.indexer_name, show.series_id)
 
         for indexer, shows in iteritems(results):
             indexer_api = indexerApi(indexer)
@@ -228,7 +229,7 @@ class InternalHandler(BaseRequestHandler):
                         show['firstaired'] or 'N/A',
                         show.get('network', '').encode('utf-8') or 'N/A',
                         sanitize_filename(show['seriesname']).encode('utf-8'),
-                        show_id in all_show_ids.get(indexer, []) and (indexer, show_id)
+                        all_show_ids.get((indexer, show_id), False)
                     )
                 )
 
--- a/themes-default/slim/views/addShows_newShow.mako
+++ b/themes-default/slim/views/addShows_newShow.mako
@@ -370,8 +370,8 @@ const startVue = () => {
                         alreadyAdded = (() => {
                             if (!alreadyAdded) return false;
                             // Extract existing show info
-                            const [ mIndexerName, mShowId ] = alreadyAdded;
-                            return 'home/displayShow?indexername=' + mIndexerName + '&seriesid=' + mShowId;
+                            const [ matchIndexerName, matchShowId ] = alreadyAdded;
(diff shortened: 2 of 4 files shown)


medusa/server/api/v2/internal.py
```diff
@@ -208,10 +208,11 @@ def resource_search_indexers_for_show_name(self):
...
         for show in app.showList:
-            for external in show.externals:
-                indexer_id = reverse_mappings.get(external)
-                show_id = show.externals[external]
-                all_show_ids.setdefault(indexer_id, []).append(show_id)
+            for ex_indexer_name, ex_show_id in iteritems(show.externals):
+                ex_indexer_id = reverse_mappings.get(ex_indexer_name)
+                if not ex_indexer_id:
+                    continue
+                all_show_ids[(ex_indexer_id, ex_show_id)] = (show.indexer_name, show.series_id)
```
sharkykh: The changes made here are:
 - Iterating over the show externals more efficiently (using `iteritems` instead of `external` + `show.externals[external]`)
 - The dictionary structure is now:
```py
all_show_ids = {
    # Key: Show external value: (indexer id, show id)
    # Value: Show in library: (indexer name, show id) -- notice: name
    # For a show that was added from TVDB:
    (1, 123456) : ('tvdb', 123456),
    (3, 456123) : ('tvdb', 123456),
    (4, 516243) : ('tvdb', 123456),
    # ... for each show in the library
}
```

medusa/server/api/v2/internal.py
```diff
@@ -228,7 +229,7 @@ def resource_search_indexers_for_show_name(self):
                         show['firstaired'] or 'N/A',
                         show.get('network', '').encode('utf-8') or 'N/A',
                         sanitize_filename(show['seriesname']).encode('utf-8'),
-                        show_id in all_show_ids.get(indexer, []) and (indexer, show_id)
+                        all_show_ids.get((indexer, show_id), False)
```
sharkykh: The dictionary structure above makes it easy to get the key using the current search result.
If found, the value is returned, e.g. `('tvdb', 123456)`, to be used when generating the link to the current show page.
If not found, False is returned - meaning the show is not already in the library.

themes-default/slim/views/addShows_newShow.mako
```diff
@@ -370,9 +370,8 @@ const startVue = () => {
                         alreadyAdded = (() => {
                             if (!alreadyAdded) return false;
                             // Extract existing show info
-                            const [ mIndexerId, mShowId ] = alreadyAdded;
-                            const indexerIdentifier = indexers[mIndexerId] ? indexers[mIndexerId].identifier : mIndexerId;
-                            return 'home/displayShow?indexername=' + indexerIdentifier + '&seriesid=' + mShowId;
+                            const [ mIndexerName, mShowId ] = alreadyAdded;
+                            return 'home/displayShow?indexername=' + mIndexerName + '&seriesid=' + mShowId;
```
sharkykh: Just updating the code here to account for the changes in backend.
No need to get the indexer name, as we already know it.

themes-default/slim/views/addShows_newShow.mako
```diff
@@ -370,9 +370,8 @@ const startVue = () => {
                         alreadyAdded = (() => {
                             if (!alreadyAdded) return false;
                             // Extract existing show info
-                            const [ mIndexerId, mShowId ] = alreadyAdded;
-                            const indexerIdentifier = indexers[mIndexerId] ? indexers[mIndexerId].identifier : mIndexerId;
-                            return 'home/displayShow?indexername=' + indexerIdentifier + '&seriesid=' + mShowId;
+                            const [ mIndexerName, mShowId ] = alreadyAdded;
```
OmgImAlexis: What's up with the `m` at the start of the var name?

sharkykh: I think it was `m` for matched.
I just wanted to distinguish those vars from the ones up top.
I realize it's not really needed, but I feel like it's a bit more readable like that.

OmgImAlexis: Since we're in a different scope maybe change them over or use `match` since it's really that obvious the `m` is for?

PR #4479, opened 2018-06-27
remove unused isMeta

OmgImAlexis:
I can't find any references to this in our code so we should be fine to remove it.


OmgImAlexis:
@sharkykh sync and lint commits added.

PR #4480, opened 2018-06-27
Fix SceneTime and add missing pubdate

duramato:

Fixes SceneTime parsing and adds missing pubdate.
Also adds missing cats.

### DIFF
--- a/medusa/providers/torrent/html/scenetime.py
+++ b/medusa/providers/torrent/html/scenetime.py
@@ -33,7 +33,7 @@ class SceneTimeProvider(TorrentProvider):
         self.url = 'https://www.scenetime.com'
         self.urls = {
             'login': urljoin(self.url, 'login.php'),
-            'search': urljoin(self.url, 'browse_API.php'),
+            'search': urljoin(self.url, 'browse.php'),
             'download': urljoin(self.url, 'download.php/{0}/{1}'),
         }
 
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/html/scenetime.py
```diff
@@ -131,9 +132,9 @@ def parse(self, data, mode):
                     continue
 
                 try:
-                    link = cells[labels.index('Name')].find('a')
-                    torrent_id = link['href'].replace('details.php?id=', '').split('&')[0]
-                    title = link.get_text(strip=True)
+                    link = cells[labels.index('Name')]
+                    torrent_id = link.find('a')['href'].replace('details.php?id=', '').split('&')[0]
+                    title = link.find('a').get_text(strip=True)
```
medariox: Why this change?

medusa/providers/torrent/html/scenetime.py
```diff
@@ -112,7 +113,7 @@ def parse(self, data, mode):
         items = []
 
         with BS4Parser(data, 'html5lib') as html:
-            torrent_rows = html.find_all('tr')
+            torrent_rows = html.find('div', {'id": 'torrenttable'}).find_all('tr')
```
medariox: I think the preferred syntax is: `.find('div', id='torrenttable')`

medusa/providers/torrent/html/scenetime.py
duramato: That way I can use the `link`, already defined before, for the lineage. And isn't defined again twice.

medusa/providers/torrent/html/scenetime.py
sharkykh: Also note the quotation marks don't match.

duramato: Changed.

medusa/providers/torrent/html/scenetime.py
medariox: I see it now, all good.

PR #4075, opened 2018-04-22
Fixes for editShow

sharkykh:
To do:
- [x] Fix show location file browser, autocomplete, saving location.
       Not every file browser needs a localStorage key. Show location doesn't (removed)
       Closes #4072 
- [x] Sync changes to slim + dark themes
- [x] Fix #4076 
- [x] API has a patch for `series.allowed_qualities` and `series.preferred_qualities`, yet those properties have no `@[property].setter` function in `medusa/tv/series.py@Series`
- [x] Fix saving black/white lists.
- [x] Fix saving default episode status - currently not being saved at all.
- [x] Restore POST functionally to editShow because [**massEditSubmit still uses it**](https://github.com/pymedusa/Medusa/blob/130cfca37762342537d7d19720e42635a2fbe7b9/medusa/server/web/manage/handler.py#L681-L688) ~(all themes), and 
because **legacy theme** still uses that method for saving shows.~

PR #4486, opened 2018-06-28
Convert editShow to Vue store and SFC

OmgImAlexis:
# TODO
- [x] Fix anidb UI
- [x] ~double check some weird bug in massUpdate when checking multiple shows and running update~ Can't reproduce it anymore ??


themes/dark/templates/editShow.mako
```diff
@@ -43,121 +42,133 @@ const startVue = () => {
...
-            const url = 'series/' + seriesSlug;
-            try {
-                const response = await api.get('series/' + this.seriesSlug);
-                this.series = Object.assign({}, this.series, response.data);
-                this.series.language = response.data.language;
-            } catch (error) {
-                console.debug('Could not get series info for: '+ seriesSlug);
+        computed: Object.assign(store.mapState(['shows']), {
+            params() {
+                return location.search.slice(1).split('&').reduce((obj, pair) => {
```
sharkykh: @OmgImAlexis @p0psicles 
A really cool thing I recently found out about:
You can use VueRouter for this, even if it's not a route!
Add `router` to this Vue instance (like you added store above),
then you can use: `this.$route.query` instead of this. :tada: 
image

p0psicles: Ooh cool!

sharkykh:
@OmgImAlexis 
The state's [`defaults.show`](https://github.com/pymedusa/Medusa/blob/38d911e0d5151527c8e77ac1003ce1a668e739fa/themes-default/slim/static/js/store.js#L288-L317) needs to be updated to the following:
<details>

```js
            show: {
                id: {
                    slug: null
                },
                rating: {
                    imdb: {
                        rating: null,
                        votes: null
                    }
                },
                country_codes: [], // eslint-disable-line camelcase
                network: null,
                airs: null,
                config: {
                    qualities: {
                        allowed: [],
                        preferred: []
                    },
                    defaultEpisodeStatus: null,
                    dvdOrder: null,
                    seasonFolders: null,
                    scene: null,
                    sports: null,
                    paused: null,
                    location: null,
                    airByDate: null,
                    release: {
                        requiredWords: [],
                        ignoredWords: []
                    },
                    aliases: [],
                    subtitlesEnabled: null,
                    anime: null
                }
            }
```
</details>

themes-default/slim/views/editShow.mako
```diff
@@ -43,121 +42,133 @@ const startVue = () => {
...
+            id() {
+                return this.params.seriesid;
+            },
+            indexer() {
+                return this.params.indexername;
+            },
+            // @TODO: Enable this once we remove this.series
+            // show() {
+            //     const { $store } = this;
+            //     return this.shows.length === 0 ? $store.defaults.show : this.shows.find(show => show.indexer === this.indexer && Number(show.id[show.indexer]) === Number(this.id));
```
sharkykh: 1. This wouldn't work as is (`$store` is supposed to be `$store.state`).
2. I'd prefer if you split it into multiple lines.
3. I have a suggestion.
```js
            show() {
                const { $store, id, indexer } = this;
                const { state } = $store;
                if (this.shows.length === 0 || !indexer || !id) return state.defaults.show;
                const slug = indexer + id.toString();  // or String(id)
                return this.shows.find(show => show.id.slug === slug);
            },
```

And lastly: Wouldn't we still need some form of a `show` object anyway?
What object will be updated when the users changes the values?
I'm guessing here that it would be the new `show` object in the store state.
But if that's the case you won't be able to have 2 properties with the same name... So, how?

themes-default/slim/views/editShow.mako
```diff
@@ -43,121 +42,133 @@ const startVue = () => {
...
+            },
+            displayShowUrl() {
+                // @TODO: Change the URL generation to use `this.show`. Currently not possible because
+                // the values are not available at the time of app-link component creation.
+                return window.location.pathname.replace('editShow', 'displayShow') + window.location.search;
             }
-            this.seriesLoaded = true;
+        }),
+        mounted() {
+            const { $store, indexer, id} = this;
```
sharkykh: Missing a space here.

themes-default/slim/views/editShow.mako
```diff
@@ -43,121 +42,133 @@ const startVue = () => {
...
+                return this.saving === false ? 'Save Changes' : 'Saving...';
+            },
+            displayShowUrl() {
+                // @TODO: Change the URL generation to use `this.show`. Currently not possible because
+                // the values are not available at the time of app-link component creation.
+                return window.location.pathname.replace('editShow', 'displayShow') + window.location.search;
             }
-            this.seriesLoaded = true;
+        }),
+        mounted() {
```
sharkykh: Would `created()` be better for this? It might speed it up a little.

themes-default/slim/views/editShow.mako
```diff
@@ -171,180 +182,180 @@ const startVue = () => {
 <input type="hidden" id="series-slug" value="${show.slug}" />
```
sharkykh: These are still here, but they're unused now.

themes-default/slim/views/editShow.mako
```diff
@@ -11,24 +11,23 @@ const startVue = () => {
         metaInfo: {
             title: 'Edit Show'
         },
+        store,
         data() {
             return {
                 seriesSlug: $('#series-slug').attr('value'),
```
sharkykh: You may remove that too, it's unused.

themes/dark/templates/editShow.mako
OmgImAlexis: Oh I knew that but I didn't want to add the router as we don't need it right now and didn't want to get ahead of myself again.

p0psicles:
@OmgImAlexis could you finish this one. As im having some issues with store in the postprocessing pr. Like select-list-ui component is not updating its list after its mounted. Looks like the watch is not working on values that come from a computed.

sharkykh:
I believe most of my review still applies, but it's now hidden because of the rebase.
https://github.com/pymedusa/Medusa/pull/4486#pullrequestreview-132918689

p0psicles:
I'm hijacking this PR.

themes-default/slim/src/components/edit-show.vue
```diff
@@ -0,0 +1,363 @@
...
+                language: this.show.language
+            };
+
+            if (data.config.anime) {
+                data.config.release.blacklist = this.show.config.release.blacklist;
+                data.config.release.whitelist = this.show.config.release.whitelist;
+            }
+
+            const { indexer, id, $store } = this;
+            $store.dispatch('setShow', { indexer, id, data, save: true }).then(() => {
```
p0psicles: @sharkykh I'm not totally sure. Should we call an action in vuex, and have that 1. update store then 2. update the show using apiv2? Or update the show from here, like it's now?

themes-default/slim/src/components/edit-show.vue
```diff
@@ -0,0 +1,363 @@
...
+        showUrl() {
+            // @TODO: Change the URL generation to use `this.series`. Currently not possible because
+            // the values are not available at the time of app-link component creation.
+            return window.location.pathname.replace('editShow', 'displayShow') + window.location.search;
+        },
+        globalIgnored() {
+            return this.$store.state.search.filters.ignored.map(x => x.toLowerCase());
+        },
+        globalRequired() {
+            return this.$store.state.search.filters.required.map(x => x.toLowerCase());
```
p0psicles: This was previously `this.$store.state.search.filters.ignored.map(x => x.toLowerCase());`

medusa/server/web/home/handler.py
```diff
@@ -1131,281 +1131,21 @@ def check_show_for_language(series_obj, language):
...
             return True
 
-    def editShow(self, indexername=None, seriesid=None, location=None, allowed_qualities=None, preferred_qualities=None,
-                 exceptions_list=None, season_folders=None, paused=None, directCall=False,
-                 air_by_date=None, sports=None, dvd_order=None, indexer_lang=None,
-                 subtitles=None, rls_ignore_words=None, rls_require_words=None,
-                 anime=None, blacklist=None, whitelist=None, scene=None,
-                 defaultEpStatus=None, quality_preset=None):
+    def editShow(self, indexername=None, seriesid=None):
+        t = PageTemplate(rh=self, filename='index.mako')
```
sharkykh: If you want to remove this code from here, move it to `massEdit`

Reminding you this: #4075
> * [x]  Restore POST functionally to editShow because [**massEditSubmit still uses it**](https://github.com/pymedusa/Medusa/blob/130cfca37762342537d7d19720e42635a2fbe7b9/medusa/server/web/manage/handler.py#L681-L688)

and this commit: https://github.com/pymedusa/Medusa/pull/4075/commits/2400d77f78082f7101000501e4de505f3799b89c

p0psicles: Tnx for the reminder.
My main focus now was to get the page ~~working~~ loading. So that's working.
Next will be getting the saving working. But I need to know how we're going to update store/api.

themes-default/slim/src/components/edit-show.vue
sharkykh: TL;DR: The action should call the patch and fetch the show again. I don't see a way around this at the moment.

I didn't look at the rest of the code, but since some of the patch data can be ignored by the API handler (invalid values and such), and whatever you send does not necessarily become the new show config, I'd say we can't update the store using the same data that we're sending, and we must fetch the show after the patch is complete.

medusa/server/web/home/handler.py
p0psicles: I've choosen to create a new method `massEditShow` and move it to there.
Felt a little bit less messy for now.

themes-default/slim/src/store/modules/shows.js
```diff
@@ -40,7 +40,27 @@ const mutations = {
...
+     * vue will not detect any changes in `state.shows`. We would have to add another watch, for that. While the computed
+     * should be able to handle that.
+     * @param {object} _ - state by is not used
+     * @param {array} shows - array with shows
+     * @param {number} id - show's id
+     * @param {string} indexer - show's indexer
+     * @returns {object} show object or null
+     */
+    getShowByIdFromShows: _ => ({ shows, id, indexer }) => {
+        const show = shows.find(show => Number(show.id[indexer]) === Number(id));
```
p0psicles: @sharkykh passing in the shows, I can have vue computed take care of the reactiveness.
I worked around the "issue" before by added the `currentShow` https://github.com/pymedusa/Medusa/pull/4486/files#diff-4327d3f74ed18de447b2fc16ffe81b5eR7 object.
But this feels better. What do you think?

themes-default/slim/src/components/edit-show.vue
```diff
@@ -0,0 +1,355 @@
...
+            const showRequired = this.show.config.release.requiredWords.map(x => x.toLowerCase());
+
+            if (this.show.config.release.requiredWordsExclude) {
+                return arrayExclude(globalRequired, showRequired);
+            }
+
+            return arrayUnique(globalRequired.concat(showRequired));
+        },
+        show() {
+            const { defaultShow, id, indexer, getShowByIdFromShows, shows } = this;
```
p0psicles: show is now a small computed property.

p0psicles:
@sharkykh editShow and displayShow share a lot of common solutions.
Shall we focus on this for now, bump the quality, and apply the same in vueify display-show?

themes-default/slim/src/store/modules/shows.js
sharkykh: Okay, umm, you created a getter that doesn't use the store state, which seems counter-intuitive (to me).
But I don't know what reactiveness issue you are trying to work around.

My point is, the way you've used this getter in `edit-show.vue` makes it pretty much **identical** to `getShowById` defined above.
I don't see a reason to pass the "shows" state around - using `mapState` to get it in `edit-show.vue`, and then passing it back to the getter that **already has access to it** from the `state` parameter.

You can just use `getShowById` IMO.

p0psicles: Ok Lets user getShowById as An example.
Just swap it in. Then at mount time state.shows.shows is empty. So we'll start with An empty array. And show Will be `null`.

How would you fix that?

themes-default/slim/src/components/edit-show.vue
```diff
@@ -0,0 +1,436 @@
...
+            this.show.config.release.ignoredWords = items.map(item => item.value);
+        },
+        onChangeRequiredWords(items) {
+            this.show.config.release.requiredWords = items.map(item => item.value);
+        },
+        onChangeAliases(items) {
+            this.show.config.aliases = items.map(item => item.value);
+        },
+        onChangeReleaseGroupsAnime(groupNames) {
+            this.series.config.release.whitelist = groupNames.whitelist;
```
p0psicles: This.show

themes-default/slim/src/components/edit-show.vue
```diff
@@ -0,0 +1,437 @@
...
+                return arrayExclude(globalRequired, showRequired);
+            }
+
+            return arrayUnique(globalRequired.concat(showRequired));
+        }
+    },
+    created() {
+        const { getShows } = this;
+        // Without getting any specific show data, we pick the show needed from the shows array.
+        getShows();
```
sharkykh: @p0psicles Are we getting all the shows in order to show the `ShowSelector`?
It's currently not visible [because it's limited to `displayShow` only:

https://github.com/pymedusa/Medusa/blob/63a6b28789bebb3f8621bfecf816d5bedfe61455/themes-default/slim/src/components/sub-menu.vue#L43-L46

p0psicles: No we're getting the shows just for that one show we want to edit. So maybe we shouldn't do that

sharkykh: Okay, although I'm keep having an reproducible issue - the show is not available at render time because we're fetching from an API.
I will have to continue this tomorrow... but I'll probably go with a `v-if` and a loading spinner.

p0psicles: Mm how is that different from getting all shows?

sharkykh: Couldn't let this go and turns out it was just bad `QualityChooser` design on my part...
But to answer your question, I think it's just a simple race condition.

p0psicles:
@sharkykh can you do a yarn dev for me? I still have my build env messed up

sharkykh:
Do you think we should merge it now?
Hopefully we didn't miss anything major...

p0psicles:
yeah lets merge

PR #4488, opened 2018-06-28
Add PATCH capabilities for APIv2 episodes

medariox:

This PR adds `PATCH` support for single and multiple episodes.
For now, only `quality` and `status` is supported.

Example (multiple):
```
PATCH /api/v2/series/tvdb123/episodes
{
	 "s01e02": {"status": 3, "quality": 64},
	 "s02e0333": {"status": 2, "quaality": 4},
	 "s02e5": {"status": 1, "quality": 32}
}
Response (207):
{
	"s02e5": {
		"status": 200
	},
	"s01e02": {
		"status": 200
	},
	"s02e0333": {
		"status": 400
	}
}
```
Example (single):
```
PATCH /api/v2/series/tvdb123/episodes/s1e5
{
	 "status": 1, 
	 "quality": 2
}
Response (200):
{
	"status": 1,
	"quality": 2
}
```

Other changes:
- Rename the `episode` endpoint to `episodes`.
- Use BraceAdapter throughout the APIv2.


sharkykh:
Are you sure it's the best course of action to replace `episode` with episodes?
I'd prefer to have 2 endpoints, where `episodes` only has PATCH enabled, at least for now.
I just don't feel like `episodes/sXeX` is right.

@medariox @OmgImAlexis @p0psicles 
Thoughts?

medariox:
Why don't you feel like `episodes/sXeY` is right? It makes perfect sense to me. What didn't make any sense is using `episode` to get all episodes of a X season, that's why I've changed it. I'm not sure what you mean with the route you posted? We need to get the episodes, not the season. Keeping `episode` also feels redundant and unnecessary.

sharkykh:
1. I didn't realize `/series/:seriesSlug/episode` returned all episodes when the episode slug was omitted.
2. `/series/:seriesSlug/episodes/:episodeSlug` just looked weird at at first glance, don't know how to explain. As I understood it, I thought the routes should be resources and not collections ('episode' rather than 'episodes') but I guess I missed something.
3. The `operation` endpoint?
AFAIK It's not for the season... It currently only has POST for archiving episodes.

OmgImAlexis:
Is there any reason you'd need to update episodes on multiple shows at once?
Otherwise I'd suggest something like this.

`/series/:seriesSlug/seasons/:seasonIdentifer/episodes/:episodeIdentifer`

---

@sharkykh an endpoint can be composed of a resource or a collection of resources.
Episodes happen to be a sub-resource of a show/season.

Think of this as the API nowhere does it make sense for episodes to be outside of show nor does it make much sense to call it "episode".
```js
const show = {
    seasons: [{
        identifier: 's01',
        episodes: [
            identifier: 'e01',
            ...
        ]
    }, {
        identifier: 's01',
        episodes: [
            identifier: 'e01',
            ...
        ]
    }]
};
```

---

@medariox I'm not too happy with how single and multiple return the same key but different results. That'll need to change before this is accepted.

What happens with shows using absolute numbering, etc.? I'm guessing if it doesn't contain `sXeY` you'd fall back to absolute?
```js
const show = {
    seasons: [{
        identifier: '001',
        episodes: []
    }, {
        identifier: '002',
        episodes: []
    }]
};
```

medariox:
> The operation endpoint?
AFAIK It's not for the season... It currently only has POST for archiving episodes.

Oh, I see what you mean now. That doesn't really look like it's what we need tho. We need to be able to update multiple single episodes, so an operation on the whole show is probably not ideal.

> Is there any reason you'd need to update episodes on multiple shows at once?

The code is to patch multiple episodes of a single show. It is useful to update multiple episodes with a single request, I belive it's something we will use quite often.

> @medariox I'm not too happy with how single and multiple return the same key but different results. That'll need to change before this is accepted.

The idea was to allow both type of requests. The response is also a bit diferent. Multiple results will always return `207` with a list of status codes, while single results return the "usual" status codes (200, 400, 404,...). This seems to coherent with how it is described [here](https://apihandyman.io/api-design-tips-and-tricks-getting-creating-updating-or-deleting-multiple-resources-in-one-api-call/) (one of the few sources I could find about this). But I'm really open to changes and other approches, especially since multi-status doesn't seem to be very popular.

 > What happens with shows using absolute numbering, etc.? I'm guessing if it doesn't contain sXeY you'd fall back to absolute?

Any valid show identifier can be used, meaning: `sXeY`, `eXX` or `YYYY-MM-DD`. Currently there is no fallback, the response would just return 404.

labrys:
@OmgImAlexis 

> nowhere does it make sense for episodes to be outside of show nor does it make much sense to call it "episode".

If each episode had a unique ID in the database it would make sense to access an episode outside of a show.  For example some episodes can be contained in multiple shows (e.g. crossovers where its an episode in one show but a special for the other).  Also the unique ID would be fixed regardless of episode numbering changes, date changes, or absolute numbering conventions making it a true canonical reference for that episode.

medariox:
Anyone wants to review this?

medusa/server/api/v2/base.py
```diff
@@ -420,10 +425,9 @@ def patch(self, target, value):
                     setattr(target, self.attr, self.converter(value))
             except AttributeError:
                 log.warning(
-                    'Error trying to change attribute %s on target %s, you sure'
-                    ' you are allowed to change this attribute?',
-                    self.attr,
-                    target
+                    'Error trying to change attribute {attr} on target {target}'
```
sharkykh: I think `{target!r}` is better since it's usually an object.
For example when patching as series it would show up as:
```
Error trying to change attribute status on target indexerid: 1234
indexer: 1
name: Show Name
```
etc...

medusa/server/api/v2/episodes.py
```diff
@@ -0,0 +1,169 @@
...
+
+        episode = Episode.find_by_series_and_episode(series, episode_number)
+        if not episode:
+            return self._not_found('Episode not found')
+
+        accepted = {}
+        ignored = {}
+        patches = {
+            'status': IntegerField(episode, 'status'),
+            'quality': IntegerField(episode, 'quality'),
```
sharkykh: I don't like the fact the patches are declared twice.
Maybe make a `_patch_episode` function that returns true if accepted, false otherwise?

Issue #4499, opened 2018-06-29
[APP SUBMITTED]: IOError: [Errno 13] Permission denied: u'/mnt/TvShows/TvShows/Criminal Minds/tvshow.nfo'


chest069:
### INFO
**Python Version**: `2.7.15rc1 (default, Apr 15 2018, 21:51:34) [GCC 7.3.0]`
**Operating System**: `Linux-4.15.0-23-generic-x86_64-with-Ubuntu-18.04-bionic`
**Locale**: `UTF-8`
**Branch**: [master](../tree/master)
**Database**: `44.11`
**Commit**: pymedusa/Medusa@c65ca2607fae74405dcceaaf8c02f3d29d0313ea
**Link to Log**: https://gist.github.com/b09aecb00dc875891c420d42fd68f3fb
### ERROR
<pre>
2018-06-29 04:22:27 ERROR    SHOWQUEUE-REFRESH :: [c65ca26] Unable to write file to /mnt/TvShows/TvShows/Criminal Minds/tvshow.nfo - are you sure the folder is writeable? error b'13' : error b'Permission denied'
Traceback (most recent call last):
  File "/opt/medusa/<a href="../blob/c65ca2607fae74405dcceaaf8c02f3d29d0313ea/medusa/metadata/generic.py#L281">medusa/metadata/generic.py</a>", line 281, in update_show_indexer_metadata
    showXML.write(nfo_file_path, encoding=u'UTF-8')
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 802, in write
    file = open(file_or_filename, "wb")
IOError: [Errno 13] Permission denied: u'/mnt/TvShows/TvShows/Criminal Minds/tvshow.nfo'
</pre>
---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

PR #4505, opened 2018-06-29
Bugfix/metadata ioerror

sharkykh:
"Fix" #4499 & #4500 
Permission errors are user errors, log them as warnings.

Also use errno package instead of comparing to integers.

### DIFF
--- a/medusa/helpers/__init__.py
+++ b/medusa/helpers/__init__.py
@@ -346,7 +346,7 @@ def hardlink_file(src_file, dest_file):
         link(src_file, dest_file)
         fix_set_group_id(dest_file)
     except OSError as msg:
-        if hasattr(msg, 'errno') and msg.errno == 17:
+        if hasattr(msg, 'errno') and msg.errno == errno.EXIST:
             # File exists. Don't fallback to copy
             log.warning(
                 u'Failed to create hardlink of {source} at {destination}.'
@@ -399,7 +399,7 @@ def move_and_symlink_file(src_file, dest_file):
         fix_set_group_id(dest_file)
         symlink(dest_file, src_file)
     except OSError as msg:
-        if hasattr(msg, 'errno') and msg.errno == 17:
+        if hasattr(msg, 'errno') and msg.errno == errno.EXIST:
             # File exists. Don't fallback to copy
             log.warning(
                 u'Failed to create symlink of {source} at {destination}.'
--- a/medusa/exception_handler.py
+++ b/medusa/exception_handler.py
@@ -2,6 +2,7 @@
 """Exception Handler."""
 from __future__ import unicode_literals
 
+import errno
 import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


medusa/helpers/__init__.py
```diff
@@ -346,7 +346,7 @@ def hardlink_file(src_file, dest_file):
         link(src_file, dest_file)
         fix_set_group_id(dest_file)
     except OSError as msg:
-        if hasattr(msg, 'errno') and msg.errno == 17:
+        if hasattr(msg, 'errno') and msg.errno == errno.EXIST:
```
labrys: @sharkykh generally you don't LBYL in python.  `OSError` should always have an `errno` so the `hasattr` call is extraneous.

sharkykh: _(Ignoring the fact I didn't write that code)_
I couldn't find any documentation that clearly states OSError / IOError always have an `errno` attr.
Any chance you have a source for that claim?

labrys: @sharkykh From the Python 2 docs for [Built-in Exceptions](https://docs.python.org/2/library/exceptions.html):

#### For `OSError`:

> *exception* __OSError__
This exception is derived from `EnvironmentError`. It is raised when a function returns a system-related error (not for illegal argument types or other incidental errors). The `errno` attribute is a numeric error code from `errno`, and the `strerror` attribute is the corresponding string, as would be printed by the C function `perror()`. See the module `errno`, which contains names for the error codes defined by the underlying operating system.

#### For `IOError`:
> *exception* __IOError__
Raised when an I/O operation (such as a print statement, the built-in `open()` function or a method of a file object) fails for an I/O-related reason, e.g., “file not found” or “disk full”.
> This class is derived from `EnvironmentError`. See the discussion above for more information on exception instance attributes.

#### For the relevant `EnvironmentError` mentioned in both of the above:
> *exception* __EnvironmentError__
The base class for exceptions that can occur outside the Python system: `IOError`, `OSError`. When exceptions of this type are created with a 2-tuple, the first item is available on the instance’s `errno` attribute (it is assumed to be an error number), and the second item is available on the `strerror` attribute (it is usually the associated error message). The tuple itself is also available on the `args` attribute.

## TLDR:
> When exceptions of this type are created with a 2-tuple, the first item is available on the instance’s `errno` attribute

labrys: @sharkykh I know you didn't write the line, but I'm a firm believer that if you're editing code then you should take the time to understand and review the lines your modifying.

sharkykh: Thanks.

> if you're editing code then you should take the time to understand and review the lines your modifying.

That is precisely what I did (tried to do).
I tried to find info about it and somehow completely missed it (was probably tired).

labrys: Just to clarify, since its difficult to express tone in text, my comment wasn't meant to imply that you don't but as a clarification of the reason I tagged you specifically since you were the most recent editor of the lines.  I know from personal I experience I miss stuff like this all the time.  You tend to get code-blind when looking at it for so long 😁

PR #4508, opened 2018-06-30
Feature/add service worker

OmgImAlexis:
This will only work on `http(s)://localhost` and all other HTTPS sites.
Self signed certs don't seem to work at all.


p0psicles:
Sorry. What can we do with this?

OmgImAlexis:
I'm using it to add proper caching to all assets. By using the service worker to look up the file first the browser can skip going to the server.

themes-default/slim/static/js/lib/globby-to-regexp.js
```diff
@@ -0,0 +1,130 @@
+const globbyToRegExp = function (glob, opts) {
```
sharkykh: Why not a minified file?

sharkykh:
I'm pretty bummed that this doesn't work on self-signed certificates.
I keep reading that localhost should always be accepted as secure, but it never worked for me.

Seems like we currently have to choose between having push notifications (HTTPS on) or having the ServiceWorker (HTTPS off). :disappointed:

labrys:
Disabling security should never be considered as an option.

OmgImAlexis:
@sharkykh https://www.chromium.org/blink/serviceworker/service-worker-faq

>  I get an error message about "Only secure origins are allowed". Why?

> A: Service workers are only available to "secure origins" (HTTPS sites, basically) in line with a policy to prefer secure origins for powerful new features. However http://localhost is also considered a secure origin, so if you can, developing on localhost is an easy way to avoid this error.

OmgImAlexis:
Ah... I'm testing on `http://localhost` and it works.

PR #4509, opened 2018-06-30
Fix proper search downloading multiple propers. Fixes #4495

### LINKED ISSUE #4495: Single proper snatched multiple times
There seems to be some issue that started the last month or so where medusa will download multiple copies of the same proper from different search provider within seconds of each other.  Once I got two from the same search provider only 7 seconds apart as well.  The worst was last night where it snatched the same episode 10 times.

medariox:

2 issues here:
- Multi episodes would get propers downloaded over and over
- There was no check to make sure that the existing release wasn't already proper, that means that it kept downloading proper releases over and over (not the same, but e.g. by different groups/name)

### DIFF
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -341,12 +341,10 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
                 b'AND episode IN ({episodes}) '
                 b'AND quality = ? '
                 b'AND date >= ? '
-                b'AND action IN (?, ?, ?, ?) '
-                b'LIMIT {amount}'.format(
+                b'AND action IN (?, ?, ?, ?)'.format(
                     episodes=','.join(
                         text_type(ep) for ep in candidate.actual_episodes
                     ),
-                    amount=len(candidate.actual_episodes) * 2,
                 ),
                 [candidate.indexerid, candidate.actual_season, candidate.quality,
                  history_limit.strftime(History.date_format),
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -9,13 +9,12 @@ import logging
 import operator
 import re
 import threading
-import time
 from builtins import map
 from builtins import object
 from builtins import str
 
 from medusa import app, db, helpers
-from medusa.common import DOWNLOADED, SNATCHED, SNATCHED_BEST, SNATCHED_PROPER, SUBTITLED, cpu_presets
+from medusa.common import DOWNLOADED, SNATCHED, SNATCHED_BEST, SNATCHED_PROPER, SUBTITLED
 from medusa.helper.common import enabled_providers
 from medusa.helper.exceptions import AuthException, ex
 from medusa.logger.adapters.style import BraceAdapter
--- a/medusa/search/proper.py
+++ b/medusa/search/proper.py
@@ -228,7 +228,7 @@ class ProperFinder(object):  # pylint: disable=too-few-public-methods
 
             # check if we have the episode as DOWNLOADED
             main_db_con = db.DBConnection()
-            sql_results = main_db_con.select(b"SELECT status, quality, release_name "
+            sql_results = main_db_con.select(b"SELECT quality, release_name "
                                              b"FROM tv_episodes WHERE indexer = ? "
                                              b"AND showid = ? AND season = ? "
                                              b"AND episode = ? AND status = ?",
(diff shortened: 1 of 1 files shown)


medusa/search/proper.py
```diff
@@ -312,8 +314,8 @@ def _get_proper_results(self):  # pylint: disable=too-many-locals, too-many-bran
             # if the show is in our list and there hasn't been a proper already added for that particular episode
             # then add it to our list of propers
             if best_result.indexerid != -1 and (
-                best_result.indexerid, best_result.actual_season, best_result.actual_episodes[0]
-            ) not in list(map(operator.attrgetter('indexerid', 'actual_season', 'actual_episode'), final_propers)):
+                best_result.indexerid, best_result.actual_season, best_result.actual_episodes
+            ) not in list(map(operator.attrgetter('indexerid', 'actual_season', 'actual_episodes'), final_propers)):
```
medariox: ^ This was the reason for muli-episodes repeated downloads.
For a multi-ep:
`actual_episode = None`
while `actual_episodes[0]` would be the first episode of the multi-episode release.

medusa/search/proper.py
```diff
@@ -334,37 +336,52 @@ def _download_propers(self, proper_list):
...
                 b'WHERE showid = ? '
                 b'AND season = ? '
-                b'AND episode = ? '
+                b'AND episode IN ({episodes}) '
                 b'AND quality = ? '
                 b'AND date >= ? '
-                b'AND action in (?, ?, ?, ?)',
-                [cur_proper.indexerid, cur_proper.actual_season, cur_proper.actual_episode, cur_proper.quality,
+                b'AND action IN (?, ?, ?, ?) '
+                b'LIMIT {amount}'.format(
```
medariox: `LIMIT` because we want Downloaded + the Snatched prior to that. There is no need to get the whole history of an episode.

medusa/search/proper.py
```diff
@@ -334,37 +336,52 @@ def _download_propers(self, proper_list):
...
-                    log.debug('This proper {result!r} is already in history, skipping it', {'result': cur_proper.name})
-                    continue
+            proper_tags_len = len(cur_proper.proper_tags)
+            proper_name = self._canonical_name(cur_proper.name, clear_extension=True)
+            proper_name_ext = self._canonical_name(cur_proper.name)
+
+            for result in history_results:
+
+                proper_tags = result[b'proper_tags']
+                if proper_tags and len(proper_tags.split('|')) >= proper_tags_len:
```
medariox: This check makes sure that the current release isn't already proper and in case it is, it will only get the new one if it has a higher proper priority (e.g. PROPER vs REAL.PROPER).

medusa/search/proper.py
sharkykh: Wow, that seems like a bad design.
Not for this PR, but I think `actual_episode` should just point to the first item in `actual_episodes`.
To make sure it's always correct though, it should raise an error if there are multiple episodes in `actual_episodes`.

medusa/search/proper.py
sharkykh: Are you sure that's not needed?
What if you snatched / downloaded [PP'ed] the same episode multiple times in a row?
And it stops at '30 days ago' anyway.

medusa/search/proper.py
```diff
@@ -328,43 +329,59 @@ def _download_propers(self, proper_list):
...
+                        'This proper {result!r} is already in history, skipping it',
+                        {'result': candidate.name},
+                    )
+                    break
 
+            else:
+                candidate.create_episode_object()
                 # snatch it
-                snatch_episode(cur_proper)
-                time.sleep(cpu_presets[app.CPU_PRESET])
```
sharkykh: Isn't this in place to keep CPU usage low (or high), per user preference?

medusa/search/proper.py
medariox: It was an attempt to make this chunk of slow code somewhat faster. Checking the history is unreliable anyway, as history data is volatile. I will remove the `LIMIT`.

medusa/search/proper.py
medariox: No, time.sleep() just stops code execution and nothing else. It doesn't give "a break" to the CPU, more likely the contrary. This kind of things shouldn't be implemented manually in high-level programming languages.

PR #4404, opened 2018-06-18
add more fields to config including patch

OmgImAlexis:
@sharkykh @p0psicles 

From Slack:
> OmgImAlexis [10:30]
> I want to remove some of the fields from the API such as `PLEX_SERVER_PASSWORD` and other API keys, etc. I'm thinking instead we return a boolean depending on if it's set or not?

PR #4519, opened 2018-07-02
Vueify header

sharkykh:
- Convert `partials/header.mako` to a Vue component ~(as much as possible)~
- Remove all `topmenu` arguments :tada:
- Implement mobile view notifications badge (#3921)

~I still need to add some values to the API, and hopefully remove all the Python conversions.~


themes-default/slim/views/vue-components/app-header.mako
```diff
@@ -0,0 +1,216 @@
...
+    from medusa import app
+%>
+<script>
+Vue.component('app-header', {
+    template: '#app-header-template',
+    data() {
+        return {
+            // Python conversions
+            <% has_emby_api_key = json.dumps(app.EMBY_APIKEY != '') %>
+            hasEmbyApiKey: ${has_emby_api_key},
```
sharkykh: @OmgImAlexis 
I think this the last blocker to finishing this component (and maybe moving to `.vue` file).
I can't add the API key to the endpoint (#4404), and we never decided on a solution for the passwords / API keys.

OmgImAlexis: We should allow `PATCH` for them for now. I'm really not okay with sending the API keys and passwords back to the client if we don't have to.

sharkykh: Here I need to know if the API key is set, not patch it.
So what I'm saying is, there was a discussion on Slack about how to handle situations where you need to know if it's set (excluding the config pages where we currently have the value in the HTML).
That discussion never reached its conclusion - so we need to figure it out.

I've decided to just remove the emby API key check for now, just so I can move forward with this.
Left a comment in place though.

OmgImAlexis: Why is the value needed? We may be able to work around this in another way.

sharkykh:
>Why is the value needed? We may be able to work around this in another way.

It was needed to ensure the button to update Emby was only visible if you had all the settings set (`enabled`, `host` and `apikey`).
It can work without it, but we're gonna hit that issue again soon with the config pages and we'll need a solution.

medusa/server/web/core/base.py
```diff
@@ -88,19 +87,13 @@ def __init__(self, rh, filename):
             'sbDefaultPage': app.DEFAULT_PAGE,
             'loggedIn': rh.get_current_user(),
             'sbStartTime': rh.startTime,
-            'numErrors': len(classes.ErrorViewer.errors),
-            'numWarnings': len(classes.WarningViewer.errors),
             'sbPID': str(app.PID),
```
OmgImAlexis: Can this be removed too?

sharkykh: No, it's still being used everywhere.

sharkykh:
Waiting on #4666

PR #4520, opened 2018-07-02
Vueify footer

sharkykh:


sharkykh:
To do:

- [ ] Update `/stats` endpoint to include the "overall" library stats used by this component.
- [ ] Update `/config` endpoint to include a `system` section that will include, at the very least:
  - The information on the schedulers, will be needed for the status page anyway.
  - The memory usage information.

After that, this component can be converted into an SFC and then it'll be completed.

sharkykh:
Lastly, waiting on #6811.

setup.py
```diff
@@ -102,7 +102,7 @@ def packages():
     },
     cmdclass={'test': PyTest},
     tests_require=tests_runtime_require + [
-        'flake8>=3.5.0',
+        'flake8>=3.7.7',
```
sharkykh: Version 3.7.0 added native support for `per-file-ignores`.
http://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores

PR #4522, opened 2018-07-02
Feature/misc vue changes

OmgImAlexis:
Check commit messages.


themes/light/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "manifest_version": 1,
   "version": "0.1.0",
-  "name": "light",
+  "name": "dark",
```
OmgImAlexis: @sharkykh any idea why this keeps changing when using `yarn gulp sync`?

sharkykh: Because of the way the sync is set up. It copies the `package.json` from `slim/package.json`.
And the default name in that file is `dark`.

I tried to add in a json editor, and it worked but it kept messing up the json formatting...

p0psicles: Ill fix it ok?

OmgImAlexis: @p0psicles open a new branch for it please.

PR #4564, opened 2018-07-05
add missing addShows template

OmgImAlexis:
Missed adding this back in the last vue PR.


sharkykh:
Looks like you're giving the `gulp sync` command way too much credit... :sweat_smile:
My point is the files you renamed didn't get renamed in light and dark templates, just got duped.
Also with the previous PR you needed to edit the handler in the Python code. Don't you need to do that here? Pretty sure it won't work without that change.

sharkykh:
Also need to fix `/addRecommended` (`addShows_recommended.mako`) with this PR.

OmgImAlexis:
> Looks like you're giving the gulp sync command way too much credit... 😅

What do you mean?

sharkykh:
>What do you mean?

I mean that it's not that smart.
If you rename or remove a file you need to manually do that for the dark and light themes.
As I said - 
> the files you renamed didn't get renamed in light and dark templates, just got duped.

image

sharkykh:
Also:

image
image

themes-default/slim/static/js/templates/add-recommended.vue
```diff
@@ -1,6 +1,5 @@
 <template>
     <div>
```
sharkykh: Now you don't need that wrapper div :)

themes-default/slim/static/js/templates/add-shows.vue
```diff
@@ -1,6 +1,5 @@
 <template>
     <div>
```
sharkykh: Same.

themes-default/slim/static/js/templates/add-recommended.vue
OmgImAlexis: Didn't want to remove them incase we had any styles relying on the div depth. Haven't had a chance to check.

sharkykh: I did, it's okay, can be removed.
Keep the `<div id="addShowPortal">` though.

PR #4565, opened 2018-07-05
Remove `/ui/` route, add fallback 404/500

sharkykh:
- `/ui` is no longer used (not to be confused with `/ws/ui`.
- Add 404/500 fallback routes for Vue error pages (500 error isn't on the router yet)


sharkykh:
Rebased.

PR #4580, opened 2018-07-05
add test for app-link to

OmgImAlexis:
This is allowed to fail since it's not used anywhere yet. @sharkykh should I just remove the `to` from this and we can start using `router-link` where needed?


sharkykh:
Let me see if I can figure it out on this branch, 5 minutes tops.

PR #4595, opened 2018-07-05
Remove `gulp-xo`, use native `xo`

sharkykh:
Hope no one was using the watch feature. @OmgImAlexis 

Related to #4587


themes-default/slim/gulpfile.js
```diff
@@ -20,14 +19,18 @@ const argv = require('yargs').argv;
...
+
 // Const postcss = require('gulp-postcss');
 // const sass = require('gulp-sass');
 const cssnano = require('cssnano');
 const autoprefixer = require('autoprefixer');
 const reporter = require('postcss-reporter');
 
 const PROD = process.env.NODE_ENV === 'production';
-const config = require('./package.json').config;
+const { config, xo: xoConfig } = require('./package.json');
```
OmgImAlexis: Didn't know you could rename a deconstructed param.

sharkykh: You and me both. That's pretty neat. Hopefully it's readable enough...

PR #4601, opened 2018-07-05
Fix gulp sync for now

sharkykh:
@OmgImAlexis 
If we remove the `async`/`await` for the `sync` task, it doesn't work correctly anymore - only syncs the first theme.

I spent 2 hours trying to get it to work with Promises, and couldn't.
I think this has something to do with how `runSequence` works..

### DIFF
--- a/themes-default/slim/gulpfile.js
+++ b/themes-default/slim/gulpfile.js
@@ -303,11 +303,11 @@ const syncTheme = (theme, sequence) => {
  *
  * Do not run the xo build, as this takes a lot of time.
  */
-gulp.task('sync', () => {
+gulp.task('sync', async () => {
     // Whe're building the light and dark theme. For this we need to run two sequences.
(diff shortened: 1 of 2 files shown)


sharkykh:
I must say it's annoying not being able to use `gulp sync`.
@OmgImAlexis Are you okay with ignoring the `gulpfile.js` lint for the time being?

OmgImAlexis:
@sharkykh merge?

p0psicles:
Aaaahh you broke my sync task 😭

sharkykh:
It's pretty fragile to be honest :cold_sweat: 
Also, looks I forgot to actually sync these changes to light/dark. Oh, the irony.

sharkykh:
@p0psicles I think we can fix it with gulp v4 - it has [`gulp.series` and `gulp.parallel`](https://fettblog.eu/gulp-4-parallel-and-series/)

PR #4623, opened 2018-07-07
Fix dependency listing

sharkykh:
@medariox 

Made some mistakes :stuck_out_tongue:

### DIFF
--- a/ext/readme.md
+++ b/ext/readme.md
@@ -2,7 +2,7 @@
  Status  |  Package  |  Version / Commit  | Usage | Notes
 :------: | :-------: | :----------------: | :---- | :----
 :: | <code><b>appdirs</b>.py</code> | [1.4.3](https://pypi.org/project/appdirs/1.4.3/) | `subliminal` (cli only), `simpleanidb` | -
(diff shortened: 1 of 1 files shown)


ext/readme.md
```diff
@@ -2,7 +2,7 @@
  Status  |  Package  |  Version / Commit  | Usage | Notes
 :------: | :-------: | :----------------: | :---- | :----
 :: | <code><b>appdirs</b>.py</code> | [1.4.3](https://pypi.org/project/appdirs/1.4.3/) | `subliminal` (cli only), `simpleanidb` | -
-:: | `attr` | [18.1.0](https://pypi.org/project/attr/18.1.0/) | `imdbpie` | -
+:: | `attrs` | [18.1.0](https://pypi.org/project/attr/18.1.0/) | `imdbpie` | Module: `attr`
```
medariox: Link also needs updating and why Module: `attr`?

sharkykh: >Link also needs updating

Oh, right.

>why Module: `attr`?

If the name of the package (on PyPI) is not the same as the actual module name (how you import it), I add "Module: `name`" for 2 reasons:
1. If it's a VCS package I need that for the `#egg=name` in the URL
2. For convenience - just to know what each package consists of.

PR #4629, opened 2018-07-07
Display a soft "Loading..." mask until page is ready

### LINKED ISSUE #4618: History flashes an unsorted list for a second before showing sorted list
**Describe the bug**
Whenever I click on "History", Medusa briefly shows a blank page then for about 1 second it flashes an unsorted history list, and then it shows the sorted history list (I have it sorted by date/time).

**To Reproduce**
Click on history and watch the screen carefully.

**Expected behavior**
The unsorted history list should not be shown. It is disorienting. Medusa should show the blank page until it can display the sorted history list.

Medusa Info: | Branch:                     master                      
Commit:                     e7c4ff3bc9a48b47c932b0e40a539d62e3a62f64                      
Version:                     v0.2.6                      
Database:                     44.11
Python Version: | 2.7.15 (default, May  1 2018, 20:16:04)  [GCC 7.3.1 20180406]
SSL Version: | OpenSSL 1.1.0h  27 Mar 2018
OS: | Linux-4.17.2-1-ARCH-x86_64-with-glibc2.2.5
Locale: | en_US.UTF-8

**Additional context**
I have my initial page set to "History" but I do not think it is relevant to this bug because
it happens even when I am already in Medusa and specifically click on "History".

I think this behavior only started with the latest release (v0.2.6). I did not notice it with previous versions.

sharkykh:
Fixes #4618.

Also fixes all the other pages that are not ready when the Vue cloak (`v-cloak`) is lowered,
Help & Info for example.

I've also tested what happens if there was a javascript error on the page, and the mask was still lowered.
Just in case, I added the handler to manually remove the mask.

### DIFF
--- a/themes-default/slim/views/layouts/main.mako
+++ b/themes-default/slim/views/layouts/main.mako
@@ -181,8 +181,8 @@
                         setTimeout(() => {
                             const { store } = window;
                             store.dispatch('login');
-                            store.dispatch('getConfig');
-                            this.$emit('loaded');
+                            store.dispatch('getConfig')
+                                .then(() => this.$emit('loaded'));
                         }, 1000);
                     }
 
--- a/themes-default/slim/views/layouts/main.mako
+++ b/themes-default/slim/views/layouts/main.mako
@@ -62,7 +62,12 @@
             % endif
             <%include file="/partials/alerts.mako"/>
                <div id="content-row" class="row">
-                    <div id="content-col" class="${'col-lg-10 col-lg-offset-1 col-md-10 col-md-offset-1' if not app.LAYOUT_WIDE else 'col-lg-12 col-md-12'} col-sm-12 col-xs-12">
+                    <div v-if="globalLoading" class="text-center ${'col-lg-10 col-lg-offset-1 col-md-10 col-md-offset-1' if not app.LAYOUT_WIDE else 'col-lg-12 col-md-12'} col-sm-12 col-xs-12">
+                        <h3>Loading....</h3>
+                        If this is taking too long,<br>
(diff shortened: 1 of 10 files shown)


sharkykh:
@OmgImAlexis 
Thanks.

This PR should be merged regardless of our Slack discussion about the `setTimeout` (for the first mutations not showing up).
We still have to wait for the `getConfig` action to finish before showing the page.

medariox:
I must say the one with the logo is pretty awesome! But I'm fine with either way 👍

themes-default/slim/views/layouts/main.mako
```diff
@@ -62,7 +62,12 @@
             % endif
             <%include file="/partials/alerts.mako"/>
                <div id="content-row" class="row">
-                    <div id="content-col" class="${'col-lg-10 col-lg-offset-1 col-md-10 col-md-offset-1' if not app.LAYOUT_WIDE else 'col-lg-12 col-md-12'} col-sm-12 col-xs-12">
+                    <div v-if="globalLoading" class="text-center ${'col-lg-10 col-lg-offset-1 col-md-10 col-md-offset-1' if not app.LAYOUT_WIDE else 'col-lg-12 col-md-12'} col-sm-12 col-xs-12">
```
p0psicles: Could we maybe incorporate this as a separate component?

sharkykh: I could try,
but I don't think it's worth the effort because we don't plan on keeping it.

p0psicles: Why not keep it? You always need to load data right?

sharkykh: I think it will be much less noticeable when it's a true single page application.

PR #4630, opened 2018-07-07
Remove root gulp, use simple yargs script

sharkykh:
`yarn [build|install]` will still lint themes.


OmgImAlexis:
You shouldn't need node beforehand btw as long as you change the first line to `# /bin/node`.

package.json
```diff
@@ -11,8 +11,7 @@
   "homepage": "https://github.com/PyMedusa/Medusa#readme",
   "scripts": {
     "test": "yarn run build && yarn run test-css && yarn run test-api",
-    "gulp": "gulp",
-    "build": "gulp lintthemes",
+    "build": "node ./themes-default/helper.js lint",
```
OmgImAlexis: You should need this if the `helper.js` starts with `# /bin/node`.

medariox:
Can someone solve the conflicts please?

Issue #4624, opened 2018-07-07
Javascript errors on login page


sharkykh:
**Describe the bug**
Doesn't affect usability, just need to fix some stuff to not run on the `/login` route, like we do elsewhere.

![image](https://user-images.githubusercontent.com/10238474/42410893-62a89288-81fa-11e8-8f7b-330f898d9684.png)

**To Reproduce**
1. Set username and password for the Web UI
2. Log out of the Web UI
3. See console errors

**Medusa:**
 - OS: Windows 7 x64
 - Branch: `vueify/header` (probably applies to `develop`)
 - Commit: ef0e51161cfcfc1f0b9d90ba60f50467ac3740c5

**Logs:**
<details>

```
15:26:14.066 api.js:8 Uncaught TypeError: Cannot read property 'replace' of undefined
    at api.js:8
(anonymous) @ api.js:8
15:26:14.243 store.js:501 Uncaught TypeError: Cannot read property 'replace' of undefined
    at websocketUrl (store.js:501)
    at store.js:504
websocketUrl @ store.js:501
(anonymous) @ store.js:504
15:26:14.259 router.js:11 Uncaught TypeError: Cannot read property 'replace' of null
    at router.js:11
(anonymous) @ router.js:11
15:26:14.260 localhost/:1528 Loading Vue with router since window.app is missing.
15:26:14.263 vue.js:597 [Vue warn]: Error in created hook: "TypeError: Cannot set property 'onmessage' of undefined"

(found in <Root>)
warn @ vue.js:597
logError @ vue.js:1739
globalHandleError @ vue.js:1734
handleError @ vue.js:1723
callHook @ vue.js:2923
Vue._init @ vue.js:4617
Vue @ vue.js:4716
(anonymous) @ localhost/:1529
15:26:14.263 vue.js:1743 TypeError: Cannot set property 'onmessage' of undefined
    at Vue.created (localhost/:1483)
    at callHook (vue.js:2921)
    at Vue._init (vue.js:4617)
    at new Vue (vue.js:4716)
    at localhost/:1529
logError @ vue.js:1743
globalHandleError @ vue.js:1734
handleError @ vue.js:1723
callHook @ vue.js:2923
Vue._init @ vue.js:4617
Vue @ vue.js:4716
(anonymous) @ localhost/:1529
15:26:14.549 vue.js:8553 You are running Vue in development mode.
Make sure to turn on production mode when deploying for production.
[...]

PR #4634, opened 2018-07-08
vueify login page

OmgImAlexis:
We have to expose the `api-root` to non-authed users. I can't see this being an issue since they'll need to know this once we move to using the API for authentication anyways.

Closes https://github.com/pymedusa/Medusa/issues/4624


sharkykh:
I'm okay with exposing the `api-root`.
We should instead expose the `web-root`, though.
Then update all the base paths to stop doing the `.replace`, and only add `/api/v2` to the base path for the `api` axios object.

themes-default/slim/static/js/router.js
```diff
@@ -11,6 +12,14 @@ if (!window.router) {
         base: document.body.getAttribute('api-root').replace('api/v2/', ''),
         mode: 'history',
         routes: Object.assign(routes || [], [{
+            path: '/login',
+            name: 'login',
+            meta: {
+                title: 'Login',
+                header: 'Medusa'
```
sharkykh: No header for the login page. It has its own header.
It's not currently an issue, because it doesn't use the VueRouter template (`index.mako`),
but it will be when it does.

themes-default/slim/views/layouts/main.mako
```diff
@@ -49,7 +49,7 @@
         <link rel="stylesheet" type="text/css" href="css/lib/vue-snotify-material.css?${sbPID}"/>
         <%block name="css" />
     </head>
-    <body ${('data-controller="' + controller + '" data-action="' + action + '" api-key="' + app.API_KEY +'"  api-root="' + app.WEB_ROOT + '/api/v2/"', '')[title == 'Login']}>
+    <body ${('data-controller="' + controller + '" data-action="' + action + '" api-key="' + app.API_KEY + '"', '')[title == 'Login']} api-root="${app.WEB_ROOT}/api/v2/">
```
sharkykh: Let's expose `web-root="${app.WEB_ROOT}"` instead.
(see non-review comment above)

themes-default/slim/static/js/templates/login.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+    <div class="login">
+        <form action="" method="post">
+            <h1>Medusa</h1>
+            <div class="ctrlHolder"><input class="inlay" name="username" type="text" placeholder="Username" autocomplete="off" autocomplete="no" /></div>
```
sharkykh:
```
  static/js/templates/login.vue:5:128
  ✖  5:128  Duplicate attribute autocomplete.    vue/no-duplicate-attributes
  ✖  5:128  Parsing error: duplicate-attribute.  vue/no-parsing-error
```

`off` is the correct value.

sharkykh:
Uhh...
image

Look at the elements, Vue didn't parse them.

sharkykh:
Latest commit (cherry picked from #4656) should fix that issue.

PR #4639, opened 2018-07-09
Fix unhandled request error on existingShow

### LINKED ISSUE #4637: Medusa fails to load existing shows with rclone cache mount
When adding existing shows through a rclone cache, the spinning folder icon never advances past loading folders. The folders do appear when navigating through the Manage Directories popup.

**To Reproduce**
1. Create rclone cache of rclone remote
2. Mount rclone cache
3. Launch Medusa (must be done after mounting if rclone is mounted as user-mode mount)
4. Add existing show.
5. See error

**Expected behavior**
Folders should load, add existing shows workflow should continue.

Branch: master 
Commit: e7c4ff3bc9a48b47c932b0e40a539d62e3a62f64 
Version: v0.2.6 
Database: 44.11
Python Version: | 2.7.13 (default, Nov 24 2017, 17:33:09) [GCC 6.3.0 20170516]
SSL Version: | OpenSSL 1.1.0f 25 May 2017
OS: | Linux-4.9.0-6-amd64-x86_64-with-debian-9.4
Locale: | en_US.UTF-8

**Logs:**
No pertinent information available through info or debug logs.

sharkykh:
Increase timeout

Fixes #4637


themes/dark/templates/addShows_addExistingShow.mako
```diff
@@ -104,8 +113,23 @@ const startVue = () => {
...
-                const { data } = await api.get('internal/existingSeries', { params });
+                let data = null;
+                const config = {
+                    params: {
+                        'rootDirs': indices.join(',')
+                    },
+                    timeout: this.requestTimeout
+                };
+                try {
+                    const response = await api.get('internal/existingSeries', config)
```
OmgImAlexis: If you need a try/catch please use `.then` and `.catch`

```js
api.get('internal/existingSeries', config).then(response => {
    data = response.data;
}).catch(error => {
    this.errorMessage = error.message;
    this.isLoading = false;
    this.dirList = [];
    return;
});
```

sharkykh: Also, I know have to put everything inside the `then` block, or move it to a new function, because the function continues to run without waiting for the response...

OmgImAlexis:
Is the local `loading` still needed since we now have the global loading?

sharkykh:
Yes. The local `loading` is used every time we're waiting for the folder data from the server.

PR #4658, opened 2018-07-12
Add ability to manually change episode quality. Fixes #4474

### LINKED ISSUE #4474: Manual Quality/Status has been Removed
I'm sure removing the ability to manually set the quality/status per episode was removed for a good reason but I know I'm not the only one missing that feature. Sure it's not a huge issue but it was helpful nonetheless.

First use case I've come across since it being removed.  I know there are more, some brought up in irc.
Medusa falsely is identifying a file as 720p web-dl. I would like to set it correctly to sdtv and archive it.
I think the above situation came about when medusa snatched a prefered quality (720p web-dl) to replace the existing sdtv. The snatched file failed and left the state in snatch 720p web-dl. I didn't want medusa to try an contine snatching that episode, so set it as archive. As archive sdtv now does not exist. Now the file is falsely labeled as 720p web-dl archived.

medariox:

img1
img2


medusa/server/web/manage/handler.py
```diff
@@ -297,7 +297,7 @@ def downloadSubtitleMissed(self, *args, **kwargs):
                 to_download[(cur_indexer_id, cur_series_id)] = [str(x[b'season']) + 'x' + str(x[b'episode']) for x in all_eps_results]
 
             for epResult in to_download[(cur_indexer_id, cur_series_id)]:
-                season, episode = epResult.split('x')
+                season, episode = epResult.lstrip('s').split('e')
```
sharkykh: I see a problem here, look at line 297, it's still using the `SNxEP` format

themes-default/slim/views/partials/showheader.mako
```diff
@@ -265,14 +266,23 @@
...
+                        % endif
+                        % for cur_status in statuses:
+                            <option value="${cur_status}">${statusStrings[cur_status]}</option>
+                        % endfor
+                    </select>
+                    <select id="qualitySelect" class="form-control form-control-inline input-sm-custom input-sm-smallfont">
+                        <option selected value="">Change quality to:</option>
+                        <% qualities = sorted(Quality.qualityStrings.items(), key=operator.itemgetter(0)) %>
+                        % for quality, name in qualities:
+                            % if quality > 1:  # Skip N/A (0) and Unknown (1)
```
sharkykh: I prefer:
```py
% if quality not in (Quality.NA, Quality.UNKNOWN):
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 **New Features**
 
 - Hot-swap themes: No need to restart the Medusa after changing the theme ([#4271](https://github.com/pymedusa/Medusa/pull/4271))
+- Added ability to manually change episode quality ([#4658](https://github.com/pymedusa/Medusa/pull/4658))
```
sharkykh: I wouldn't call it a new feature...

medariox: It was never possible to only change the quality of an episode... so it definitely is IMO.

medariox: But we can also move it to improvements, it's the same for me.

sharkykh: I feel like improvements would be a better description.

themes-default/slim/static/js/home/display-show.js
```diff
@@ -414,4 +424,19 @@ MEDUSA.home.displayShow = function() { // eslint-disable-line max-lines
             log.error(error.data);
         });
     });
+
+    function setQuality(quality, seriesSlug, episodes) {
```
OmgImAlexis: Please use `const setQuality = (quality, seriesSlug, episodes) => {`

themes-default/slim/static/js/home/display-show.js
```diff
@@ -414,4 +424,19 @@ MEDUSA.home.displayShow = function() { // eslint-disable-line max-lines
...
+        const patchData = {};
+        episodes.forEach(episode => {
+            patchData[episode] = { quality: parseInt(quality, 10) };
+        });
+
+        api.patch('series/' + seriesSlug + '/episodes', patchData)
+            .then(response => {
```
OmgImAlexis: `.then`  should be on the same line as the `api.patch`.

OmgImAlexis:
Strings can be checked `if (var_name)` no need for the check against an empty string.

sharkykh:
I hope this won't cause more confusion.

For example I changed from `Snatched <SD>` to `Skipped`,
and got `Skipped <SD>`. Doesn't really make sense.

themes-default/slim/static/js/home/display-show.js
medariox: Are you sure? That's a really ugly syntax IMO 😕

themes-default/slim/static/js/home/display-show.js
medariox: If I change it the XO lint won't pass.

sharkykh: Try
```js
api.patch('series/' + seriesSlug + '/episodes', patchData).then(response => {
    log.info(response.data);
    window.location.reload();
}).catch(error => {
    log.error(error.data);
});
```

themes-default/slim/static/js/home/display-show.js
p0psicles: It feels like that at first. But it gets better when your getting used to it

p0psicles:
I think where moving to a situation where status and quality should be independent of one another. It gives more flexibility. I think the almost in all places we use the quality `or` the status of an episode. But not both at the same time. That we display them together shouldnt matter much.

Ad an example the `Skipped (SD) ` could occur. As you can change the status of an already downloaded ep to Skipped. Then i dont mind knowing whats already on disk. 

Do we really want to make all these decisions for our users? Like removing the quality when its changed to the status skipped? Users already voiced they want to have that flexibility.

PR #4663, opened 2018-07-13
Convert titles and headers to VueRouter

OmgImAlexis:
Sorry for the single commit.


themes-default/slim/views/home.mako
```diff
@@ -14,16 +14,13 @@ window.app = {};
...
-        data() {
-            return {
-                header: 'Show List'
-            };
-        },
-        computed: Object.assign({
+        computed: {
+            isRoute() {
+                const { matched } = this.$route;
+                return matched.filter(route => route.components.default !== undefined).length >= 1;
```
sharkykh: What's this?

OmgImAlexis: It's for the future. It returns `true` if we have a matched component vs a matched route.

sharkykh:
- [x] /home
- [x] /home/displayShow
- [ ] /home/snatchSelection
- [x] /home/editShow
- [ ] /home/testRename
- [ ] /home/postprocess
- [ ] /home/status
- [ ] /home/restart
- [ ] /home/shutdown
- [ ] /home/update
- [x] /login
- [x] /config
- [x] /config/anime
- [x] /config/backuprestore
- [x] /config/general
- [x] /config/notifications
- [x] /config/postProcessing
- [x] /config/providers
- [x] /config/search
- [x] /config/subtitles
- [x] /addShows
- [x] /addShows/newShow
- [x] /addShows/existingShow
- [ ] /addShows/trendingShows
- [ ] /addShows/popularShows
- [ ] /addShows/popularAnime
- [x] /addRecommended
- [x] /schedule
- [x] /history
- [x] /manage
- [x] /manage/backlogOverview
- [x] /manage/episodeStatuses
- [x] /manage/failedDownloads
- [x] /manage/manageSearches
- [x] /manage/massEdit
- [x] /manage/subtitleMissed
- [x] /manage/subtitleMissedPP
- [ ] /errorlogs
- [ ] /errorlogs/viewlog
- [ ] /news
- [ ] /IRC
- [ ] /changes
- [x] /not-found
- [ ] /server-error
- [ ] /apibuilder

themes-default/slim/views/home.mako
sharkykh: For the home "partials"?

OmgImAlexis:
@sharkykh how's that?

Thanks for the list. I was having a hard time trying to keep track of which ones I had completed.

sharkykh:
Much easier, yes, thanks!
Are we still going to use `window.routes`, because now it's unused.

sharkykh:
`/addShows` is empty.

OmgImAlexis:
No, I think we should just move the routes to the router instead of adding them to the window.routes and then trying to merge it all together.

I'm pretty sure I got all the pages that were using it so we should be okay. I've done a quick scan through and can't find any references to it.

`/addShows` will be fixed. Give me 2 mins.

OmgImAlexis:
@sharkykh try commenting in the `// component: { template: '<router-view/>' },` for `addShows`. It should fix it. I think it may break the other nested routes like `addExistingShows` and `newShow`.

sharkykh:
It works, `newShow` and `existingShow` work too.

sharkykh:
I think it's working because if the page is not a component, then it loads using the local Vue and not the global Vue + Router.

themes-default/slim/views/home.mako
OmgImAlexis: No for determining if you're in a Vue component or a mako template.

OmgImAlexis:
The anonymous component is the parent element that we render so we can have all the not nested yet still keep a nested path. We could set the parent name instead of the route with `path: ''` like we do now.

For example without this `/home/addShows` would render inside of the `/home` template, same goes for all the manage pages, etc.

sharkykh:
The config pages are now broken on this branch.
I think this has something to do with the fact `startVue` function was removed.

```
(index):548 Uncaught ReferenceError: router is not defined
    at (index):548
```
image

sharkykh:
@OmgImAlexis I'll push a rebased branch to this PR in the next hour or so.

sharkykh:
I recreated the branch on top of current develop, because we had so many changes, so I went over the files and manually made the same changes this PR had before.
Hopefully I didn't make a mistake. A quick glance at the pages, especially home, displayShow and editShow, looks like everything is still good.

themes-default/slim/views/home.mako
```diff
@@ -13,16 +13,13 @@
...
-    data() {
-        return {
-            header: 'Show List'
-        };
-    },
-    computed: Object.assign({
+    computed: {
+        isRoute() {
+            const { matched } = this.$route;
+            return matched.filter(route => route.components.default !== undefined).length >= 1;
```
sharkykh: Unfortunately [I still don't really understand](https://github.com/pymedusa/Medusa/pull/4663#discussion_r202223963) why it's needed.
If you go to `/home/displayShow` it shouldn't load the home component.

themes-default/slim/views/errorlogs.mako
```diff
@@ -4,12 +4,24 @@
     from medusa import classes
 %>
 <%block name="scripts">
+
+<%
+    if logLevel == logger.WARNING:
```
OmgImAlexis: Is this not available via URL params?

sharkykh: Hmm, yes, I can use that instead.

sharkykh:
The only thing that really bugs me about this PR is the extra router-view component it addes for nested routes.
Maybe we should just drop the children as it seems to be intended for actual nested views...

OmgImAlexis:
@sharkykh if you can find a way to do nested routes without it then by all means suggest away. The way I've used is what has been recommended over and over by the devs.

themes-default/slim/views/errorlogs.mako
OmgImAlexis: Just gets us closer to SFC.

sharkykh:
@OmgImAlexis Last commit 'fixes' it.
I'm sure we can also find a way to not write the path prefix for each route group (`/home/`, `/config/`), if you don't like it.

themes-default/slim/test/specs/router.spec.js
```diff
@@ -0,0 +1,6 @@
+import test from 'ava';
+import router from '../../src/router';
+
+test('router compiles', t => {
+    t.truthy(router);
```
OmgImAlexis: I think [`notThrows`](https://github.com/avajs/ava#notthrowsfn-message) would work better here.

sharkykh: `notThrows` requires a function, but the import happens at the top of the file.

PR #4667, opened 2018-07-13
Fix app-header simple render test

sharkykh:


themes-default/slim/test/helpers/setup.js
```diff
@@ -6,6 +6,11 @@ const hooks = require('require-extension-hooks');
 // Setup browser environment
 browserEnv();
 
+// Setup document variables
+const baseElement = document.createElement('base');
+baseElement.setAttribute('href', 'http://localhost:8081');
+document.head.appendChild(baseElement);
```
OmgImAlexis: How are we going to test a `webRoot` since the base will be setup before anything else?

sharkykh: `AppLink doesn't use `web-root`,
`Asset` will, so in that case we should be able to just add:
```js
document.body.setAttribute('web-root', '/medusa');
document.body.setAttribute('api-key', 'abcdef123456');
```

PR #4672, opened 2018-07-14
Move scroll-buttons into a Vue component

sharkykh:
Includes a simple render test for the component.


sharkykh:
@OmgImAlexis 
image

PR #4679, opened 2018-07-16
Fix bj-share animes

DarkSupremo:
Fix bj-share animes, for some reason Anime EXX stopped being recognized as Absolute format, so changed it to Anime XX. Added boruto as absolute format, it also started having the wrong season on the tracker

### DIFF
--- a/medusa/providers/torrent/html/bjshare.py
+++ b/medusa/providers/torrent/html/bjshare.py
@@ -51,8 +51,8 @@ class BJShareProvider(TorrentProvider):
         # Cache
         self.cache = tv.Cache(self, min_time=30)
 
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/html/bjshare.py
```diff
@@ -63,6 +63,7 @@ def __init__(self):
         # until they or the source from where they get that info fix it...
         self.absolute_numbering = [
             'One Piece'
+            'Boruto'
```
medariox: Missing comma

DarkSupremo: Fixed

PR #4692, opened 2018-07-18
Bundle using Webpack

OmgImAlexis:
I know this is a crappy single commit with loads of changes but I just wanted feedback before I continue.

This should work out of the box in the dark theme. Do not run any gulp commands, etc. or it'll break as I've had to manually change a few things after the sync.


themes-default/slim/package.json
```diff
@@ -7,6 +7,7 @@
     "gulp": "gulp",
     "lint": "xo",
     "build": "cross-env NODE_ENV=production gulp build",
+    "webpack": "webpack-cli && cp dist/js/*.js ../../themes/dark/assets/js && cp dist/js/*.js ../../themes/light/assets/js",
```
OmgImAlexis: You can just use `webpack` here.

themes-default/slim/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "compilerOptions": {
+        "allowJs": true,
+        "noImplicitAny": true
+    },
+    "include": [
+        "./static/js/*.js"
```
OmgImAlexis: This isn't including any subdirs?

sharkykh: That's your code, What's this even for?

OmgImAlexis: Oh shoot, it is.

It's for enabling typescript typechecking and intellisense in vscode. That allows us to use the jsdoc syntax to add types for all of our functions, etc. and then it'll use those to create our type file so we don't need todo it ourselves.

OmgImAlexis:
Use this now. The `cp` wasn't being run for some reason.

```
cd themes-default/slim
yarn webpack && yarn webpack-copy && yarn gulp sync
```

OmgImAlexis:
@sharkykh can we try and get this merged today?

themes-default/slim/package.json
```diff
@@ -4,10 +4,10 @@
   "name": "slim",
   "author": "p0psicles",
   "scripts": {
-    "gulp": "gulp",
```
OmgImAlexis: We don't need this as `yarn gulp` is the same even if you don't have this.

p0psicles:
Please don't forget to update the WIKI.
It get's overlooked every time. And after the PR is merged, nobody seems to have time to update it.

PR #4696, opened 2018-07-19
convert x-templates to vue SFCs

OmgImAlexis:
I had to turn off linting for vue files right now as most of the new ones have issues. I'll fix them and re-enable linting in another PR after this.


themes/dark/assets/js/templates/root-dirs.vue
```diff
@@ -0,0 +1,225 @@
...
+            this.$emit('update:root-dirs-value', this.rootDirsValue, this.rootDirs);
+            this.$nextTick(() => {
+                // @FIXME: Temporarily trigger a regular event as well
+                $('#rootDirText').trigger('change');
+            });
+        }
+    }
+};
+</script>
+<style>
```
p0psicles: Should we make sfc css scoped?

OmgImAlexis: If you don't want the outside CSS to be able to override the component then yes.

p0psicles:
EditShow broke with the branch.
Can you fix that first?

OmgImAlexis:
@p0psicles any other pages you can see with an issue?

OmgImAlexis:
It should be working fine now.

OmgImAlexis:
I would suggest disabling whitespace when you review. Should bring the whole PR down to less than `100+/-`.

Issue #4683, opened 2018-07-16
[APP SUBMITTED]: GithubException: 500 None


OmgImAlexis:
### INFO
**Python Version**: `2.7.14 (default, Dec 14 2017, 15:51:29) [GCC 6.4.0]`
**Operating System**: `Linux-4.14.49-unRAID-x86_64-with`
**Locale**: `None`
**Branch**: [master](../tree/master)
**Database**: `44.11`
**Commit**: pymedusa/Medusa@1a8dc66c7ec9231be0d3444710e476437c9ae1c8
**Link to Log**: https://gist.github.com/0646a9888763d9d15ce5e4693b708d93
### ERROR
<pre>
2018-07-17 03:46:39 ERROR    CHECKVERSION :: [1a8dc66] Exception generated: 500 None
Traceback (most recent call last):
  File "/app/medusa/<a href="../blob/1a8dc66c7ec9231be0d3444710e476437c9ae1c8/medusa/scheduler.py#L93">medusa/scheduler.py</a>", line 93, in run
    self.action.run(self.force)
  File "/app/medusa/<a href="../blob/1a8dc66c7ec9231be0d3444710e476437c9ae1c8/medusa/version_checker.py#L67">medusa/version_checker.py</a>", line 67, in run
    self.list_remote_branches()
  File "/app/medusa/<a href="../blob/1a8dc66c7ec9231be0d3444710e476437c9ae1c8/medusa/version_checker.py#L357">medusa/version_checker.py</a>", line 357, in list_remote_branches
    app.GIT_REMOTE_BRANCHES = self.updater.list_remote_branches()
  File "/app/medusa/<a href="../blob/1a8dc66c7ec9231be0d3444710e476437c9ae1c8/medusa/version_checker.py#L981">medusa/version_checker.py</a>", line 981, in list_remote_branches
    gh = get_github_repo(app.GIT_ORG, app.GIT_REPO)
  File "/app/medusa/<a href="../blob/1a8dc66c7ec9231be0d3444710e476437c9ae1c8/medusa/github_client.py#L101">medusa/github_client.py</a>", line 101, in get_github_repo
    return gh.get_organization(organization).get_repo(repo)
  File "/app/medusa/<a href="../blob/1a8dc66c7ec9231be0d3444710e476437c9ae1c8/ext/github/Organization.py#L652">ext/github/Organization.py</a>", line 652, in get_repo
    "/repos/" + self.login + "/" + name
[...]

PR #4704, opened 2018-07-21
Fix Github error 500 exception

sharkykh:
Fixes https://github.com/pymedusa/Medusa/issues/4683


OmgImAlexis:
That looks a lot nicer. 👍

PR #4715, opened 2018-07-22
vueify header hover

OmgImAlexis:
Trying to remove more of the jQuery.


themes-default/slim/static/js/templates/app-header.vue
```diff
@@ -112,14 +119,20 @@ module.exports = {
...
             topMenuMapping: [
-                ['system', ['/home/restart', '/home/status', '/errorlogs', '/changes', '/news', '/IRC']],
                 ['home', ['/home', '/addShows', '/addRecommended']],
-                ['config', ['/config']],
-                ['history', ['/history']],
                 ['schedule', ['/schedule']],
+                ['history', ['/history']],
                 ['manage', ['/manage']],
+                ['config', ['/config']],
+                ['system', ['/home/restart', '/home/status', '/errorlogs', '/changes', '/news', '/IRC']],
```
sharkykh: The order has meaning.
If you change the order, it will match the current path to an incorrect `topMenu`.
For example `/home/status` will match `home` instead of `system` with the change you made.

themes-default/slim/static/js/templates/app-header.vue
```diff
@@ -193,21 +206,6 @@ module.exports = {
     mounted() {
         const { $el } = this;
 
-        // Hover Dropdown for Nav
-        $($el).on({
-            mouseenter(event) {
-                const $target = $(event.currentTarget);
-                $target.find('.dropdown-menu').stop(true, true).delay(200).fadeIn(500, () => {
-                    $target.find('.dropdown-toggle').attr('aria-expanded', 'true');
```
sharkykh: Setting `aria-expanded` on the `.dropdown-toggle` child to `'true'` is still needed.
You should be able to do that in `toggleDropdown` if you pass `$event` to it as a parameter.

themes-default/slim/static/js/templates/app-header.vue
```diff
@@ -13,24 +13,26 @@
         </div>
         <div v-if="auth.isAuthenticated" class="collapse navbar-collapse" id="main_nav">
             <ul class="nav navbar-nav navbar-right">
-                <li id="NAVhome" class="navbar-split dropdown" :class="{ active: topMenu === 'home' }">
-                    <app-link href="home/" class="dropdown-toggle" aria-haspopup="true" data-toggle="dropdown" data-hover="dropdown"><span>Shows</span>
-                    <b class="caret"></b>
+                <li @mouseover="toggleDropdown('home')" @mouseout="toggleDropdown('home')" id="NAVhome" class="navbar-split dropdown" :class="{ active: topMenu === 'home' }">
```
sharkykh: Just a thought: Would using the `.passive` modifier on all these events perform better?

themes-default/slim/static/js/templates/app-header.vue
```diff
@@ -13,24 +13,26 @@
...
-                        <template v-if="recentShows.length > 0">
-                        <li role="separator" class="divider"></li>
-                        <li v-for="recentShow in recentShows" :key="recentShow.link">
-                            <app-link :href="recentShow.link">
-                                <i class="menu-icon-addshow"></i>&nbsp;{{ recentShow.name }}
-                            </app-link>
-                        </li>
-                        </template>
-                    </ul>
+                    <transition name="fade" :duration="500">
```
sharkykh: Once you add the transition effect to CSS, you can remove the `duration` property, It's optional.
```css
.fade-enter-active, .fade-leave-active {
    transition: opacity .5s;
}

.fade-enter, .fade-leave-to {
    opacity: 0;
}
```

themes-default/slim/static/js/templates/app-header.vue
sharkykh: I'd say hooks are the way to go:
https://vuejs.org/v2/guide/transitions.html#JavaScript-Hooks

sharkykh:
It doesn't work good on mobile.

PR #4717, opened 2018-07-22
Prepare v0.2.7 release

sharkykh:
I'm still unsure about what we're going to do with the `CHANGELOG.md` file,
but if it stays updated as PRs get merged, it should make the changelog easier to read.

So maybe we just list the changes here, but on release copy them over to [CHANGES.md](https://github.com/pymedusa/medusa.github.io/blob/master/news/CHANGES.md)...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
-## 0.2.7 (unreleased)
+## Unreleased
 
-**New Features**
+-----
 
-- Hot-swap themes: No need to restart the Medusa after changing the theme ([#4271](https://github.com/pymedusa/Medusa/pull/4271))
-- _Simple message describing the new feature, and a link to the pull request._
+## 0.2.7 (2018-07-26)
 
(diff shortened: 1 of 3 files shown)


p0psicles:
Cant we just start to load changelog in the iframe instead of changes?

sharkykh:
The issue with that is, it's going to require users to update Medusa before they can see the actual changelog, while currently they may see the changelog before they update (if the CDN updates in that window).

sharkykh:
Or... we can load it directly from Github instead of the local file :thinking: ...

OmgImAlexis:
Why don't we keep a news file for "human text" and the changelog for the automations?

OmgImAlexis:
We could even have a link to the develop branch changelog so people can look at the to be added features and see if their issue has been fixed already.

sharkykh:
Preview:
image

p0psicles:
Wow, thats some clean changelog!

PR #4718, opened 2018-07-22
Update root dirs

sharkykh:
- Remove reference to `MEDUSA.config`, use Vuex.
- Add JSDocs.
- Use object destructuring where applicable.
- Simplify the events.
- **Add tests for basic functionality.**


themes-default/slim/test/specs/root-dirs.spec.js
```diff
@@ -0,0 +1,60 @@
+import test from 'ava';
+import Vuex from 'vuex';
+import VueRouter from 'vue-router';
+import { createLocalVue, mount } from '@vue/test-utils';
+import fixtures from '../__fixtures__/root-dirs';
+
+// Needs to be required otherwise nyc won't see it
+const RootDirs = require('../../static/js/templates/root-dirs.vue');
```
sharkykh: In case this PR gets merged after #4833, need to change this import to:
```js
import { RootDirs } from '../../static/js/templates';
```

PR #4739, opened 2018-07-24
Convert submenu to Vue

sharkykh:
This is set for v0.2.7 because it fixes the bug with the confirmation dialogs for removing shows, clearing history, etc.

## To do:
- [x] Remove the check for `controller !== 'schedule' && action !== 'previewRename'`.
		For `previewRename` - either remove the submenu argument from [the handler](https://github.com/pymedusa/Medusa/blob/c304860c5360b2e9fa118717a639c28578e5a27e/medusa/server/web/home/handler.py#L2084-L2090),
		or keep it - if we want it. Make sure removing that submenu has no effect.
- [x] Show selector - check `location.pathname` has `displayShow`?
- [x] Show selector - get current show slug from the url?
- [x] Cherry-picked from #4663 - VueRouter: Disable `not-found` redirection for now


sharkykh:
This is as far as I can go with this at the moment. The last Python conversion requires that all the routes be listed on the VueRouter (#4663).
This is good enough for now, though.

themes-default/slim/views/layouts/main.mako
```diff
@@ -61,7 +61,7 @@
 
             <app-header></app-header>
             % if submenu:
-            <%include file="/partials/submenu.mako"/>
+            <submenu></submenu>
```
OmgImAlexis: All non page/root components should use a 2 section name like `sub-menu`, `app-header`, etc.

themes-default/slim/views/vue-components/submenu.mako
```diff
@@ -0,0 +1,138 @@
+<script type="text/x-template" id="submenu-template">
+<%text>
+<div id="sub-menu-container" class="row shadow">
+    <div id="sub-menu" class="submenu-default hidden-print col-md-12">
+        <template v-for="menuItem in submenu">
+            <app-link :key="menuItem.title" :href="menuItem.path" class="btn-medusa top-5 bottom-5"
+                      v-if="!menuItem.confirm">
+                <span :class="`pull-left ${menuItem.icon || ''}`"></span> {{ menuItem.title }}
```
OmgImAlexis: Please use `:class="['pull-left', menuItem.icon]"` instead.

PR #4749, opened 2018-07-24
Fix manage/manageSearches

### LINKED ISSUE #4738: Refreshing scene exceptions on Manage Searches broken
**Describe the bug**
See title.

**To Reproduce**
1. Go to Manage > Manage Searches
2. Click on 'Force' under `Scene Exceptions`
3. Nothing happens

**Screenshots**
image

**Medusa:**
develop @ 1bc94568403e573fe3f84b464a268a4782ecc64a

sharkykh:
- Fix #4738
- Fix disabled buttons weren't actually disabled
- Convert more of the page to Vue format.

### DIFF
--- a/themes-default/slim/views/manage_manageSearches.mako
+++ b/themes-default/slim/views/manage_manageSearches.mako
@@ -1,156 +1,197 @@
 <%inherit file="/layouts/main.mako"/>
 <%!
+    import json
     from medusa import app
-    from medusa.common import SKIPPED, WANTED, UNAIRED, ARCHIVED, IGNORED, SNATCHED, SNATCHED_PROPER, SNATCHED_BEST, FAILED
-    from medusa.common import Quality, qualityPresets, statusStrings, qualityPresetStrings, cpu_presets
 %>
 <%block name="scripts">
 <script>
 window.app = {};
-const startVue = () => {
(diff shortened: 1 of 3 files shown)


themes-default/slim/views/manage_manageSearches.mako
```diff
@@ -1,156 +1,197 @@
...
+        }
+    },
+    methods: {
+        /**
+         * Trigger the force refresh of all the exception types.
+         */
+        forceSceneExceptionRefresh() {
+            const { updateExceptionData, sceneRefresh } = this;
+            // Start a spinner.
+            sceneRefresh.showSpinner = true;
```
OmgImAlexis: Should be using `this.` for setting vars, not updating the const.

sharkykh: Are you sure? `sceneRefresh` is an object. I've been using this technique right from the start.

p0psicles: Yeah this also feels like an edge case to me. Should we use let when changing mutables?

OmgImAlexis: Since it's updating the vue data object it shouldn't be using let or const. It should be using `this.`

PR #4758, opened 2018-07-26
Fix API error when an episode airdate is `datetime.date.min`

### LINKED ISSUE #4633: Edit show is broken, at least for Anime
Version : 0.2.6 (through docker)
Basic stuff tried : Ive tried to reboot, rebuild the container from the last image, same result.

TL;DR : After adding a new anime show, if I edit the show, parameters are set to something else, and if corrected and saved, File location is lost.

Step 1 : I've added a new anime show, with "HD1080p" as quality and anime and scene numbering options, and unchecked folder for seasons  (like 20 sec before the screen shot) I didn't choose to whitelist a release group since it's currently bugged (in another issue) : 
medusa 2018-07-08 00-35-28

 Step2 : I click on "Edit" on the show page : 
> On tab "MAIN" :  there "HD1080p" is replaced by "Custom" without any quality option checked, "default status episode" is empty. Also, the name of the Show is missing. 
> On tab "Format" : almost everything is wrong 
edit show 2018-07-08 00-36-55

edit show 2018-07-08 00-37-34

Step3 : once I've corrected everything and click "Save", I go back to the show page and now Location is missing ...
medusa - isekai maou to shoukan shoujo no dorei majutsu 2018-07-08 00-38-57

I've tried it with 3 different anime, same result.

Bonus : if you can  whitelisted release group while adding the show (if you can because it's currently with the old UX which is bugged), for exemple "HorribleSubs" you could end up like me with "H","o","r","i","b","l","e","S","u","b","s" :)

sharkykh:
Fixes #4633

See: https://github.com/pymedusa/Medusa/issues/4633#issuecomment-408007736 for the explanation.
In short, it affected users with timezones smaller than UTC.

Changelog updated in #4756

### DIFF
--- a/dredd/api-description.yml
+++ b/dredd/api-description.yml
@@ -1463,10 +1463,12 @@ definitions:
         type: string
       season:
         type: integer
-        minimum: 0
+        minimum: 1
+        x-nullable: true
       type:
         type: string
-        enum: [local]
+        enum: [null, local]
+        x-nullable: true
     example:
       series: tvdb301824
       name: TheBig
--- a/dredd/api-description.yml
+++ b/dredd/api-description.yml
@@ -942,6 +942,7 @@ definitions:
         type: string
         format: date-time
         description: Next episode air date
+        x-nullable: true
       runtime:
         type: integer
         minimum: 1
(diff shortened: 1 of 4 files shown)


sharkykh:
The tests are failing because `airDate` is can now be null.
Before I remove it, I'm going to try to allow for `nullable` values again...

sharkykh:
I've fixed all the nullable for `episode`, `series` and `alias` endpoints.
Needs testing, hold on.

I'll do config in another PR since it's the biggest change.

PR #4764, opened 2018-07-27
Fix home tabs

sharkykh:
It seems to work. Javascript seems to still work.


sharkykh:
Don't know what's wrong with the build, let's give Travis some time to recover and I'll restart it again later.

sharkykh:
Ready to merge.

PR #4836, opened 2018-08-02
Remove keys from config object.

p0psicles:
We don't want to patch all the fields that are provided by the api. Therefor these keys are removed just before the patch call is send.
Using destructuring because we don't want to remove the key from the vue data object.

This will remove some api warnings, when saving config_postProcessing.

### DIFF
--- a/themes-default/slim/views/config_postProcessing.mako
+++ b/themes-default/slim/views/config_postProcessing.mako
@@ -124,14 +124,22 @@ window.app = new Vue({
             // Disable the save button until we're done.
             this.saving = true;
 
-            let config = {
+            const config = {
                 postProcessing: this.postProcessing,
                 metadata: {
                     metadataProviders: this.metadataProviders
                 }
             };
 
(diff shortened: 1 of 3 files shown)


themes-default/slim/views/config_postProcessing.mako
```diff
@@ -124,14 +124,22 @@ window.app = new Vue({
...
                     metadataProviders: this.metadataProviders
                 }
             };
 
-            $store.dispatch('setConfig', {section: 'main', config: config}).then(() => {
+            // Clone the config into a new object
+            let configCopy = Object.assign({}, config);
+
+            // Use destructuring to remove the unwanted keys.
+            const { multiEpStrings, reflinkAvailable, ...rest} = configCopy.postProcessing
```
OmgImAlexis: Won't this cause any issue with `xo` or `ava`?

p0psicles: Ooh i dont know. Its valid destructuring. I checked. And our test are okay for now.

PR #4848, opened 2018-08-05
Vueify add show options

### LINKED ISSUE #4525: Black/White lists on Add New Show don't work
develop / e37859f7a3cadac6c72df8511fe754360f4e350b

Release groups black / white lists on the Add New Show page don't work (the buttons).
It used to not be visible due to a bug (I think), but now it is, and it needs to be fixed.

### LINKED ISSUE #4838: Adding Anime mess up release group initial  whitelist
Version : last master

When : while adding an Anime type show with whitelisted release groups. 
What : while adding a show and selecting one or more release group into the whitelist everything  looks fine. But after the show is created release group is not well set. 

Example : if I add HorribleSubs during creation it appears as 12 release group one per letter like that : H,o,r,r,i,b,l,e,S,u,b,s. Which make medusa download all wrong group at first scan :)

p0psicles:
Fixes #4838 
Fixes #4525 

### TODO:
- [x] finish `add-show-options.vue` component
- [x] release black/white list groups
- [x] Update `addShows_newShow.mako` to use the new component
- [x] save default options. We're using vue data now that changes.
- [x] The save default button should only be enabled if an option has changed.
- [ ] error handling when getting black/white lists. For example we cache the bl/wh list in dogpile. So often the api will use that. But when the bl/wh are not available from cache, they are requested through the anidb udp api, which is not that reliable. Requests can timeout. So we should handle that. Maybe give some feedback.

~~To properly use black/white lists from anidb, we need merge another PR, where I implemented dogpile caching for the anidb udp api.~~

A followup (new PR) will be to change all other pages, that use inc_addShowOptions.mako.
- [ ] addShows_addExistingShow.mako
- [ ] addShows_recommended.mako
- [ ] addShows_trendingShows.mako


themes-default/slim/static/js/store.js
```diff
@@ -563,5 +570,5 @@ Vue.use(VueNativeSock, websocketUrl, {
     reconnectionAttempts: 2, // (Number) number of reconnection attempts before giving up (Infinity),
     reconnectionDelay: 1000 // (Number) how long to initially wait before attempting a new (1000)
 });
-
+debugger;
```
OmgImAlexis: Missed debugger.

p0psicles: WIP. Already removed local

themes/dark/assets/js/templates/add-show-options.vue
```diff
@@ -0,0 +1,258 @@
+<template>
```
sharkykh: Notice you added that SFC to light and dark themes, it is no longer needed, so just remove the files and be sure to merge develop.

p0psicles: I noticed your pr

themes-default/slim/views/addShows_newShow.mako
```diff
@@ -81,16 +97,6 @@ window.app = new Vue({
                 }
             });
 
-            setTimeout(() => {
-                this.updateBlackWhiteList();
-
-                const { providedInfo } = this;
-                const { use, showId, showDir } = providedInfo;
-                if (use && showId !== 0 && showDir) {
```
sharkykh: Don't remove this part please, it is used to automatically go to the options section when you are adding a show from a folder, with provided metadata from `tvshow.nfo`

medusa/server/api/v2/config.py
```diff
@@ -296,6 +296,13 @@ def data_main():
         section_data['subtitles']['enabled'] = bool(app.USE_SUBTITLES)
         section_data['recentShows'] = app.SHOWS_RECENT
 
+        section_data['default'] = {}
+        section_data['default']['status'] = common.statusStrings[app.STATUS_DEFAULT]
+        section_data['default']['statusAfter'] = common.statusStrings[app.STATUS_DEFAULT_AFTER]
+        section_data['default']['seasonFolders'] = app.SEASON_FOLDERS_DEFAULT
+        section_data['default']['anime'] = app.ANIME_DEFAULT
+        section_data['default']['scene'] = app.SCENE_DEFAULT
```
sharkykh: Not sure if default is the most descriptive name. Maybe showDefaults? Open to discussion of course...

themes-default/slim/src/components/config-toggle-slider.vue
```diff
@@ -37,7 +37,7 @@ module.exports = {
     },
     data() {
         return {
-            localChecked: null
+            localChecked: null,
```
sharkykh: You could just revert this change

themes-default/slim/views/addShows_newShow.mako
```diff
@@ -212,13 +218,33 @@ window.app = new Vue({
...
 
             this.otherShows.forEach(nextShow => formData.append('other_shows', nextShow));
 
+            // Because we're using te toggle-button.vue component, we don't have valid form input's for these.
+            // Therefor we need to add these values manually to the form data.
+
+            formData.append('subtitles', subtitles ? '1' : '0');
+            formData.append('anime', anime ? '1' : '0');
+            formData.append('scene', scene ? '1' : '0');
+            formData.append('season_folders', seasonFolder ? '1' : '0');
```
sharkykh: All of these can use `Number(value)` instead of the conditionals, however it **should** (need to check add_shows.py code) also be working with any truthy value, and I think supplying booleans is the best option in the long term.

themes-default/slim/views/addShows_newShow.mako
```diff
@@ -212,13 +218,33 @@ window.app = new Vue({
...
+
+            for (name of release.whitelist) {
+                formData.append('whitelist', name);
+            }
+
+            for (name of release.blacklist) {
+                formData.append('blacklist', name);
+            }
+
+            const statusToDesc = { 'Wanted': 3, 'Skipped': 5, 'Ignored': 7 }
```
sharkykh: There's a module in the store dedicated specifically for those, `statuses`, please use it instead of hardcoded values in javascript. We should only hardcode in Python.

medusa/server/api/v2/config.py
p0psicles: I think showDefaults is also fine.

themes-default/slim/views/addShows_newShow.mako
p0psicles: Yeah I'd also rather pass booleans. And have the backend translate (as late as possible). Which in the long term should be done by the ORM, when there is no other way supported by db engine.

themes-default/slim/views/addShows_newShow.mako
p0psicles: No idea how to use that, sorry.

themes-default/slim/views/addShows_newShow.mako
p0psicles: Changed them to Number(subtitles).. etc.

themes-default/slim/views/addShows_newShow.mako
OmgImAlexis: `this.$store.state.statuses` I think?

sharkykh:
Sorry about the messy commits :disappointed:

ngosang:
If I press enter in the textbox... boom!
imagen

sharkykh:
@ngosang Umm yeah, this PR is still unfinished. What do you mean by "boom"?

ngosang:
> What do you mean by "boom"?

Sorry I was joking. The page crash, it shows a json file.

sharkykh:
@ngosang Thanks, fixed. I would have probably missed it.

OmgImAlexis:
@sharkykh pretty sure there's still the issue of hitting "search" while a search is going just cancels instead of starting the new one.

sharkykh:
@OmgImAlexis That's the first time I'm reading about it... On `develop`?

OmgImAlexis:
@sharkykh it's been there for a while now. If you click the "search" again it works. Pretty sure master has the same issue.

medusa/common.py
```diff
@@ -260,6 +260,21 @@ def split_quality(quality):
...
+    def is_valid_combined_quality(quality):
+        """
+        Check quality value to make sure it is a valid combined quality.
+
+        :param quality: Quality to check
+        :type quality: int
+        :return: True if valid, False if not
+        """
+        for cur_qual in Quality.qualityStrings:
+            if not (cur_qual & quality) and (cur_qual << 16 & quality):
```
sharkykh: Note to self: This is incorrect.

sharkykh:
While this is possibly stable enough to be merged to `develop`,
I don't think it should be included in the next version (0.2.9), so maybe push this to 0.2.10 as today is the due date for 0.2.9.

sharkykh:
@p0psicles About that last "to-do", is it just you want to handle errors when calling `home/fetch_releasegroups` and show a message to the user?

p0psicles:
More like what i recently implemented. Its not relevant anymore

p0psicles:
Lets merge?

sharkykh:
Sure, hopefully I didn't miss anything big.

p0psicles:
Faster we merge faster we notice.

PR #4833, opened 2018-08-02
Webpackify most of app

OmgImAlexis:
- The store has been split into modules to allow easier handling.
- The `displayShow` template has been moved to a new `.vue` file.
- All of the .vue files are now loaded via webpack instead of `httpVueLoader` which allows nyc to properly see all of our coverage in files and speeds up page loading times by quite a bit.
- Added 2 new tests for .vue components
- Moved ava require to import

https://github.com/pymedusa/Medusa/blob/1daa223128679ed16bb5ad3ff78d450e5bdc6a6f/themes-default/slim/views/displayShow.mako#L379-L402

PR #4259, opened 2018-05-27
Vueify config-postprocessing

p0psicles:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

Depends on: 
- [x] ~#4258~
- [x] ~#4288~
- [x] #4779 

Todo: 
- [x] fix metadata tab
- [x] Use the same saved message as in editShow
- [x] Move vue components to vue-loader
- [x] Double check responsiveness
- [x] Update openApi specs
- [x] update changelog

PR #4851, opened 2018-08-05
UI bugfixes

sharkykh:
- **Fixed:** Login page leaking the username and trying to call the API (it shouldn't)
   I believe this was caused by the webpack change, so no need for a changelog update.
- **Fixed:** Config pages (like notifications) a little malformed due to the change I suggested in #4259.
- **Fixed:** `SelectList` components CSV mode switch was broken (trying to mutate a prop).
   @p0psicles I hope my fix is okay with you.
   I believe this was caused by one of the changes in #4259, so no need for a changelog update.
- Don't copy Vue SFCs when building themes
   Since the webpack bundle now contains the components, we don't need the raw files in the built themes.

Some small meta changes:
  - Trim whitespaces from `config_postProcessing.mako`
  - Remove empty `core.js` (emptied in #4833)
  - Update app.js import (add the `?${sbPID}` for caching purposes)
  - Fix `RootDirs` test import (from require)
  - Remove `http-vue-loader` (now unused - ~@OmgImAlexis hopefully you're not planning on using it anymore~)
  - Remove unused `topImageHtml`.


themes-default/slim/gulpfile.js
```diff
@@ -59,7 +59,8 @@ const webpackedJsFiles = [
     'static/js/index.js',
     'static/js/api.js',
     'static/js/router.js',
-    'static/js/store/**/*.js'
+    'static/js/store/**/*.js',
+    'static/js/templates/**/*'
```
OmgImAlexis: While you have this open. I'm thinking we move templates to components since we now store components in there and we don't really need that exposes as a static file handler anymore either then since we don't have `httpVueLoader`.

sharkykh: Oh I see, move them to the root folder of the `slim` theme?

themes-default/slim/views/index.mako
```diff
@@ -1,6 +1,6 @@
 <%inherit file="/layouts/main.mako"/>
 <%block name="scripts">
-<script src="./js/app.js"></script>
+<script type="text/javascript" src="js/app.js?${sbPID}"></script>
```
OmgImAlexis: `type` isn't needed in HTML5.

themes-default/slim/gulpfile.js
OmgImAlexis: I was thinking `static/js/components` for now but removing the templates static file handler so they don't get exposed anymore. Actually now that you removed the copy that's probably fixed that anyways.

sharkykh: > I was thinking `static/js/components` for now

Yeah but theme will end up being a bundled project, so the `js` folder would only hold the webpack bundles, IMO

> but removing the templates static file handler

It's actually a static file handler for the `js` folder. I don't think there's another way to not serve the `templates` subfolder other than to split the handler, move the folder itself.

> Actually now that you removed the copy that's probably fixed that anyways.

Yes, there won't be any files to serve from that folder. Just need make sure older PRs don't reintroduce the files.

PR #4869, opened 2018-08-07
More linting

OmgImAlexis:
This disables a LOT of the jQuery stuff we use but only in the `src` dir.

I'll be fixing most if not all before merging.

PR #4871, opened 2018-08-07
Bower => Webpack

sharkykh:
This PR moves the following packages from the Bower file (`vender[.min].js`) to Webpack (`vendors.js`):
- `lodash` - stripped to only import `debounce`, the only function that is currently used
- `underscore.string` - removed completely. doesn't look like it was used.
- `jQuery`
- `Bootstrap` - limited to version `< 4` (3.3.7)
- Open Sans fonts

@OmgImAlexis 
I made some changes that will affect / be affected by #4869. ~What do you want to merge first?~
**This PR needs to be merged before the other PR.**


themes-default/slim/src/index.js
```diff
@@ -114,7 +121,17 @@ $.fn.extend({
     }
 });
 
-if (!document.location.pathname.includes('/login')) {
+// @FIXME: Workaround just for the time being!
+const log = {
```
OmgImAlexis: What's this needed for?

sharkykh: `log` is defined, I chose to make a workaround instead of changing the calls to `console.[..]`, like you did in #4869.

sharkykh:
I'm going to try to look at using [`mini-css-extract-plugin`](https://github.com/webpack-contrib/mini-css-extract-plugin#minimal-example) instead of injecting the style. It would be less prone to errors.

themes-default/slim/src/index.js
OmgImAlexis: I actually think the `log` should have been removed a while ago. I removed the old logger months and months ago.

sharkykh:
Last commit made it possible to easily import CSS files (see bootstrap example).
It should automatically add all the styles together into `css/vendors.css`.

PR #4874, opened 2018-08-07
Update Store/WebSocket code

sharkykh:
- [x] Utilize `passToStoreHandler` instead of a global mixin
- [x] Fix store being created twice (or rather, two separate stores are created) when loading the main application (from `app.js`).
       As a result, two separate WebSocket connections are created,
       and each socket is linked to a different store object.
- [x] Fix config `layout` getter (unused, but wasn't working)


sharkykh:
Regarding the store issue:
- Edit `./app.js`, before the `export default app` statement, add: `window.app = app;`
- Go to a page that doesn't use the application from `./app.js` (home page for example)
- Open console, run: `window.store === app.$store`, should be **`true`**
- Go to a page that **does** use the application from `./app.js` (Help & Info page for example)
- In the console, run: `window.store === app.$store`, should be **`false`**

I believe that's the issue right there - it's not using the same store object.

themes-default/slim/src/store/modules/config.js
```diff
@@ -263,7 +263,7 @@ const mutations = {
 };
 
 const getters = {
-    layout: layout => state => {
+    layout: state => layout => {
```
OmgImAlexis: This isn't right. This a higher level function, the state is the most inner function.

sharkykh: https://vuex.vuejs.org/guide/getters.html#method-style-access
```js
getters: {
  // ...
  getTodoById: (state) => (id) => {
    return state.todos.find(todo => todo.id === id)
  }
}
```

OmgImAlexis: I'm an idiot.

PR #4886, opened 2018-08-10
move shows to store module

OmgImAlexis:
As title states.


themes-default/slim/src/store/modules/shows.js
```diff
@@ -0,0 +1,50 @@
+import { ADD_SHOW } from '../mutation-types';
+
+const state = {
+    shows: []
```
sharkykh: I believe it will result in a store state of:
```js
store: {
	state: {
		shows: { // module
			shows: [] // module state
		}
	}
}
```

OmgImAlexis: This will still work so maybe it's not too bad.

```js
...mapState('shows', [
   'shows'
]),
```

sharkykh: The working solution is:
```js
    computed: {
        ...mapState({ shows: state => state.shows.shows }),
        localComputed() { ... }
    }
```

PR #4888, opened 2018-08-10
Vueify display show and snatch selection

OmgImAlexis:
Adds `show` to Vue object so we can start using it on that page as well as some helpers. I've also moved the snatch selection over the Vue as need that done to convert `showheader.mako` to a component.


OmgImAlexis:
Yes, this leads to another PR I have locally. I'll be moving more of it over there.

PR #4889, opened 2018-08-10
Move JS to SFC: home, status, manual postprocess

OmgImAlexis:
This moves the home, status and manual post processing pages over to vue templates.


OmgImAlexis:
@p0psicles did you clear cache? That's working fine here.

sharkykh:
@OmgImAlexis I can confirm the home page bug. Banner and Small Poster are broken as well.

PR #4892, opened 2018-08-10
Add missing commas

p0psicles:
Fixes the episode obj queries.

### DIFF
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -23,7 +23,7 @@ MIN_DB_VERSION = 40  # oldest db version we support migrating from
 MAX_DB_VERSION = 44
 
 # Used to check when checking for updates
-CURRENT_MINOR_DB_VERSION = 11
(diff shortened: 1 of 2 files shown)


medariox:
Now that I think about it... Shouldn't we also upgrade the minor DB version?

sharkykh:
@medariox I think you're right.
_Beginning to despise that migration system._

p0psicles:
@medariox is this what you mean?
@sharkykh whell be rid of it in a while. Just have to continue on the sqlalchemy PR with alembic support.

sharkykh:
Also keep in mind there are users who had issues with the previous upgrade and now they have database version `44.12` already. **This will be an issue.**!

p0psicles:
I've started on a branch with sqlalchemy. #3792 
I've also worked with peewee. It's fast. But beside that I can't really compare the two. I know that alembic works nice with sqlal. Don't know allot about the migration support for peewee.

PR #4949, opened 2018-08-23
convert python truncate to vue

OmgImAlexis:
The displayShow and snatch selection are the only two pages affected.


themes-default/slim/views/partials/showheader.mako
```diff
@@ -155,15 +155,11 @@
...
-                                % if len(show.plot) < 250:
-                                    <div>${show.plot}</div>
-                                % else:
-                                    <div>${show.plot[0:250]}<span style="display:none;">${show.plot[250:len(show.plot)]}</span>
-                                    <span class="imdbPlot" style="color:#6ae;cursor:pointer">show more..</span>
-                                    </div>
-                                % endif
-                                    </i></td></tr>
+                            <tr>
+                                <td colspan=2 style="padding-bottom:15px">
```
sharkykh:
```html
<td colspan="2" style="padding-bottom: 15px;">
```

themes-default/slim/views/partials/showheader.mako
```diff
@@ -155,15 +155,11 @@
...
+                                    <truncate v-on:toggle="reflowLayout()" clamp="..." :length="250" less="Show Less" :text="`${show.plot}`"></truncate>
```
sharkykh: 1. Why `v-on:toggle` and not `@toggle`?
2. IMO you shouldn't change the casing and the words used.
   `clamp="show more..." less="show less..."`

themes-default/slim/src/app.js
```diff
@@ -69,6 +70,9 @@ const app = new Vue({
     name: 'App',
     store,
     router,
+    components: {
+        Truncate
```
sharkykh: There should be an option to import the component as a plugin (look at the source code).
So it wouldn't use the `components` property, but rather use `Vue.use` like `toggle-button`. It's much cleaner IMO.

themes-default/slim/views/layouts/main.mako
```diff
@@ -188,6 +188,7 @@
                 // Global components
                 Vue.use(ToggleButton);
                 Vue.use(Snotify);
+                Vue.component('truncate', Truncate);
```
sharkykh: See comment on `app.js`.

sharkykh: See comment on `app.js`.

themes-default/slim/src/app.js
OmgImAlexis: It'd make more sense to use components than register it globally. I'd like to remove all global registrations eventually.

sharkykh: I agree about the global components, since you'd (probably) only use it in the `showheader` component (exactly like you do here) once it has been Vueified.
My point is, for now I think it's much cleaner to use `Vue.use(...)` for vendored components if possible.

sharkykh: It's should produce exactly the same result BTW - `Vue.use` registers globally, just like putting the `components` property on the main Vue instance.

PR #4959, opened 2018-08-25
Fix unsafe html in truncate component

sharkykh:
Fixes the displayShow / snatchSelection bug with unsafe HTML (reported on Slack).


themes-default/slim/views/partials/showheader.mako
```diff
@@ -154,13 +154,11 @@
                 <div id="summary" class="col-md-12">
                     <div id="show-summary" class="${'summaryFanArt' if app.FANART_BACKGROUND else ''} col-lg-9 col-md-8 col-sm-8 col-xs-12">
                         <table class="summaryTable pull-left">
-                            % if show.plot:
-                            <tr>
+                            <tr v-if="show.plot">
```
OmgImAlexis: Is `show` the js var available on all pages that use this?

PR #4961, opened 2018-08-25
Fix adba error handling

p0psicles:
The show `How not to summon a demon lord` fails to get release group info from anidb.
This is an anidb issue. As they are sending a bad response.

The problem was that 1. the adba lib, can't properly handle the UnicodeDecodeException.
And 2. the to_json of the Series class, was not handling the previously raised exception.

Note it still blocks the apiv2 tornado thread (which we currently only have 1) for a while. But after a few minutes it's given free again. The release groups retrieved (none), are cached in anidb.dbm, so you'll not have this error often (at least until cache expires).

Still need to have a proper solution for this.


p0psicles:
@sharkykh need to (merge) rebase on develop, after your PR is merged.

medusa/tv/series.py
```diff
@@ -2021,7 +2022,12 @@ def to_json(self, detailed=True):
         if self.is_anime:
             data['config']['release']['blacklist'] = bw_list.blacklist
             data['config']['release']['whitelist'] = bw_list.whitelist
-            data['config']['release']['allgroups'] = get_release_groups_for_anime(self.name)
+            try:
+                data['config']['release']['allgroups'] = get_release_groups_for_anime(self.name)
+            except AnidbAdbaConnectionException:
```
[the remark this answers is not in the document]
p0psicles: I thought a warning is fitting. Then the user at least knows why its not getting any release groups when they are available in anidb.

medusa/tv/series.py
```diff
@@ -2021,7 +2022,15 @@ def to_json(self, detailed=True):
...
+                data['config']['release']['allgroups'] = []
+                log.debug(
+                    'An anidb adba exception occurred when attempting to get the release groups for the show {show}',
+                    {'show': self.name}
```
sharkykh: Well it's not really what I meant,
I suggested you add the actual error to the log...

ext/readme.md
```diff
@@ -1,7 +1,7 @@
 ## ext
  Status  |  Package  |  Version / Commit  | Usage | Notes
 :------: | :-------: | :----------------: | :---- | :----
-:: | `adba` | pymedusa/[119b9d3](https://github.com/pymedusa/adba/tree/119b9d30a30feb97bcea5e1dc742c82340300091) | **`medusa`** | -
+:: | `adba` | pymedusa/[119b9d3](https://github.com/pymedusa/adba/tree/70bc381a75e20e1813c848c1edb7c6f16987397e) | **`medusa`** | -
```
sharkykh:
```diff
- pymedusa/[119b9d3](https://github.com/pymedusa/adba/tree/70bc381a75e20e1813c848c1edb7c6f16987397e)
+ pymedusa/[70bc381](https://github.com/pymedusa/adba/tree/70bc381a75e20e1813c848c1edb7c6f16987397e) 
```

Maybe I should just change the link to say `pymedusa/[commit]` instead, it would be easier to update...

p0psicles:
@sharkykh better?

sharkykh:
We might have to revisit it later - as the `Series.to_json()` is going to be called a lot. Maybe we don't call anidb unless the data is requested:
`displayShow` (?) / `editShow` need it,
but the home page doesn't need it to create the show list.

p0psicles:
Yeah thats in line with what i said before. That we should move this to a dedicated call.

PR #4956, opened 2018-08-24
Improve Webpack copy and watch

sharkykh:
- [x] Replace `filemanager-webpack-plugin` with `copy-webpack-plugin`: Closes #4954
- [x] Copy templates (mako) using Webpack: Closes #4953
- [x] Copy 'old' javascript using Webpack

This means:
1. `yarn gulp sync` should be used much less frequently now.
**In any case** Travis will alert us if we missed something, as it runs `yarn gulp sync` each build.
2. The `--watch` feature of Webpack now watches these files as well, so any changes are automatically copied.

Issue #4954, opened 2018-08-24
Webpack: Replace `filemanager` plugin with `copy` plugin


sharkykh:
[copy-webpack-plugin](https://www.npmjs.com/package/copy-webpack-plugin) simply has more options for copying (which we may need soon).

sharkykh:
As it turns out, we shouldn't **replace** the `filemanager-webpack-plugin` plugin,
but use `copy-webpack-plugin` **as well**.
This created a bug where it would copy and outdated bundle on the first build.
#4962 will fix this issue.

PR #4962, opened 2018-08-25
Use FileManagerPlugin to copy bundled assets

sharkykh:
Related: #4956, #4954

- [x] Use FileManagerPlugin to copy bundled assets
	* FileManagerPlugin runs **after** the bundle has finished (on `--watch` as well).
	* CopyWebpackPlugin runs **as the bundle is generated**, so it copies the old files.
	  It's fine for the files from `./static` and `./views` as they are not created using Webpack.
- [x] Clean `./dist` folder before build.
Note that when using `--watch`, this plugin will only clean the folder on the initial build.
(I initially thought this was the plugin that caused an issue, but it wasn't).

### DIFF
--- a/themes-default/slim/webpack.config.js
+++ b/themes-default/slim/webpack.config.js
@@ -2,6 +2,7 @@ const path = require('path');
 const { ProvidePlugin } = require('webpack');
 const VueLoaderPlugin = require('vue-loader/lib/plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const FileManagerPlugin = require('filemanager-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const pkg = require('./package.json');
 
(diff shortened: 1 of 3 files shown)


themes-default/slim/webpack.config.js
```diff
@@ -15,17 +17,16 @@ const { cssThemes } = pkg.config;
 const perTheme = action => Object.values(cssThemes).map(theme => action(theme));
 
 /**
- * Helper function to simplify CopyWebpackPlugin configuration when copying assets from `./dist`.
+ * Helper function to simplify FileManagerPlugin configuration when copying assets from `./dist`.
  * To be used in-conjunction-with `perTheme`.
  * @param {string} type - Asset type (e.g. `js`, `css`, `fonts`). Must be the same as the folder name in `./dist`.
  * @param {string} [search] - Glob-like string to match files. (default: `**`)
- * @returns {function} - A function the recieves the theme object from `perTheme`.
+ * @returns {function} - A function the receives the theme object from `perTheme`.
```
OmgImAlexis: "that receives"

sharkykh: Oh man the typos :| Let's just ignore "the" (that) one for now

Issue #3351, opened 2017-11-30
[ FEATURE ] Make use of a Post Processing queue


Rouzax:
Would it be possible to implement a Post Process queue like SickRage is currently doing?
In Medusa currently when it is busy post processing in the background the web GUI does not respond anymore.
I mean this pull from SickRage https://github.com/SickRage/SickRage/pull/2839

labrys:
Closing this as it has been added to the  #282 **Master Feature List** to track all feature requests.  Discussion specific to this feature can continue in this thread.

Rouzax:
This also for fills https://github.com/pymedusa/Medusa/issues/3351
Does this mean that when I call the API it will give an acknowledge directly or will it still wait for the process to complete and it can just handle more "threads" at a time?

PR #4970, opened 2018-08-26
Multi-threaded API v2 handler

### LINKED ISSUE #4963: Implement multi-threading in the API v2 handler
For reference, the current WebHandler:
https://github.com/pymedusa/Medusa/blob/1b62217a9ea7dbceae924d8f3dd2bcb46caaecda/medusa/server/web/core/base.py#L211-L262

sharkykh:
Fixes #4963


medusa/server/api/v2/base.py
```diff
@@ -167,12 +234,13 @@ def create_app_handler(cls, base):
 
     def _handle_request_exception(self, e):
```
p0psicles: can `e` be changed to `error`?

p0psicles:
Let's keep it up2date with develop?

sharkykh:
There's still an issue that the errors generated inside the routes are not being handled correctly.
To test this just raise an exception inside one of the routes, `GET /config` for example..

sharkykh:
https://github.com/pymedusa/Medusa/blob/688f4a67a00f47253524b54b2a417b5c88122ef9/medusa/server/api/v2/base.py#L89-L95

I still think that solution is a bit crude, but I don't have any better ideas at the moment.
That said, I do believe it's stable now, so I guess you can go ahead and merge...

Rouzax:
@sharkykh 
This also fulfills https://github.com/pymedusa/Medusa/issues/3351
Does this mean that when I call the API it will give an acknowledge directly or will it still wait for the process to complete and it can just handle more "threads" at a time?

p0psicles:
It can handle more threads. It will always wait for the process thats providing its result

p0psicles:
Oh this is unrelated to 3351

Rouzax:
Is the GUI still unresponsive when post processing? 
If it is responsive that actually fulfils my request in the original #3351

Rouzax:
I do it when a download is done by calling the API. So in occasion when I browse Medusa it will hang because it is post processing in the back end.
Not a big issue. Not sure with the new threaded post processing if that is still the case.

labrys:
One thing to watch for is locked threads due to failed post processing.  They will lock up resources.  This happens most commonly when trying to postprocess an invalid rar file.

Rouzax:
Medusa will never encounter a rare in my environment as I already unrar files with a post process script that is kicked off when qbitorrent finishes the download, the last step is calling the Medusa api to post process.

PR #4979, opened 2018-08-27
Improve shows module

OmgImAlexis:
- Added `getSeason` and `getEpisode` getters
- Replaced `getShowByName` with getShowByTitle
- Improved comments
- Fixed merging detailed show into existing


themes-default/slim/src/store/modules/shows.js
```diff
@@ -6,19 +7,40 @@ const state = {
...
     [ADD_SHOW](state, show) {
-        const { shows } = state;
-        const showExists = shows.filter(({ id, indexer }) => id[indexer] === show.id[indexer]).length === 1;
-        if (showExists) {
-            state.shows[shows.indexOf(showExists)] = show;
-        } else {
+        const existingShow = state.shows.find(({ id, indexer }) => Number(show.id[show.indexer]) === Number(id[indexer]));
+
+        if (!existingShow) {
+            console.debug(`Adding ${show.title || show.indexer + String(show.id)} as it wasn't found in the shows array`, show);
```
sharkykh: I personally don't like that much console spam, but that's your call.

sharkykh:
You can avoid looping over `state.shows` twice if you get the index of the existing show when you loop over it the first time in `state.shows.find(...)`.

themes-default/slim/src/store/modules/shows.js
OmgImAlexis: I'd like to try and add more debug logging where possible so we can get easier feedback from users when things go wrong by them simply enabling "debug" in the console area.

Issue #4975, opened 2018-08-26
Python3 support / migration


ngosang:
What's stopping you from using python3?

* In 1 year and half python2 will be unsupported => https://pythonclock.org/
* In most common distributions Ubuntu, Archlinux... python3 is the default version, so to use python2 you have to install a lot of dependencies.

p0psicles:
Precious dev time

p0psicles:
Moving to py3 is not allot of work. But supporting py2 and py3 is. Wanna chip in?

ngosang:
> Moving to py3 is not allot of work. But supporting py2 and py3 is.

It's. I did some test and there are a lot of code not working in python3 and a lot of hacks specific for python2. At some point python2  could be totally removed in future releases.

The more painful issue is external dependencies. **Why do you copy 3rd party modules in ext package instead of installing the packages with pip?**

The external modules in ext package are old versions non compatible with python3. The first step imho is to remove the "ext" packages, install latest version from pip and check if it's working in python2.
I guess most latest versions of all dependencies are already compatible with python2/3. So you can keep launching python2 releases with latest dependencies and the future migration to python3 will be easier.

ngosang:
I made a lot of changes and I have medusa running with python3. There are still minor issues but it's functional.
I can open a massive PR so you can test it and give feedback, but I don't want to maintain python2 compatibility. What do you think?

OmgImAlexis:
@ngosang we move very fast here, I wouldn't recommend a massive PR as it'll just make more work for us. If you'd like to help maybe start small.

@medariox should we start an issue to track what's needed to get to full python 3 support just as we do with vue?

ngosang:
I uploaded my changes here => https://github.com/ngosang/Medusa/commit/161c9364d270226b66dcdc629ac9904fa501e18b
95% of the features are working

p0psicles:
It needs to also run on py2. We cant break it for the userbase which are currently running py2. So if the module has a py3 only version, we need to add both. And select the correct lib. 

If we need to monkey patch it, we might need to use something like six.

p0psicles:
not just too big, it's focused to run explicitly on python3.

like:
```python
if from_encoding and isinstance(markup, str):
            warnings.warn("You provided Unicode markup but also provided a value for from_encoding. Your from_encoding will be ignored.")	            warnings.warn("You provided Unicode markup but also provided a value for from_encoding. Your from_encoding will be ignored.")
[...]

PR #4980, opened 2018-08-27
[Python3] Add packages compatible only with python3 (part 2)

ngosang:
This is the first PR to make pymedusa compatible with python2 and python3 at the same time.
**This PR is the 2nd in [Python3] series.** Code still non working in python3.
Main discussion: #4975

**You wanna help?**
* Test with python2. Everything should work as usual. Especially test strings/series with non ascii characters. Code still non working in python3.

TODO:
- [x] Add folders ext2, lib3 and ext3 into Python PATH. Remove old package folders in `GitUpdateManager.clean` instead of adding new code. Rework after #5145
- [x] Move backported python2 libs (future) to ext2 folder. Check, there is future and futures. Complete list https://github.com/pymedusa/Medusa/blob/develop/ext/readme.md
- [x] Update `GitUpdateManager.clean` with the new folders. https://github.com/pymedusa/Medusa/blob/1435fcbfe0412abca06a18a5ab7ee56ec60370fe/medusa/version_checker.py#L708
- [x] Update dependencies/requirements.txt lists https://github.com/pymedusa/Medusa/blob/develop/ext/readme.md https://github.com/pymedusa/Medusa/blob/develop/lib/readme.md
- [x] Add `ext3/sgmllib` package
- [x] Add `ext3/bs4` package (same version as `ext/bs4`)
- [x] Add `ext3/github` package (same version as `ext/github`)
- [x] Update `ext3/yaml` after #5140 is merged.
~~Add `ext3/yaml` package (same version as `ext/yaml`)~~
- [x] Remove `ext3/httplib2` after #5091 is merged.
~~Add `ext3/httplib2` package~~
- [x] Remove `lib3/bencode` after #5140 is merged.
~~Add `lib3/bencode` package~~
- [x] Update `lib/unrar2`.  Done in #5096 
~~I updated `lib/unrar2` lib for both python2 and python3. Please test it still works in python2.~~
- [x] Update `ext/pythontwitter` . Done in #4990
~~The `ext/pythontwitter` package is really old and not compatible with python3. I added 2 new packages in `ext3/pythontwitter` and `ext3/requests_oauthlib`. These packages are only used in python3 (because I can't test in python2) but should work with python2 also. Please copy:
`ext3/pythontwitter => ext/pythontwitter` (overwrite)
`ext3/requests_oauthlib => ext3/requests_oauthlib`~~

Libraries non compatible with python3 but the code works if you don't use that features (will be fixed in other PRs):
* lib/synchronousdeluge =>  #5097
* lib/rtorrent

Disclaimer:
> I'm testing careful with both python2 and python3 interpreters. With python2 you should expect the same functionally than development branch. Python3 is a long term goal, first versions won't be functional or even won't run.


ngosang:
PR rebased. Removed changes related to unrar2 and twitter as they have been fixed in #5096 and #4990

@sharkykh I think this is ready for review. I didn't change any package for python2, just added new packages for python3 in a different folder. Python2 should work as it does now. Code still non working in python3. The first PR with python3 support is #4982

sharkykh:
We will probably give the current develop some time to settle before introducing more Python 3 changes.

But for now what I can see from the commits messages:
1. `sgmllib`
	Should be in `ext3`.
2. `lib/bencode` is modified (commented out exception handler)
	I need to implement a [fix I already implemented in SickRage](https://github.com/SickRage/SickRage/pull/3871/files) in order to be able to install a vanilla `bencode` that works for both major versions of Python.
3. Why did you add `httplib2`?
	I would appreciate it if you could list reasons for any new packages you added
	(ones that are not currently in `ext` or `lib`).
4. All the following modules you moved are part of the `future` package.
	- `_markupbase`
	- `html`
	- `http`
	- `queue`
	- `socketserver`
	- `xmlrpc`
	>
	See the entry [on the readme](https://github.com/pymedusa/Medusa/blob/develop/ext/readme.md) for the full list.
	I believe it should be moved in whole from `ext` to `ext2`.
	I am not 100% sure, but I think it shouldn't be in the `PATH` for Python 3.
5. `GitUpdateManager.clean` method on `version_checker.py` needs to be updated for the new folders.
https://github.com/pymedusa/Medusa/blob/1435fcbfe0412abca06a18a5ab7ee56ec60370fe/medusa/version_checker.py#L708-L726

sharkykh:
And finally I want to have the same lists I added to `ext/` and `lib/` because it makes it so much more easier to track versions and ultimately it will help a lot in moving stuff to `requirements*.txt`.
I have a tool I wrote (locally) to convert the markdown list to a `requirements.txt` file.

I will create the lists, but I'll appreciate your help with keeping them true to what's actually in the folders :)

sharkykh:
I edited my comment above about `httplib2`, so please remove `ext3/httplib2` because it's now unused and I'm removing it in #5091.

ngosang:
@sharkykh I think you already noticed, but the main idea is:
* `ext` and `lib` contains code compatible py2/py3
* `ext2` and `lib2` contains code compatible only py2. Makes py3 to crash. Code from `futures` (already built in in python3), and the packages I listed below in point 3)
* `ext3` and `lib3` contains code compatible only py3. Makes py2 to crash. The packages I listed below in point 3)

In python2 we import `ext`, `lib`, `ext2` and `lib2`. Those packages are not changed, just moved from lib to lib2...

> We will probably give the current develop some time to settle before introducing more Python 3 changes.

Up to you.  :smile:  If it helps I can split it in smaller PRs, but I need this merged to keep working

> 1 sgmllib Should be in ext3.

Sure, I will move.

> 2 lib/bencode is modified (commented out exception handler)

I also prefer to have only 1 lib for py2/py3. When you open the other PR I can remove it from here. By now I need it.

> 3 Why did you add httplib2? I would appreciate it if you could list reasons for any new packages you added

* `ext3/bs4` if you open `setup.py` in the package, you will see `use_2to3 = True` that's mean than the code is written in python2 and the installer converts de code to python3 if you are in python3 using ( https://docs.python.org/3/library/2to3.html ) I converted the code to python3 with that tool so we have python2 in ext/bs4 and python3 in ext3/bs4. Now I think the python2 should be in ext2/bs4
* `ext3/github package` Same as `ext3/bs4` https://github.com/PyGithub/PyGithub/blob/master/setup.py
* `ext3/httplib2` if you open the package you will see python2 and python3 folders. Instead of converting the code, they maintain 2 different codes, something like I'm doing in this PR. https://github.com/httplib2/httplib2
* `ext3/yaml` Same as `ext3/httplib2` https://github.com/yaml/pyyaml/releases/tag/3.13

> 4 All the following modules you moved are part of the future package. 

I moved only the packages crashing in python3. I'm going to move all of them and see what happens.

> 5 GitUpdateManager.clean method on version_checker.py needs to be updated for the new folders

Sure, I will do

> I will create the lists, but I'll appreciate your help with keeping them true to what's actually in the folders :)

I can do it, but maybe we need some changes in the table. My proposal is to split column `package` in ext, ext2 and ext3. Most of the packages will be in ext and the others should be both in ext2 and ext3 https://github.com/pymedusa/Medusa/blob/develop/ext/readme.md

> I edited my comment above about httplib2, so please remove ext3/httplib2

I will do

sharkykh:
>I think you already noticed, but the main idea is:
>* `ext` and `lib` contains code compatible py2/py3
>* `ext2` and `lib2` contains code compatible only py2. Makes py3 to crash. Code from `futures` (already built in in python3), and the packages I listed below in point 3)
>* `ext3` and `lib3` contains code compatible only py3. Makes py2 to crash. The packages I listed below in point 3)
>
>In python2 we import `ext`, `lib`, `ext2` and `lib2`. Those packages are not changed, just moved from lib to lib2...

Yep, you and I have the same way of thinking. That's what I would have done too.


> If it helps I can split it in smaller PRs, but I need this merged to keep working

I will let you know if this is too big, but currently this PR is only getting smaller :wink: 

> * `ext3/bs4` if you open `setup.py` in the package, you will see `use_2to3 = True` that's mean than the code is written in python2 and the installer converts de code to python3 if you are in python3 using ( https://docs.python.org/3/library/2to3.html ) I converted the code to python3 with that tool so we have python2 in ext/bs4 and python3 in ext3/bs4. Now I think the python2 should be in ext2/bs4

Thank you for all the explanations!
Yep, it should. But that kind of thing (moving python 2 packages from `ext/` -> `ext2/`) should be in a separate PR though, IMO.

> * `ext3/github package` Same as `ext3/bs4` https://github.com/PyGithub/PyGithub/blob/master/setup.py

Which version did you install? It should be 1.40 as that's the last one that was published.

> * `ext3/yaml` Same as `ext3/httplib2` https://github.com/yaml/pyyaml/releases/tag/3.13

Did you install v3.13 ?
Then I probably need to update `ext/yaml` to 3.13 too since it's currently on 3.12.

> I can do it, but maybe we need some changes in the table. My proposal is to split column `package` in ext, ext2 and ext3. Most of the packages will be in ext and the others should be both in ext2 and ext3 https://github.com/pymedusa/Medusa/blob/develop/ext/readme.md

That's a neat idea. I was actually thinking I'd make new a list for each folder, but I think I like your idea better. Might get a little cramped in there :stuck_out_tongue:

sharkykh:
Heads up: #5140 will move `lib/bencode` to `ext/bencode` and it will be Python 2/3 compatible.

medusa/init/__init__.py
```diff
@@ -11,6 +11,10 @@
 import site
 import sys
 
+import six
```
sharkykh: Remove this import
1. It's unused
2. It isn't available for import here because the vendored libs are still not in `sys.path`.

medusa/init/__init__.py
```diff
@@ -11,6 +11,10 @@
...
+
+from .. import app
```
sharkykh: Please do not use relative imports, the project uses `from medusa import app`.

medusa/init/__init__.py
```diff
@@ -38,12 +42,21 @@ def _check_python_version():
         sys.exit(1)
 
 
-def _lib_location():
-    return os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..', 'lib'))
+def _remove_dead_python_bytecode():
```
sharkykh: Please remove this method.
1. It's unused.
2. It removes files when the `medusa` package is first imported (per python session), that is not a good idea IMO.
	(you already have an item in your to-do list to update the `UpdateManager` code as I suggested)

sharkykh:
> - [ ] Update `ext3/yaml` after #5140 is merged.
~~Add `ext3/yaml` package (same version as `ext/yaml`)~~

Sorry if I wasn't clear:
Don't remove `ext3/yaml`, just move `ext/yaml` to `ext2/yaml`.

medusa/init/__init__.py
ngosang: done

medusa/init/__init__.py
ngosang: done

medusa/init/__init__.py
ngosang: done

ngosang:
@sharkykh I did all requested changes. I tested all in py2 & py3
The only remaining thing is the task to update the lib/ext tables in the doc. Could you help?
Current versions for py2 and py3:
* bs4 4.6.3
* github 1.43.1
* yaml 3.13

Remember, after this PR you should keep 2 versions of these packages. #5200

ngosang:
@sharkykh in the first commit I duplicated the `_configure_syspath()` code in the test __init__ file. Where we can move the shared code?

sharkykh:
> @sharkykh I did all requested changes. I tested all in py2 & py3
> The only remaining thing is the task to update the lib/ext tables in the doc. Could you help?
> Current versions for py2 and py3:
> 
> * bs4 4.6.3
> * github 1.43.1
> * yaml 3.13
> 
> Remember, after this PR you should keep 2 versions of these packages. #5200

Sorry, I still don't have time to work on the list...

> @sharkykh in the first commit I duplicated the `_configure_syspath()` code in the test **init** file. Where we can move the shared code?

I don't think you need to go that far. All you have to do is:

```
sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))

import medusa
```

Once you import `medusa` (or anything from `medusa`), all the patches are applied.
This is done in: https://github.com/pymedusa/Medusa/blob/develop/medusa/__init__.py
The `medusa` package usually gets imported in the actual tests anyway.

ngosang:
@sharkykh Updated PR all DONE!
* Updated PyGithub to v1.43.2 #5200
* I followed your advice in `tests init`
* Updated lib and ext readme

@p0psicles I saw your changes in #5232 but I think it's not useful. After this PR all packages are compatible with python2 and python3. Even all packages in `lib` folder. I prefer to show the folders in the table so, when you update `github` package. You remember to do it in the 2 folders.

ngosang:
@OmgImAlexis done

p0psicles:
Creating merge commit, as the commits are just to clean.

PR #5087, opened 2018-08-30
vueify start of showheader

OmgImAlexis:
This converts a majority of the `showheader.mako` to Vue.

The two props are used to render the component dynamically without needing to worry about the query params for example when testing or when the "show selector" changes.


OmgImAlexis:
@sharkykh I built the themes?

```
--------------------------------------------
 M themes/dark/assets/js/app.js
 M themes/dark/assets/js/index.js
 M themes/dark/assets/js/vendors.js
 M themes/light/assets/js/app.js
 M themes/light/assets/js/index.js
 M themes/light/assets/js/vendors.js
```

sharkykh:
@OmgImAlexis 
I believe your dependencies are outdated.
Try `yarn install` and then rebuild.

themes-default/slim/yarn.lock
```diff
@@ -5127,9 +5131,18 @@ is-valid-glob@^0.3.0:
...
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/is-whitespace-character/-/is-whitespace-character-1.0.2.tgz#ede53b4c6f6fb3874533751ec9280d01928d03ed"
+=======
+is-visible@2.2.0:
+  version "2.2.0"
+  resolved "https://registry.yarnpkg.com/is-visible/-/is-visible-2.2.0.tgz#722b3acff3694a30074927e1156f58ae87db0e9c"
+  dependencies:
+    iselement "^1.1.3"
+    style-properties "^1.3.1"
+>>>>>>> Stashed changes
```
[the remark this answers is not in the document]
OmgImAlexis: Not sure how I missed that. 🤔

themes-default/slim/src/components/display-show.vue
```diff
@@ -265,25 +285,20 @@ export default {
...
-        $('.collapse.toggle').on('hide.bs.collapse', function() {
+        // Changes the button when clicked for collapsing/expanding the season to show/hide episodes
+        document.addEventListener('hide.bs.collapse', function() {
             const reg = /collapseSeason-(\d+)/g;
             const result = reg.exec(this.id);
             $('#showseason-' + result[1]).text('Show Episodes');
             $('#season-' + result[1] + '-cols').addClass('shadow');
         });
-        $('.collapse.toggle').on('show.bs.collapse', function() {
+        document.addEventListener('show.bs.collapse', function() {
```
OmgImAlexis: There's a good chance this is broken.

sharkykh:
>The two props are used to render the component dynamically without needing to worry about the query params for example when testing or when the "show selector" changes.

Just for the record, looks like you *can* mock `$route` in tests.
But it is probably easier / better just to use props. :+1: 
https://vue-test-utils.vuejs.org/guides/using-with-vue-router.html#mocking-route-and-router

themes-default/slim/views/partials/showheader.mako
```diff
@@ -223,16 +203,16 @@
...
-                            % if app.USE_SUBTITLES:
-                            <tr><td class="showLegend">Subtitles: </td><td><img src="images/${("no16.png", "yes16.png")[bool(show.subtitles)]}" alt="${("N", "Y")[bool(show.subtitles)]}" width="16" height="16" /></td></tr>
-                            % endif
+                            <template v-if="config.subtitles.enabled">
+                                <tr><td class="showLegend">Subtitles: </td><td><img :src="'images/' + (show.config.subtitlesEnabled ? 'yes' : 'no') + '16.png'" :alt="show.config.subtitlesEnabled ? 'Y' : 'N'" width="16" height="16" /></td></tr>
+                            </template>
                             <tr><td class="showLegend">Season Folders: </td><td><img src="images/${("no16.png", "yes16.png")[bool(show.season_folders or app.NAMING_FORCE_FOLDERS)]}" alt="${("N", "Y")[bool(show.season_folders or app.NAMING_FORCE_FOLDERS)]}" width="16" height="16" /></td></tr>
```
sharkykh:
```vue
<tr><td class="showLegend">Season Folders: </td><td><img :src="'images/' + ((show.config.seasonFolders || config.namingForceFolders) ? 'yes' : 'no') + '16.png'" :alt="(show.config.seasonFolders || config.namingForceFolders) ? 'Y' : 'N'" width="16" height="16" /></td></tr>
```

themes-default/slim/views/partials/showheader.mako
```diff
@@ -152,7 +136,7 @@
             <div class="row">
                 <!-- Show Summary -->
                 <div id="summary" class="col-md-12">
-                    <div id="show-summary" class="${'summaryFanArt' if app.FANART_BACKGROUND else ''} col-lg-9 col-md-8 col-sm-8 col-xs-12">
+                    <div id="show-summary" :class="[{ summaryFanArt: config.fanArtBackground }, 'col-lg-9', 'col-md-8', 'col-sm-8', 'col-xs-12']">
```
sharkykh: `config.fanartBackground`
[Lowercase `a`.](https://github.com/pymedusa/Medusa/blob/feature/vueify-showheader/themes-default/slim/src/store/modules/config.js#L73)

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,45 +77,46 @@
...
+                    </app-link>
                  % if xem_numbering or xem_absolute_numbering:
                      <app-link href="http://thexem.de/search?q=${show.name | h}" title="http://thexem.de/search?q-${show.name | h}">
                          <img alt="[xem]" height="16" width="16" src="images/xem.png" style="margin-top: -1px; vertical-align:middle;"/>
                      </app-link>
                  % endif
-                     <app-link href="https://fanart.tv/series/${show.indexerid}" title="https://fanart.tv/series/${show.name | h}"><img alt="[fanart.tv]" height="16" width="16" src="images/fanart.tv.png" class="fanart"/></app-link>
+                    ##  @TODO: Is this meant to use show.title for title but show.id[show.indexer] for href?
+                    <app-link :href="'https://fanart.tv/series/' + show.id[show.indexer]" :title="'https://fanart.tv/series/' + show.title"><img alt="[fanart.tv]" height="16" width="16" src="images/fanart.tv.png" class="fanart"/></app-link>
```
sharkykh: 1. `show.id[show.indexer]` for both.
2. I think `fanart.tv` only supports TVDB ids so it should only be shown for shows that have a TVDB id
3. The id could possibly be in `show.externals` non-TVDB shows, but that's not available through the API right now...

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,45 +77,46 @@
...
+                        </app-link>
+                    </template>
+                    <app-link v-if="show.id.trakt" :href="'https://trakt.tv/shows/' + show.id.trakt" :title="'https://trakt.tv/shows/' + show.id.trakt">
                         <img alt="[trakt]" height="16" width="16" src="images/trakt.png" />
                     </app-link>
-                % endif
-                 <app-link href="${indexerApi(show.indexer).config['show_url']}${show.indexerid}" title="${indexerApi(show.indexer).config["show_url"] + str(show.indexerid)}">
-                     <img alt="${indexerApi(show.indexer).name | h}" height="16" width="16" src="images/${indexerApi(show.indexer).config["icon"]}" style="margin-top: -1px; vertical-align:middle;"/>
-                 </app-link>
+                    ## @TODO: What is this?
```
sharkykh: The link to the show on the indexer website.
image

themes-default/slim/views/partials/showheader.mako
```diff
@@ -10,64 +10,47 @@
 <%namespace file="/inc_defs.mako" import="renderQualityPill"/>
 
 <div class="row">
-    <div id="showtitle" class="col-lg-12" data-showname="${show.name | h}">
+    <div id="showtitle" class="col-lg-12" :data-showname="show.title">
```
sharkykh: Can't find any usage for `data-showname` (or `data('showname')`) in the code.
Can probably be removed.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -10,64 +10,47 @@
...
         <div>
-            <h1 class="title" data-indexer-name="${show.indexer_name}" data-series-id="${show.indexerid}" id="scene_exception_${show.indexerid}">
-                <app-link href="home/displayShow?indexername=${show.indexer_name}&seriesid=${show.indexerid}" class="snatchTitle">${show.name | h}</app-link>
+            <h1 class="title" :data-indexer-name="show.indexer" :data-series-id="show.id[show.indexer]" :id="'scene_exception_' + show.id[show.indexer]">
```
sharkykh:
```
:data-indexer-name="show.indexer"
:data-series-id="show.id[show.indexer]"
:id="'scene_exception_' + show.id[show.indexer]"
```

As far as I can tell, you can remove these three props.
The first two **are** used in [scene-exceptions-tooltip.js](https://github.com/pymedusa/Medusa/blob/feature/vueify-showheader/themes-default/slim/static/js/scene-exceptions-tooltip.js#L2-L4), however the file isn't imported anywhere ATM.
That file is used to put a 'qtip' tooltip on the show title, that lists the show's scene names.
If we want that, we should convert that to Vue **in a separate PR**.

OmgImAlexis:
@sharkykh yeah... if Ava wouldn't break when you try and change routes that'd be great.

That's why I opted for props.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -10,64 +10,47 @@
...
-            ##There is a special/season_0?##
-            % if int(season_results[-1]["season"]) == 0:
-                <% season_special = 1 %>
-            % else:
-                <% season_special = 0 %>
-            % endif
-            % if not app.DISPLAY_SHOW_SPECIALS and season_special:
-                <% lastSeason = season_results.pop(-1) %>
-            % endif
+        <template v-if="$route.name !== 'snatchSelection' && show.seasons && show.seasons.length >= 1">
```
sharkykh: Unless I'm reading this wrong,
you can do without the `<template>` wrapper, and just put the `v-if` on the next element?

themes-default/slim/views/partials/showheader.mako
```diff
@@ -10,64 +10,47 @@
...
-                                <app-link href="#season-${seasonNum["season"]}">Specials</app-link>
-                            % else:
-                                <app-link href="#season-${seasonNum["season"]}">${str(seasonNum["season"])}</app-link>
-                            % endif
-                            % if seasonNum != season_results[-1]:
-                                <span class="separator">|</span>
-                            % endif
-                        % endfor
-                    % endif
+                        <template v-if="show.seasons && show.seasons.length >= 13">
```
sharkykh: You already checked for `show.seasons` above, so you know it's not `undefined` / `null`.
Plus, same note as above, no need for the `<template>` wrapper, put it on the `<select>` element,
**or better yet**, the `<span>` element it's currently in.
As long as you don't use `<element v-for="..." v-if="...">` you should be good to go.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -10,64 +10,47 @@
...
+                            <select @change="jumpToSeason($event)" id="seasonJump" class="form-control input-sm" style="position: relative">
+                                <option value="jump">Jump to Season</option>
+                                <option v-for="season in show.seasons" :value="'#season-' + season.season" :data-season="season.season">
+                                    Season {{ season.season === 0 ? 'Specials' : season.season }}
+                                </option>
+                            </select>
+                        </template>
+                        <template v-else-if="show.seasons && show.seasons.length >= 1">
```
sharkykh: You already checked for `show.seasons` above, so you know it's not `undefined` / `null`.
If you use my previous suggestion to use `<span>`, then you can also drop the `<template>` wrapper, and just put the `if` on the `<span>` element.

Even if you have to add another empty `<span v-else></span>` element, I still feel like it's much more readable.

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: Couldn't remember where `namingForceFolders` was.

sharkykh: I generally just do a search for `app.VAR_NAME` in the codebase and look for `medusa/server/api/v2/config.py` or open up that file and look for the variable.

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: > If we want that, we should convert that to Vue in a separate PR.
Sounds good.

themes-default/slim/src/components/display-show.vue
sharkykh: I suggest:
```js
document.querySelectorAll('.collapse.toggle').forEach(element => {
    element.addEventListener('hide.bs.collapse', function() {
        // on hide
    });
    element.addEventListener('show.bs.collapse', function() {
        // on show
    });
));
```

And to test this you need to disable `Show all seasons` in General Config -> Interface.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,45 +77,46 @@
...
-                                     <img alt="[imdb]" height="16" width="16" src="images/imdb.png" style="margin-top: -1px; vertical-align:middle;"/>
-                                 </app-link>
-                 % endif
-                % if show.externals.get('trakt_id'):
-                    <app-link href="https://trakt.tv/shows/${show.externals.get('trakt_id')}" title="https://trakt.tv/shows/${show.externals.get('trakt_id')}">
+                    <span v-if="show.rating.imdb && show.rating.imdb.rating"
+                        class="imdbstars"
+                        :qtip-content="show.rating.imdb.rating + ' / 10 Stars<br> ' + show.rating.imdb.votes + ' Votes'"
+                    >
+                        <span :style="{ width: (Number(show.rating.imdb.rating) * 12) + '%' }"></span>
```
sharkykh: You changed the original value here, it should be pixels, not percentage.
See http://api.jquery.com/width/#width-value
>_**.width( value )**_
**value**
Type: String or Number
An integer representing the number of pixels, or an integer along with an optional unit of measure appended (as a string).

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,45 +77,46 @@
...
+                    </span>
+                    <template v-if="!show.rating.imdb">
```
sharkykh:
```vue
<template v-if="!show.rating.imdb">
```
is not the same as
```mako
% if not show.imdb_id:
```

I think you thought it was
```mako
% if not show.imdb_info:
```

It should probably be something like:
```vue
<template v-if="!show.id.imdb">
```

themes-default/slim/src/components/display-show.vue
```diff
@@ -158,24 +174,28 @@ export default {
...
-        $(document.body).on('click', '.seriesCheck', () => {
-            $('.epCheck:visible').each((index, element) => {
-                element.checked = true;
-            });
-            $('.seasonCheck:visible').each((index, element) => {
-                element.checked = true;
-            });
+        // Selects all visible episode checkboxes
+        document.addEventListener('click', event => {
+            if (event.target && event.target.className.includes('clearAll')) {
```
sharkykh: That class is incorrect, should be `seriesCheck`.

themes-default/slim/src/components/display-show.vue
```diff
@@ -158,24 +174,28 @@ export default {
...
+                [...document.getElementsByClassName('epCheck')].filter(isVisible).forEach(element => {
+                    element.checked = true;
+                });
+                [...document.getElementsByClassName('seasonCheck')].filter(isVisible).forEach(element => {
+                    element.checked = true;
+                });
```
sharkykh: You *could* use this instead of the two loops, but it's probably going to be replaced later..
```js
[...document.querySelectorAll('.epCheck, .seasonCheck')].filter(isVisible).forEach(element => {
    element.checked = true;
});
```

themes-default/slim/src/components/display-show.vue
```diff
@@ -158,24 +174,28 @@ export default {
...
 
         // Clears all visible episode checkboxes and the season selectors
-        $(document.body).on('click', '.clearAll', () => {
-            $('.epCheck:visible').each((index, element) => {
-                element.checked = false;
-            });
-            $('.seasonCheck:visible').each((index, element) => {
-                element.checked = false;
-            });
+        document.addEventListener('click', event => {
```
sharkykh: You can join this with the `click` event listener above...

themes-default/slim/views/partials/showheader.mako
sharkykh: I could be misreading that, but to me it looks like you changed the logic here:
```py
len(season_results) > 14
```
is not the same as
```js
show.seasons.length >= 13
```

```
14 > 14 : false
14 >= 13 : true
```

OmgImAlexis: So it should be `>= 15`. Will change.

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: Nope, it's `%`. It's for the width of the imdb stars. Otherwise it won't line up properly with the stars.

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: Btw an `||` doesn't need `()` around it in a ternary statement.

```js
((show.config.seasonFolders || config.namingForceFolders) ? 'yes' : 'no')
```

vs 

```js
(show.config.seasonFolders || config.namingForceFolders ? 'yes' : 'no')
```

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: Shall I skip this for now? I could add a `@TODO` comment.

sharkykh: Sure, go ahead with a `@TODO`.

sharkykh:
Some of the changes to `showheader.mako` may have affected `snatch-selection`'s code.
Need to make sure any updates to the showheader code **in `display-show`** are ported to `snatch-selection` as well.

themes-default/slim/src/components/display-show.vue
```diff
@@ -504,6 +517,36 @@ export default {
                     $('#' + seasonNo + '-cols').show();
                 }
             });
+        },
+        jumpToSeason(event) {
+            const { id } = event.currentTarget;
```
sharkykh: That is incorrect, it uses `#jumpToSeason` which is the `<select>` id, not the target id.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,45 +75,45 @@
...
-                % if show.externals.get('trakt_id'):
-                    <app-link href="https://trakt.tv/shows/${show.externals.get('trakt_id')}" title="https://trakt.tv/shows/${show.externals.get('trakt_id')}">
+                    <span v-if="show.rating.imdb && show.rating.imdb.rating"
+                        class="imdbstars"
+                        :qtip-content="show.rating.imdb.rating + ' / 10 Stars<br> ' + show.rating.imdb.votes + ' Votes'"
+                    >
+                        <span :style="{ width: (Number(show.rating.imdb.rating) * 12) + '%' }"></span>
+                    </span>
+                    <template v-if="!show.id.imdb">
+                        <span v-if="show.year && show.year.start">({{ show.year.start }}) - {{ show.runtime }} minutes - </span>
```
sharkykh: This would hide the runtime if the start year is missing.
Plus, `show.year` should always be available as an object. you should be able to just check for `show.year.start`.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,45 +75,45 @@
...
+                    </template>
+                    <template v-else>
+                        <template v-if="show.country_codes && show.country_codes.length >= 1 ">
```
sharkykh: `show.country_codes` should always be an array.
You can drop the `<template v-if="...">` completely IMO.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,45 +75,45 @@
...
+                            <img v-for="country in show.country_codes" src="images/blank.png" :class="['country-flag', 'flag-' + country]" width="16" height="11" style="margin-left: 3px; vertical-align:middle;" />
+                        </template>
+                        <span>
+                        % if show.imdb_info.get('year'):
```
sharkykh: I don't understand why IMDB info overrides the indexer info on the API data...
I made one change (to the show plot) - #4353, but I'm not even sure that was the correct way to fix it.
In this case, [it's merged into `show.year.start`](https://github.com/pymedusa/Medusa/blob/feature/vueify-showheader/medusa/tv/series.py#L1986)

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,45 +75,45 @@
...
+                            (${show.imdb_info['year']}) -
+                        % endif
+                            ${show.imdb_info.get('runtimes') or show.runtime} minutes
+                        </span>
+                        <app-link :href="'http://www.imdb.com/title/' + show.id.imdb" :title="'http://www.imdb.com/title/' + show.id.imdb">
```
sharkykh: Please change to `https`.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -223,16 +200,16 @@
...
-                            <tr><td class="showLegend">Anime: </td><td><img src="images/${("no16.png", "yes16.png")[bool(show.is_anime)]}" alt="${("N", "Y")[bool(show.is_anime)]}" width="16" height="16" /></td></tr>
-                            <tr><td class="showLegend">DVD Order: </td><td><img src="images/${("no16.png", "yes16.png")[bool(show.dvd_order)]}" alt="${("N", "Y")[bool(show.dvd_order)]}" width="16" height="16" /></td></tr>
-                            <tr><td class="showLegend">Scene Numbering: </td><td><img src="images/${("no16.png", "yes16.png")[bool(show.scene)]}" alt="${("N", "Y")[bool(show.scene)]}" width="16" height="16" /></td></tr>
+                            <template v-if="config.subtitles.enabled">
+                                <tr><td class="showLegend">Subtitles: </td><td><img :src="'images/' + (show.config.subtitlesEnabled ? 'yes' : 'no') + '16.png'" :alt="show.config.subtitlesEnabled ? 'Y' : 'N'" width="16" height="16" /></td></tr>
+                            </template>
```
sharkykh:
```vue
<tr v-if="config.subtitles.enabled"><td class="showLegend">Subtitles: </td><td><img :src="'images/' + (show.config.subtitlesEnabled ? 'yes' : 'no') + '16.png'" :alt="show.config.subtitlesEnabled ? 'Y' : 'N'" width="16" height="16" /></td></tr>
```

themes-default/slim/src/components/display-show.vue
```diff
@@ -265,25 +279,24 @@ export default {
...
+                $('#showseason-' + result[1]).text('Hide Episodes');
+                $('#season-' + result[1] + '-cols').removeClass('shadow');
+            });
         });
 
-        // Generate IMDB stars
-        $('.imdbstars').each((index, element) => {
-            $(element).html($('<span/>').width($(element).text() * 12));
-        });
         attachImdbTooltip(); // eslint-disable-line no-undef
```
sharkykh: Tooltip is not working. It's possible you need to put this in `reflowLayout()`.

themes-default/slim/views/partials/showheader.mako
```diff
@@ -10,64 +10,45 @@
...
-                            % endif
-                            % if seasonNum != season_results[-1]:
-                                <span class="separator">|</span>
-                            % endif
-                        % endfor
-                    % endif
-                    </span>
-                </div>
+                        <template v-for="(season, $index) in reverse(show.seasons)">
+                            <app-link :href="'#season-' + season[0].season">{{ season[0].season === 0 ? 'Specials' : season[0].season }}</app-link>
```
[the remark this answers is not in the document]
OmgImAlexis: That may be the best option.

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: `country_codes` won't be `>= 1` when we first load. Gotta keep default state in mind. I can remove the `show.country_codes` check though.

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: I'm not seeing the difference?

themes-default/slim/views/partials/showheader.mako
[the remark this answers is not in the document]
OmgImAlexis: Come to think of it if the season has no episodes would it still show in the UI? I thought most indexer(tvdb, themoviedb) needed at least a single episode.

themes-default/slim/src/components/display-show.vue
OmgImAlexis: I'm confused as this works...

themes-default/slim/src/api.js
```diff
@@ -8,7 +8,7 @@ const apiKey = document.body.getAttribute('api-key');
  */
 const apiRoute = axios.create({ // eslint-disable-line no-unused-vars
     baseURL: webRoot + '/',
-    timeout: 10000,
+    timeout: 30000,
```
OmgImAlexis: The timeouts are needed for shows with 20+ seasons.

Waiting on https://github.com/pymedusa/Medusa/pull/4970

themes-default/slim/views/partials/showheader.mako
sharkykh: Remove the `template` wrapper and put the condition on the `tr`

themes-default/slim/views/partials/showheader.mako
```diff
@@ -121,14 +121,12 @@
                  </div>
                  <div id="tags" class="pull-left col-lg-9 col-md-9 col-sm-12 col-xs-12">
                      <ul class="tags">
-                         % if show.imdb_info.get('genres'):
+                         %if show.genre:
+                            <app-link v-for="genre in dedupeGenres(show.genres)" :href="'https://trakt.tv/shows/popular/?genres=' + genre.toLowerCase().replace(' ', '-')" :title="'View other popular ' + genre + ' shows on trakt.tv.'"><li>{{ genre }}</li></app-link>
```
sharkykh: Unfortunately `show.genres` (from the API) contains both indexer genres and imdb genres (no duplicates).
Unlike `show.genre` (Directly from Python) which only contained the indexer genre...
So I don't know how well this is going to work..

themes-default/slim/views/partials/showheader.mako
sharkykh: But it doesn't matter that it won't be `>= 1`. If the array is empty, it won't have any items to loop over anyway.
So no country flags will be visible.

themes-default/slim/src/components/display-show.vue
```diff
@@ -265,26 +283,23 @@ export default {
...
+                const reg = /collapseSeason-(\d+)/g;
+                const result = reg.exec(this.id);
+                $('#showseason-' + result[1]).text('Hide Episodes');
+                $('#season-' + result[1] + '-cols').removeClass('shadow');
+            });
         });
-        attachImdbTooltip(); // eslint-disable-line no-undef
 
         // Get the season exceptions and the xem season mappings.
         getSeasonSceneExceptions();
```
sharkykh: Need to fix this function now,
this is the console error, on the first line of the function.

```js
        getSeasonSceneExceptions() {
            // this line VVV
            const indexerName = document.querySelector('#indexer-name').value;
            const seriesId = document.querySelector('#series-id').value;
```

```
[Vue warn]: Error in mounted hook: "TypeError: Cannot read property 'value' of null"

found in

---> <DisplayShow> at src/components/display-show.vue
       <Root>
warn @ webpack:///./node_modules/vue/dist/vue.esm.js?:616

TypeError: Cannot read property 'value' of null
    at VueComponent.getSeasonSceneExceptions (webpack:///./src/components/display-show.vue?./node_modules/babel-loader/lib!./node_modules/vue-loader/lib??vue-loader-options:488)
    at VueComponent.mounted (webpack:///./src/components/display-show.vue?./node_modules/babel-loader/lib!./node_modules/vue-loader/lib??vue-loader-options:326)
    at callHook (webpack:///./node_modules/vue/dist/vue.esm.js?:2959)
    at Object.insert (webpack:///./node_modules/vue/dist/vue.esm.js?:4146)
    at invokeInsertHook (webpack:///./node_modules/vue/dist/vue.esm.js?:5962)
    at Vue.patch [as __patch__] (webpack:///./node_modules/vue/dist/vue.esm.js?:6187)
    at Vue._update (webpack:///./node_modules/vue/dist/vue.esm.js?:2686)
    at Vue.updateComponent (webpack:///./node_modules/vue/dist/vue.esm.js?:2811)
    at Watcher.get (webpack:///./node_modules/vue/dist/vue.esm.js?:3182)
    at new Watcher (webpack:///./node_modules/vue/dist/vue.esm.js?:3169)
logError @ webpack:///./node_modules/vue/dist/vue.esm.js?:1776
```

themes-default/slim/src/components/display-show.vue
```diff
@@ -536,6 +536,9 @@ export default {
         },
         reverse(array) {
             return array.slice().reverse();
+        },
+        dedupeGenres(genres) {
+            return [...new Set(genres.slice(0).map(genre => genre.replace('-', ' ')))];
```
sharkykh:
```
[Vue warn]: Error in render: "TypeError: Cannot read property 'slice' of null"

found in

---> <DisplayShow> at src/components/display-show.vue
       <Root>
warn @ webpack:///./node_modules/vue/dist/vue.esm.js?:616

TypeError: Cannot read property 'slice' of null
    at VueComponent.dedupeGenres (webpack:///./src/components/display-show.vue?./node_modules/babel-loader/lib!./node_modules/vue-loader/lib??vue-loader-options:582)
    at Proxy.eval (eval at createFunction (webpack:///./node_modules/vue/dist/vue.esm.js?), <anonymous>:3:6130)
    at VueComponent.Vue._render (webpack:///./node_modules/vue/dist/vue.esm.js?:4520)
    at VueComponent.updateComponent (webpack:///./node_modules/vue/dist/vue.esm.js?:2805)
    at Watcher.get (webpack:///./node_modules/vue/dist/vue.esm.js?:3182)
    at new Watcher (webpack:///./node_modules/vue/dist/vue.esm.js?:3169)
    at mountComponent (webpack:///./node_modules/vue/dist/vue.esm.js?:2825)
    at VueComponent.Vue.$mount (webpack:///./node_modules/vue/dist/vue.esm.js?:8518)
    at VueComponent.Vue.$mount (webpack:///./node_modules/vue/dist/vue.esm.js?:10870)
    at init (webpack:///./node_modules/vue/dist/vue.esm.js?:4128)
logError @ webpack:///./node_modules/vue/dist/vue.esm.js?:1776
```

themes-default/slim/views/partials/showheader.mako
sharkykh:
```
[Vue tip]: <app-link v-for="genre in dedupeGenres(show.genres)">: component lists rendered with v-for should have explicit keys. See https://vuejs.org/guide/list.html#key for more info.

found in

---> <DisplayShow> at src/components/display-show.vue
       <Root>
tip @ webpack:///./node_modules/vue/dist/vue.esm.js?:622
```

themes-default/slim/views/partials/showheader.mako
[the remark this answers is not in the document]
sharkykh: I can't answer that with certainty...

themes-default/slim/views/partials/showheader.mako
sharkykh: I don't know the answer to that. I wasn't here when the decision was made to join them. I don't know why it was done in the first place...

themes-default/slim/views/partials/showheader.mako
```diff
@@ -94,55 +79,53 @@
...
+                        class="imdbstars"
+                        :qtip-content="show.rating.imdb.rating + ' / 10 Stars<br> ' + show.rating.imdb.votes + ' Votes'"
+                    >
+                        <span :style="{ width: (Number(show.rating.imdb.rating) * 12) + '%' }"></span>
+                    </span>
+                    <template v-if="!show.id.imdb">
+                        <span v-if="show.year.start">({{ show.year.start }}) - {{ show.runtime }} minutes - </span>
+                    </template>
+                    <template v-else>
+                        <template v-if="show.country_codes && show.country_codes.length >= 1 ">
```
sharkykh: >`country_codes` won't be `>= 1` when we first load. Gotta keep default state in mind. I can remove the `show.country_codes` check though. 

But it doesn't matter that it won't be `>= 1`. If the array is empty, it won't have any items to loop over anyway.
So no country flags will be visible.

sharkykh:
@OmgImAlexis 
Is there any specific reason you didn't start this branch by converting the `show-header` to a vue component? (`x-template` like `quality-chooser`)?

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: Did you want to check git blame? Depending on who added it we may be able to work out the reason.

For example if it was me then who the hell knows why I did that.

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: I haven't need any that don't work with the trakt link and the `dedupeGenres` function ensures it removes ones that are the same but have a hyphen instead of space.

OmgImAlexis:
@sharkykh yes I knew I wouldn't be able to completely do the conversion for now as it uses quality pill, etc. so I wanted to just move what I could.

themes-default/slim/views/partials/showheader.mako
OmgImAlexis: Okay yeah let's split them, if you're okay with that? I think the current indexer's data should really be used there as we can fall back to either null or IMDB in the frontend. I don't think we should hardcode it in the API.

themes-default/slim/views/partials/showheader.mako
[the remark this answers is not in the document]
OmgImAlexis: Shall we leave this for now or ?

themes-default/slim/views/partials/showheader.mako
p0psicles: I think we should take this one step further. Lets add a Structure where each attribute is prepended with the indexer id.

That way in future users can choose which indexer they want to use for, for example the genre, summary or poster.

p0psicles: But it should be additional if we choose for it, so we dont risk refactoring for now?

themes-default/slim/src/api.js
sharkykh: That PR won't change anything about the time it takes to generate the data... Just saying.

themes-default/slim/views/partials/showheader.mako
sharkykh: I said "(no duplicates)", meaning the backend already dedupes the list.
If it works, that's okay. It's not really a priority.

themes-default/slim/src/api.js
OmgImAlexis: It seems to be directly related to multiple requests. If I move the request for the detailed show to a domain(bypassing the x request per domain/ip limit) I don't see this issue.

OmgImAlexis:
@sharkykh I ran `git merge origin/develop` and then it said I had to update the vendors so I re-ran `yarn dev`. Hope that's right.

sharkykh:
Yes that is correct.
You had to do that because I edited `snatch-selection.vue` on #5107, and it had a conflict.

`yarn install ; yarn dev` after a merge or rebase I always my go-to action.

themes-default/slim/src/components/display-show.vue
```diff
@@ -447,17 +447,17 @@ export default {
...
                 console.warn('Unable to get season scene exceptions: Unknown series identifier');
                 return;
             }
-            const params = {
-                indexername: indexerName,
-                seriesid: seriesId
-            };
-            apiRoute.get('home/getSeasonSceneExceptions', { params }).then(response => {
+
+            apiRoute.get('home/getSeasonSceneExceptions', {
```
sharkykh: You must put the parameters in an object like `{ params: { a: 1, b: 2 } }` (or like it was before `{ params }`), because the second argument is `config`.
[`axios.post`, `axios.put` and `axios.patch` support what you're after, but the other methods don't](https://github.com/axios/axios/tree/v0.18.0#request-method-aliases).

themes-default/slim/views/partials/showheader.mako
```diff
@@ -24,7 +24,10 @@
         <div v-if="$route.name === 'snatchSelection'" id="show-specials-and-seasons" class="pull-right">
             <span class="h2footer display-specials">
                 Manual search for:<br>
-                <app-link href="home/displayShow?indexername=${show.indexer_name}&seriesid=${show.series_id}" class="snatchTitle">{{ show.title }}</app-link> / Season {{ season }}<template v-if="episode"> Episode {{ episode }}</template>
+                <app-link
+                    href="home/displayShow?indexername=${show.indexer_name}&seriesid=${show.series_id}"
```
sharkykh: The `href` is still using Python values, change to:
`:href="'home/displayShow?indexername=' + show.indexer + '&seriesid=' + show.id[show.indexer]"`

themes-default/slim/views/partials/showheader.mako
```diff
@@ -24,7 +24,10 @@
...
+                    class="snatchTitle"
+                    >{{ show.title }}</app-link> / Season {{ season }}<template v-if="episode && !$route.query.manual_search_type"> Episode {{ episode }}</template>
```
sharkykh: Probably should move `$route.query.manual_search_type` into a computed property, right?

themes-default/slim/views/partials/showheader.mako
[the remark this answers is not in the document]
OmgImAlexis: @sharkykh did you want to open an issue to track this so I can mark as resolved here?

OmgImAlexis:
@sharkykh would you mind a last review and then merge?

OmgImAlexis:
Should we add `show_message` to `show.message` in the show python dict?

sharkykh:
Still a small bug. I think I may have created it.
```
2018-09-08 16:45:00 WARNING  TORNADO :: [6354111] 404 GET /images/null (127.0.0.1) 8.00ms
```
Looking into it now.

> Should we add `show_message` to `show.message` in the show python dict?

:+1:

OmgImAlexis:
That may be from the new components I added. Did you want to quickly fix that?

sharkykh:
@OmgImAlexis No :) You're confusing the two PRs you have.
Already found and fixed it, it was my indexer icon code. Forgot a condition.
Will push in a minute..

PR #5095, opened 2018-08-31
Move date-fns to Webpack and use selective import

sharkykh:


sharkykh:
@OmgImAlexis If you could re-approve that, just rebased onto develop because of a small conflict.

PR #1881, opened 2017-01-02
Fix #1843: Medusa does not start with Python 2.7.13

labrys:
http://bugs.python.org/issue29082

- [ ] PR is based on the DEVELOP branch
- [ ] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [ ] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

PR #5096, opened 2018-08-31
Update lib unrar2 (python2 and python3 supported)

ngosang:
Current version of unrar2 lib has several problems:
* Don't support python3
* Non working in my linux (python2). `2018-08-31 11:49:38 WARNING  POSTPROCESSOR :: [f5973b2] Failed unpacking archive XXXXX.rar: error b'Unsupported RAR version, expected 4.x or 5.x, found: archive'`
* Not updated since 2008

I used master version which fix all issues => https://github.com/kyegupov/py-unrar2

Tested in windows-python2, windows-python3, linux-python2 and linux-python3 I did 2 test for each platform:
* Download a rar episode and enable postprocessor rar uncompress.
* This test => https://github.com/pymedusa/Medusa/blob/master/medusa/server/web/config/post_processing.py#L223


lib/unrar2/windows.py
```diff
@@ -62,145 +80,160 @@
...
-volume_naming3 = re.compile("\.part([0-9]+).rar$")
+volume_naming1 = re.compile("[.]r([0-9]{2})$")
+volume_naming2 = re.compile("[.]([0-9]{3})[.]rar$")
+volume_naming3 = re.compile("[.]part([0-9]+)[.]rar$")
 
 try:
-    dll_filename = os.path.join(os.path.split(__file__)[0], 'UnRARDLL', dll_name)
-
-    if sys.version_info[:3] == (2, 7, 13):
-        # http://bugs.python.org/issue29082
```
sharkykh: Pretty sure we still need these fixes.
https://github.com/pymedusa/Medusa/pull/1881

The fact that this library is modified was listed on [the readme file](https://github.com/pymedusa/Medusa/blob/develop/lib/readme.md), last entry.

ngosang: @sharkykh fixed

sharkykh:
**FYI**
1. When installed through `pip`, the `UnRARDLL` folder doesn't get included.
2. The **module** name (as in, how you import it) was changed from `unrar2` to `UnRAR2`.

ngosang:
@sharkykh I already know that, thank you.

I didn't rename the package because it can cause weird troubles. Remember that when you update pymedusa, the package source code moves to the new folder, but the old python bytecode files remains in the old folder. Maybe the updater should remove that files...

About the test.rar. https://github.com/pymedusa/Medusa/blob/master/medusa/server/web/config/post_processing.py#L223 This class has a lot of validations for when you save the post processing configuration. One of them is test if there is RAR support in the system. I think that class is not used anymore. When you vueified the UI you removed the validations. Anyway I keeped the file.

sharkykh:
That's okay, I wrote the comment just so everyone else who sees it knows about it.

About the bytecode, the updater uses git clean on the ext and lib folders.

PR #5103, opened 2018-09-02
Vueify: Quality Pill

sharkykh:
- [x] Convert the `inc_defs.mako` function `renderQualityPill` into a Vue component
- [x] Replace "simple" `renderQualityPill` calls with component
- [x] Replace "complex" `renderQualityPill` calls with component
- [x] Remove all quality pill styles from global
	Still used outside of component in:
	- [x] [`./static/js/ajax-episode-search.js`](https://github.com/pymedusa/Medusa/blob/ebba2e69f2874a4b53205f337f13e632dba6fe46/themes-default/slim/static/js/ajax-episode-search.js#L56-L60)
	- [x] [`./views/history.mako`](https://github.com/pymedusa/Medusa/blob/ebba2e69f2874a4b53205f337f13e632dba6fe46/themes-default/slim/views/history.mako#L160)
	- [x] ~Possibly in other places as well...~
- [x] Remove `inc_defs.mako`
- [x] Add [extensive] tests for the `quality-pill` component
- [x] Update changelog


themes-default/slim/src/components/quality-pill.vue
```diff
@@ -0,0 +1,292 @@
...
+         */
+        isSubsetOf(set1, set2) {
+            return set1.every(value => set2.includes(value));
+        }
+    }
+};
+</script>
+
+<style scoped>
+span.quality {
```
OmgImAlexis: Since you're using scoped and only a single element you can drop the `span` for all the styles.

themes-default/slim/static/js/ajax-episode-search.js
```diff
@@ -55,7 +55,9 @@ function updateImages(data) {
                 // Update Status and Quality
                 let qualityPill = '';
                 if (ep.quality_style !== 'na') {
-                    qualityPill = ' <span class="quality ' + ep.quality_style + '">' + ep.quality_name + '</span>';
+                    // @FIXME: (sharkykh) This is a hack to get the scoped style to work.
+                    const qualityPillScopeId = window.components.find(c => c.name === 'quality-pill')._scopeId;
+                    qualityPill = ' <span ' + qualityPillScopeId + ' class="quality ' + ep.quality_style + '">' + ep.quality_name + '</span>';
```
sharkykh: It's a hack, but it can be removed once this file gets implemented in Vue.
I think implementing an `EpisodeRow` component would be best suited for this, BTW.

OmgImAlexis: I could see the `EpisodeRow` component being used a few places.

themes-default/slim/src/components/quality-pill.vue
```diff
@@ -0,0 +1,288 @@
...
+
+<script>
+import { mapState } from 'vuex';
+
+/**
+ * @typedef {Object} Quality
+ * @property {number[]} allowed - Allowed qualities
+ * @property {number[]} preferred - Preferred qualities
+ */
+
```
OmgImAlexis: Remove the extra line between the jsdoc and the export.

themes-default/slim/src/components/quality-pill.vue
```diff
@@ -0,0 +1,288 @@
+<template>
+    <span :title="title" :class="pill.class">{{ pill.text }}</span>
+</template>
+
+<script>
+import { mapState } from 'vuex';
+
+/**
```
OmgImAlexis: Change from this
```
/**
 * @typedef {Object} Quality
```

to this

```
/**
 * Quality pill
 *
 * @typedef {Object} Quality
```

sharkykh: It doesn't represent the quality pill, it represents a split quality:
`{ allowed: [ 1, 2, 4 ], preferred: [ 128, 256 ] }`

sharkykh: How's that?
```js
/**
 * An object representing a split quality.
 * @typedef {Object} Quality
 * @property {number[]} allowed - Allowed qualities
 * @property {number[]} preferred - Preferred qualities
 */
```

sharkykh:
1. I need to add more tests to the component.
	I want to have a test for most if not all quality combinations to test every possible result.
2. If possible I want to try and make the `pill` computed property less complex...

themes-default/slim/test/specs/quality-pill.spec.js
```diff
@@ -16,72 +16,101 @@ test.beforeEach(t => {
...
         store,
-        propsData: {
-            quality: 128
-        }
+        propsData
+    }).html(), message);
+};
+
+test('renders quality pills correctly', t => {
+    pillTestCase(t, 'No qualities', {
```
OmgImAlexis: This would come in handy in quite a few places. Why don't we move this to a helper and make it a higher level function?

```js
const generatePropTest = component => (t, message, propsData) => {
    const { localVue, store } = t.context;

    t.snapshot(mount(component, {
        localVue,
        store,
        propsData
    }).html(), message);
};
const pillTestCase = generatePropTest(QualityPill);
```

sharkykh: That's a good point.

1. I'm not so sure how it should be structured tree wise though, I mean, in `test/helpers/[what here?]`.
2. I can't find a way of documenting the inner function using JSDoc without assigning it to an inner variable.

This seemed to work best:
```js
/**
 * Generate a simple component prop test case.
 *
 * @param {*} component - A Vue component.
 * @returns {propTest} Test case function.
 */
const generatePropTest = component => {
    /**
     * A simple component prop test case.
     *
     * @param {*} t - AVA test with a context containing `localVue` and `store` properties.
     * @param {string} message - Test message.
     * @param {object} propsData - Props to pass to the component.
     */
    const propTest = (t, message, propsData) => {
        const { localVue, store } = t.context;

        t.snapshot(mount(component, {
            localVue,
            store,
            propsData
        }).html(), message);
    };
    return propTest;
};
```

OmgImAlexis: Aren't the tests failing early as we have the `failFast` flag enabled?

OmgImAlexis: `propTest` may be better outside of that function so the js-doc resolution works properly.

sharkykh: > `propTest` may be better outside of that function so the js-doc resolution works properly.

If I put `propTest` outside the function I can't reference `component`. AFAIK it must be an inner function to work.

> Aren't the tests failing early as we have the `failFast` flag enabled?

I'll check that.

> 1. I'm not so sure how it should be structured tree wise though, I mean, in `test/helpers/[what here?]`.

Can you please answer that? I don't know where to put the function. :confused:

sharkykh: `failFast` flag doesn't affect this scenario. If one assertion fails inside a `test`, it fails the whole test.
I'm not saying that's wrong, I'm just saying we may want to have it be multiple `test`s, instead of one `test` with multiple assertions.

themes-default/slim/test/specs/snapshots/quality-pill.spec.js.md
```diff
@@ -0,0 +1,149 @@
...
+      1080p BluRay" class="quality HD1080p">1080p</span>`
+
+> Both quality lists are of 4K UHD resolution, with show title
+
+    `<span title="Allowed Quality:␊
+      4K UHD TV␊
+      4K UHD WEB-DL␊
+    ␊
+    Preferred Quality:␊
+      4K UHD BluRay" class="quality HD720p">4K-UHD</span>`
```
sharkykh: Uhhh that's incorrect. The class should be `quality UHD-4K`.

themes-default/slim/test/specs/snapshots/quality-pill.spec.js.md
```diff
@@ -0,0 +1,149 @@
...
+
+> Both quality lists are of 8K UHD resolution, with show title
+
+    `<span title="Allowed Quality:␊
+      8K UHD TV␊
+      8K UHD WEB-DL␊
+    ␊
+    Preferred Quality:␊
+      8K UHD BluRay" class="quality HD720p">8K-UHD</span>`
```
sharkykh: And this one should be `quality UHD-8K`.

themes-default/slim/test/specs/quality-pill.spec.js
OmgImAlexis: > If I put propTest outside the function I can't reference component. AFAIK it must be an inner function to work.

@sharkykh change `const propTest =` to `return`. That'll be the same as what I was trying to achieve.

> I'm not saying that's wrong, I'm just saying we may want to have it be multiple tests, instead of one test with multiple assertions.

We could generate the test as well with that function, these are called [test macros](https://github.com/avajs/ava#test-macros).

sharkykh: jsdoc doesn't recognize the return arrow function as a function. that's why I used a variable.
I said it in my first reply:
>I can't find a way of documenting the inner function using JSDoc without assigning it to an inner variable.

It just didn't acknowledge the fact the wrapper function returns a function with 3 arguments.

OmgImAlexis: Curried functions do work with jsdoc, see [this](https://github.com/jsdoc3/jsdoc/issues/1286#issuecomment-341911947).

sharkykh: It's not perfect, as you can't document the parameters of the inner function:
image
:disappointed:

OmgImAlexis: So we'll go with `#2` then?

sharkykh: yeah I guess...

and `#2` is the current code

PR #5117, opened 2018-09-05
Misc changes

sharkykh:
- [x] Remove unused images - some images were as old as the initial SickRageTV commit...
- [x] Remove `fix-broken-ie.js` shim - [`endsWith` is supported in all browsers except IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#Browser_compatibility)
- [x] Make `apiBuilder.mako` more like `main.mako` (Review in `No Whitespace` mode)
		`apiBuilder.mako` should be using `layouts/main.mako`, and in the future, it might (low priority).
- [x] Remove unused `browserconfig.xml`
- [x] Remove unused `sbThemeName` variable
- [x] Update `restart.mako` to remove `sbThemeName` `try...except` block.
- [x] Update `webpack.config.js` - This should fix issues with other scripts that try to read `webpack.config.js`.
- [x] Remove duplicate `<meta name="theme-color">` tags


themes-default/slim/views/apiBuilder.mako
```diff
@@ -4,156 +4,149 @@
...
-    <link rel="apple-touch-icon" sizes="76x76" href="images/ico/favicon-76.png">
-    <link rel="apple-touch-icon" sizes="72x72" href="images/ico/favicon-72.png">
-    <link rel="apple-touch-icon" href="images/ico/favicon-57.png">
+    <head>
+        <meta charset="utf-8">
+        <meta name="robots" content="noindex, nofollow">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <!-- These values come from css/dark.css and css/light.css -->
+        % if app.THEME_NAME == "dark":
```
OmgImAlexis: Since these use the same content is there much point in keeping this?

sharkykh: Good point.
I don't think that meta var is even used, so I'll remove those from both files.
Plus `dark.css` / `light.css` are not really a thing anymore (they are now `themed.css` for each theme respectively)

OmgImAlexis: I wonder if there are other parts of the CSS we could now look at removing since the themes are "separate".

themes-default/slim/views/apiBuilder.mako
```diff
@@ -9,12 +9,6 @@
         <meta name="robots" content="noindex, nofollow">
         <meta http-equiv="X-UA-Compatible" content="IE=edge">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <!-- These values come from css/dark.css and css/light.css -->
-        % if app.THEME_NAME == "dark":
-        <meta name="theme-color" content="#333333">
```
OmgImAlexis: I figured you were leaving one copy of this as I think this is still used for mobile apps when you pin the site to your homescreen.

sharkykh: Oh I thought it wasn't used at all, didn't even occur to me to search Google for this meta tag.
I'll add one back.

PR #5127, opened 2018-09-06
Fix UI bugs

sharkykh:
- [x] Show list bug when using tabs (after adding show) 
	jQuery UI tabs widget was fetching the whole page using AJAX.
- [x] Pre-emptive fix for recommended shows (for the same tabs bug)
- [x] Fix status page with 3 or more root dirs is malformed:
chrome_2018-09-06_22-08-36

We should wait a little more to see if more bugs pop up, if they do we should probably cherry pick fixes into a hotfix release.


sharkykh:
Note to self: Add Changelog for #5126

PR #5132, opened 2018-09-07
Fix watched null values.

p0psicles:

@sharkykh still not sure if this PR is needed at all. But throwing it out here, just in case.


medusa/databases/main_db.py
```diff
@@ -221,6 +222,10 @@ def fix_subtitles_codes(self):
     def fix_show_nfo_lang(self):
         self.connection.action("UPDATE tv_shows SET lang = '' WHERE lang = 0 OR lang = '0';")
 
+    def fix_tv_episodes_watched_field(self):
+        """A mistake was made when we added this field. Migration didn't went very well."""
```
OmgImAlexis: Might want to swap "went" with "go".

sharkykh:
Where have you seen `null` values?

sharkykh:
@p0psicles 
IMO we really have to put a hold on migrations until the migration issues are solved...
I still not sure about this fix too. I mean it should fix it...

p0psicles:
For new users yes. Users that are allready on db 4.12 not.

sharkykh:
@p0psicles 
Why? I think sanity checks run every time Medusa starts up?

sharkykh:
I don't know why I didn't think of this before.
`Episode.watched` should be boolean anyway, so we should be using `bool` on the value.

p0psicles:
I also cant merge it

p0psicles:
Yeah the bool does the trick

PR #4982, opened 2018-08-27
[Python3] Several changes to migrate the code to python3 (part 3)

ngosang:
**This PR contains the commits from #4985 and #4980.  After other PRs are merged this PR will be rebased.**
**This PR is the 3rd in [Python3] series and the first one with python3 support.**
Main discussion: #4975

**You wanna help?**
* Test with python2. Everything should work as usual.
* Test with python3. Main features are working. Report all problems you found (trace + python version)
* Request other people to test!

**Instructions to test**
* `git clone https://github.com/ngosang/Medusa.git`
* `cd Medusa`
* `git pull` and `git checkout feature/py3_2`
* edit `config.ini` and change `debug = 1`
* `python -V` or `python2 -V` or `python3 -V`
* `python start.py`  or `python2 start.py` or `python3 start.py`
* Try to enable/configure all features in the webui

**Known bugs**
* python2/python3 unicode in sqlite3 => #5139

Disclaimer:
> I'm testing careful with both python2 and python3 interpreters. With python2 you should expect the same functionally than development branch. Python3 is a long term goal, first versions won't be functional or even won't run.

PR #5136, opened 2018-09-07
CI: Test on Python 3.6

sharkykh:
Runs pytest tests on Python 3.6.
Dredd tests disabled for now, because they take too long (13 minutes ... to fail).
Re enable them when Medusa can *actually* start on Python 3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ before_install:
 install: false
 script: false
 jobs:
+  fast_finish: true
+  allow_failures:
+    - python: '3.6'
   include:
     # test stage + frontend tests start here
     - stage: test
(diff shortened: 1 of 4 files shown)


ngosang:
Nice. I run the tests in local for branch #4982 and:
* dredd is working fine
* backend requieres several changes in python test code. I will open a PR soon.

labrys:
@ngosang You should really test python 3.7 due to the changes in ssl.

OmgImAlexis:
@ngosang we'll likely be shipping `py@3.7` or `py@latest` in our Docker container so please make sure it's working for that.

sharkykh:
So should I change this to test on 3.7 instead of 3.6?

p0psicles:
I would go for 3.7 as thats the latest release

p0psicles:
Hmm cant find the 3.7 python package

.travis.yml
```diff
@@ -43,9 +43,9 @@ jobs:
         - pip install --upgrade tox
       script:
         - tox -v --recreate
-    # backend tests (py3.6) start here
-    - name: 'Backend tests (py3.6)'
-      python: '3.6'
+    # backend tests (py3.7) start here
+    - name: 'Backend tests (py3.7)'
+      python: '3.7'
```
OmgImAlexis: This should be `3.7-dev`.

OmgImAlexis: https://docs.travis-ci.com/user/languages/python/

OmgImAlexis: > Recent Python branches require OpenSSL 1.0.2+. As this library is not available for Trusty, 3.7, 3.7-dev, 3.8-dev, and nightly do not work (or use outdated archive).

sharkykh: https://github.com/travis-ci/travis-ci/issues/9831
```
language: python
python: 3.7
dist: xenial
sudo: true
```

`sudo: true` though... It's gonna be slow I think.

OmgImAlexis: Stick with `3.6` till full release then?

sharkykh: I think so... We're waiting on Travis to implement something and it could take a while... /jk

p0psicles:
Yeah for now 3.6 is also fine

ngosang:
There are minor differences between 3.6 and 3.7. We should support both since most used linux distributions ship 3.6 and don't plan to update to 3.7 soon.

OmgImAlexis:
@ngosang most of our users use Docker so the OS's version doesn't really matter all that much.

dredd/dredd_hook.py
```diff
@@ -89,7 +93,7 @@ def configure_transaction(transaction):
         new_url = path
         for name, value in params.items():
             value = evaluate(value)
-            new_url = new_url.replace('{' + name + '}', str(value))
+            new_url = new_url.replace('{' + name + '}', binary_type(value))
```
ngosang: @sharkykh replace `binary_type(value)` with `str(value)`. This is the reason because dreed is taking 13 minutes to fail. Please, enable dreed for py3.

ngosang:
@sharkykh I tested this PR with the changes in #4982. Still doesn't pass but looks good.

p0psicles:
@ngosang can you keep an eye out for slack? For when we have questions?

dredd/dredd_hook.py
sharkykh: Done, however the reason it takes 13 minutes to fail is because Medusa cannot be loaded using Python 3.6 yet, so the API tests try to run but the backend server (Medusa) isn't available, so each test times out, which takes time.

PR #5165, opened 2018-09-08
Post Processing config saving `select-list` values incorrectly

### LINKED ISSUE #5155: [APP SUBMITTED]: AttributeError: 'dict' object has no attribute 'replace'
so my Auto Post Processing is not working and when I try to do Manual PP, this happens:

### INFO
**Python Version**: `2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]`
**Operating System**: `Windows-7-6.1.7601-SP1`
**Locale**: `cp1252`
**Database**: `44.12`
**Commit**: pymedusa/Medusa@db3d9d767bbf3fe7c365df20d92099b0cbbd91b4
**Link to Log**: https://gist.github.com/9676ce5b08d10ad0969db89b2764bc06
### ERROR
<pre>
2018-09-08 01:07:49 ERROR    Thread_7 :: [db3d9d7] Exception generated: 'dict' object has no attribute 'replace'
  File "C:\Medusa\Medusa\<a href="../blob/db3d9d767bbf3fe7c365df20d92099b0cbbd91b4/medusa/server/web/core/base.py#L259">medusa\server\web\core\base.py</a>", line 259, in async_call
    result = function(**kwargs)
  File "C:\Medusa\Medusa\<a href="../blob/db3d9d767bbf3fe7c365df20d92099b0cbbd91b4/medusa/server/web/home/post_process.py#L49">medusa\server\web\home\post_process.py</a>", line 49, in processEpisode
    ignore_subs=argToBool(ignore_subs)
  File "C:\Medusa\Medusa\<a href="../blob/db3d9d767bbf3fe7c365df20d92099b0cbbd91b4/medusa/process_tv.py#L138">medusa\process_tv.py</a>", line 138, in process
    postpone = app.POSTPONE_IF_SYNC_FILES and any(sync_files)
  File "C:\Medusa\Medusa\<a href="../blob/db3d9d767bbf3fe7c365df20d92099b0cbbd91b4/medusa/process_tv.py#L135">medusa\process_tv.py</a>", line 135, in <genexpr>
    if is_sync_file(filename))
  File "C:\Medusa\Medusa\<a href="../blob/db3d9d767bbf3fe7c365df20d92099b0cbbd91b4/medusa/helper/common.py#L162">medusa\helper\common.py</a>", line 162, in is_sync_file
    any(fnmatch(filename, match) for match in app.SYNC_FILES))
  File "C:\Medusa\Medusa\<a href="../blob/db3d9d767bbf3fe7c365df20d92099b0cbbd91b4/medusa/helper/common.py#L162">medusa\helper\common.py</a>", line 162, in <genexpr>
    any(fnmatch(filename, match) for match in app.SYNC_FILES))
  File "C:\Medusa\Python\lib\fnmatch.py", line 42, in fnmatch
    pat = os.path.normcase(pat)
  File "C:\Medusa\Medusa\<a href="../blob/db3d9d767bbf3fe7c365df20d92099b0cbbd91b4/medusa/init/filesystem.py#L113">medusa\init\filesystem.py</a>", line 113, in patched_output
    return handle_output(f(*args, **kwargs))
  File "C:\Medusa\Python\lib\ntpath.py", line 47, in normcase
    return s.replace("/", "\\").lower()
AttributeError: 'dict' object has no attribute 'replace'
</pre>
---

sharkykh:
Fixes #5155 

The tests for `select-list` are how they should be, however the components needs to be changed to allow for them to pass, and it cannot be changed without changing the usages of it, which will cause an infinite loop at the moment...


sharkykh:
Alright, made the changes to do this in the backend. This is actually better since `ast.literal_eval` is safe, and it will automatically fix the issue without having to ask the users to saving their config again.

medusa/__main__.py
```diff
@@ -95,6 +95,34 @@
 logger = logging.getLogger(__name__)
 
 
+def fix_incorrect_list_values(data):
```
p0psicles: I would put a method in medusa/config.py. 
And just fix it for the suspected fields there.
Then call the method from main somewhere here:
https://github.com/pymedusa/Medusa/blob/release/hotfix-0.2.10/medusa/__main__.py#L1119

That way we don't clutter __main__.py more then necessary?

sharkykh: I thought since we are removing this in a future version it would be best to put it all in one place?

p0psicles: It's technically not a config migration. But it's easy to reuse the self.config_obj property. So maybe just add a function there named fix_config() ?

p0psicles: Yeah it's really good either way

PR #5166, opened 2018-09-08
Update `/series` API to only get AniDB data if requested

sharkykh:
- [x] Add `fetch` argument to `/series` API endpoint
  Only fetch AniDB info when it is set to `True` (default is `False`)
- [x] Update `editShow` to get AniDB data
- [x] Update store/shows actions
- [x] Update changelog

This will fix the calls to AniDB (for every anime show in the library) when navigating to `displayShow` for a show of any type (normal or anime).

Closes https://github.com/pymedusa/Medusa/issues/5170


p0psicles:
I dont see this as the most desirable solution. I think we should eventually move the anidb fetch logic to a separate endpoint. Meaning we can always load the page fast. But for now its good enough

sharkykh:
>I think we should eventually move the anidb fetch logic to a separate endpoint.

Haha, you mean like `home/fetch_releasegroups`? :wink:

p0psicles:
Haha. Yeah but preferebly restfull

OmgImAlexis:
@p0psicles you seem to put RPC calls and REST into the same basket. Fetching would be an action. REST endpoint can have actions that run in the background and return data over another transport(websocket) or on the next `GET`.

If you want to just add RPC calls then do that but please don't mix them.

p0psicles:
I dont mix them. I said i prefer a restfull solution. Didnt say we should use 'fetch' restfull

OmgImAlexis:
I know that but you're talking about moving an ACTION not a RESOURCE to an endpoint.

OmgImAlexis:
Oh okay so the fetch + cache results would be there not just the fetching?

PR #5178, opened 2018-09-09
[WIP] Change Series JSON data structure

### LINKED ISSUE #5121: API: Change `show.seasons` structure
Related to [this discussion](https://github.com/pymedusa/Medusa/pull/5087#discussion_r215536755).

### Option 1:
An object where the keys are the season numbers (as strings because JSON doesn't support key indices)
```js
"show": {
    "seasons": {
        "0": [
            { episode1 (object) },
            // ...
            { episodeN (object) }
        ],
        "1": {
            { episodeN (object) }
        },
        // ...
    }
}
```

### Option 2:
An array of objects
```js
"show": {
    "seasons": [
        {
            "season": 0,
            "episodes": [
                { episode1 (object) },
                // ...
                { episodeN (object) }
            ]
        },
        { season1 (object) },
        // ...
    ]
}
```

sharkykh:
- [ ] Decouple shows' indexer and IMDB data - as per [this discussion](https://user-images.githubusercontent.com/10238474/45270314-dbcbc180-b4a4-11e8-950f-4b1c7606db88.png) from #5087 
- [ ] Fix #5121 - `show.seasons` structure
- [ ] ?

### DIFF
--- a/dredd/api-description.yml
+++ b/dredd/api-description.yml
@@ -953,24 +953,25 @@ definitions:
         items:
           type: string
         example: [Drama, Romance]
-      rating:
+      imdbInfo:
         type: object
         properties:
-          imdb:
-            type: object
-            properties:
-              stars:
-                type: string
-                description: "IMDB's star rating from 0 to 10"
-              votes:
-                type: integer
-                format: int32
-                minimum: 1
-                description: "Total number of votes"
+          runtime:
+            type: integer
+            minimum: 1
+            format: int32
+            description: Episodes runtime in minutes
+          stars:
+            type: string
+            description: "IMDB's star rating from 0 to 10"
+          votes:
+            type: integer
+            format: int32
+            minimum: 1
+            description: "Total number of votes"
         example:
-          imdb:
-            stars: '8.3'
-            votes: 558507
+          stars: '8.3'
+          votes: 558507
       classification:
         type: string
         enum: ['', 'TV-Y', 'TV-Y7', 'TV-G', 'TV-PG', 'TV-14', 'TV-MA']
@@ -989,7 +990,7 @@ definitions:
         type: array
         items:
           type: string
-          description: Country codes
+          description: Country names
       config:
         type: object
         description: Series configuration and preferences
--- a/themes-default/slim/src/store/modules/defaults.js
+++ b/themes-default/slim/src/store/modules/defaults.js
@@ -1,7 +1,7 @@
 const state = {
     show: {
         airs: null,
-        akas: null,
+        akas: {},
         cache: null,
         classification: null,
         config: {
@@ -19,11 +19,14 @@ const state = {
             sports: null,
             subtitlesEnabled: null
         },
-        countries: null,
-        country_codes: null, // eslint-disable-line camelcase
-        genres: null,
+        countries: [],
+        countryCodes: [],
+        genres: [],
         id: {
             tvdb: null,
+            tmdb: null,
+            tvmaze: null,
+            imdb: null,
             slug: null
         },
         indexer: null,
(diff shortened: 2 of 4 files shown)


dredd/api-description.yml
```diff
@@ -989,7 +990,7 @@ definitions:
         type: array
         items:
           type: string
-          description: Country codes
+          description: Country names
```
OmgImAlexis: Does this still pass `en`, etc?

themes-default/slim/src/store/modules/defaults.js
```diff
@@ -19,30 +19,39 @@ const state = {
...
         language: null,
         network: null,
         nextAirDate: null,
         plot: null,
-        rating: {
-            imdb: {
-                rating: null,
-                votes: null
-            }
+        imdbInfo: {
```
OmgImAlexis: Maybe this would be better as `indexer: { imdb, trakt, ... }` since Trakt, etc. can also have different ratings, etc. vs IMDB. Personally I'd prefer Trakt data over `IMDB`.

dredd/api-description.yml
sharkykh: The documentation was just incorrect.
So, from the updated `Series.to_json`:
```py
        data['countries'] = self.countries  # e.g. ['ITALY', 'FRANCE']
        data['countryCodes'] = self.country_codes  # e.g. ['it', 'fr']
```

themes-default/slim/src/store/modules/defaults.js
sharkykh: `indexer` is already in use as the indexer name, plus, it's not really indicative of the data it contains
:thinking: 

P.S.
Trakt users *always* rate the movies with low scores even before they come out, so annoying.

p0psicles: Im confused. In the other discussion you mentioned that having a multi indexer structure would get messy. I get that. But then why add imdInfo here?

Then just decouple it into a dedicated call? I mean next we could finish the imdb indexer. Wouldn't we need to change this eventually?

p0psicles: To have it clean we should only return tvdb rating for a tvdb show?

sharkykh: > Im confused. In the other discussion you mentioned that having a multi indexer structure would get messy. I get that. But then why add imdbInfo here?

@p0psicles 
Alright maybe I didn't fully get the point you were trying to make in the other discussion.
Maybe we were talking about the same thing, and just had different structures in mind?
Did you mean calling `GET /api/v2/series/tvdb12345` would return:
```json5
{
	"tvdb": {
		/* data from tvdb */
	},
	"imdb": {
		/* data from imdb */
	},
	"tmdb": {
		/* data from tmdb */
	}
}
```
Because that's what I thought was messy.

> Then just decouple it into a dedicated call?

I hate that part of the UI change,
because I thought everything we need would be within the reach of **one** API call.
I thought everything will be like the `/config` endpoint where you can get different sections of data, or all of them together.

> I mean next we could finish the imdb indexer. Wouldn't we need to change this eventually?

<sub>Personally, I don't see how imdb can be a better indexer than any of the other 3.</sub>
I mean, yes, `imdbInfo` would be useless for the shows that would use imdb as an indexer...

> To have it clean we should only return tvdb rating for a tvdb show?

It's this whole "consolidating information from different sources" thing that has me confused.
We serve the data from the show's main indexer,
and then some of the information is collected from other indexers (TMDB - countries/country codes),
and some of it from "external" indexers (IMDB - rating, votes, ...)
It was fine for using in mako because you can always access all the information stored in the Series object, but now, you have to decide where it's going to go and how to structure all the possible info,
when sometimes the info comes **directly** from the main indexer
and sometimes it doesn't...
:confused: :confounded:

p0psicles:
Whats the status of this pr?

PR #5222, opened 2018-09-16
convert showheader's status changer to vue

OmgImAlexis:
Shouldn't need to update snatch selection as it doesn't use this even though it uses `showheader.mako`.


themes-default/slim/src/components/display-show.vue
```diff
@@ -47,7 +48,8 @@ export default {
     computed: {
         ...mapState({
             shows: state => state.shows.shows,
-            indexerConfig: state => state.config.indexers.config.indexers
+            indexerConfig: state => state.config.indexers.config.indexers,
+            statusStrings: state => flip(state.statuses.strings)
```
sharkykh: just a suggestion, isn't this simpler?
https://github.com/pymedusa/Medusa/blob/7597778ce0c19be4af039877127342eba7e4c316/themes-default/slim/src/components/add-show-options.vue#L239-L248

themes-default/slim/src/components/display-show.vue
```diff
@@ -89,6 +91,31 @@ export default {
             const id = show.id[show.indexer];
             const indexerUrl = indexerConfig[show.indexer].showUrl;
             return `${indexerUrl}${id}`;
+        },
+        statuses() {
+            const { statuses, config } = this.$store.state;
```
sharkykh: you're destructuring from `this.$store.state`, but I think you meant to use `mapState` for these

OmgImAlexis: Nope, read the function.

p0psicles: I don't understand this. A little bit more explanation would be nice.

themes-default/slim/package.json
```diff
@@ -60,6 +60,7 @@
     "imagemin-pngquant": "6.0.0",
     "is-visible": "2.2.0",
     "jquery": "3.3.1",
+    "just-flip-object": "1.1.0",
```
sharkykh: https://github.com/sharkykh/Medusa/commits/feature/vueify-showheader-statuschanger-updated

sharkykh:
Replaced by #6189

PR #5357, opened 2018-10-07
use router-link when both pages are vueified

OmgImAlexis:
If both pages are 100% using Vue then we can load the next page via a `router-link` instead of needing to use a normal anchor which avoids the reload.

So far I've only enabled `/addShows` and `/config`.

Waiting on https://github.com/standard-things/esm/issues/621


OmgImAlexis:
@sharkykh any ideas why the build is failing? I'm getting a weird error for a non0-existent line. `helpers/index.js` is only 15 lines long.

```
  Uncaught exception in test/specs/add-recommended.spec.js

  /Users/xo/code/Medusa/themes-default/slim/src/components/helpers/index.js:35

  TypeError: Cannot read property 'default' of undefined

  Object.get [as AppLink] (src/components/helpers/index.js:35:21)
  Object.<anonymous> (src/components/add-shows.vue:448:23)
  Object.<anonymous> (node_modules/esm/esm.js:1:214605)
  Object.o._compile (node_modules/esm/esm.js:1:210603)
  compile (node_modules/require-extension-hooks/src/hook.js:75:10)
  hook (node_modules/require-extension-hooks/src/hook.js:32:3)
  Zl (node_modules/esm/esm.js:1:209533)
  Object.<anonymous> (node_modules/esm/esm.js:1:209726)
  node_modules/esm/esm.js:1:211678
  node_modules/esm/esm.js:1:211732
```

sharkykh:
I don't know either, but keep in mind you're importing `router`, which in turn imports other vue files. It could be a number of things...
As for the non-existent line - the source code is transpiled...

sharkykh:
For what it's worth, switching the imports on `./src/router.js` to dynamic imports seems to fix the test...

Followed https://router.vuejs.org/guide/advanced/lazy-loading.html#grouping-components-in-the-same-chunk
1. Requires [`@babel/plugin-syntax-dynamic-import`](https://babeljs.io/docs/en/babel-plugin-syntax-dynamic-import/) to be installed and configured in `.babelrc`
2. XO lint doesn't pass due to a parsing error (`import()` is unsupported by `eslint` because it's still experimental), may require [`babel-eslint`](https://www.npmjs.com/package/babel-eslint) to work....
3. It seems to work but it's still mostly untested...

I'll push the commit so you can play with it...

sharkykh:
There - the unit tests are working, and as I said - XO/eslint throws parsing error due to the `import()` functions...
I wasn't able to make `babel-eslint` work. It's possible it's not compatible with eslint 5.

sharkykh:
Looks like it's fixed now.
For reference: https://github.com/babel/babel-eslint/issues/530#issuecomment-433926548

themes-default/slim/src/router.js
```diff
@@ -1,13 +1,12 @@
...
-    AddRecommended,
-    AddShows,
-    Config,
-    ConfigPostProcessing,
-    IRC,
-    Login,
-    NotFound
-} from './components';
+
+const AddRecommended = () => import('./components/add-recommended.vue');
```
OmgImAlexis: Can these not return the promise directly?

PR #5461, opened 2018-10-20
Added support for Pushover notification priority configuration.

andofrjando:
Pushover supports five levels of notification priority, from very low to emergency. This is detailed https://pushover.net/api#priority and support for this is the main reason for this PR.

Also fixed a bug where failed notifications were not properly detected, and clarified which menu option will use the default sound.

The failed notifications bug was caused by the fact that 4XX HTTP response codes don't raise an exception, which is how they were trying to be detected previously. This just checks the response code rather than try and catch an exception that is never raised.

- [X] PR is based on the DEVELOP branch
- [X] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [X] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

PR #4913, opened 2018-08-18
Feature/fully vueify notifications

p0psicles:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

### TODO:
- [x] create component config-toggle-slider
- [x] create component config-textbox
- [x] kodi
- [x] plex media server
- [x] plex home
- [x] emby
- [x] nmj
- [x] nmjv2
- [x] synology
- [x] synology notifier
- [x] pyTivo
- [x] growl
- [x] prowl
- [x] libnotify
- [x] pushover
- [x] boxcar2
- [x] pushalot
- [x] pushbullet (jquery removed & tested)
- [x] freemobile
- [x] telegram (jquery not removed & tested)
- [x] twitter (jquery removed & tested)
- [x] trakt (jquery not removed)
- [x] email ((jquery not fully removed)
- [x] slack (jquery not removed)
- [x] save config
- [x] Test if all the jquery still works. As allot of $('id') needs to be replaced with $('id').find('input')
- [x] update api-description
- [x] fix build (need to update the test fixtures)
- [x] add snotify

PR #5512, opened 2018-10-26
Added ability to set priority of Pushover notifications.

andofrjando:
Also fixed a bug where failed notifications would not be caught.

This PR is the same as #5461 which was close pending some tidy up and fixed the way it was rebased. This is now correctly rebased on #4913


p0psicles:
Please rebase on develop now.

andofrjando:
Wilco.

PR #5518, opened 2018-10-26
Utilize the SnatchQueueItem

p0psicles:
…ueueItem), to snatch the torrent/nzb.

- [x] Replace the ManaulSnatchQueueItem with generic SnatchQueueItem.
- [x] Update SearchResult.search_type docstring.


p0psicles:
I added enum because at first, I thought it was a default lib. And it's simple, and you can repr() or str() it. So no need to do all kinds of reverse mappings. But then noticed enum was introduced in py3.4. So I added a very small enum package to /ext.

medusa/search/__init__.py
```diff
@@ -3,7 +3,9 @@
 """Search module for all Medusa searches."""
 from __future__ import unicode_literals
 
-from enum import Enum
+import sys
+if sys.version_info < (3, 4):
+    from ext2.enum import Enum
```
labrys: You never import enum in Py >= 3.4

labrys: You should only add ext2 in Python < 3.  Also Python 3 < 3.4 shouldn't be supported.  Finally your version never imports under Python > 3.4.

p0psicles: Good point

labrys: Also consider using IntEnum since they were traditionally ints, although im not sure if the backport supports int.  You could create your own IntEnum if it doesnt.  `class IntEnum(builtins.int, Enum)`.

PR #5522, opened 2018-10-26
Send episode object to download and subtitle notifications

MrTimscampi:

This prepares the terrain for tackling a proper Discord notifier. 
Sending the episode object allows to use embeds in the Discord notifier, as we'll have easy access to season, episode number, quality and other information, and will probably allow us to make other notifiers nicer in the future as well.

### DIFF
--- a/medusa/notifiers/boxcar2.py
+++ b/medusa/notifiers/boxcar2.py
@@ -62,15 +62,15 @@ class Notifier(object):
         if app.BOXCAR2_NOTIFY_ONSNATCH:
             self._notify_boxcar2(title, ep_name)
 
-    def notify_download(self, ep_name, title=common.notifyStrings[common.NOTIFY_DOWNLOAD]):
+    def notify_download(self, ep_obj, title=common.notifyStrings[common.NOTIFY_DOWNLOAD]):
         """Send the download message."""
         if app.BOXCAR2_NOTIFY_ONDOWNLOAD:
-            self._notify_boxcar2(title, ep_name)
+            self._notify_boxcar2(title, ep_obj._format_pattern('%SN - %Sx%0E - %EN - %QN'))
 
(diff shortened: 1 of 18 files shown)


medusa/notifiers/boxcar2.py
```diff
@@ -62,15 +62,15 @@ def notify_snatch(self, ep_name, is_proper):
         if app.BOXCAR2_NOTIFY_ONSNATCH:
             self._notify_boxcar2(title, ep_name)
 
-    def notify_download(self, ep_name, title=common.notifyStrings[common.NOTIFY_DOWNLOAD]):
+    def notify_download(self, ep_obj, title=common.notifyStrings[common.NOTIFY_DOWNLOAD]):
         """Send the download message."""
         if app.BOXCAR2_NOTIFY_ONDOWNLOAD:
-            self._notify_boxcar2(title, ep_name)
+            self._notify_boxcar2(title, ep_obj._format_pattern('%SN - %Sx%0E - %EN - %QN'))
```
medariox: This should be something like `pretty_name()` or maybe just an extension of it to include the quality.

MrTimscampi: The _format_pattern() is already what was used higher in the execution (ep_name was ep_obj._format_pattern() for this download notification and ep_obj.pretty_name() for the subtitles)

I can whip up a public function for it if need be, though (I don't like using private functions outside of the class).

medariox: > I can whip up a public function for it if need be, though (I don't like using private functions outside of the class).

This is exactly what I meant actually, sorry for being unclear about it.

MrTimscampi: @medariox I added a pretty_name_with_quality() public function to Episode.
It produces the same format the old code was generating, with the added benefit of supporting absolute episodes and date-based (it's essentially pretty_name() with the quality added at the end).

Every notifier has been modified to use it. Should be good to go like this, I think.

PR #5543, opened 2018-10-29
Fix database always trying to update

medariox:

### DIFF
--- a/medusa/databases/main_db.py
+++ b/medusa/databases/main_db.py
@@ -369,7 +369,6 @@ class AddMinorVersion(AlterTVShowsFieldTypes):
         self.inc_minor_version()
 
 
-
 class TestIncreaseMajorVersion(AddMinorVersion):
     """
     This tests the inc_major_version function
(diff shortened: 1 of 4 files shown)


medusa/databases/main_db.py
```diff
@@ -346,26 +346,28 @@ def inc_major_version(self):
...
         minor_version += 1
         self.connection.action('UPDATE db_version SET db_version = ?, db_minor_version = ?;',
                                [major_version, minor_version])
+        log.info(u'[MAIN-DB] Updated minor version to: {}.{}', *self.connection.version)
+
         return self.connection.version
 
     def execute(self):
-        utils.backup_database(self.connection.path, self.checkDBVersion())
+        utils.backup_database(self.connection.path, self.checkMajorDBVersion())
```
sharkykh: I don't these (the backup calls) should be changed.
Don't you need the tuple so that the database backup contains the minor version too?
image

medariox: I think that's okay, because these backups are going to be of very old versions (< 42) and won't have a minor version anyway. The new versions are using `self.connection.version` which contains the major and minor version.

PR #5556, opened 2018-10-30
Feature/vueify backuprestore

OmgImAlexis:
Converts backuprestore to vue SFC.


themes-default/slim/src/router.js
```diff
@@ -297,8 +300,8 @@ const changesRoute = {
 };
 
 const ircRoute = {
-    path: '/IRC',
-    name: 'IRC',
+    path: '/irc',
```
sharkykh: Note that this will break the IRC page link. The link is case sensitive, so if you want to change it, change the route as well.
https://github.com/pymedusa/Medusa/blob/98aeff745c8338ebaa4fdd433adb801031c10542/medusa/server/web/home/irc.py#L12

themes-default/slim/src/components/index.js
```diff
@@ -5,6 +5,7 @@ export { default as AnidbReleaseGroupUi } from './anidb-release-group-ui.vue';
 export { default as AppHeader } from './app-header.vue';
 export { default as Backstretch } from './backstretch.vue';
 export { default as Config } from './config.vue';
+export { default as ConfigBackuprestore } from './config-backuprestore.vue';
```
sharkykh: This makes more sense in my opinion... :thinking: 
```suggestion
export { default as ConfigBackupRestore } from './config-backup-restore.vue';
```

p0psicles:
@OmgImAlexis what's the status of this? Can we have this merged asap? Might wanna take a look at @sharkykh's comments.

PR #5623, opened 2018-11-04
Fix show-selector for libraries with more than 1k shows

### LINKED ISSUE #5240: Show count over 1000 gives some issues.
When you have db with more than 1000 shows the following issues arise:

* displayShow page does not display the shows that are above 1000 count (when sorted "alphabetically")
  - so for example if "Still Game" is your #1000 show the show "The Young Ones" does not show data on displayShow page, but the show "A Touch of Cloth" does work.

* Also an issue is the dropdown box on displayShow does not render all shows, but rather stops at 1000 entries.

When I found what would be show #1001 in the list and removed the #1000 the next show in the list (#1001) took #1000s place in the dropdown. I tried this 4 times by removing of the of shows that was in the list. For each I removed the next in line after #1000 would be visible and displayShow worked.

The index page does however display all show entries if > 1000 in total.
If this issue is affecting other aspects of the medusa enviroment I am unsure as I noticed this today.

Tested in both chrome and firefox.
No visible bugs noted in console or log files as far as I can see.
No errors reported inside medusa error reporting page.

sharkykh:
Fixes #5240


medusa/server/api/v2/base.py
```diff
@@ -319,7 +319,7 @@ def _get_page(self):
         except ValueError:
             self._raise_bad_request_error('Invalid page parameter')
 
-    def _get_limit(self, default=20, maximum=1000):
+    def _get_limit(self, default=20, maximum=10000):
```
OmgImAlexis: @sharkykh the one problem with this approach that we're now skipping pagination. Maybe we should open an issue about this so we don't forget?

PR #5680, opened 2018-11-09
Enable non-anime search on Nyaa

MrTimscampi:

Nyaa has a Live Action category, this allows the provider to make use of it by adapting the RSS URL depending if the series has the is_anime flag or not.

### DIFF
--- a/medusa/providers/torrent/rss/nyaa.py
+++ b/medusa/providers/torrent/rss/nyaa.py
@@ -32,7 +32,7 @@ class NyaaProvider(TorrentProvider):
 
         # Miscellaneous Options
         self.supports_absolute_numbering = True
-        self.anime_only = True
+        self.anime_only = False
         self.confirmed = False
 
         # Torrent Stats
@@ -48,7 +48,7 @@ class NyaaProvider(TorrentProvider):
 
         :param search_strings: A dict with mode (key) and the search value (value)
         :param age: Not used
-        :param ep_obj: Not used
+        :param ep_obj: An episode object
         :returns: A list of search results (structure)
         """
         results = []
@@ -56,7 +56,7 @@ class NyaaProvider(TorrentProvider):
         # Search Params
         search_params = {
             'page': 'rss',
-            'c': '1_0',  # All Anime
+            'c': '1_0' if ep_obj.series.is_anime else '4_0',  # All Anime
             'f': 0,  # No filter
             'q': '',
         }
--- a/medusa/providers/torrent/rss/nyaa.py
+++ b/medusa/providers/torrent/rss/nyaa.py
@@ -32,7 +32,6 @@ class NyaaProvider(TorrentProvider):
 
         # Miscellaneous Options
         self.supports_absolute_numbering = True
-        self.anime_only = False
         self.confirmed = False
 
         # Torrent Stats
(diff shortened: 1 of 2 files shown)


medusa/providers/torrent/rss/nyaa.py
```diff
@@ -48,15 +48,15 @@ def search(self, search_strings, age=0, ep_obj=None, **kwargs):
...
+        :param ep_obj: An episode object
         :returns: A list of search results (structure)
         """
         results = []
 
         # Search Params
         search_params = {
             'page': 'rss',
-            'c': '1_0',  # All Anime
+            'c': '1_0' if ep_obj.series.is_anime else '4_0',  # All Anime
```
p0psicles: Test is failing, because we have tests where we test an rss/daily search. Those don't pass an ep_obj. So you have to take that into account for both of your PR's.

MrTimscampi: Just did a commit on the AniDex one. Doing this one now.

I opted to handle the None case in the provider, instead of fixing the test case to pass an object, since it seems sensible to handle all the cases in the function itself (And I failed my attempt at making the test work :p )

medusa/providers/torrent/rss/nyaa.py
```diff
@@ -32,7 +32,7 @@ def __init__(self):
 
         # Miscellaneous Options
         self.supports_absolute_numbering = True
-        self.anime_only = True
+        self.anime_only = False
```
medariox: You can just remove this as it defaults to `False`.

medusa/providers/torrent/rss/nyaa.py
```diff
@@ -48,15 +48,21 @@ def search(self, search_strings, age=0, ep_obj=None, **kwargs):
...
         :param search_strings: A dict with mode (key) and the search value (value)
         :param age: Not used
-        :param ep_obj: Not used
+        :param ep_obj: An episode object
         :returns: A list of search results (structure)
         """
         results = []
 
         # Search Params
+        if ep_obj:
```
medariox: I think this can be a bit more DRY like:
```
category = '1_0'
if ep_obj and ep_obj.series.is_anime:
    category = '4_0'
```

medusa/providers/torrent/rss/nyaa.py
```diff
@@ -48,15 +48,21 @@ def search(self, search_strings, age=0, ep_obj=None, **kwargs):
...
+            category = '1_0' if ep_obj.series.is_anime else '4_0',  # All Anime if is_anime, else All Live Action
+        else:
+            # If ep_obj is None, assume we want anime
+            category = '1_0'
```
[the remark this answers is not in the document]
MrTimscampi: Their RSS doesn't handle multiple categories, so in the case an ep_obj is given (outside of a test case), it'll either return category 4_0 (Live Action) or 1_0 (Anime) if the series is an anime.

And I also needed to keep the compatibility with the current behaviour by defaulting to anime if ep_obj=None.

medusa/providers/torrent/rss/nyaa.py
MrTimscampi: Indeed, it makes more sense, but it'd be something more like:

```
category = '1_0'
if ep_obj and not ep_obj.series.is_anime:
    category = '4_0'
```

Since 4_0 is Live Action :)

PR #5681, opened 2018-11-09
Enable non-anime search on Anidex

MrTimscampi:

Anidex has a Live Action category, this allows the provider to make use of it by adapting the RSS URL depending if the series has the is_anime flag or not.

### DIFF
--- a/medusa/providers/torrent/html/anidex.py
+++ b/medusa/providers/torrent/html/anidex.py
@@ -56,8 +56,14 @@ class AniDexProvider(TorrentProvider):
         """
         results = []
 
+        if ep_obj:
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/html/anidex.py
```diff
@@ -36,7 +36,7 @@ def __init__(self):
 
         # Miscellaneous Options
         self.supports_absolute_numbering = True
-        self.anime_only = True
+        self.anime_only = False
```
p0psicles: Same here. You can remove to make it opt-in like the other providers.

MrTimscampi: Already done ;)

PR #5691, opened 2018-11-11
Added new function runs_in_docker.

p0psicles:
Used for checking if medusa is running in a docker container.

### DIFF
--- a/medusa/version_checker.py
+++ b/medusa/version_checker.py
@@ -89,6 +89,29 @@ class CheckVersion(object):
 
         self.amActive = False
 
+    def run_in_docker(self):
+        """"
+        Check if medusa is run in a docker container.
+
+        If run in a container, we don't want to use the auto update feature, but just want to inform the user
+        there is an update available. The user can update through getting the latest docker tag.
+        """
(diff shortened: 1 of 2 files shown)


medusa/version_checker.py
```diff
@@ -89,6 +89,29 @@ def run(self, force=False):
 
         self.amActive = False
 
+    def run_in_docker(self):
```
medariox: I think this might be better called `runs_in_docker`.

medusa/version_checker.py
```diff
@@ -89,6 +89,29 @@ def run(self, force=False):
...
+        there is an update available. The user can update through getting the latest docker tag.
+        """
+        path = '/proc/{pid}/cgroup'.format(pid=os.getpid())
+        try:
+            if not os.path.isfile(path):
+                return False
+
+            with open(path) as f:
+                for line in f:
+                    if re.match("\d+:[\w=]+:/docker(-[ce]e)?/\w+", line):
```
medariox: Probably better to use `r` and and escape the slashes: `r'\d+:[\w=]+:\/docker(-[ce]e)?\/\w+'`

Issue #5198, opened 2018-09-12
The database migrations bug


sharkykh:
### The bug
During the past 2 database migrations, the databases of some of the users got updated past the "max" database version and some got incorrect values inserted into them.
For example, between `v0.2.4` and `v0.2.5` we [bumped the main DB version from `44.9` to `44.11`](https://github.com/pymedusa/Medusa/pull/4221).
However, during the migration to shift the quality values, some of the users got incorrect values inserted into the databases.
In addition, some users database versions actually got bumped past `44.11`, to `44.12`, which caused an issue for users when the [next database migration](https://github.com/pymedusa/Medusa/pull/4825) landed on `v0.2.9`.

### What happens when Medusa updates?
Currently, after Medusa updates its code either using `git pull` or downloading the archive from GitHub, it stops the web server and other threads, closes open files, etc.
After that's done, **and unless the user had enabled the `No Restart` option**, it spawns a new process of Medusa with the same arguments as the one that was running. ([source code](https://github.com/pymedusa/Medusa/blob/ae29b17e1e489c7e0067f4bbd162965c8761b451/medusa/__main__.py#L2097-L2105))

### How does the restart process affect the database migrations?
Due to the use of process monitors or service restarters, which automatically restart a process if it dies, in-conjunction-with Medusa's restart process (with `No Restart` off),
Medusa is being loaded **twice, at the same time, by two separate processes**.
The database migrations (try to) execute on both processes at the same time, and in doing so, causing unexpected changes to database entries.

The current Docker configuration (ours, [@linuxserver's](https://github.com/linuxserver/docker-medusa) and possibly others) uses [`S6 overlay`](https://github.com/just-containers/s6-overlay#writing-an-optional-finish-script) to automatically restart the server inside the container when it dies ([source code](https://github.com/pymedusa/Medusa/blob/ae29b17e1e489c7e0067f4bbd162965c8761b451/.docker/root/etc/services.d/medusa/run)).
This means that as soon as the original Medusa process dies out, `S6` spawns a new one, but Medusa does too, which causes the scenario explained above.
This can be **clearly** seen in the Docker logs, if you restart Medusa from the web UI, you can see duplicate log messages being logged, and the server tries to bind to the port twice - one of the processes worked, and the other says [it can't bind to the port because it's in use, and exits](https://github.com/pymedusa/Medusa/blob/ae29b17e1e489c7e0067f4bbd162965c8761b451/medusa/server/core.py#L292-L295).

[...]

PR #5700, opened 2018-11-12
Feature/enable check runs in docker

p0psicles:

Step 2.
Enable the check. And visualize the result in the Help & Info page.

Next step will be to utilize the check, to prevent the automatic updating and restarting in docker containers. Need to figure out how that works first.

Should address some of the issues in: #5198 

What I noticed is that there are two docker images widely used.

1. pymedusa/medusa
2. linuxserver/medusa

### pymedusa/medusa
Does not configure a git repository on the app source within the container.
Therefor it uses the SourceUpdater.

### linuxserver/medusa
Uses git, and uses the GitUpdater.

For both these containers, users where getting notifications when an update was available.
They could click the `update now` link, to start the update process.

This will now not be available to them anymore. So allot of users will be confused.
For now i've added the following message. But that can be modified of course: `There is an update available: please pull the latest docker image, and rebuild your container to update`

I understand if some users think this is a step back. But this is the only way for now, to perform a reliable upgrade.


medusa/version_checker.py
```diff
@@ -896,6 +896,12 @@ def set_newest_text(self):
         # if we're up to date then don't set this
         app.NEWEST_VERSION_STRING = None
 
+        if self.runs_in_docker() and (not self._cur_commit_hash or self._num_commits_behind > 0):
+            log.debug(u'There is an update available, medusa is running in a docker container, so auto updating is disabled.')
+            app.NEWEST_VERSION_STRING = 'There is an update available: please pull the latest docker image, ' \
+                                        'and rebuild your container to update'
```
OmgImAlexis: Maybe we should link to something here?

p0psicles: Step by step. First prio, is to not have installations break.
I'd rather have users come over here, to ask how to update in docker for now.

p0psicles: We could link to wiki?

p0psicles: Okay, this works: 
image

p0psicles:
@sharkykh thoughts?

medusa/version_checker.py
```diff
@@ -46,6 +46,35 @@
 log.logger.addHandler(logging.NullHandler())
 
 
+def runs_in_docker():
```
medariox: Why not keep this in `UpdateManager` and run it once on init?

p0psicles: And then only use the global after?

medariox: Possibly, yes.

p0psicles: implemented the changes.

medusa/version_checker.py
```diff
@@ -405,9 +382,53 @@ def get_github_repo():
...
+        try:
+            if not os.path.isfile(path):
+                return False
+
+            with open(path) as f:
+                for line in f:
+                    if re.match(r'\d+:[\w=]+:/docker(-[ce]e)?/\w+', line):
+                        log.debug(u'Running in a docker container')
+                        app.RUNS_IN_DOCKER = True
+                        app.instance.save_config()
```
p0psicles: Ooh this shouldn't be in here

PR #5701, opened 2018-11-12
Do not allow to set anime option(s) when adding shows through tmdb or…

p0psicles:
… tvmaze.

Update changelog.


p0psicles:
@sharkykh @OmgImAlexis @MrTimscampi what do you guys think?
I see allot issues reported now from users that use tvmaze or tmdb to add anime.
The issue is, that the scene numbering column in the db is not properly populated. You can see that because it tries to use episode `00` to search.

So there are a number of underlying bugs / restrictions, why tvmaze and tmdb should be omitted for anime. Like for ex, that we can't map tmdb to scene numbering from xem.

I think best is now, to disallow to set anime for these indexers.
@MrTimscampi when you split the anime from the numbering, this should get attention.

themes-default/slim/views/addShows_newShow.mako
```diff
@@ -609,7 +615,7 @@ window.app = new Vue({
                 <fieldset class="sectionwrap">
                     <legend class="legendStep">Customize options</legend>
                     <div class="stepDiv">
-                        <add-show-options :show-name="showName" enable-anime-options @change="updateOptions" @refresh="refreshOptionStep"></add-show-options>
+                        <add-show-options :show-name="showName" :enable-anime-options="enableAnimeOptions" @change="updateOptions" @refresh="refreshOptionStep"></add-show-options>
```
OmgImAlexis: Let's use `v-bind` for the ones with the same name.

Issue #5729, opened 2018-11-16
Match release group for subtitles


Thilas:
I have a problem with the automatic subtitle downloading. If I am not using "perfect match" option, subtitles are most of the time incorrect (mainly because of a wrong release group). And if I am using it, subtitles are almost never automatically downloaded (at least in my case for French language using only addic7ed source).

Looking at the code, I saw that:

- by default, Medusa tries to match: series + year + season + episode
- and with perfect match, it tries to match: hash - resolution - video_codec - audio_codec

While perfect match score is fine to me according to its definition ("perfect" match), I think the default match should also take the release group into account. I think this is a safe way to ensure that Medusa is downloading valid subtitles.

An other way we could see it is to allow to specify whatever score we want through Medusa configuration (with multiple checkboxes or a multiple select for instance) but I guess this would make the thing more complex to define.

Thank you!

Thilas:
From what I saw in the code, yes.

p0psicles:
Then youll have to report there.

p0psicles:
https://github.com/Diaoul/subliminal

medariox:
The perfect match is something we have implemented, will look into it when I have some time.

Thilas:
@p0psicles, I misunderstood your question. The issue comes from how Medusa uses subliminal and not from subliminal itself.

If my proposition makes sense, I can provide a pull request (I mean match the release group even for non-perfect match).

Thilas:
@medariox, what do you have in mind? Leave non-perfect match as is and change perfect match to: series + year + season + episode + release_group.

Thilas:
Still waiting for #5784 to be released, but #5740 is ok to me. Closing this ticket.

PR #5740, opened 2018-11-18
Perfect match for subtitles is too strict

Thilas:
Change perfect match in order to make it less strict as discussed in #5729.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### New Features
 
 #### Improvements
+- Improved perfect match for subtitles downloading by making it a bit less strict ([#5729](https://github.com/pymedusa/Medusa/issues/5729))
 
 #### Fixes
 - Fixed ImportError when using Download Station client ([#5748](https://github.com/pymedusa/Medusa/pull/5748))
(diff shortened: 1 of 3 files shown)


Thilas:
@medariox, could you review this PR?

medariox:
Thanks for the PR!
The problem with this is that some providers like `LegendasTV` group the release groups together, like: img
I understand your reasoning behind this, but it will cause a lot of users that use `LegandasTV` to complain about subs not being downloaded. Maybe there is some other way to do this e.g. using the language instead of release group?

Thilas:
I'm confused: I only changed the perfect match option to make it less strict. I mean that with the current perfect match, from what I understand, it already didn't download subtitles in such cases like the one you mention for `LegendasTV`. And I guess it also wouldn't with my change. So the only way for users is to use non-perfect match here, which I didn't change. Am I wrong?

medariox:
Maybe it was there other way around, lowering the score makes it download wrong subtitles even with perfect match? Oh well, I would say we try it out and see how it turns out.

medariox:
@Thilas
Can you update the changelog please? I would add this to improvements.

Thilas:
@medariox, @p0psicles, here is the changelog.

PR #5829, opened 2018-12-05
Begin fixes for storing commit hash dynamically

a10kiloham:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,3 @@
-.git
 .gitignore
-.github
-.gitattributes
 READMETEMPLATE.md
 README.md


p0psicles:
@a10kiloham you have to explain me. Are we or are we not including the .git folder?
As when we add the .git folder, medusa will be able to determine the cur commit hash just fine. So no further changes required.

For example the image https://github.com/linuxserver/docker-medusa already adds the .git folder.
That image doesn't have the issue we're trying to solve.

But, i'm no docker expert here. So I don't know what's best practice here.

So we first need to decide what we want. Do we add the .git folder or not?
Personally, I don't think we should want that. And just add the source, as we don't want users to provide the option of running commands like `git pull`.

p0psicles:
@OmgImAlexis care to pitch in? As you also use docker?

a10kiloham:
The Linux server version should be ignored imo. Not relevant since I added our own docker file.
Someone else's call whether to keep the git dir. I'd prefer to remove it after we extract the head but that's above my pay grade. My commit was a suggestion how to do exactly that. Then just rm the git directory so people don't try to pull inside a container.

OmgImAlexis:
Is there a reason we're not just reading the version via an env that we set in the dockerfile?

PR #5831, opened 2018-12-05
Check version if /git-commit exists

a10kiloham:
For use in Docker scenarios - checks existence of newly created /git-commit file which has the hash for the Docker version's git pull. Not sure if this is v good but please do feel free to edit if not up to snuff.

### DIFF
--- a/medusa/version_checker.py
+++ b/medusa/version_checker.py
@@ -585,6 +585,10 @@ class GitUpdateManager(UpdateManager):
             self._cur_commit_hash = cur_commit_hash
             app.CUR_COMMIT_HASH = str(cur_commit_hash)
             return True
+        elif os.path.isfile('/git-commit'):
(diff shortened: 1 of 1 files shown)


p0psicles:
You've made the check part of the GitUpdateManager. But that's not an issue.
The issue is when we don't have a .git folder, with the source. So we should have a way in the SourceUpdateManager, to get the commit.

a10kiloham:
Taking .git out of the .dockerignore file leaves a .git directly but I don't know how to get the commit out of that.

p0psicles:
Something like this could work. Not tested:
https://gist.github.com/p0psicles/87e648522470270b9374af5aad24cf93

p0psicles:
Okay i understand. 
But Its not a problem that we dont have the git folder.

Issue #3679, opened 2018-01-28
[Feature] Allow option to download early episodes (before air date)


jtwill:
Currently, if an episode is available before the air date that Medusa thinks it should have, Medusa will not download, and does not even log the fact that it is declining to download the episode. The only evidence in the log is that the episode gets added to the cache, but there is no message saying why the episode is not downloaded.

Medusa very much needs to have an option to download episodes early. A global option would be okay with me. Alternatively, make it an option specific to each provider. All of the providers I use are trustworthy and almost never get fake episodes. But I know some users probably have usenet or public tracker providers where fake episodes could be a problem, so in that case they would probably want to disable "early episode download" for certain providers.

And Medusa really needs to explicitly log why it is declining to download an episode ("not downloading because it is before episode air date")

p0psicles:
We could allow it only using the cache filled by rss. And define an offset per provider.

jtwill:
I don't understand the restriction "allow it only using the cache filled by rss". How would this affect which early episodes are downloaded?

I have noticed that Medusa fails to download two types of early episodes:

(1) Episodes that are available up to a week early. The most common type is when the episode is available on-demand in the morning before the evening air date, but sometimes it can be a day or two early, and once in a while, almost a week early.

(2) The first one or two episodes of a new show are sometimes available weeks or months early, as a preview. Maybe an explicit check could be added to download early episodes when a new show is added to Medusa.

duramato:
I'm not a big fan or allowing to search for episodes before air date, the downsides outweigh the _**possibily**_ of an episode "airing" before it's announced air date.

The solution I see is add the checkboxes for unaired episodes, so that if an user is inclined to set an episode as wanted he can do so.

![image](https://vgy.me/tzuN4P.png)

jtwill:
No, having to identify an early episode manually completely eliminates the whole benefit of having Medusa download episodes automatically.

Also, it is a lot more than a "possibility" of early episodes. It happens quite regularly. Several shows I watch are available early every week. And the "downsides" are almost non-existent for the providers that I use, since they almost never get fake episodes.

For people with reliable providers, it makes no sense to force them to wait until the air date to download early episodes. By all means make the default not to download early episodes, but for people who want to enable that option, it should be available.

ripper1967:
With over 400 shows i it can be very time consuming going through cleaning
fakes out.  i use good providers as well but they still get fakes through
from time to time. I use both Nzb and torrents. once people realize the
[...]

PR #5874, opened 2018-12-10
Feature configure per show daily search airdate offset

p0psicles:
…ate offset.

* Implemented check in dailsearch for searching episodes early.

### TODO:
- [x] add api spec
- [x] update changelog
- [x] add field to tv_shows db
- [x] add field to series class
- [x] add logic to dailsearch, for searching earlier
- [x] add logic to dailysearch, for search later
- [x] add UI control to editShow
- [ ] ~~add UI control to providers~~
- [x] add UI indicator to displayShow/snatchSelection header
- [x] Test by users

implement feature #3679 

### Update:
I've scratched the feature to enable/disable the feature per provider. As for that I need to make structural changes throughout medusa's code. And i'm not going to do that now.


rafi-d:
I've tested this with both per-search (negative offsets) and post-search (positive offsets), and everything seems to work OK!

Still, this offset needs to  be visible on the show page, as well as the manual search page - in the top/header - show-info part, at least in cases it is *not* 0.

Thanks!

Rafi

https://github.com/pymedusa/Medusa/issues/3679#issuecomment-446289920

rafi-d:
...a small typo in series.py - 

# The offset in **ours**, for when we want to start searching episodes for this show. -1 is search 1 hour early. 1, 1 hour later.

p0psicles:
> ... one small typo in series.py -
tnx, will correct

> Also, be aware that once the early offset-time arrives, the status for a specific episode (changed to "wanted" when the offset time arrives) is not/cannot be reverted back to "unaired" even if the user corrects this offset, or sets it to 0 again. No a big deal I believe...

Yes I'm aware. Also, if the backlog search is triggered. It will also search for these early episodes if they're already set to wanted by the daily search.

rafi-d:
This feature (dealing with offsets/time-limits) reminds  of another desired feature - an autofiles size filter (min/max). Actually I have it working here locally great (tho, w/o changing anything in the DB yet). So, since you are changing the DB for this feature, it might be a good point in time to consider adding two more edit-boxes for this file-size feature if you are interested ... :)

p0psicles:
> it might be a good point in time to consider adding two more edit-boxes for this file-size feature if you are interested ... :)

No i'm not interested. I already got like 6 other PR's on my backlog to finish. I did this one cause it's rather easy to implement.

rafi-d:
Sure, no problem. Works for me... 
For this feature here, I think you still need to add some "external" visibility for users .

medariox:
@rafi-d 
We are aware of it. Once the PR is done we will label it as **concluded**.

rafi-d:
@medariox 
What I meant, was have the offset ( if set ) *visible* on the show's page among the other show's details, so not to make it more clear why the status shows  as "wanted"  earlier or late relative to the air-time...

p0psicles:
Yes i updated the todo checklist of this pr.

rafi-d:
Ok, Thanks. Actually these two are pretty much done IMHO... 👍 
v add logic to dailysearch, for search later
v Test by users

One more thing to consider - one might see a strange show list like the attached (after a show has been snatched early, sorted by 'Next Ep', where future date(s) show in the 'Prev Ep' column). So, maybe adding another possible column - for "offset" can be helpful here?

rafi-d:
I have been testing this for the last couple of weeks, and results are very good, but there is this one symptom which I've noticed, and not sure if it is related or not:
After an episode has been downloaded, say, -24 hours ahead of time as designed, I  sometimes see an extra/second double download of the same episode for that same  show with -24 offset, with a "Proper" notation, and it seems to occur after the original airtime. It also does not delete the previous downloaded episode (by uTorrent). 
Is this something that can be expected? or maybe needs to be tweaked with regards to this new setting?

p0psicles:
Sounds like a "proper" was released after. But that should replace the original?

rafi-d:
Yes, I guess it was released later on. And yes, if you decided to download same release twice, it should have deleted the first one. Honestly, I have never seen this happen. 

Also, be aware that delete cleanly,  is not that easy, since you have to remove it from the client (WebAPI to  uTorrent) as well. So, in the worse case, when the client is still seeding it, you'd also need to stop it first.

p0psicles:
Okay but that doesn't sound as anything related to this pr. More something with your post-processing setup.

rafi-d:
>that doesn't sound as anything related to this pr

If so, let's forget  it for now over here.

BTW,  I do not have any post processing set, and didn't know that deleting of double download  episodes even requires post-processing. I thought if it  does an extra download, it should just clean/erase the previous one ....

I hope you release this feature soon, I am already one version behind...  :)

p0psicles:
I'll merge develop into it.

rafi-d:
maybe you can add minimal UI for it, so to make this easier to use/test?...

rafi-d:
Got it. Looks good, thanks. I guess it is ready for release...

p0psicles:
Yeah, just needs to get review

PR #5892, opened 2018-12-15
Store the python version on startup in config

p0psicles:
This can be used to check when a user starts medusa the next time with a different version.

### TODO:
- [x] Check on startup if where running python 3.0 > 3.4, then exit
- [x] Check if we up/down grade python version. Last version is saved to config
- [x] Replace dogpile cache files, when down/up grade.

### DIFF
--- a/medusa/__main__.py
+++ b/medusa/__main__.py
@@ -425,7 +425,7 @@ class Application(object):
             app.GIT_PASSWORD = check_setting_str(app.CFG, 'General', 'git_password', '', censor_log='low')
             app.GIT_TOKEN = check_setting_str(app.CFG, 'General', 'git_token', '', censor_log='low', encrypted=True)
             app.DEVELOPER = bool(check_setting_int(app.CFG, 'General', 'developer', 0))
-            app.PYTHON_VERSION = check_setting_int(app.CFG, 'General', 'python_version', 0)
+            app.PYTHON_VERSION = check_setting_list(app.CFG, 'General', 'python_version', [], transform=int)
 
             # debugging
             app.DEBUG = bool(check_setting_int(app.CFG, 'General', 'debug', 0))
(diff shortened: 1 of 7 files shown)


p0psicles:
@medariox this way we can check on version switches, and act on it?

medusa/__main__.py
```diff
@@ -1077,9 +1078,15 @@ def initialize(self, console_logging=True):
                 # Disable flag to erase cache
                 app.SUBTITLES_ERASE_CACHE = False
 
+            python_up_down_graded = False
+            # run some sanitation when switching from python versions
+            if app.PYTHON_VERSION and app.PYTHON_VERSION != sys.version_info.major:
```
medariox: I think this can all go into migrate_python_version()

medusa/__main__.py
```diff
@@ -1108,6 +1115,10 @@ def initialize(self, console_logging=True):
             migrator = ConfigMigrator(app.CFG)
             migrator.migrate_config()
 
+            if not app.PYTHON_VERSION:
+                app.PYTHON_VERSION = sys.version_info.major
+                self.save_config()
```
medariox: Same with this, should be in migrate migrate_python_version()

medusa/__main__.py
```diff
@@ -1247,6 +1258,27 @@ def path_leaf(path):
...
+        :param new_version: currently started python version
+        :type new_version: int
+        """
+        # TODO: Leaving this here as a template for when we upgrade merge the python3 changes.
+        if new_version == 3:
+            pass
+        else:
+            pass
+
+        app.PYTHON_VERSION = new_version
```
medariox: If this is only going to save the major version, I'd rather have this named `MAJOR_PYTHON_VERSION`. Otherwise I'd like to have the full version saved (even better actually).

p0psicles: Okay, i can store as a list

medusa/__main__.py
```diff
@@ -1247,6 +1258,27 @@ def path_leaf(path):
...
+
+        # save config
+        logger.info(u'Saving config file to disk')
+        self.save_config()
```
medariox: I don"t think this is needed? It's not done anywhere else and everything seems to be saved on shutdown.

p0psicles:
Ive tested this on the py3 branch. And now i dont have any more issues with the dogpile cache.

medusa/__main__.py
p0psicles: Well. No in not all cases its saved on shutdown nor restart

medariox: How so? It looks like the config is always saved on shutdown, see: https://github.com/pymedusa/Medusa/blob/3644aecf38a6689f1cfb76a8d0833b5d9376988d/medusa/__main__.py#L2158-L2168

p0psicles: okay, will depend on that

labrys:
@p0psicles What action do you envision taking upon version change or what issues are you experiencing that would require this?

Ideally Python code should be written to work with any supported version.

p0psicles:
@labrys for its only the dogpile dbm that needs to be repaced on some os. And we need to clean up pyc files. But im not sure if this is automatable

labrys:
@p0psicles you can skip creation of `.pyc` and `.pyo` files by running python with the `-b` option or the `PYTHONDONTWRITEBYTECODE` environment variable.

Python 3 stores `.pyc` and `.pyo` in the `__pycache__` directory so there should be no collisions between that and the Py2 files.

If its for issues with git updates, there are various ways to address the Py2 files.

PR #5917, opened 2018-12-25
Added a break, to prevent snatching multiple results.

p0psicles:

@medariox i'm really confused here.
Did we change something which resulted in this function being able to return multiple results for the same episode?
https://github.com/pymedusa/Medusa/blob/master/medusa/search/core.py#L634

Because this comment throws me off. As it says we only use the first. But there is no code, that makes sure of that.
https://github.com/pymedusa/Medusa/blob/master/medusa/search/queue.py#L551

I tested this. And the backlog search, returns an array with multiple search_result objects, for the same episode.

### DIFF
--- a/medusa/search/queue.py
+++ b/medusa/search/queue.py
@@ -405,6 +405,7 @@ class ForcedSearchQueueItem(generic_queue.QueueItem):
 
                     # Give the CPU a break
                     time.sleep(common.cpu_presets[app.CPU_PRESET])
+                    break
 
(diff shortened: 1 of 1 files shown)


medariox:
Iirc `search_providers` should only return one result per episode?

p0psicles:
Yeah that's not the case

p0psicles:
No don't merge it. The break is not the solution. Especially not for backlog search

PR #5920, opened 2018-12-26
Patched the combine_results function, to reduce the double single_ep …

p0psicles:
…results.

We only need one (best) result using the sort_results function.
* Added __eq__ to episode, so the == checks work properly on episode objects.

### DIFF
--- a/medusa/search/core.py
+++ b/medusa/search/core.py
@@ -705,9 +705,9 @@ def search_providers(series_obj, episodes, forced_search=False, down_cur_quality
                     cache_search_results = cur_provider.search_results_in_cache(episodes)
                     if cache_search_results:
                         # From our provider multi_episode and single_episode results, collect candidates.
-                        found_cache_results = list_results_for_provider(cache_search_results, found_results, cur_provider)
+                        cache_found_results = list_results_for_provider(cache_search_results, found_results, cur_provider)
                         # We're passing the empty lists, because we don't want to include previous candidates
-                        cache_multi, cache_single = collect_candidates(found_cache_results, cur_provider, [],
+                        cache_multi, cache_single = collect_candidates(cache_found_results, cur_provider, [],
                                                                        [], series_obj, down_cur_quality)
 
                 # For now we only search if we didn't get any results back from cache,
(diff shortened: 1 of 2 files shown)


medusa/search/core.py
```diff
@@ -906,6 +906,15 @@ def combine_results(multi_results, single_results):
     """Combine single and multi-episode results, filtering out overlapping results."""
     log.debug(u'Combining single and multi-episode results')
     result_candidates = []
+    return_single_results = []
+    return_multi_results = []
+
+    # Only return one result per episode for the single results and multi-result result_candidates
+    if single_results:
```
p0psicles: @medariox this is now added to go back from a list like: `[61, 52, 61, 11, 11, 61, 61]` to `[61, 52, 11]` (the numbers are now episode objects of course, which can be compared to one another.

medusa/search/core.py
```diff
@@ -951,16 +960,18 @@ def combine_results(multi_results, single_results):
...
                     log.debug(
                         u'A needed multi-episode result overlaps with a single-episode result'
                         u' for episode {0}, removing the single-episode results from the list',
                         ep_obj.episode,
                     )
-                    del single_results[i]
+                    del return_single_results[i]
 
-    return single_results + result_candidates
+    return return_single_results + return_multi_results
```
p0psicles: If this method does its work. A result for a specific shows episode, should only be in `return_single_results` *or* in `return_multi_results` but never in both. Which was the case now.

medusa/search/core.py
```diff
@@ -938,8 +947,8 @@ def combine_results(multi_results, single_results):
         # see how many of the eps that this result covers aren't covered by single results
         needed_eps = []
         not_needed_eps = []
-        for result in single_results:
-            if result in multi_result.episodes:
+        for result in return_single_results:
+            if result.episodes[0] in multi_result.episodes:
```
p0psicles: Previously you where comparing are TorrentSearch object with a list of episode objects. Which was always `False` of course.

medusa/search/core.py
```diff
@@ -792,6 +778,7 @@ def collect_candidates(found_results, provider, multi_results, single_results, s
                   if result in (SEASON_RESULT, MULTI_EP_RESULT))
     candidates = list(itertools.chain(*candidates))
     if candidates:
+        # FIXME: collect_multi_candidates can potentially result in downloading multiple multi-ep results.
```
medariox: This needs to be handled in collect_results as we need all single episode results and all multi episode results to determine the needed episode results.

medariox: And it should be already working like that.

medariox: Yes, because the filtering is done in combine_results as we need the single episode results.

p0psicles: I'm saying that for example, nyaa returns one release that's multi_ep. But after this method, it's duplicated? Why?

PR #5948, opened 2019-01-01
[WIP] Add initial pagination to the states getShows method.

p0psicles:

In this PR i've:
* Implemented pagination to get the list of shows using the apiv2
* Refactored show.vue so it doesn't use a computed anymore for the show. It now uses a watch on the shows state (lists with all shows). Every time that list changes, it refreshes the show.

Next step will be fully vueifying displayShow.


p0psicles:
@wimpyrbx I might need your help in testing this.

themes-default/slim/src/components/show.vue
```diff
@@ -72,7 +72,7 @@ export default {
 
             // Not detailed
             if (!show.seasons) {
-                $store.dispatch('getShow', { id, indexer, detailed: true });
+                $store.dispatch('getShow', { id, indexer, detailed: false });
```
p0psicles: @OmgImAlexis @sharkykh I need your help on this.
I'm trying to disable the detailed = true flag "for now". But it's causing an indefinite loop.
This made me try to understand how the show.vue component fits into all this. But I can't figure it out.

What i'm trying to accomplish is, to not use the series route with the detailed = true flag, to get the episodes, but to create alls on the /series/tvdb.../episodes route with the season flag, to get them per season.

themes-default/slim/src/store/modules/shows.js
```diff
@@ -81,20 +81,32 @@ const actions = {
      * @param {ShowParameteres[]} shows - Shows to get. If not provided, gets the first 10k shows.
      * @returns {(undefined|Promise)} undefined if `shows` was provided or the API response if not.
      */
-    getShows(context, shows) {
+    async getShows(context, shows) {
```
OmgImAlexis: `async` + `while` is likely causing this. 🤔

themes-default/slim/src/store/modules/shows.js
```diff
@@ -81,20 +81,32 @@ const actions = {
...
-                const shows = res.data;
-                return shows.forEach(show => {
-                    commit(ADD_SHOW, show);
-                });
-            });
+
+            const records = [];
+            let keepGoing = true;
+            while (keepGoing) {
+                const response = await api.get('/series', { params });
```
OmgImAlexis: I'd suggest instead look at the header you get back as it should have the pagination info in it. Instead of doing the while loop try this. The header should also have the total pages so you can then easily know the end result instead of relying on a boolean being toggled.

- Get Page 1
- `commit(ADD_SHOW, show)`
- Check pagination header
- call `api.get` again with new page param
- rinse and repeat

themes-default/slim/webpack.config.js
```diff
@@ -5,6 +5,7 @@ const VueLoaderPlugin = require('vue-loader/lib/plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const FileManagerPlugin = require('filemanager-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
+require('babel-polyfill');
```
OmgImAlexis: What's the need for the polyfill?

p0psicles: @omgimalexis it needed for the async

themes-default/slim/src/store/modules/shows.js
p0psicles: Then I'd still have a loop. A for loop instead of while.

Btw I got the infinite loop when I disabled the detailed flag in the show.vue component?

OmgImAlexis: A while loop can cause issues inside of an async function. 

Have a look at the answer from this.

https://stackoverflow.com/questions/49155139/while-loop-inside-async-await

p0psicles:
@omgimalexis can you give this another look?

themes-default/slim/src/components/show.vue
```diff
@@ -70,11 +73,12 @@ export default {
...
+            // With the current mako there is no reason to enable this IMO.
             // Not detailed
-            if (!show.seasons) {
-                $store.dispatch('getShow', { id, indexer, detailed: true });
-                return getShowById({ indexer, id });
-            }
+            // if (!show.seasons) {
+            //     $store.dispatch('getShow', { id, indexer, detailed: false });
+            //     return getShowById({ indexer, id });
+            // }
```
sharkykh: It is used by the "jump to season"
https://github.com/pymedusa/Medusa/blob/918cfe7ce01808499585ffe6c1356fd5ed369f76/themes-default/slim/views/partials/showheader.mako#L32-L56

themes-default/slim/package.json
```diff
@@ -26,7 +26,9 @@
       }
     }
   },
-  "dependencies": {},
+  "dependencies": {
+    "babel-polyfill": "6.26.0"
```
sharkykh: 1. This is an old package name, new name is `@babel/polyfill`
2. All dependencies should be `devDependencies` (added using `yarn add -D <pkg>`)
3. `@babel/polyfill` is already a `devDependency` (line 37)

themes-default/slim/webpack.config.js
sharkykh: According to [babel docs](https://babeljs.io/docs/en/babel-polyfill#usage-in-node-browserify-webpack) the proper way to add the polyfills is using `.babelrc`:
https://github.com/pymedusa/Medusa/blob/2d21ecd593d184f8fa5e45e07ae128e9ebc6a9b4/themes-default/slim/.babelrc#L3
change to:
```json
    ["@babel/preset-env", {
      "useBuiltIns": "usage"
    }]
```

Keep in mind that adding the polyfills is adding a big chunk of code, that might not actually be needed by the browsers we support...

themes-default/slim/src/store/modules/shows.js
```diff
@@ -84,17 +85,40 @@ const actions = {
...
+                const limit = 1000;
+                const page = 1;
+                const params = {
+                    limit,
+                    page
+                };
+
+                // Get first page
+                try {
+                    const response = await api.get('/series', { params });
```
sharkykh: Unlike Python, `response` will only be available within the `try { }` block.
If you need to access response outside of that block, you need to define `let response = null;` before the block, and assign the `await`-ed value inside the block.

themes-default/slim/src/store/modules/shows.js
```diff
@@ -84,17 +85,40 @@ const actions = {
...
+                const totalPages = Number(response.headers['x-pagination-total']);
+                response.data.forEach(show => {
                     commit(ADD_SHOW, show);
                 });
-            });
+
+                // Optionally get additional pages
+                for (let x = 2; x <= totalPages; x++) {
+                    params.page = x;
+                    api.get('/series', { params }).then(response => {
```
sharkykh: Disclaimer - I did make some of the changes I proposed with `@babel/polyfill` above, but nothing on to this part of the code.
I just wanted to let you know it doesn't seem to work for me.

image

themes-default/slim/yarn.lock
```diff
@@ -5,12 +5,10 @@
 "@ava/babel-plugin-throws-helper@^3.0.0":
   version "3.0.0"
   resolved "https://registry.yarnpkg.com/@ava/babel-plugin-throws-helper/-/babel-plugin-throws-helper-3.0.0.tgz#2c933ec22da0c4ce1fc5369f2b95452c70420586"
-  integrity sha512-mN9UolOs4WX09QkheU1ELkVy2WPnwonlO3XMdN8JF8fQqRVgVTR21xDbvEOUsbwz6Zwjq7ji9yzyjuXqDPalxg==
```
sharkykh: I believe your `yarn` version is out of date.

themes-default/slim/src/components/show.vue
p0psicles: Tnx allot! I'll look into it.

themes-default/slim/webpack.config.js
p0psicles: honestly i'm still confused why I can use await in vue files but not js?

sharkykh: > honestly i'm still confused why I can use await in vue files but not js?

No vue **files** currently use `async`/`await`. This is an important distinction, because vue & js files from the `themes-default/slim/src/` folder are transpiled using babel.
The places that currently use async/await are vue components in **mako** files.

```js
// themes-default/slim/views/addShows_addExistingShow.mako
  177,30:             const response = await apiRoute.post('addShows/addExistingShows', formData);

// themes-default/slim/views/addShows_newShow.mako
  291,30:             const response = await apiRoute.post('addShows/addNewShow', formData);
  378,34:                 const response = await api.get('internal/searchIndexersForShowName', config);

// themes-default/slim/views/config_notifications.mako
  356,30:             const response = await apiRoute('home/loadShowNotifyLists')
  364,30:             const response = await apiRoute('home/loadShowNotifyLists')
  381,30:             const response = await apiRoute('home/getPushbulletDevices', { params: { api: pushbulletApiKey }});
  406,30:             const response = await apiRoute('home/testPushbullet', { params: { api: pushbulletApiKey }});
  421,30:             const response = await apiRoute('home/testJoin', { params: { api: joinApiKey }});
  431,30:             const response = await apiRoute('home/twitterStep1');
  442,34:                 const response = await apiRoute('home/twitterStep2', { params: { key: twitter.key } });
  451,34:                 const response = await apiRoute('home/testTwitter');

// themes-default/slim/views/config_search.mako
  319,26:             const resp = await apiRoute.get('home/testTorrent', { params });
  337,26:             const resp = await apiRoute.get('home/testNZBget', { params });
  355,26:             const resp = await apiRoute.get('home/testSABnzbd', { params });

// themes-default/slim/views/vue-components/quality-chooser.mako
  154,30:             const response = await api.get(url);
  189,30:             const response = await api.post(url, { type: 'ARCHIVE_EPISODES' });
```

Basically what this means is the browsers support `async`/`await` natively, but if those components were converted to `vue` SFCs right now, you would hit the exact same error from babel.

That said, you can choose to disable the babel plugins that transpile `async` functions.
I'm **not sure how good** of a solution this is or if it's even recommended.

But change:
https://github.com/pymedusa/Medusa/blob/2d21ecd593d184f8fa5e45e07ae128e9ebc6a9b4/themes-default/slim/.babelrc#L3
To:
```json
    ["@babel/preset-env", {
      "exclude": [
        "transform-regenerator",
        "transform-async-to-generator"
      ]
    }]
```

themes-default/slim/src/store/modules/shows.js
p0psicles: Yeah i've fixed this. IT's also now using promises. So I can park the async/await discussion for now.

themes-default/slim/webpack.config.js
p0psicles: Tnx a million for the detailed explanation. I really appreciate it. Especially now that we're kinda dead in the water regarding the entire frontend. So i'm tryig to make the best of it.

sharkykh:
image

have you noticed this?

p0psicles:
No haven't seen that. I'll try to reproduce.

PR #5967, opened 2019-01-07
Fix progressbars (home)

sharkykh:
Fixes:
image => image


tests/providers/test_search.py
```diff
@@ -6,7 +6,7 @@
 import functools
 import os
 
-from tests.providers.conftest import providers
+from .conftest import get_provider_data
```
medariox: Any specific reason to make the import relative?

Issue #5184, opened 2018-09-10
Preferred words not working


cyberundertaker:
**Describe the bug**
Medusa not downloading the release according to preferred/undesired words

**To Reproduce**
1. Preferred words set as screenshot 1
2. At the time of snatching there was no release with preferred word so medusa snatched best release (convoy) (screenshot 2)
3. Later a better release (AMZN) was published and i did a force search later hoping medusa will snatch AMZN release
4. Medusa didn't snatch anything
5. Manual search finds the AMZN release (screenshot 3)

**Expected behavior**
After force search on episode medusa should recognize that one of the releases has preferred word in it and automatically download and replace the episode.

Medusa should have an option for automatic search for better release according to set criteria within 24-48 hours after snatching any release

**Screenshots**
https://imgur.com/a/2sx1DYy

**Medusa (please complete the following information):**
 - OS: [Ubuntu 14.04 x64]
Branch: master 
Commit: 4614efc77151ded92ef458a09dec39f8bd5acfc6 
Version: v0.2.10 

**Logs:**
<details>

```
2018-09-10 17:11:21 INFO     SEARCHQUEUE-FORCED-348841 :: [4614efc] Unable to find forced results for: Kidding - S01E01 - Green Means Go
2018-09-10 17:11:21 DEBUG    SEARCHQUEUE-FORCED-348841 :: [4614efc] 348841: Formatting pattern: %SN - S%0SE01 - %EN -> Kidding - S01E01 - Green Means Go
2018-09-10 17:11:21 DEBUG    SEARCHQUEUE-FORCED-348841 :: [4614efc] Found codec for Kidding Green Means Go
2018-09-10 17:11:21 DEBUG    SEARCHQUEUE-FORCED-348841 :: [4614efc] Parsed kidding.s01e01.internal.1080p.web.h264-convoy into title: kidding, season: 1, episode: [1], other: Internal, screen_size: 1080p, format: WEB-DL, video_codec: h264, release_group: convoy, type: episode, parsing_time: 0.0225958824158, total_time: 0.0226321220398, absolute_episode: [], quality: 1080p WEB-DL
2018-09-10 17:11:21 DEBUG    SEARCHQUEUE-FORCED-348841 :: [4614efc] Matched release kidding.s01e01.internal.1080p.web.h264-convoy to a series in your database: Kidding
[...]

PR #5987, opened 2019-01-11
Concept version of upgrading based on preferred words.

p0psicles:
Using a score configured per show.

POC for #5184


medusa/classes.py
```diff
@@ -16,14 +16,15 @@
 # You should have received a copy of the GNU General Public License
 # along with Medusa. If not, see <http://www.gnu.org/licenses/>.
 """Collection of generic used classes."""
-from __future__ import unicode_literals
+from __future__ import division, unicode_literals
```
p0psicles: rtms: division can be removed again.

medusa/classes.py
```diff
@@ -132,6 +133,122 @@ def __init__(self, episodes=None, provider=None):
...
+        preferred_words = [word.lower() for word in app.PREFERRED_WORDS]
+        self._preferred_words_score = round((len([word for word in preferred_words if word in self.name.lower()]) / float(len(preferred_words))) * 100)
+        return self._preferred_words_score
+
+    @staticmethod
+    def __qualities_to_string(qualities=None):
+        return ', '.join([Quality.qualityStrings[quality] for quality in qualities or []
+                          if quality and quality in Quality.qualityStrings]) or 'None'
+
+    def want_episode(self, season, episode, quality):
```
p0psicles: @medariox i've moved want_episode from the Series class to Result class. Imo it looks allot cleaner, with less parameter. And it just feels like this is a better home.

Issue #5648, opened 2018-11-05
Ignored words need to be changeable/setable on a show level


udochrist:
**Is your feature request related to a problem? Please describe.**
you cannot override "global" ignored words. when for 99% of the shows you don't want "language" in the title, but just for one you want to allow it you must add "language" to all 99% of the shows.

**Describe the solution you'd like**
When no ignored words are set at the show level use the global ones, otherwise the shows.

**Describe alternatives you've considered**
editing 99% of shows :-( which _really_ is a lot

billotronic:
To add to this it would be also handy to be able to set Require at a show level too. Say for getting the Uncensored versions of a show or maybe a different resolution etc.

p0psicles:
Ok I'll check this out

p0psicles:
It's not that easy to change the behavior. As that will change how it now works for allot of users.

udochrist:
Depends on the approach chosen to implement that. If you keep a delta list on the show level it should not change anything to current users.

medariox:
Added.

PR #6033, opened 2019-01-17
Feature/add ignored required exclude option

p0psicles:

This feature will add flags that change the behavior of the required words and ignored words fields for the tv_show configuration. Enabling the flag will make sure that the words provided in the shows configuration, will be excluded from the global configuration.

Will implement #5648 

### TODO:
- [x] Implement unit tests
- [x] Implement logic in Series.show_words() method
- [x] Add new options in db main.db tv_shows
- [x] Add new options in UI
- [x] Update api spec
- [x] Update api unit test
- [x] Update changelog

image


udochrist:
Looking at the GUI this seems ok. 
My use case is to remove a single word from the global ignore list on a single show.

udochrist:
hmmm. in medusa 3,o the GUI is missing the select boxes for excluded/ignored. is the final implementation. different from this feature?

medariox:
Which boxes are you referring to? Final implementation should be the same. Try a hard cache refresh with `CTRL + F5`.

udochrist:
YES. that fixed  it. after clearing the cache the toggle buttons appeared.

great work with 0.3.0 and thanks for the fast reply!

scott451:
hello. i looked at this new feature. in my install it looks like the global ignore words are used for ignore words and required words as well. though adding exceptions works correctly.

medariox:
@scott451 
Please open a new issue, this is not the right place.

PR #6172, opened 2019-02-05
update tmdb _request to avoid 429

matthewfiore:
new user with big library to import. getting lots of 429 errors when scanning existing folders.
found this and tested myself. no longer erroring out when scanning large collections.

### DIFF
--- a/lib/tmdbsimple/base.py
+++ b/lib/tmdbsimple/base.py
@@ -13,7 +13,7 @@ Created by Celia Oakley on 2013-10-31.
 
 import json
 import requests
-
+import time
 
 class APIKeyError(Exception):
     pass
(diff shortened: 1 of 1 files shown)


p0psicles:
@medariox what to do with this? I'm not 100% sure. But I think the lib is already monkey patched. 
@matthewfiore did you try  getting it merged with tmdbsimple project? We actually want to get rid of all the monkey patched libs.

p0psicles:
@medariox what to do with this? I'm not 100% sure. But I think the lib is already monkey patched. 
@matthewfiore did you try  getting it merged with tmdbsimple project? We actually want to get rid of all the monkey patched libs.

matthewfiore:
ok, i found what looks like the lib that medusa is using and submitted the same pull request on that repo.
https://github.com/celiao/tmdbsimple/pull/56
hope dev is active.

p0psicles:
Looking at other pr's I doubt you'll get reaction soon. We make use of the  requests module. You could look into the possibility of writing a requests hook.

Thats only if tmdbsimple has the option of using an external requests session. I'm not sure if we do that.

matthewfiore:
ok, i see now.
thanks for looking into this.
i'm excited to be using medusa even more now that i see how active you are.

what can I do now to help get this resolved?

medariox:
@matthewfiore 
The best way is to make pressure over at https://github.com/celiao/tmdbsimple
It's in the interested of everyone to get this issue fixed upstream. Thanks!

PR #6225, opened 2019-02-18
Fix snatch selection loading detailed show

sharkykh:


p0psicles:
What is the bug here?

sharkykh:
Sorry for not providing context. There were two `getShow` calls. The one on `created()` was the first to be called and it was getting the detailed data, and the one on `mounted()` should be the only one, as it is getting the non-detailed data.

I found out about this because I kept seeing "Size: #.## GB" even though you moved size to be under the detailed flag in `Series.to_json()`.

PR #6285, opened 2019-03-09
Fix username retrieval inside Docker container

jpiron:
When running inside a Docker container with the `--user` Docker run option,
username might not exist. See: https://docs.docker.com/engine/reference/run/#user.
Fallback on the uid in this case.

### DIFF
--- a/medusa/__main__.py
+++ b/medusa/__main__.py
@@ -969,6 +969,10 @@ class Application(object):
             try:
                 import pwd
                 app.OS_USER = pwd.getpwuid(os.getuid()).pw_name
+            except KeyError:
+                # In the case of a usage with Docker `user` feature, the username might not exist.
+                # See: https://docs.docker.com/engine/reference/run/#user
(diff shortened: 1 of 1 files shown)


medariox:
@jpiron 
This is a great first contribution, thank you! There seems to be something wrong with Travis atm, the PR will be merged as soon as we get that sorted.

medariox:
@jpiron 
I wrote Travis CI about your issue and this is their reply:

> Thanks for reaching out and I'm sorry for the troubles.
> 
> Looking at the account (jpiron), it seems that they've never logged into Travis CI and this is why we are currently blocking their requests. Sorry for the inconvenience.
> 
> To confirm, do you think jpiron could try to log into Travis CI at least once at https://travis-ci.org/ and push a new commit?
> 
> Please let us know the result and we would be happy to have another look if it's still not working.

Do you think this is something you could do? Please let me know.

medariox:
We did it 👍

PR #6424, opened 2019-03-29
Docker image support Python3.6

a10kiloham:
Not sure this is ready for primetime unless we feel 3.6 support is. I don't notice much of a speed change but feel free to test and send notes. This probably could be more elegantly refactored into a tag for Docker to choose :latest or :python3 or :python2 but that's beyond me at the moment time wise :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,20 @@ ARG BUILD_DATE
 ARG VERSION
 LABEL build_version="Version:- ${VERSION} Build-date:- ${BUILD_DATE}"
 
+RUN apk add --no-cache python3 && \
+    python3 -m ensurepip && \
+    rm -r /usr/lib/python*/ensurepip && \
+    pip3 install --upgrade pip setuptools && \
+    if [ ! -e /usr/bin/pip ]; then ln -s pip3 /usr/bin/pip ; fi && \
+    if [[ ! -e /usr/bin/python ]]; then rm /usr/bin/python; fi && \
+    ln -sf /usr/bin/python3.6 /usr/bin/python && \
+    rm -r /root/.cache
+
 # install packages
 RUN \
  apk add --no-cache \
-	--repository http://nl.alpinelinux.org/alpine/edge/community \
-	mediainfo gdbm py-gdbm
+        --repository http://nl.alpinelinux.org/alpine/edge/community \
+        mediainfo
 
 # install app
 COPY . /app/medusa/


sharkykh:
Just a thought...
Have you considered using [`lsiobase/alpine.python3`](https://hub.docker.com/r/lsiobase/alpine.python3) instead of installing `python3` on top of `python2` present in the current `lsiobase/alpine.python` image?

a10kiloham:
That seems like a good idea. I'll try it soon and see how it works. Probably more efficient-I didn't know that was there.

p0psicles:
Isn't there a way of using py3.7? As with 3.7 there could be some performance gains.

a10kiloham:
i don't see an alpine image yet for that. is there a better base distro for docker you know of? I just always use alpine because it's what i'm familiar with

sharkykh:
I think 3.6 is fine for now. And because it's Docker, users can easily get the update to 3.7 later.

> i don't see an alpine image yet for that. is there a better base distro for docker you know of? I just always use alpine because it's what i'm familiar with

There isn't a LinuxServerIO image,
but there is `python:3.7-alpine3.9` ([from the official python image](https://hub.docker.com/_/python/)), which is still based on Alpine 3.9.
However, it may have some missing dependencies ([some dependencies are added on the `lsiobase/alpine.python3` layer](https://github.com/linuxserver/docker-baseimage-alpine-python3/blob/master/Dockerfile)).

If you ask me, I think the _even better_ course of action is to make Medusa's Dockerfile based on `python:3.7-alpine3.9` and then add the stuff/dependencies that are needed.

sharkykh:
I'm still very new to Docker, but maybe something like this:
https://github.com/pymedusa/Medusa/compare/develop...feature/docker

It works (server starts up), not sure if anything else is needed.
I added more folders to `.dockerignore` just to make it a little lighter.

Not sure why the database says version `.0`, it should be `44.14`, but maybe it's an unrelated bug.

Here are some of the commands I used:
```sh
# Build image
docker build --tag medusa:new-docker

# Run once and remove when done (interactive)
docker run --rm -p 8081:8081 --name medusa01 -it medusa:new-docker

# Run once and remove when done (detached)
docker run --rm -p 8081:8081 --name medusa01 -d medusa:new-docker

# Shell into container
docker exec -it medusa01 /bin/sh
```

image

a10kiloham:
@sharkykh gdbm and py-gdbm are required for some post processors. please don't yank out things unless you're sure. linux-headers and GCC probably can go though and the other bugs you note are fine to remove.

sharkykh:
@a10kiloham I only removed them because you did.
Linux-headers and gcc are needed for installing psutil. They are cleaned from the image

a10kiloham:
@sharkykh oh my bad sorry, just glanced at the diffs and vaguely recalled removing things before :)

a10kiloham:
FWIW it seems to be working fine w/o errors now so that's a plug. Just need to figure out the best Docker image etc.

medariox:
Any chance we could get this concluded anytime soon?
Using `lsiobase/alpine.python3:3.9` + `mediainfo` sounds like the way to go.

sharkykh:
Closing in favor of #6852

PR #6463, opened 2019-04-03
Fix TNTVillage search

Superpiffer:
For some reason right now there are two "showrelease_tb" classes in the page, so maybe it's better to select the one with a meaningful <td> label.

- [X ] PR is based on the DEVELOP branch
- [X ] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [X ] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/medusa/providers/torrent/html/tntvillage.py
+++ b/medusa/providers/torrent/html/tntvillage.py
@@ -96,11 +96,8 @@ class TNTVillageProvider(TorrentProvider):
         items = []
 
         with BS4Parser(data, 'html5lib') as html:
-            torrent_table = html.find('div', class_='showrelease_tb')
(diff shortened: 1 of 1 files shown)


Superpiffer:
Sorry, I'm not a python expert and I forgot to try with python2. Maybe someone can provide a more elegant solution.

medariox:
Thanks for the bug report! You are right, they have added an additional `showrelease_tb` div and it's not possible to determine which one is the one we want to use. The best approach for now would be to just use the last `showrelease_tb`. It can be done with using `.find_all()` and selecting the last item `[-1]`.

PR #6537, opened 2019-04-19
Add nCore torrent provider

medariox:

Big thanks to @ExPeacer for providing access to the provider.


medusa/providers/torrent/json/ncore.py
```diff
@@ -0,0 +1,176 @@
...
+from medusa.providers.torrent.torrent_provider import TorrentProvider
+
+from requests.utils import dict_from_cookiejar
+
+log = BraceAdapter(logging.getLogger(__name__))
+log.logger.addHandler(logging.NullHandler())
+
+
+class NcoreProvider(TorrentProvider):
+    """Norbits Torrent provider."""
```
p0psicles: Norbits..

medusa/providers/torrent/json/ncore.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # Proper Strings
+        self.proper_strings = ['PROPER', 'REPACK', 'REAL', 'RERIP']
+
+        # Miscellaneous Options
+
+        # Cache
+        self.cache = tv.Cache(self, min_time=20)
+
+    def search(self, search_strings, age=0, ep_obj=None, **kwargs):
```
p0psicles: You can remove age and ep_obj. I don't think you need to put *args. The **kwargs should pick them up

medariox: It just shadows how it is implemented in generic_provider. If we need to change it, we should change it for every other provider as well.

PR #6657, opened 2019-05-07
New url for Yggtorrent provider

Wqntedpw:
Since 2 days Yggtorrent have a new domain (.ch) and .to is unreachable

### DIFF
--- a/medusa/providers/torrent/html/yggtorrent.py
+++ b/medusa/providers/torrent/html/yggtorrent.py
@@ -36,7 +36,7 @@ class YggtorrentProvider(TorrentProvider):
         self.password = None
 
         # URLs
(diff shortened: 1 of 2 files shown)


medariox:
Thanks for the PR! Actually the .to domain should work just fine. Our code was not meant to handle a triple redirect, that got fixed with https://github.com/pymedusa/Medusa/pull/6656. That doesn't mean we can't update the domain now.

Wqntedpw:
> Thanks for the PR! Actually the .to domain should work just fine. Our code was not meant to handle a triple redirect, that got fixed with #6656. That doesn't mean we can't update the domain now.

Hello, I just realized it was dns block from my isp.

PR #6679, opened 2019-05-13
Update cfscrape to v2.0.0-Custom

medariox:

Also updates Js2Py to v0.63


sharkykh:
I don't know if you had a chance to test it, but looks like it the newest CF challenge can't be solved using Js2Py. The problem seems to be `String.prototype.italics`, or as it's used in the challenge code:
```js
Function("return escape")()(("")["italics"]())[2]
```
It's just not implemented.

medariox:
Yes, I've tested it. It seems to work just fine for me (tested with `http://itorrents.org`).

sharkykh:
Ok. I had tested the new code against a saved html file of a CF landing page ("checking your... etc) and it didn't work for me. And all the regexp patterns fit correctly, just the actual js execution raised an exception

medariox:
@sharkykh 
Maybe CF uses different landing pages? I would say let's merge this and watch for any any issues.

Issue #6021, opened 2019-01-16
Failed to modify date of episode


Shootersss:
**Describe the bug**
I got a warning that Medusa wasn't able to mody the date of an Episode.

**To Reproduce**
1. Go to 'config/postProcessing/'
2. Scroll down to 'Change File Date' and Enable it
3. See error

**Expected behavior**
The file date of the episode should be changed to Jan 15, 2019 (22:00)

**Medusa (please complete the following information):**
 - OS: [Synology, DS216play, DSM 6.2.1-23824 Update 4]
 - Branch: [master]
 - Commit: [latest]

**Logs:**
<details>

```
2019-01-16 16:26:03 DEBUG    Thread_8 :: [918cfe7] 350665: About to modify date of /volume1/TV-Series/The Rookie/Season 01/S01E10 - The Rookie - Flesh and Blood - 720p WEB-DL - BAMBOOZLE.mkv to series air date Jan 15,2019 (22:00)
2019-01-16 16:26:03 WARNING  Thread_8 :: [918cfe7] 350665: Failed to modify date of S01E10 - The Rookie - Flesh and Blood - 720p WEB-DL - BAMBOOZLE.mkv to series air date Jan 15,2019 (22:00)
```
</details>

ThomasDK81:
Gettting this too.

-     OS: [FreeNAS 11.2 U3 in iocage with 11.2-release]
-     Branch: [master]
-     Commit: [latest]

TheHolyRoger:
@ThomasDK81 @Shootersss please see the PR above.

The issue here is that we need CAP_FOWNER capability for the medusa process.

Linux permissions for UTIME and CHOWN are disabled for non-owners even if they have the correct group permissions.

Either have everything running under the same owner, or grant CAP_FOWNER capability to the medusa process. The PR above only applies to systemd, other methods need to be used if you're not using systemd.

potyt:
One question I have, if you subsequently manually fix the file permissions/ownership, will Medusa know to revisit the file and try and update the date/time on the file once it's already in it's final destination directory?
If not, any way to safely nudge it to do that?

PR #6699, opened 2019-05-19
CAP_FOWNER capability for utime

TheHolyRoger:
Fix for changing timestamps on downloads not owned by the Medusa process when started in Systemd.

Causes "permission denied" when trying to modify airdates

Relevant issue: #6021


medariox:
`CapabilityBoundingSet` will limit the capabilities, so it can be removed. All that's needed is `AmbientCapabilities`.

PR #6703, opened 2019-05-20
Fix API v2 exception on incorrect API key. Fixes #5425

### LINKED ISSUE #5425: API errors should always be JSON
I'm currently getting this when trying to use the API without a valid key. This should be returning JSON even if I'm not asking for it.

```
  File "/Users/xo/code/Medusa/ext/tornado/web.py", line 1571, in _execute
    result = yield result
  File "/Users/xo/code/Medusa/ext/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/Users/xo/code/Medusa/ext/tornado/concurrent.py", line 261, in result
    raise_exc_info(self._exc_info)
  File "/Users/xo/code/Medusa/ext/tornado/gen.py", line 1221, in handle_yield
    self.future = convert_yielded(yielded)
  File "/Users/xo/code/Medusa/ext/singledispatch.py", line 210, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/xo/code/Medusa/ext/tornado/gen.py", line 1357, in convert_yielded
    return multi(yielded)
  File "/Users/xo/code/Medusa/ext/tornado/gen.py", line 773, in multi
    return multi_future(children, quiet_exceptions=quiet_exceptions)
  File "/Users/xo/code/Medusa/ext/tornado/gen.py", line 868, in multi_future
    children = list(map(convert_yielded, children))
  File "/Users/xo/code/Medusa/ext/singledispatch.py", line 210, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/xo/code/Medusa/ext/tornado/gen.py", line 1363, in convert_yielded
    raise BadYieldError("yielded unknown object %r" % (yielded,))
BadYieldError: yielded unknown object u'No authorization token.'
```

sharkykh:
Fixes #5425, added 2 small tests for that
Fixes RESTful `HTTPError`s causing real exceptions.

### DIFF
--- a/medusa/server/api/v2/base.py
+++ b/medusa/server/api/v2/base.py
@@ -154,8 +154,8 @@ class BaseRequestHandler(RequestHandler):
         exc_info = kwargs.get('exc_info', None)
 
         if exc_info and isinstance(exc_info[1], HTTPError):
-            error = exc_info[1]
-            response = self.api_response(status=status_code, error=error.reason)
+            error = exc_info[1].log_message or exc_info[1].reason
+            response = self.api_response(status=status_code, error=error)
         elif app.DEVELOPER and exc_info:
             self.set_header('content-type', 'text/plain')
             self.set_status(500)
--- a/medusa/server/api/v2/base.py
+++ b/medusa/server/api/v2/base.py
@@ -25,10 +25,9 @@ from medusa.logger.adapters.style import BraceAdapter
 from six import iteritems, string_types, text_type, viewitems
 
 from tornado.gen import coroutine
-from tornado.httpclient import HTTPError
 from tornado.httputil import url_concat
 from tornado.ioloop import IOLoop
-from tornado.web import RequestHandler
+from tornado.web import HTTPError, RequestHandler
 
 log = BraceAdapter(logging.getLogger(__name__))
 log.logger.addHandler(logging.NullHandler())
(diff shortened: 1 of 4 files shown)


sharkykh:
Working on implementing a couple of tests, and another small bug:
```
  File "C:\Dev\Medusa\ext\tornado\web.py", line 1559, in _execute
    raise HTTPError(405)
tornado.web.HTTPError: HTTP 405: Method Not Allowed
```

p0psicles:
Ooh then ignore my approval

sharkykh:
Sorry, took me a while to figure all of this stuff out. I think the code for the multi-threaded API looks much better now.
For the record -
#5425 - The issue was with the `prepare` method.
`HTTPError` - A combination of things, I rewrote the error handling code. Like I said above, should be much better now.

sharkykh:
All good now.

medusa/server/api/v2/base.py
```diff
@@ -90,13 +86,11 @@ def async_call(self, name, *args, **kwargs):
...
-            raise HTTPError(405, '{name} method is not allowed'.format(name=name.upper()))
+            raise HTTPError(405)
 
         def blocking_call():
-            try:
-                return method(*args, **kwargs)
-            except Exception as error:
-                self._handle_request_exception(error)
+            result = self._check_authentication()
+            return method(*args, **kwargs) if result is None else result
```
p0psicles: Your kinda asuming that the method call cant raise exceptions?

sharkykh: I'll try to explain as best I can.
Basically - yes.
But the difference is the `RequestHandler` class has an exception handler, and we handle everything in `write_error` now.
https://github.com/pymedusa/Medusa/blob/91140e1694a2264affb964e9ffefb729018538f6/medusa/server/api/v2/base.py#L151-L165

`HTTPErrors` are handled as RESTful responses,
And anything other than `HTTPError` is handled as an actual exception and sends "Internal Server Error" as the response (or the traceback, if `app.DEVELOPER` is enabled, like before)

sharkykh: >Your kinda asuming that the method call cant raise exceptions?

Sorry, I actually misread that.
Quite the opposite, I plan on it, but as I explained in the comment above, it's all being handled

sharkykh:
Retry Deepcode

PR #6707, opened 2019-05-20
Further improvements to API v2 multithreading

sharkykh:
- Better exception logging
- Async code is more DRY and more dynamic,
as the async function is bound for every request.

### DIFF
--- a/medusa/server/api/v2/base.py
+++ b/medusa/server/api/v2/base.py
@@ -57,6 +57,9 @@ class BaseRequestHandler(RequestHandler):
 
     def _check_authentication(self):
         """Check if JWT or API key is provided and valid."""
+        if self.request.method == 'OPTIONS':
+            return
+
         api_key = self.get_argument('api_key', default=None) or self.request.headers.get('X-Api-Key')
         if api_key and api_key == app.API_KEY:
             return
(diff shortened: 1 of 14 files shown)


sharkykh:
Retry Deepcode

sharkykh:
Retry Deepcode

sharkykh:
@medariox @p0psicles 
This is what I wanted to have from the beginning.
Please let me know how you feel about this.
I tried to document it as best I can, and I believe the code is also easier to read.

medusa/server/api/v2/base.py
```diff
@@ -81,72 +114,21 @@ def _check_authentication(self):
...
-        content = self.async_call('patch', *args, **kwargs)
-        if content is not None:
-            content = yield content
-        if not self._finished:
-            self.finish(content)
+        This function is called for each request.
+        """
+        name = self.request.method.lower()
+        # Wrap the original method with the code needed to run it asynchronously
+        method = make_async(self, getattr(self, name))
```
sharkykh: By the way, this means `OPTIONS` will also be asynchronous.
It doesn't mean much, I'm just saying it because currently it is synchronous.

PR #6723, opened 2019-05-24
Replace `xo` with `eslint`

sharkykh:


sharkykh:
I'm actually not sure whether to add `--cache` to the `lint` script or not. XO had (its own?) caching enabled by default. If things feel slow we can always add it later.

PR #6727, opened 2019-05-25
Add subtitle search SFC

p0psicles:
- [x] Add jest snapshot test

Dependent on #6726


p0psicles:
@sharkykh can you look at this one?

sharkykh:
Yes I will do that tomorrow.
One thing I can already say is that we probably have to update our own `date-fns` to the alpha version because if we don't we'll be bundling 2 versions until that version is out..

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+</template>
+<script>
+
+import { mapState } from 'vuex';
+import { VueGoodTable } from 'vue-good-table';
+
+export default {
+    name: 'subtitle-search',
+    components: {
+        VueGoodTable
```
sharkykh: It's a good idea to import all the components you use. In the future we might not have `StateSwitch` registered globally.
```suggestion
        StateSwitch,
        VueGoodTable
```

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+        ...mapState({
+            config: state => state.config
+        })
+    },
+    mounted() {
+        this.displayQuestion = true;
+    },
+    destroyed() {
+        // Remove the element from the DOM
+        this.$el.parentNode.removeChild(this.$el);
```
sharkykh: 😬 
Okay... two things.
1. **I think** that when a component gets destroyed, it means it removes anything it rendered.
	So... you shouldn't need to do that. If you encountered an issue with that logic, let me know and we'll try to work around it.

2. I would not do it (hide/remove the component) this way.
	Vue uses a VDOM, as in *Virtual* DOM.
	If you remove an element that is tied to a component or a state, my guess is as soon as the state changes, or when the component needs to update the DOM, it will just re-create it.

	My suggestion?
	This component should emit an event to let its parent know that it is no longer "needed",
	and the parent component should use `v-if="subtitleSelectionVisible"` on this component.
	This will trigger an automatic (and more reliable IMO) destruction of this component.
	Plus it makes this component more detached from its parent, in the sense that it doesn't manipulate it.

P.S.
Actually I look at this and I think of how complex the DOM manipulation is on [`ajax-episode-subtitles.js`](https://github.com/pymedusa/Medusa/blob/develop/themes-default/slim/static/js/ajax-episode-subtitles.js)/[`ajax-episode-search.js`](https://github.com/pymedusa/Medusa/blob/b1a1a86ed02787ee1a365b2356c5031b1bb582fc/themes-default/slim/static/js/ajax-episode-search.js), and we don't want that, do we? 😄

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+                });
+        },
+        manualSearch() {
+            const { show, season, episode } = this;
+
+            this.displayQuestion = false;
+            this.loading = true;
+            this.loadingMessage = 'Searching for subtitles... ';
+            const url = `home/manualSearchSubtitles?indexername=${show.indexer}&seriesid=${show.id[show.indexer]}&season=${season}&episode=${episode}`;
+            apiRoute(url) // eslint-disable-line no-undef
```
sharkykh: Same as above

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+    destroyed() {
+        // Remove the element from the DOM
+        this.$el.parentNode.removeChild(this.$el);
+    },
+    methods: {
+        autoSearch() {
+            const { episode, season, show } = this;
+
+            this.displayQuestion = false;
+            const url = `home/searchEpisodeSubtitles?indexername=${show.indexer}&seriesid=${show.id[show.indexer]}&season=${season}&episode=${episode}`;
```
sharkykh: Let `Axios` handle the heavy lifting.
```suggestion
            const params = {
                indexername: show.indexer,
                seriesid: show.id[show.indexer],
                season,
                episode
            };
```
Also, I suggest that instead of doing that *every time* you need the `params`,
just make a computed property that returns this object, and use it in all three locations.

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+                });
+        },
+        pickSubtitle(subtitleId) {
+            // Download and save this subtitle with the episode.
+            const { show, season, episode } = this;
+
+            this.displayQuestion = false;
+            this.loadingMessage = 'downloading subtitle... ';
+            this.loading = true;
+            const url = `home/manualSearchSubtitles?indexername=${show.indexer}&seriesid=${show.id[show.indexer]}&season=${season}&episode=${episode}&picked_id=${subtitleId}`;
```
sharkykh: Same as above, but with a note:
Let's say you follow my suggestion and make a computed property for the URL parameters.
Here you have an extra parameter - `picked_id`, so you extend the computed property's value:
```suggestion
            const params = {
                ...episodeParams, // This is the computed property
                picked_id: subtitleId // eslint-disable-line camelcase
            };
```

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+                    }
+                })
+                .catch(error => {
+                    console.log(`Error trying to search for subtitles. Error: ${error}`);
+                })
+                .finally(() => {
+                    // Destroy this component.
+                    this.loadingMessage = '';
+                    this.loading = false;
+                    this.$destroy();
```
sharkykh: As I explained above in the `destroyed` section, I don't think you should call `$destroy` on an element.
Plus, the docs say it's not preferable.
https://vuejs.org/v2/api/#vm-destroy
> In normal use cases you shouldn’t have to call this method yourself. Prefer controlling the lifecycle of child components in a data-driven fashion using `v-if` and `v-for`.

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+                });
+        }
+    }
+};
+</script>
+<style>
```
sharkykh: You should probably use `<style scoped>`.
I assume the styling doesn't apply to the `vue-good-table` child component when you do?
Try https://vue-loader.vuejs.org/guide/scoped-css.html#child-component-root-elements

For example, you could change:
```css
table.subtitle-table tr {
```
to
```css
.subtitle-search-wrapper >>> table.subtitle-table tr {
```

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+.v--modal-overlay .v--modal-box {
```
sharkykh: Remove this?

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+            <div class="question">
+                <p>Do you want to manually pick subtitles or let us choose it for you?</p>
+            </div>
+            <div class="options">
+                <button type="button" class="btn-medusa btn-info" @click="autoSearch">Auto</button>
+                <button type="button" class="btn-medusa btn-success" @click="manualSearch">Manual</button>
+            </div>
+        </div>
+
+        <vue-good-table v-if="subtitles.length"
```
sharkykh:
```suggestion
        <vue-good-table v-if="subtitles.length > 0"
```
For clarity that `subtitles` is an array

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
+<template>
+<!-- template for the subtitle-search component -->
+<tr class='subtitle-search-wrapper'>
```
sharkykh: I think the root element should have an indent (makes it easier to read the template)

themes-default/slim/src/components/subtitle-search.vue
```diff
@@ -0,0 +1,258 @@
...
+    <td colspan='9999' transition="expand">
```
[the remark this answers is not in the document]
p0psicles: 2. It's the only of having the td behaving in the <tr> full with. If I don't set it, it will limit the with the episodes table first column.

Moving the hack to a parent node, doesn't full any less hacky to me, tbh? It's more a CSS problem. I don't want to make it a `vue problem`.

themes-default/slim/src/components/subtitle-search.vue
p0psicles: Except it's used as a dynamic mounted component. So it's setup at runtime within a specific table (episode) row. I need to destroy the component. But i'll try if a bare `this.$destroy` is sufficient.

p0psicles: A little bit of context. This is how i'm creating the component in display-show.vue
```
const { id, indexer, getEpisodes, show, subtitleSearchComponents } = this;
            const SubtitleSearchClass = Vue.extend(SubtitleSearch);  // eslint-disable-line no-undef
            const instance = new SubtitleSearchClass({
                propsData: { show, season, episode, key: rowIndex },
                parent: this
            });

            // Update the show, as we downloaded new subtitle(s)
            instance.$on('update', event => {
                // This could be replaced by the generic websocket updates in future.
                if (event.reason === 'new subtitles found') {
                    getEpisodes({ id, indexer, season });
                };
            });

            
        // This was just added, to try to move the destroying to the calling component. That could provide for some more options when using the components. But else I don't see much difference compared to destroying it from within the component.
        instance.$on('close', vm => {
                // Close the component.
                vm.$destroy();
            });

            const node = document.createElement('div');
            this.$refs['table-seasons'].$refs[`row-${rowIndex}`][0].after(node);
            instance.$mount(node);
            subtitleSearchComponents.push(instance);
        },
```

sharkykh: What I still don't understand is, why do you need to create the component like that?

themes-default/slim/src/components/subtitle-search.vue
p0psicles: no, I do need to overwrite the vue-good-table styling. Like text color.
For example these two are important:
```
.v--modal-overlay .v--modal-box {
    overflow: inherit!important;
}
table.subtitle-table tr {
    background-color: rgb(190, 222, 237);
}
```

sharkykh: Okay, then please try that `>>>` selector with `<style scoped>`. According to the docs it should let you do exactly that, while still keeping the component styles in a scope.

p0psicles: never mind. Your suggestion worked. I'm using that.

themes-default/slim/src/components/subtitle-search.vue
p0psicles: image
because I only want to create the component after I click on the subtitle button, in the episode row. If a show as 400 episodes. I don't want to create 400 subtitle modals, if I might not even need them. Now I create it, and it will slide in the table.

themes-default/slim/src/components/subtitle-search.vue
p0psicles: it's really needed. I can call $destory() as much as I want. But it won't remove the node.
I really need to have it removed. I tested, it won't recreate it.

themes-default/slim/src/components/subtitle-search.vue
sharkykh: A simple `v-if` on the component would prevent that. `v-if` only creates / renders the component when the expression in it evaluates to true.
If I understand you correctly, then all you need to have is (on the parent component), something like an array that holds episode slugs, that you add a slug of an episode the user clicked on the subtitle search.... 
Eh. I'm tired right now but maybe a code will do better:
```js
data() {
    return {
        subtitleSearchVisible: []
    };
},
methods: {
    closeSubtitleSearch(slug) {
        const index = this.subtitleSearchVisible.indexOf(episode.indentifier);
        if (index > -1) {
            this.$delete(this.subtitleSearchVisible, index);
        }
    }
}
```
```vue
<!-- episode row (inside a v-for loop) -->
<a @click="searchSubtitles(episode.identifier)"><img ... /></a>
<subtitle-search v-if="subtitleSearchVisible.includes(episode.identifier)" @close="closeSubtitleSearch(episode.identifier)"></subtitle-search>
```

p0psicles: I'm pretty much done for now. I'll try that in the coming days

p0psicles: this:
```
<!-- episode row (inside a v-for loop) -->
<a @click="searchSubtitles(episode.identifier)"><img ... /></a>
<subtitle-search v-if="subtitleSearchVisible.includes(episode.identifier)" @close="closeSubtitleSearch(episode.identifier)"></subtitle-search>
```
is not possible. As i'm using vue-good-table. We don't have a v-for loop.
And i'm currently injecting a <tr></tr> after the episodes row.

Any other ideas?

p0psicles:
@sharkykh can you give this another look?

PR #4104, opened 2018-04-25
convert viewlogs to vue

OmgImAlexis:
This more or less converts the whole page over. There are a few places that're using Python vars that I'd like to replace in the review.

PR #6738, opened 2019-05-28
Vueify log viewer (and use Backstretch)

sharkykh:
Supersedes #4104

**It's basically done!**
Looks like I just need to fix the tests.
_Edit:_ Also push [the fix for this bug](https://i.imgur.com/tP21MW2.jpg).

- I added log polling (updates the page every 1.5 seconds), with a button to pause the refreshing.
- Backstretch component was not being used, fixed it and hopefully got it going with a temporary API variable for a random show slug. It should ultimately be from the list of shows in the store, but they're really available right now.


sharkykh:
Retry Deepcode

sharkykh:
image
Need to fix that (when selecting `Thread` as filter)

p0psicles:
nice PR!

sharkykh:
Many thanks

Issue #5311, opened 2018-09-26
Loading home page seems very slow


h3llrais3r:
**Describe the bug**
I can be wrong, but since the application becomes more and more vueified, I have the impression that page loading is becoming slower.
For example: loading my home page takes more than 15s while in the past this was a lot faster.
Is this normal?

If you need more info/logs, please let me know and I'll provide it.

**Screenshots**

![image](https://user-images.githubusercontent.com/1945295/46089725-73bcf100-c1af-11e8-84a0-0db55ab42f7d.png)

**Medusa (please complete the following information):**
- OS: Windows-8-6.2.9200
- Branch: master 
- Commit: 4614efc77151ded92ef458a09dec39f8bd5acfc6 
- Version: v0.2.10 
- Database: 44.12

OmgImAlexis:
@p0psicles this is part of the reason I was suggesting [vuex-persistedstate](https://github.com/robinvdvleuten/vuex-persistedstate). I think we should look at caching the shows array and having it update on load but in the background. Since new shows are just shown with a `loading...` right now anyways I don't see this being a huge issue.

@h3llrais3r would you mind checking the network tab, the rendering of vue isn't the issue it's the network requests, we currently block the whole UI from rendering while we fetch a large resource. I have a feeling if we moved this to a service worker([polyfill](https://github.com/mathiasbynens/cache-polyfill) and just loaded a cached list from the first load we'd have a lot better results.

For example try running a new copy of medusa seperate from your main install and look at how quick the page load without any shows.

p0psicles:
15s is a long time. Lets first analyze what does this.
@h3llrais3r can you give some more info?
System your running on. Number of shows. Any debug logs at time of load? 
Maybe check your systems resource usage at time of load. Cpu, mem, swapping?

h3llrais3r:
@OmgImAlexis, @p0psicles 
Running it on my dev machine (latest develop branch) with only 2 anime shows and 1 normal show, it loads really fast.

![image](https://user-images.githubusercontent.com/1945295/46128130-e3c08b00-c232-11e8-8e90-2f6dd7060c56.png)

I notice the slow loading on my server installation:
![image](https://user-images.githubusercontent.com/1945295/46128238-35691580-c233-11e8-87c7-a5a0c434ec82.png)

@OmgImAlexis, these are the details of the network tab during a refresh of the page:
[...]

PR #6740, opened 2019-05-29
Optimize show_statistics()

sharkykh:
I don't know if and how much it will help for #5311 (in the meantime at least)

### DIFF
--- a/medusa/server/api/v2/stats.py
+++ b/medusa/server/api/v2/stats.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 from datetime import date
+from textwrap import dedent
 
 from medusa import db
 from medusa.common import (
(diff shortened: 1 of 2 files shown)


rafi-d:
I've modified only medusa/server/web/home/handler.py . Sadly, no real change in load time of main page (  #5311 ) . About the same ~20 seconds on a slow C2D PC. 

Test clip:
https://drive.google.com/file/d/13z6O6vPUOsVY5ZDT-QSBbPpSmEhLBobM/view?usp=sharing

FYI

PS: still on older 0.29 (modified) waiting on a fix on show->manual search GUI.

Medusa Info: 
 Branch:                 master
Commit:                 54da0c77d705a23f1ce67945ba4af3e1b978dcfc                  
Version:                 Unknown                  
Database:                 44.13
-- | --
Python Version: | 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)]
SSL Version: | OpenSSL 1.0.2j  26 Sep 2016
OS: | Windows-10-10.0.17763
Locale: | en_US.cp1255
  |

sharkykh:
@rafi-d 	Thanks.

medariox:
Retry DeepCode

rafi-d:
> 
> 
> Besides the possibly moderate performance improvement, this is a nice code quality improvement. We should probably add a dedicated table about the stats to DB if we want to have an efficient way to retrieve them.

Sadly, as you can see in the test-clip, there is no speed improvement... 
Just a thought: why not keep a "cache"-copy of the last viewed screen, and just immediately display it (with some "pending update" message) ,  while you run an update job for it (like today) in the background, and refresh it when done?  

Most of the times, the user just wants to click-navigate to  some show... Maybe you can do the same thing in the show view page (most of the data are old anyway, except of the last episode).

Rafi

sharkykh:
This DB query is only part of the problem.
The bigger issue is that Vue needs to parse the HTML provided by the server and essentially re-render everything from scratch as "Virtual DOM". There is no "easy fix". We have to complete the transition to Vue/API.

rafi-d:
I've just tried to remove 20 out of my 71 shows that were actually marked as "Ended". To my surprise - I have managed to cut 50% of home page refresh (20 sec to 10 sec) .

Does it imply you may be doing some unnecessary checks for "static" shows (Ended/Not-active ATM) ?  

Rafi

sharkykh:
No, no, I simply stated some facts.
And if you mean about the "there's a chance some stats won't be needed",
I just meant that there's a lot of legacy code, and maybe we'll be able to come up with a better solution.

PR #6746, opened 2019-05-30
Restructure API qualities and statuses

sharkykh:
- Combined the two into `consts` both on the API and on the store.
- On the API side it will only generate this data once. It makes no sense to do that every call, because these values can't be changed by the user (that's why I renamed it to `consts`, as in constants)
- See commit messages for the rest
- The new schema:
```json
{
  "qualities": {
    "values": [
      // ...
      {
        "value": 8,
        "key": "hdtv",
        "name": "HDTV",
        "cssClass": "HD720p"
      }
      // ...
    ],
    "anySets": [
      // ...
      {
        "value": 40,
        "key": "anyhdtv",
        "name": "ANYHDTV",
        "cssClass": "any-hd"
      }
      // ...
    ],
    "presets": [
      // ...
      {
        "value": 65518,
        "key": "any",
        "name": "ANY",
        "cssClass": "Any"
      }
      // ...
    ]
  },
  "statuses": [
      // ...
      {
        "value": 3,
        "key": "wanted",
        "name": "Wanted"
      }
      // ...
  ]
}
```


medusa/server/api/v2/config.py
```diff
@@ -596,72 +599,74 @@ def data_main():
...
+            in iteritems(common.Quality.__dict__)
+            if isinstance(value, int) and not key.startswith('__')
+        }
+
+        def make_quality(value, name, key=None):
+            return {
+                'value': value,
+                'key': key or quality_key_map.get(value),
+                'name': name,
+                'cssClass': common.Quality.cssClassStrings.get(value) if key is None else name,
```
p0psicles: I think the cssClassStrings is a bit of unwanted legacy. Can't we refactor this a little and use `name` for the class?

sharkykh: Hmm. Some `name`s have spaces in them. Maybe I'll use the `key`s and update the CSS for it.
I'll look into it later.
Thanks!

p0psicles: Agree, key is better

medusa/search/manual.py
```diff
@@ -34,14 +36,19 @@
...
+# Copied from `medusa/server/api/v2/config.py` @ `DataGenerator.data_consts`
+quality_key_map = {
+    value: text_type(key).lower().replace('_', '')
+    for (key, value)
+    in iteritems(Quality.__dict__)
+    if isinstance(value, int) and not key.startswith('__')
+}
+
+
 def get_quality_class(ep_obj):
```
medariox: What about putting this in Quality? I don't see a strong reason to keep this in manual search.

medusa/search/manual.py
```diff
@@ -34,14 +36,19 @@
 SEARCH_STATUS_SEARCHING = 'searching'
 
 
+# Get key names from `common.Quality` class (for the quality values and any-sets)
+# Copied from `medusa/server/api/v2/config.py` @ `DataGenerator.data_consts`
+quality_key_map = {
```
medariox: Same as below.

medusa/server/api/v2/config.py
```diff
@@ -448,7 +450,8 @@ def sections(cls):
         """Get the available section names."""
         return [
             name[5:]
-            for (name, function) in inspect.getmembers(cls, predicate=inspect.isfunction)
+            for (name, function)
+            in inspect.getmembers(cls, predicate=lambda f: inspect.isfunction(f) or inspect.ismethod(f))
```
medariox: Iirc you can pass multiple predicates as a list, so you can avoid the lambda.

medusa/server/api/v2/config.py
```diff
@@ -596,72 +599,73 @@ def data_main():
...
-
-        section_data['strings'] = {}
-        section_data['strings']['values'] = common.Quality.qualityStrings
-        section_data['strings']['anySets'] = common.Quality.combinedQualityStrings
-        section_data['strings']['presets'] = common.qualityPresetStrings
-        section_data['strings']['cssClass'] = common.Quality.cssClassStrings
+        section_data['qualities'] = {}
+
+        # Get key names from `common.Quality` class (for the quality values and any-sets)
+        quality_key_map = {
```
medariox: Same code as above, consider putting this in Quality.

medusa/server/api/v2/config.py
```diff
@@ -596,72 +599,73 @@ def data_main():
...
+            value: text_type(key).lower().replace('_', '')
+            for (key, value)
+            in iteritems(common.Quality.__dict__)
+            if isinstance(value, int) and not key.startswith('__')
+        }
+
+        def make_quality(value, name, key=None):
```
medariox: Same as above.

medusa/search/manual.py
sharkykh: If I move this to Quality class then there's no need for it.
I mean... I still need this map, but I probably would not generate it from `Quality.__dict__`, I will just make a static dict for it...
Yeah maybe I'll do that.

medusa/search/manual.py
sharkykh: I don't want this at all, but I still need it for [the AJAX episode search](https://github.com/pymedusa/Medusa/blob/d2f971afa85b7c438f47c32e28c60e51c6125152/themes-default/slim/static/js/ajax-episode-search.js#L55-L62).
(😅 everything is dependencies upon dependencies, that's why I'm trying to convert the smaller components to Vue first)
Anyway, I don't see us needing it anywhere else in the future. I believe that I removed everything related to qualities' CSS classes from Python...

medusa/server/api/v2/config.py
sharkykh:
```py
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from medusa.server.api.v2.config import DataGenerator
>>> import inspect
>>> inspect.getmembers(DataGenerator, predicate=[inspect.isfunction, inspect.ismethod])
  File "<stdin>", line 1, in <module>
  File "C:\Python\Python37\lib\inspect.py", line 354, in getmembers
    if not predicate or predicate(value):
TypeError: 'list' object is not callable
>>>
```

medusa/server/api/v2/config.py
sharkykh: I don't see the point in doing that.
1. Only used in this context, for the API data, probably won't have any other uses for it.
2. The `data_consts` function only generates the data once per Medusa instance, so it's not performance.

But prove me otherwise and I'll move it..

medariox: In that case you don't need to move it. It's just really hard to keep an overview of what is already there this way. I mean, maybe in a few moths somebody else will need to do exactly the same and will push similar code. If we keep it all in one place it will be easier to avoid duplicates. Same goes with the `quality_key_map`. Also if one day we want to refactor the terrible Quality class, it might be easier.

sharkykh: Thank you for that. I'll keep that in mind. One more thing about this specific function is that it's simply a helper function.
You know, like, when you feel like you write the same code for more than 2 times, you want it to not be this way? 😄 It just creates a dict. Makes it easier to read in the context of a list comprehension.

medariox:
LGTM! Waiting for @p0psicles to merge because I'm unsure if it has any implications for what he is working on.

p0psicles:
Yeah it will cause some conflicts. But that's all. I'll manage.

p0psicles:
Can we disable deepcode all together?

medariox:
@p0psicles I've removed it. It should be disabled for new PRs.

PR #6748, opened 2019-05-31
UI Improvements

sharkykh:
- [x] Add indexer name/id getters to store config module
    - [x] Remove old code from `src/index.js`
    - [x] Update `app-link.vue`
    - [x] Update `manage_episodeStatuses.mako` (removing `MEDUSA.config` 🎉)
- [x] Replace duplicate call to `config/main` with the data from Vue
- [x] Auto close AppHeader menus when clicking a RouterLink
- [x] Don't register components twice
- [x] Add more info to the `common` fixture


themes-default/slim/src/index.js
```diff
@@ -144,24 +144,6 @@ if (!pathname.includes('/login') && !pathname.includes('/apibuilder')) {
...
-
-        MEDUSA.config.indexers.indexerIdToName = indexerId => {
-            if (!indexerId) {
-                return '';
-            }
-            return Object.keys(MEDUSA.config.indexers.config.indexers).filter(indexer => { // eslint-disable-line array-callback-return
-                if (MEDUSA.config.indexers.config.indexers[indexer].id === parseInt(indexerId, 10)) {
-                    return MEDUSA.config.indexers.config.indexers[indexer].name;
-                }
-            })[0];
```
sharkykh: Just a small note:
I changed the logic in the new function, but it's actually still returning the same value.
`Object.keys(...).filter(callback)` returns the elements of the array that the callback function returns a truthy value for, **and not the value the function returns.**
Which in this case, was the keys of `MEDUSA.config.indexers.config.indexers`.
That's why the new function does the same thing, but with a simpler logic.
```js
        return Object.keys(indexers).find(name => indexers[name].id === parseInt(indexerId, 10));
```

sharkykh:
Sorry 😅

PR #6752, opened 2019-05-31
Fix missing sub menu on errors page

sharkykh:
@medariox 
If you have time, this just needs to be bundled (yarn dev)
If not, I'll do it late**r**


sharkykh:
thanks :)

PR #6807, opened 2019-06-09
Update API endpoints and store

sharkykh:
- [x] Update `/api/v2/stats` and add to store
- [x] Add `/api/v2/config/system` and update store

Not sure about `medusa/system/schedulers.py`... Maybe it shouldn't be in `medusa.system`...
In any case, this is for the footer component (#4520) and for the status page.


medusa/system/schedulers.py
```diff
@@ -0,0 +1,139 @@
...
+
+all_schedulers = [
+    ('dailySearch', 'Daily Search', 'daily_search_scheduler'),
+    ('backlog', 'Backlog', 'backlog_search_scheduler'),
+    ('showUpdate', 'Show Update', 'show_update_scheduler'),
+    ('versionCheck', 'Version Check', 'version_check_scheduler'),
+    ('showQueue', 'Show Queue', 'show_queue_scheduler'),
+    ('searchQueue', 'Search Queue', 'search_queue_scheduler'),
+    ('properFinder', 'Proper Finder', 'proper_finder_scheduler'),
+    ('postProcess', 'Post Process', 'auto_post_processor_scheduler'),
```
sharkykh: Note to self - #6810 affects this line
```suggestion
    ('postProcess', 'Post Process', 'post_processor_scheduler'),
```

Issue #6799, opened 2019-06-08
[APP SUBMITTED]: IOError: [Errno 2] No such file or directory: '/mnt/HD/HD_a2/Public/downloads/TV/The.Thundermans.S02E03.1080p.WEB-DL.DD5.1.x264-TrollHD/The.Thundermans.S02E03.1080p.WEB-DL.DD.5.1.x264-TrollHD.mkv'


shanevoirin:
### INFO
**Python Version**: `2.7.15 (default, May 12 2018, 06:33:51) [GCC 7.3.0]`
**Operating System**: `Linux-4.1.13-x86_64-with-glibc2.2.5`
**Locale**: `UTF-8`
**Branch**: [master](../tree/master)
**Database**: `44.14`
**Commit**: pymedusa/Medusa@6a21a6001b10827d14d450135bae61729b01edf1
**Link to Log**: https://gist.github.com/49ec40ac8dfff9c118355aa7c0a70129
### ERROR
<pre>
2019-06-07 18:12:58 ERROR    POSTPROCESSOR :: [6a21a60] Unable to move file /mnt/HD/HD_a2/Public/downloads/TV/The.Thundermans.S02E03.1080p.WEB-DL.DD5.1.x264-TrollHD/The.Thundermans.S02E03.1080p.WEB-DL.DD.5.1.x264-TrollHD.mkv to /shares/Public/Shared Videos/TV/The Thundermans/Season 02/The Thundermans - S02E03 - Max's Minions.mkv: IOError(2, 'No such file or directory')
Traceback (most recent call last):
  File "/mnt/HD/HD_a2/Nas_Prog/**********/Medusa/**********/post_processor.py", line 472, in move
    helpers.move_file(cur_file_path, new_file_path)
  File "/mnt/HD/HD_a2/Nas_Prog/**********/Medusa/**********/helpers/__init__.py", line 312, in move_file
    shutil.move(src_file, dest_file)
  File "/mnt/HD/HD_a2/Nas_Prog/**********/Medusa/**********/init/filesystem.py", line 112, in patched_output
    return handle_output(f(*args, **kwargs))
  File "/opt/lib/python2.7/shutil.py", line 316, in move
  File "/opt/lib/python2.7/shutil.py", line 144, in copy2
  File "/mnt/HD/HD_a2/Nas_Prog/**********/Medusa/lib/shutil_custom/__init__.py", line 32, in copyfile_custom
[...]

PR #6810, opened 2019-06-10
Avoid post-processor concurrency issues

medariox:

This PR makes sure that the post-processor always runs threaded (not only the auto post-processor). This allows to keep using Medusa even if the post-processor is still running. Also makes sure that only one post-process thread can run at a time. This avoids issues like #6799, where the manual post-processor was executed right after/before the auto post-processor.


medusa/process_tv.py
```diff
@@ -151,7 +194,7 @@ def process(self, resource_name=None, force=False, is_priority=None, delete_on=F
                     self.missedfiles.append('{0}: Sync files found'.format(dir_path))
 
         if self.succeeded:
-            self.log('Successfully processed.')
+            self.log('Post-processing completed.')
```
medariox: I've changed this because many people are confused by the current log.

sharkykh:
@medariox Don't forget to run `yarn install` before `yarn dev`. If you don't, it bundles outdated packages.

medariox:
> @medariox Don't forget to run `yarn install` before `yarn dev`. If you don't, it bundles outdated packages.

I did. I'm not sure why it complaints 😞

sharkykh:
Ah... that was it :)

medusa/server/web/home/post_process.py
```diff
@@ -49,14 +49,20 @@ def _decode(value):
...
-                delete_on=argToBool(delete_on), failed=argToBool(failed), proc_type=type,
+            result = app.post_processor_scheduler.action.run(
+                path=proc_dir,
+                process_method=process_method,
+                resource_name=resource_name,
+                force=argToBool(force),
+                is_priority=argToBool(is_priority),
+                delete_on=argToBool(delete_on),
+                failed=argToBool(failed),
+                proc_type=proc_type,
```
medariox: This was passing `type` 🤣

medusa/server/web/home/post_process.py
```diff
@@ -49,14 +49,20 @@ def _decode(value):
             proc_dir = _decode(proc_dir)
             resource_name = _decode(nzbName)
 
-            result = process_tv.ProcessResult(proc_dir, process_method=process_method).process(
-                resource_name=resource_name, force=argToBool(force), is_priority=argToBool(is_priority),
-                delete_on=argToBool(delete_on), failed=argToBool(failed), proc_type=type,
+            result = app.post_processor_scheduler.action.run(
```
sharkykh: Something doesn't add up for me here.
So, if you call the `action`'s `run` directly, you're not *actually* running this threaded.
You need a something like a queue, so the `action` would be something like the implementation for `ShowQueue`.

sharkykh: Okay maybe not a full-on queue, but just some way to trigger the scheduler to run with other custom arguments, (maybe also without changing the scheduled auto post-processing run)

p0psicles: Well if you want to queue pp actions you need a queue. But then you need to poll, or have a callback.

p0psicles: So the only difference with just calling a method, is that now you can ask the scheduler if the method is running.

sharkykh: >So the only difference with just calling a method, is that now you can ask the scheduler if the method is running.

@p0psicles 
Yes, I agree with that.
What @medariox and I are disagreeing about is, that it runs in a thread (= not blocking).

medusa/server/api/v1/core.py
```diff
@@ -1303,9 +1326,14 @@ def run(self):
         if not self.type:
             self.type = 'manual'
 
-        data = process_tv.ProcessResult(self.path, process_method=self.process_method).process(
-            force=self.force_replace, is_priority=self.is_priority, delete_on=self.delete_files,
-            failed=self.failed, proc_type=self.type
+        data = app.post_processor_scheduler.action.run(
```
p0psicles: Because your directly calling the run method here, it's not triggered from the scheduler. The scheduler itself is running in a thread. But by calling action.run() your bypassing that.

The periodic scheduled pp, can be called by the the scheduler. But pp called from apiv1/nzbToMedia needs to be done in sync. So that will cause a block on the webserver thread.

medariox: Thanks @p0psicles, @sharkykh and I already had a discussion about that and it is clear now. I will think about what to do tomorrow, because I don't want to add a Queue for this. The PR still fixes issues like #6799 and a bunch of other minor stuff.

p0psicles: We can reserve a real async pp solution for when we move the pp API call to apiv2. Then we can also destinct the solution in nzbToMedia. And have it poll for the pp outcome.

PR #6811, opened 2019-06-10
Simplify global component registration

sharkykh:
- [x] Moved [the `main.mako` shim](https://github.com/pymedusa/Medusa/blob/90691ab14fa1bab4b25166ba4af27ae1cc4302b3/themes-default/slim/views/layouts/main.mako#L128-L187) into [a JS file](https://github.com/pymedusa/Medusa/blob/7c765bf813717e8898878fc6eb3fee5f181812e1/themes-default/slim/src/global-vue-shim.js).
	- Consolidated the global template registration code(s) and made it so we only have to change one code to add / remove global components.
	- Added explanations and TODOs, so hopefully it's easier to understand when you should add or remove global components.
- [x] Deprecated the global `auth` & `config` store states. A warning will show up in console for components that still use it.
  I fixed ~two~ 9 locations I found, and I fixed some more of these a few months back, but it's possible that there are still a few stragglers, so deprecating it first to ensure that everything will still work.
- [x] Fix Vue plugins registrations (remove doubled registrations, `Vuex` and `VueRouter` are now registered in their modules)
- [x] Change the `snatchSelection` component name to `snatch-selection` (consistency)
- [x] Convert router to a package and extract the routes to a separate file.

All in all, UX remains unchanged.

@p0psicles 	With this, when you want to add a global component (if you can't import it and use `components: { ComponentName }`), you edit `./src/global-vue-shim.js`


sharkykh:
Found more of my "deprecation warnings"

PR #6823, opened 2019-06-12
[POC] Support theme-based styles in SFCs

sharkykh:
- [x] Change `cssThemes` to an array
- [x] Update Gulpfile
- [x] Split Webpack configurations
- [x] Add support for theme-based styles in SFCs

It slows down the build by 40-50%.
Opening this PR as a Proof-of-Concept only at the moment.


sharkykh:
Something to keep in mind:
With this configuration, we can **choose** which theme to build.
https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations
> If you pass a name to `--config-name` flag, webpack will only build that specific configuration.

For our use case:
`yarn [dev|build] --config-name [dark|light] [--watch]`

So for example, if our Medusa is running the `dark` theme, we don't need to build the light theme every time we make a change. We can instead run `yarn dev --config-name dark` to only build the dark theme.

At this point, it's probably a good idea to add shortcuts for these,
in the form of `yarn dev:dark` and `yarn dev:light`.

Before committing you will still be required to run either the full build with `yarn dev`, or just build the other theme (so if you worked on `dark`, run `yarn dev:light`)

PR #6852, opened 2019-06-18
Docker on Python 3.7

### LINKED ISSUE #6378: Docker python 3
Hi,

today I tested docker for new 0.3.0 version, but its still based on python 2.7.15 not on python 3.
Can you please update dockerfile, so it can run on python 3 as well please?

If Im doing something wrong, can you please point me to the right direction here?

Thanks

With regards Sparhawk

sharkykh:
### Scheduled for: _Friday, 2019-07-05_
Supersedes #6424 
Closes #6378

- [x] News update: https://github.com/pymedusa/medusa.github.io/pull/168

Seems to work okay,
## BUT
- [x] ~Upgrading from Python 2 still requires deletion of the cache file databases (`.dbm` files)~
	This is no longer true since [we have code that automatically removes them if you change versions](https://github.com/pymedusa/Medusa/pull/5892).
	I believe users on the current [pymedusa/medusa](https://hub.docker.com/r/pymedusa/medusa) OR [linuxserver/medusa](https://hub.docker.com/r/linuxserver/medusa) will not be affected by this, since both Docker images have `gdbm` included in them.
	The new Docker image proposed here also has `gdbm` installed [from the python layer](https://github.com/docker-library/python/blob/master/3.7/alpine3.9/Dockerfile#L47).
- [x] [Done - news explain this change] This image DOES NOT have an automatic restarter (S6-overlay) like LinuxServer.io's image has.
[Users will need to configure it themselves.](https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy)
- [x] ~Not heavily tested...~ This is what the develop branch is for. Hopefully we have users on the `develop` docker tag.


medusa/__main__.py
```diff
@@ -473,12 +473,14 @@ def initialize(self, console_logging=True):
                     app.GIT_REMOTE_URL = app.APPLICATION_URL
 
             # current commit hash
-            app.CUR_COMMIT_HASH = check_setting_str(app.CFG, 'General', 'cur_commit_hash', '')
+            commit_hash_env = os.environ.get('MEDUSA_COMMIT_HASH', None)  # This is for our Docker container
+            app.CUR_COMMIT_HASH = check_setting_str(app.CFG, 'General', 'cur_commit_hash', commit_hash_env or '')
```
sharkykh: I just realized these changes might not work as expected.
Need to look into it more, but when updating the container after pulling a new image, the old hash/branch will remain...

sharkykh:
I created a temporary tag on Docker Hub for this branch:
`docker pull pymedusa/medusa:new-image`

sharkykh:
25 megabytes are taken by [lib/native](https://github.com/pymedusa/Medusa/tree/develop/lib/native) which isn't even needed on this image. 🤦‍♂
```shell
$ du -ch $(git ls-files *.dll *.dylib)
15M     lib/native/macos/libmediainfo.0.dylib
4.3M    lib/native/windows/i386/MediaInfo.dll
5.3M    lib/native/windows/x86_64/MediaInfo.dll
164K    lib/unrar2/UnRARDLL/unrar.dll
188K    lib/unrar2/UnRARDLL/x64/unrar64.dll
25M     total
```

I need to exclude them from the image in a way that if these files move in the future, it will still exclude them. Maybe just ignore all `*.dll` and `*.dylib` files...

p0psicles:
Honestly I would just assume they will not move

sharkykh:
We should, at the very least, publish news about this change before merging (even to develop).
I don't know how many people use the official Medusa image, and of those, how many are on the develop tag...

p0psicles:
Is it an idea to keep it as a separated tag for now? `medusa-py3` ?

p0psicles:
Ok I see you guys already thought about it. I'll leave it to you. So then it's publish news. Wait a week. And merge this?

sharkykh:
Wondering 🤔 
Will this ownership change cause issues for users?
https://github.com/pymedusa/Medusa/blob/317bcb3eef632a87ad94dee2368aede12253ac9c/.docker/root/etc/cont-init.d/30-config#L4-L6
I mean, `/app` is fine, but `/config`...?

medariox:
@sharkykh 
That actually _solved_ the issue for me under Windows (Linux containers). With the old image I was getting permission issues. Let's go!

PR #6880, opened 2019-06-27
Update AniDB component

sharkykh:
More commits extracted from #4486 

Still need to test these changes in the context of current develop (rather than the context of #4486)
~Will do later.~ ✔️

- [x] Test Edit Show page
- [x] Test Add New Show page


sharkykh:
CI got stuck

Issue #5919, opened 2018-12-26
Deluge authentication failed since update


Lutaksur:
Hello,

Since I updated to the last commit I can't download anymore. I have the error in log: 

2018-12-26 17:14:05 WARNING  SNATCHQUEUE-SNATCH-293088 :: [918cfe7] Deluge: Authentication Failed

In sears settings the "test connection" button work and said "Success: Connected and Authenticated" but I can't add torrent anymore. This worked great since the update.

How I can revert and come back to a previous version ?

Thanks.

medariox:
Please enable debug logging and post the logs of the failed snatch.

Lutaksur:
2018-12-26 18:10:22 WARNING  SNATCHQUEUE-SNATCH-263365 :: [918cfe7] Deluge: Authentication Failed
2018-12-26 18:10:22 DEBUG    SNATCHQUEUE-SNATCH-263365 :: [918cfe7] Calling Deluge Client
2018-12-26 18:10:22 DEBUG    SNATCHQUEUE-SNATCH-263365 :: [918cfe7] User-Agent: Medusa/0.2.14 (Linux; 4.9.0-3-amd64; 516a861a-0928-11e9-8494-d43d7e336391)
2018-12-26 18:10:22 DEBUG    SNATCHQUEUE-SNATCH-263365 :: [918cfe7] GET URL: https://xthor.to/download.php?torrent=190411&torrent_pass=********** [Status: 200]
2018-12-26 18:09:44 DEBUG    Thread_5 :: [918cfe7] With post data: {"params": [], "method": "web.connected", "id": 10}
2018-12-26 18:09:44 DEBUG    Thread_5 :: [918cfe7] User-Agent: Medusa/0.2.14 (Linux; 4.9.0-3-amd64; 516a861a-0928-11e9-8494-d43d7e336391)
2018-12-26 18:09:44 DEBUG    Thread_5 :: [918cfe7] POST URL: http://127.0.0.1:8112/json [Status: 200]
2018-12-26 18:09:44 DEBUG    Thread_5 :: [918cfe7] With post data: {"params": ["**********"], "method": "auth.login", "id": 1}
2018-12-26 18:09:44 DEBUG    Thread_5 :: [918cfe7] User-Agent: Medusa/0.2.14 (Linux; 4.9.0-3-amd64; 516a861a-0928-11e9-8494-d43d7e336391)
[...]

PR #6886, opened 2019-06-28
Config - search: force trailing slash when method deluge is chosen

p0psicles:

* Fixed the label (add the trailing slash)
* Add a check that adds the trailing slash if missing. @medariox are there any situations where you don't want this?

As reminded here:
https://github.com/pymedusa/Medusa/issues/5919#issuecomment-506153040


medariox:
For most providers it doesn't matter since we are using `urljoin()`. Also for Deluge, that's why I'm actually wondering if this is really needed. 😕 
For qBittorrent and MLDonkey it is required.
For rTorrent it should be avoided because of `https://localhost/rutorrent/plugins/httprpc/action.php`.

p0psicles:
Wait so it doesn't matter of a user misses the trailing slash for deluge?

PR #6893, opened 2019-06-29
Update & clean up router, prepare for next stage of navigation

sharkykh:
- [x] Split router sub-menus to a separate file (no change)
- [x] Use arrow functions instead
- [x] Reformat routes (indent changes only)
- [x] Update AppLink `matchingVueRoute` (this is needed to utilize the full potential of RouterLink)
- [x] Get `base[href]` using `querySelector`
- [x] Add router legacy redirect helpers (this will come in handy soon enough... 😁)


sharkykh:
@p0psicles 	Would love to get this merged too please, if you can.

PR #6709, opened 2019-05-21
Feature/vueify display show rebased

p0psicles:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

### checklist
- [x] include show header component
- [x] Display season theads
- [x] Display episodes
- [x] include subtitle flags
- [x] use localized date format for date/times
- [x] fix the status overview checkbox buttons
- [x] fix `change status to` button
- [x] fix `change quality to` button
- [x] implement 'select columns' feature
- [x] implement season summary row. for ex. `Season contains 6 episodes with total filesize: 9.06 GB`
- [x] fix dark / light template
- [x] implement download column
- [x] implement links to forced search, manual search
- [x] implement modal to ask for failed search, when clicking forced search button
- [x] implement subtitle search auto + manual 
- [x] implement 'jump to season' feature
- [x] fix with / without fanart background option
- [x] remove `show all seasons` option
~~- [ ] implement `fuzzy dates` option~~ only applies to snatchSelection
- [x] allow option to disable farnart
- [x] CSS bug ignored blue text
- [x] fix pagination div on smaller displays
- [x] add config options for enabling/disabling pagination. And page length.
- [x] for the snatch-selection release-name coloring. Take notice of the exclude ignored/required words.
- [x] fix the display specials toggle button.
- [x] fix builds

### I've created a set of smaller PR's to offload this one. Merged in this order:
- [x] Add custom vue-good-table component to package.json #6726
- [x] update package.json separate deps and dev-deps #6758
- [x] Add queuetime and starttime to the generic_queue.py #6717 
- [x] Introduce the apiv2 search route #6716
- [x] Replace FailedSearchQueueItem with BacklogSearchQueueItem #6718
- [x] Add subtitle search Vue SFC  #6727 

This will be the (reased) base PR/branch. For the coming weeks i'll start creating smaller PR's, and try to reduce this PR in size where possible.
continued from #6315 

Tnx. @sharkykh

PR #6901, opened 2019-06-30
Update `Series.to_json()` and `ShowHeader`

### LINKED ISSUE #5121: API: Change `show.seasons` structure

p0psicles:

----

To remove weight from #6709

----

- [x] Update `Series.to_json()`:
    - [x] Add episode count per season
    - [x] Add scene exceptions
    - [x] Add scene absolute numbering
    - [x] Add scene numbering
    - [x] Add xem absolute numbering
    - [x] Add xem numbering
    - [x] Change the structure of `show.seasons` (closes #5121)
- [x] Update `/api/v2/series` route:
    - [x] Add `episodes` param to `/api/v2/series`.
    - [x] Change the default value for `detailed` query param to `False`.
- [x] Use `show.seasonCount` to render the `seasonJump`. `show.seasons` is still used for the status buttons.
- [x] Remove unused data on `jumpToSeason`
- [x] Fix missing values on manual search (on `&season=0&episode=(any)`, **or a season search**)
- [x] Fix `jumpToSeason` scrolling. It's still not perfect but it's better than it was
	(<@sharkykh> wasn't scrolling smoothly for me)
- [x] Move some CSS and fix invalid HTML5 (`<font>`)


p0psicles:
Reminder to update api-spec with series in: query params

medusa/server/api/v2/series.py
```diff
@@ -64,7 +65,8 @@ def filter_series(current):
             return self._not_found('Series not found')
 
         detailed = self._parse_boolean(self.get_argument('detailed', default=True))
-        data = series.to_json(detailed=detailed)
+        episodes = self._parse_boolean(self.get_argument('episodes', default=False))
```
sharkykh: This changes the current behavior:
Currently `detailed` defaults to `True` which means episodes are added to the response.
Was this intended?

medusa/tv/series.py
```diff
@@ -97,8 +97,11 @@
     NameParser,
 )
 from medusa.sbdatetime import sbdatetime
-from medusa.scene_exceptions import get_scene_exceptions, update_scene_exceptions
-from medusa.scene_numbering import get_xem_numbering_for_show
+from medusa.scene_exceptions import get_all_scene_exceptions, get_scene_exceptions, update_scene_exceptions
+from medusa.scene_numbering import (
+    get_scene_absolute_numbering_for_show, get_scene_numbering_for_show,
+    get_xem_absolute_numbering_for_show, get_xem_numbering_for_show
```
sharkykh: IMO, if you go multi-line, put one import per line
```py
from medusa.scene_numbering import (
    get_scene_absolute_numbering_for_show,
    get_scene_numbering_for_show,
    get_xem_absolute_numbering_for_show,
    get_xem_numbering_for_show
)
```

medusa/tv/series.py
```diff
@@ -2060,7 +2083,7 @@ def __unicode__(self):
         to_return += u'anime: {0}\n'.format(self.is_anime)
         return to_return
 
-    def to_json(self, detailed=True):
+    def to_json(self, detailed=True, episodes=False):
```
sharkykh: Same as above, was the change to the default response intended?

medusa/tv/series.py
```diff
@@ -2131,19 +2155,31 @@ def to_json(self, detailed=True):
             data['config']['release']['whitelist'] = bw_list.whitelist
 
         if detailed:
-            episodes = self.get_all_episodes()
             data['size'] = self.size
-            data['seasons'] = [list(v) for _, v in
-                               groupby([ep.to_json() for ep in episodes], lambda item: item['season'])]
+            data['showQueueStatus'] = self.show_queue_status
+            data['xemNumbering'] = [{'source': {'season': src[0], 'episode': src[1]},
```
sharkykh: Since you're doing the same thing twice now (`xemNumbering` and `sceneNumbering` keys),
wouldn't it be better to have a helper function that returns it in the structure you want, simplifying `to_json()` a bit?

medusa/tv/series.py
```diff
@@ -2131,19 +2155,31 @@ def to_json(self, detailed=True):
...
+                                    for src, dest in viewitems(self.xem_numbering)]
+            data['sceneAbsoluteNumbering'] = self.scene_absolute_numbering
+            data['allSceneExceptions'] = self.all_scene_exceptions
+            if self.is_scene:
+                data['xemAbsoluteNumbering'] = self.xem_absolute_numbering
+                data['sceneNumbering'] = [{'source': {'season': src[0], 'episode': src[1]},
+                                           'destination': {'season': dest[0], 'episode': dest[1]}}
+                                          for src, dest in viewitems(self.scene_numbering)]
+            else:
+                data['xemAbsoluteNumbering'], data['sceneNumbering'] = [], []
```
sharkykh: I think splitting this over two lines in this case is more readable.
```suggestion
                data['xemAbsoluteNumbering'] = []
                data['sceneNumbering'] = []
```

themes-default/slim/src/components/show-header.vue
```diff
@@ -21,24 +21,24 @@
...
                     </span>
 
                     <div class="h2footer display-seasons clear">
                         <span>
-                            <select v-if="show.seasons.length >= 15" v-model="jumpToSeason" id="seasonJump" class="form-control input-sm" style="position: relative">
+                            <select v-if="Object.keys(show.seasonCount) >= 15" v-model="jumpToSeason" id="seasonJump" class="form-control input-sm" style="position: relative">
                                 <option value="jump">Jump to Season</option>
-                                <option v-for="season in show.seasons" :key="'jumpToSeason-' + season[0].season" :value="'#season-' + season[0].season" :data-season="season[0].season">
-                                    {{ season[0].season === 0 ? 'Specials' : 'Season ' + season[0].season }}
+                                <option v-for="season in Object.keys(show.seasonCount)" :key="`jumpToSeason-${season}`" :value="`#season-${season}`" :data-season="season">
```
sharkykh: When you're looping, you can also use:
```suggestion
                                <option v-for="(episodeCount, season) in show.seasonCount" :key="`jumpToSeason-${season}`" :value="`#season-${season}`" :data-season="season">
```

Also note that the order of the keys on `Object.keys` is not guaranteed (applies to my code and your code):
https://vuejs.org/v2/guide/list.html#v-for-with-an-Object
>When iterating over an object, the order is based on the enumeration order of `Object.keys()`, which is not guaranteed to be consistent across JavaScript engine implementations.

That's part of the reason I use lists of objects for the qualities and statuses.

themes-default/slim/src/components/show-header.vue
```diff
@@ -21,24 +21,24 @@
...
+                                    {{ season === 0 ? 'Specials' : 'Season ' + season }}
```
sharkykh: `season` is a string, not a number.
```suggestion
                                    {{ season === '0' ? 'Specials' : 'Season ' + season }}
```

themes-default/slim/src/components/show-header.vue
```diff
@@ -21,24 +21,24 @@
...
                             </select>
-                            <template v-else-if="show.seasons.length >= 1">
+                            <template v-else-if="Object.keys(show.seasonCount).length >= 1">
                                 Season:
-                                <template v-for="(season, $index) in reverse(show.seasons)">
-                                    <app-link :href="'#season-' + season[0].season" :key="`jumpToSeason-${season[0].season}`">{{ season[0].season === 0 ? 'Specials' : season[0].season }}</app-link>
-                                    <span v-if="$index !== (show.seasons.length - 1)" :key="`separator-${$index}`" class="separator">| </span>
+                                <template v-for="(season, $index) in reverse(Object.keys(show.seasonCount))">
+                                    <app-link :href="'#season-' + season" :key="`jumpToSeason-${season}`">{{ season === 0 ? 'Specials' : season }}</app-link>
+                                    <span v-if="$index !== (Object.keys(show.seasonCount) - 1)" :key="`separator-${$index}`" class="separator">| </span>
```
sharkykh:
```suggestion
                                    <span v-if="$index !== (Object.keys(show.seasonCount).length - 1)" :key="`separator-${$index}`" class="separator">| </span>
```

themes-default/slim/src/components/show-header.vue
```diff
@@ -21,24 +21,24 @@
...
+                                    {{ season === 0 ? 'Specials' : 'Season ' + season }}
                                 </option>
                             </select>
-                            <template v-else-if="show.seasons.length >= 1">
+                            <template v-else-if="Object.keys(show.seasonCount).length >= 1">
                                 Season:
-                                <template v-for="(season, $index) in reverse(show.seasons)">
-                                    <app-link :href="'#season-' + season[0].season" :key="`jumpToSeason-${season[0].season}`">{{ season[0].season === 0 ? 'Specials' : season[0].season }}</app-link>
-                                    <span v-if="$index !== (show.seasons.length - 1)" :key="`separator-${$index}`" class="separator">| </span>
+                                <template v-for="(season, $index) in reverse(Object.keys(show.seasonCount))">
```
sharkykh: Same as above about the order of `Object.keys`.

themes-default/slim/src/store/modules/shows.js
```diff
@@ -87,16 +87,23 @@ const actions = {
...
+                timeout = 60000;
             }
 
-            api.get('/series/' + indexer + id, { params })
+            if (episodes !== undefined) {
+                params.episodes = episodes;
+                timeout = 60000;
+            }
+
+            api.get(`/series/${indexer}${id}`, { params }, { timeout })
```
sharkykh: [The second argument will be ignored by the Axios instance function](https://github.com/axios/axios/blob/master/README.md#axiosgeturl-config-1). Change to:
```suggestion
            api.get(`/series/${indexer}${id}`, { params, timeout })
```

medusa/server/api/v2/series.py
p0psicles: Yes. As afaik only show-header makes use of it. And by default is like to keep the response as small as possible.

Maybe we should also change detailed by default to false.

In the displayShow we use the episodes route to get episodes per season. So this is preparation for that

medusa/tv/series.py
p0psicles: Yes

themes-default/slim/src/components/show-header.vue
p0psicles: Shouldn't be that much of a problem when the keys are seasons. It's rather easy to order them.
Also i'd like to keep the information of the amount of episodes per season. I'll create a computed for it.

medusa/tv/series.py
```diff
@@ -2131,19 +2155,31 @@ def to_json(self, detailed=True):
...
         if detailed:
-            episodes = self.get_all_episodes()
             data['size'] = self.size
-            data['seasons'] = [list(v) for _, v in
-                               groupby([ep.to_json() for ep in episodes], lambda item: item['season'])]
+            data['showQueueStatus'] = self.show_queue_status
+            data['xemNumbering'] = [{'source': {'season': src[0], 'episode': src[1]},
+                                     'destination': {'season': dest[0], 'episode': dest[1]}}
+                                    for src, dest in viewitems(self.xem_numbering)]
+            data['sceneAbsoluteNumbering'] = self.scene_absolute_numbering
```
p0psicles: @sharkykh this will result in a dynamic key: value mapping.
Like: [1: 1, 2: 2, 3: 4].
Thing is. Swagger doesn't like this. Or at least I haven't figured out how to configure this in swagger in a way that `https://editor.swagger.io/` accepts it as valid. So maybe we should make sure they all have a similar format. As in: [src: 1, dest: 1, src: 2: dest: 2, etc..] ?

dredd/api-description.yml
```diff
@@ -953,6 +953,15 @@ definitions:
       showType:
         type: string
         enum: [series, anime, sports]
+      seasonCount:
+        type: object
```
p0psicles: This is also such a dynamic key mapping. This breaks the swagger validation.

p0psicles:
This PR is currently in a broken state.

medusa/tv/series.py
```diff
@@ -2060,7 +2083,7 @@ def __unicode__(self):
         to_return += u'anime: {0}\n'.format(self.is_anime)
         return to_return
 
-    def to_json(self, detailed=True):
+    def to_json(self, detailed=False, episodes=False):
```
p0psicles: My reasoning to have this False by default. Is that I want to have a minimal call to the series object by default. If you need more detailed info you have to opt-in.

sharkykh:
@p0psicles 	Not sure if this was done intentionally - you changed `seasonJump` to only show up when there is at least 2 seasons (before it was at least 1 season).
I guess it makes sense (you don't need to jump to season if there's only one), but I think it should be consistent - if there are any seasons, we should have the jump to season (and show/hide specials)

themes-default/slim/src/components/show-header.vue
```diff
@@ -21,24 +21,24 @@
                         >{{ show.title }}</app-link> / Season {{ season }}<template v-if="episode && manualSearchType !== 'season'"> Episode {{ episode }}</template>
                     </span>
                 </div>
-                <div v-if="type !== 'snatch-selection' && show.seasons && show.seasons.length >= 1" id="show-specials-and-seasons" class="pull-right">
-                    <span class="h2footer display-specials" v-if="show.seasons.find(season => ({ season }) => season === 0)">
+                <div v-if="type !== 'snatch-selection' && seasons.length > 1" id="show-specials-and-seasons" class="pull-right">
```
sharkykh: @p0psicles 
Not sure if this was done intentionally - you changed `seasonJump` to only show up when there are at least 2 seasons (before it was at least 1 season).

I guess it makes sense (you don't need to jump to season if there's only one), but I think it should be consistent - if there are any seasons, we should have the jump to season (and show/hide specials)

p0psicles: I didn't intend to change behavior

p0psicles:
I'm leaving this to you shark. feel free to merge

themes-default/slim/src/components/show-header.vue
```diff
@@ -281,6 +284,17 @@ import { api } from '../api';
 import { combineQualities, humanFileSize } from '../utils/core';
 import { AppLink, Asset, QualityPill, StateSwitch } from './helpers';
 
+/**
+ * Return the first item of `values` that is not `null`, `undefined` or `NaN`.
+ * @param {...any} values - Values to check.
+ * @returns {any} - The first item that fits the criteria, `undefined` otherwise.
+ */
+const resolveToValue = (...values) => {
```
sharkykh: I know that this **can be** in `./src/utils/core.js`, but for now I don't think we'll need to use it in a lot of places... it's only temporary anyway, as we _probably_ won't be using query params for these.

I have a branch that adds that to `utils`, with tests as well, in case we need it.

sharkykh: I know that this **can be** in `./src/utils/core.js`, but for now I don't think we'll need to use it in a lot of places... it's only temporary anyway, as we _probably_ won't be using query params to pass data into this component.

I have a branch that adds that to `utils`, with tests as well, in case we need it.

sharkykh: I know that this **can be** in `./src/utils/core.js`, but for now I don't think we'll need to use it in a lot of places... it's only temporary anyway, as we _probably_ won't be using query params to pass data into this component.

I have a branch that adds that to `utils`, with tests as well, in case we need it.

PR #6908, opened 2019-07-02
Fix select-list: Make sure we add list-items on mounted

p0psicles:
Just have the watch in created is not enough. As it's possible the list-items are passed before the watch is created.


sharkykh:
A few days ago while working on #6893,
I took the chance to see how `EditShow` would work with path parameters (`/home/editShow/tvdb12345`) and using VueRouter navigation, and I realized that we should make our props reactive (put a watch on them, or use a computed that re-evaluates when they change).

While this will probably cause infinite loops, we have to find the correct ways to avoid them.

If we don't, when navigating from, say, `/home/displayShow/tvdb12345` to `/home/displayShow/tvmaze54321`,
the lifecycle hooks we've been using for initialization - `mounted()`, `created()`, etc. will not be invoked, because the page component is reused (updated, not recreated).
Instead, [a VueRouter navigation guard](https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards) is triggered: `beforeRouteUpdate`, where we can react to the new "incoming" props (`tvmaze54321` will become `{ indexer: 'tvdb', id: 54321 }`) - for example fetch the show data.

So for now this solution to the problem is fine.

But I feel like something in our component designs is off because in some components we encounter horrible infinite loops if we try to keep everything reactive, and some components work great.
It's usually the components that have to emit an updated value to the parent, but that value is derived from some property value they accept.

Sorry about the rant.

p0psicles:
Not really, but I only add it on mount. That can't cause any loops. I'll check the config pages tomorrow. Leave it for time being

sharkykh:
> That can't cause any loops. 

Yeah I didn't mean to imply that. That's why it's on a separate comment.
I also said, "So for now this solution to the problem is fine.", and I meant this PR as a solution for the current issue on `EditShow`.

p0psicles:
The loops where mostly cause because I over used the watch in the past.

p0psicles:
Why is codecov failing?

medariox:
@p0psicles 
Happens sometimes, just takes some more time.

p0psicles:
@sharkykh can you check in your code. If using keep-alive icw the activated hook can provide a solution? https://vuejs.org/v2/api/#activated

PR #6918, opened 2019-07-05
Fix QualityChooser

sharkykh:
From Slack
> Add New Show quality chooser not loading the default quality (it's loading then setting it to `null` because of the `await waitFor(...)` :confused:)
>image


sharkykh:
### [Vue-Test-Utils is moving towards async tests by default](https://github.com/vuejs/vue-test-utils/issues/1137)

Issue #6967, opened 2019-07-21
Medusa detecting 1080p quality as SD


domibarton:
**Describe the bug**

Medusa is detecting the following file as `SDTV` instead of `1080p`:

```
tvs-americansDETE507/The.Americans.2013.S05E07.Hoehere.Ziele.German.DD+51.DL.1080p.NetflixHD.AVC-TVS/tvs-the-americans-2013-dd51-dl-18p-nfhd-avc-507.mkv
```

Thus Medusa is not moving those files, as `SD` is not in the allowed quality of the TV show.

Why does Medusa think this is SD? Because of the `TVS` suffix'ish extension?

**To Reproduce**

1. Download a video file an name it `tvs-americansDETE507/The.Americans.2013.S05E07.Hoehere.Ziele.German.DD+51.DL.1080p.NetflixHD.AVC-TVS/tvs-the-americans-2013-dd51-dl-18p-nfhd-avc-507.mkv`
2. Run the post processing against the file
3. See the error

**Expected behavior**

File should be detected as `1080p` (via `knowit`) and not `SDTV` (via filenae).

**Medusa (please complete the following information):**

- Docker w/ own Dockerfile / image
- OS: Debian Buster
- Python: `2.7.16`
- Git Tag: `v0.3.4`

```
+-------------------------------------------------------+
|                   KnowIt 0.3.0-dev                    |
+-------------------------------------------------------+
|                      pymediainfo                      |
|                          4.0                          |
|                   libmediainfo.so.0                   |
|                        v18.12                         |
|                                                       |
|                        ffprobe                        |
|                        v4.1.3                         |
|                                                       |
|                        enzyme                         |
|                         0.4.2                         |
+-------------------------------------------------------+
|      Please report any bug or feature request at      |
|       https://github.com/ratoaq2/knowit/issues.       |
+-------------------------------------------------------+
```

**Debug logs (at least 50 lines):**
<details>
```
2019-07-21 13:03:04 DEBUG    Thread_59 :: [] Analyzing name: tvs-americansDETE507/The.Americans.2013.S05E07.Hoehere.Ziele.German.DD+51.DL.1080p.NetflixHD.AVC-TVS/tvs-the-americans-2013-dd51-dl-18p-nfhd-avc-507.mkv
[...]

PR #6982, opened 2019-07-27
Improve postprocessing name parsing.

p0psicles:
* Allow for upgrading the found quality from filename -> full path.
* Added unit test

inspired by: #6967

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ jobs:
         - yarn coverage
     # backend tests (py2.7) start here
     - name: 'Backend tests (py2.7)'
-      python: '2.7.10'
+      python: '2.7.11'
       env:
         - TOXENV=py27
       <<: *backend_tests
@@ -71,7 +71,7 @@ jobs:
       <<: *backend_tests
     # dredd tests (py2.7) start here
     - name: 'Dredd tests (py2.7)'
-      python: '2.7.10'
+      python: '2.7.11'
       <<: *dredd_tests
     # dredd tests (py3.6) start here
     - name: 'Dredd tests (py3.6)'
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ jobs:
         - yarn coverage
     # backend tests (py2.7) start here
     - name: 'Backend tests (py2.7)'
-      python: '2.7.11'
+      python: '2.7'
       env:
         - TOXENV=py27
       <<: *backend_tests
@@ -71,7 +71,7 @@ jobs:
       <<: *backend_tests
     # dredd tests (py2.7) start here
     - name: 'Dredd tests (py2.7)'
-      python: '2.7.11'
+      python: '2.7'
       <<: *dredd_tests
     # dredd tests (py3.6) start here
     - name: 'Dredd tests (py3.6)'
--- a/medusa/common.py
+++ b/medusa/common.py
@@ -334,7 +334,7 @@ class Quality(object):
                 if ep.bluray:
                     result = Quality.UHD_8K_BLURAY if is_4320 else Quality.UHD_4K_BLURAY
                 # WEB-DL
-                elif ep.web or any([ep.amazon, ep.itunes, ep.netflix]):
+                elif ep.web:
                     result = Quality.UHD_8K_WEBDL if is_4320 else Quality.UHD_4K_WEBDL
                 # HDTV
                 else:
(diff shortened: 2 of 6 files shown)


medariox:
Few things to consider here: who says that the higher quality is actually the quality we are looking for? We can't possibly know that. Same thing for keep parsing all 3 variants: are we sure that keeping parsing will yield better results? I think that the main issue here is that we return SDTV even when we don't have any hints about qualities. I believe this is due the old name parsers that falls back when guessit is unable to find any valid quality. This is likely the underlying issue and changing that could possibly fix these edge cases. The question is if it will break other parsing that currently kinda works (thinking about some x265 releases), but we can add some tests for (some of) that.

medariox:
Found the line: https://github.com/pymedusa/Medusa/blob/74c3f124fb599ad923e23f46fc320778b8b4217c/medusa/common.py#L366

The file name contains `nfhd` so it is detected as SDTV :/

domibarton:
I switched from SickBeard to SickRage years ago, and a few months ago finally to Medusa. I've to say, I love Medusa and I love the added value / features. Kudos to that! Amazing product, amazing features, great contributors! Thanks!!!

However, for me, quality matching worked much better in SickRage than in Medusa. I've never had issues with SickBeard or SickRage, but I've many issues now with Medusa. This is quite unfortunate, because it breaks the whole idea of automating everything. Something is fetched, downloaded… then nothing happens. I'll ask myself where the EP's are and I've to login to Medusa. I manually post process the items and see errors (wrong quality, not preferred quality, and so on). Then I've to "force" it to get it right. After that I've to manually update the quality in the TV show, because it's detected as `SD`, and Medusa will overwrite it at any time. This UX isn't very nice and (at least for me) it got worse since I switched from SickRage to Medusa. But again, apart from that, it works like magic!

What I get now, and I hope I understand this correctly, is:

- Medusa uses the filename to detect the quality
- This decision was primarily based on the performance, to make post processing it faster (parsing filenames > parsing metadata / video files)
- Medusa doesn't / didn't use the full path, which led to `n/a` or `SD` qualities when the quality was declared in the directory instead of the filename / basename (see #6967)
- Medusa uses the external `guessit` module to do that parsing

What I don't get is:

- Why I've to install `mediainfo` and `ffmpeg`?
- When is `mediainfo` used over `guessit`?

Probably somebody can shed some light on it :)

Here are my thoughts. Wouldn't it be great when:

- There's a fallback to `mediainfo` when filename parsing (aka `guessit`) fails
- Or the user has a setting to select the preferred source for quality detection (`filename` vs `metadata` / `video file`)
- Probably mention that `filename` (aka `guessit`) is faster than `metadata` (aka `mediainfo`)

That's just my two cents, from a "dummy user" point of view and that's what I'd love to see :)

medariox:
> Why I've to install mediainfo and ffmpeg?

You don't have to. It's completely optional.

> When is mediainfo used over guessit?

Never. The parsing goes like this:
guessit > legacy name parser (what SB/SR uses) > knowit (mediainfo, ffmpeg, enzyme)

> Here are my thoughts. Wouldn't it be great when:
> 
> There's a fallback to mediainfo when filename parsing (aka guessit) fails

This is how it currently works.

> Or the user has a setting to select the preferred source for quality detection (filename vs metadata / video file)

Another option? Not a great idea for the UX.

> Probably mention that filename (aka guessit) is faster than metadata (aka mediainfo)

Why? It already tries the faster parsing first.


This is just to answer your questions. I'm not going to explain why guessit is better compared to the legacy parser, because I think it's obvious. You are experiencing a bug with how your files are named, I get it. We are working on it.

domibarton:
@medariox I didnt' want to offend anyone, nor did I want to stress you. As mentioned before, I love the product and this is not about "what is better". I also know what kind of endurance it needs to support FOSS. Thanks, I really appreciate your effort!

Until now I just didn't knew there's a fallback when `guessit` fails. I thought if it fails (i.e. because of the bug), the default quality is `SD` and there's no fallback. But as there's already a fallback in place, this is exactly what I tried to suggest :)

So, it was just me not knowing about the post processing… sorry!

p0psicles:
@medariox what's currently the definition of a guessit `failure` ?

p0psicles:
Tbh I was in the impression that when ffmpeg was installed it was always used.

medariox:
@domibarton 
Nothing to be sorry about! We are always looking for suggestions to improve Medusa and the issue you have raised was indeed justified. I have now made some changes to the legacy parser that allow it to parse the parent folder if the file name parsing fails, without causing any regression. This should definitively fix your issue and generally improve the parsing altogether.

@p0psicles 
`guessit` needs at least a resolution and a source to convert the information to a valid quality.
knowit will be used if ffmpeg or mediainfo is installed (for .mkv files that's not even needed, because it uses enzyme), but always as last option, because it is considerably slower and uses more resources compared to the name parsers. This is where it all happens: https://github.com/pymedusa/Medusa/blob/74c3f124fb599ad923e23f46fc320778b8b4217c/medusa/common.py#L268-L286
quality_from_name = legacy parser
_extend_quality = knowit

p0psicles:
Aah ok. Zo with this pr it will go more often to knowit. As some cases will return now unknown instead of Sd right?

medariox:
In some cases yes, it will avoid that the quality is wrongfully detected as SD. So it will try to parse the folder and finally knowit as last resort.

p0psicles:
Lgtm but I can't approve

PR #7090, opened 2019-08-26
Add healthz endpoint

DJMatus23:

This PR adds a `healthz` endpoint that responds with a 200 status code if all schedulers are alive and either doesn't respond (if Medusa isn't running) or returns a general HTTP/500 if any of the schedules have failed.

Hi, this is my first PR for Medusa, I've been using it for ages now and it's been great, then recently I started running it in my rpi kubernetes cluster, I got a dodgy file that killed the thread the post processor ran in, hence me needing to implement a health-check endpoint.  It's not the best code in the world, so probably needs some refining, but I'm keen to get some feedback and with a bit of work get this merged.

### DIFF
--- a/medusa/server/web/core/__init__.py
+++ b/medusa/server/web/core/__init__.py
@@ -20,6 +20,7 @@ from medusa.server.web.core.base import (
 from medusa.server.web.core.calendar import CalendarHandler
 from medusa.server.web.core.error_logs import ErrorLogs
 from medusa.server.web.core.file_browser import WebFileBrowser
+from medusa.server.web.core.healthz import Healthz
 from medusa.server.web.core.history import History
 from medusa.server.web.core.schedule import Schedule
 from medusa.server.web.core.token import TokenHandler
--- a/medusa/server/web/core/healthz.py
+++ b/medusa/server/web/core/healthz.py
@@ -1,0 +1,29 @@
+# coding=utf-8
+"""Route to healthcheck (healthz) endpoint."""
+
+from __future__ import unicode_literals
+
+from medusa.server.web.core.base import WebRoot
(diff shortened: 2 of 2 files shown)


DJMatus23:
I see some lint errors that I caused, I've updated my commit to hopefully fix those.  I'll check back later or tomorrow once the CI job has completed.

p0psicles:
Look here https://github.com/pymedusa/Medusa/blob/master/medusa/server/api/v2/config.py#L928

DJMatus23:
No worries, I'll update this evening so it's a sub-route of the v2 api. Is the rest of it ok? Also, should I add that to the coding standards?

p0psicles:
Can't you first check if you don't have all info in the apiv2? I mean I'm not sure we need another redundant route?

p0psicles:
I'm closing this as honestly the threads are already available in apiv2. So you have everything you need for a health check already.

DJMatus23:
@p0psicles sorry about the slow response on this, I totally agree, I've been looking at other ways to check the health of Medusa without adding it to the main routes or via the apiv2, and I reckon I can create a small python sidecar that listens on a different port and just queries the API.

This would only be useful when used in a container orchestration system like Kubernetes.  Would you be interested in the yaml manifest I end up with for this?  It's similar to a docker-compose file.

p0psicles:
Sure. You can share here.

drallgood:
@DJMatus23 Yes please! I'm trying to achieve the same thing

drallgood:
Also I disagree. There should be an explicit health check endpoint, as is recommended by the CNCF and others

p0psicles:
Well the AFDW recommends otherwise. As does the SIUFO.

Issue #7105, opened 2019-09-02
Custom numbering search reverts back


homegreewn:
I have this exact same issue, which was resolved apparently but for me it is still broken. Any advice is greatly appreciated. 

https://github.com/pymedusa/Medusa/issues/1451

When I add custom numbering it reverts back, and in the logs it is indeed searching for original numbering and not custom.

medariox:
Are you sure you are not using scene numbering? That setting overwrites your custom numbering.

homegreewn:
Yes
![image](https://user-images.githubusercontent.com/11887610/64117730-8d6d9d80-cd5b-11e9-88f5-6e677aa06417.png)

Unless there is a global scene numbering setting for ALL shows that I am unaware of.

medariox:
When does it reset exactly? Does it reset in the UI?

homegreewn:
Yes. After changing number as soon as I click elsewhere it reverts to original.

medariox:
What you need to do is:
1. Enable Scene Numbering
2. Set the numbering (set it again if it is already set)

That should make it work correctly. This will be improved with: https://github.com/pymedusa/Medusa/pull/7106

homegreewn:
Doesn't work either, I'll wait it out until #7106 and post then if it got fixed for me. Thank you.

PR #7106, opened 2019-09-02
Allow to set scene numbering even if disabled

medariox:

It will be set, but won't be used until enabled.

Closes https://github.com/pymedusa/Medusa/issues/7105

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -660,7 +660,7 @@ class GenericProvider(object):
         episode_string = show_scene_name + self.search_separator
 
         # If the show name is a season scene exception, we want to use the episode number
-        if episode.series.is_scene and show_scene_name in scene_exceptions.get_season_scene_exceptions(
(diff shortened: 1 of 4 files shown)


medusa/providers/generic_provider.py
```diff
@@ -659,13 +659,13 @@ def _create_anime_search_string(self, show_scene_name, episode, search_string, a
         """Create a search string used for as anime 'marked' shows."""
         episode_string = show_scene_name + self.search_separator
 
-        # If the show name is a season scene exception, we want to use the indexer episode number.
-        if (episode.scene_season > 0 and
-                show_scene_name in scene_exceptions.get_season_scene_exceptions(episode.series, episode.scene_season)):
```
p0psicles: I'm not sure about this. It looks kinda like your removing this logic. It was added to prevent searches with a name like `show name season II` + absolute numbering. The check for season scene exception is gone now right?

medariox: Why is it gone? I just replaced `episode.scene_season > 0` with `episode.series.is_scene` because it seemed a more explicit and correct way to make this check.

PR #7305, opened 2019-11-02
Add field `identfier` to provider cache tables.

p0psicles:
Currently by default use the url for the identifier. In future this can be overwritten per provider.
NOTE! **This change will remove all provider cache tables, you will lose your cache**

This could in the future fix providers like jacket and binsearch #7304

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -781,7 +781,8 @@ class GenericProvider(object):
 
         return title, url
 
-    def _get_identifier(self, item):
+    @staticmethod
+    def _get_identifier(item):
         """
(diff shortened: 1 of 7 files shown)


medusa/providers/generic_provider.py
```diff
@@ -781,6 +781,18 @@ def _get_title_and_url(self, item):
 
         return title, url
 
+    def _get_identifier(self, item):
```
p0psicles: @medariox if you want to try it out. You'll have to copy this method into the provider, where you want to set the identifier. I can try it for binsearch. But you know more about the jacket stuff.

Issue #6892, opened 2019-06-29
Update init scripts to use Python 3


sharkykh:
Update init scripts to use Python 3
### [`runscripts` folder](https://github.com/pymedusa/Medusa/tree/develop/runscripts)
- [ ] [`init.debian`](https://github.com/pymedusa/Medusa/blob/develop/runscripts/init.debian)
- [ ] [`init.fedora`](https://github.com/pymedusa/Medusa/blob/develop/runscripts/init.fedora)
- [ ] [`init.freebsd`](https://github.com/pymedusa/Medusa/blob/develop/runscripts/init.freebsd)
- [ ] [`init.gentoo`](https://github.com/pymedusa/Medusa/blob/develop/runscripts/init.gentoo)
- [ ] [`init.solaris11`](https://github.com/pymedusa/Medusa/blob/develop/runscripts/init.solaris11)
- [ ] [`init.systemd`](https://github.com/pymedusa/Medusa/blob/develop/runscripts/init.systemd)
- [ ] [`init.ubuntu`](https://github.com/pymedusa/Medusa/blob/develop/runscripts/init.ubuntu)
- [ ] [`init.upstart`](https://github.com/pymedusa/Medusa/blob/develop/runscripts/init.upstart)

kerosyn:
One problem is the minimum version of python3 that Medusa supports is 3.5, but the wiki suggest 3.6. Not all distros are including (or have available) the same version. So I don't know if hardcoding a specific 3.x version would make sense.

Would just using the generic /usr/bin/python3 work? Are there other checks in place warning the user that they have a non-supported version of python3?

I have started a branch with just swapping python2.7 to python3, but not sure if that would break things. Seems like a simple enough change without much programming experience otherwise.

sharkykh:
Thank you for taking the time.

> One problem is the minimum version of python3 that Medusa supports is 3.5, but the wiki suggest 3.6. Not all distros are including (or have available) the same version. So I don't know if hardcoding a specific 3.x version would make sense.
>
> Would just using the generic /usr/bin/python3 work? Are there other checks in place warning the user that they have a non-supported version of python3?

Minimum version is 3.5 (I should actually change this to 3.5.2 (see below).
`/usr/bin/python3` should be okay for most systems.
Yes, there is a check in place to make sure the correct Python version is used:
[...]

Issue #6891, opened 2019-06-29
Scheduled tasks indicate they're not running with Python 3


Superxpdude:
**Describe the bug**
The "server status" page in Medusa indicates that all of the scheduled tasks are "Dead" and not running. Although the logs (and the last run time) indicate that they are.

**To Reproduce**
1. Install Medusa on FreeNAS/FreeBSD 11.2
2. Start Medusa using python 3.7 instead of 2.7
3. Go to the "Server Status" page
4. See that all of the tasks have "False" marked in the "Alive" column.

**Expected behavior**
Ideally the tasks would correctly report their alive/dead state, as well at the next run times.

**Screenshots**
All tasks report "False" in the "Alive" field, however the Last Run times do seem to update at the correct intervals.
![image](https://user-images.githubusercontent.com/1364467/60386970-482ea600-9a62-11e9-9826-18acb0341726.png)

**Medusa (please complete the following information):**
 - OS: FreeBSD-11.2-STABLE-amd64-64bit-ELF
 - Branch: master
 - Commit: 74c3f124fb599ad923e23f46fc320778b8b4217c 
 - Python version: 3.7.3 (default, Jun 11 2019, 06:40:26) [Clang 6.0.0 (tags/RELEASE_600/final 326565)]
 - Database version: 44.14

**Debug logs (at least 50 lines):**
General > Advanced Settings > Enable debug
<details>

```
2019-06-29 11:35:47 DEBUG	Thread-14 :: [74c3f12] Broken providers already updated in the last hour
2019-06-29 11:35:44 DEBUG	Thread-6 :: [74c3f12] Broken providers already updated in the last hour
2019-06-29 11:33:49 INFO	SEARCHQUEUE-DAILY-SEARCH :: [74c3f12] No needed episodes found
2019-06-29 11:33:49 DEBUG	SEARCHQUEUE-DAILY-SEARCH :: [74c3f12] Seeing if we need anything from Chernobyl
2019-06-29 11:33:49 INFO	SEARCHQUEUE-DAILY-SEARCH :: [74c3f12] Beginning daily search for new episodes
2019-06-29 11:33:49 INFO	POSTPROCESSOR :: [74c3f12] Successfully processed.
2019-06-29 11:33:49 DEBUG	POSTPROCESSOR :: [74c3f12] No processable items found in folder: /media/post
2019-06-29 11:33:49 DEBUG	POSTPROCESSOR :: [74c3f12] Processing path: /media/post
[...]

PR #7406, opened 2019-11-24
FreeBSD - Run Medusa using the daemon utility

Superxpdude:

On FreeBSD (specifically FreeNAS in this case), changing from Python 2.7 to 3.7 resulted in an issue (#6891) where scheduled tasks would fail to report their status on the "Server Status" page. I determined this to be caused by using the `-d` flag when launching medusa. Removing this flag fixes that issue, but prevents Medusa from launching as a service.

This pull request changes the runscript for FreeBSD to use FreeBSD's included [daemon utility](https://www.freebsd.org/cgi/man.cgi?query=daemon&sektion=8) to run Medusa instead.
The daemon utility will handle automatically starting Medusa, as well as restarting it when it stops (this means that the "No Restart" setting under `General > Advanced Settings > Advanced Settings` must be used to prevent duplicate processes from starting).

This change also adds two new `sysrc` variables to use with the initscript. `medusa_python_dir` allows the user to specify which python version shoud be used to run medusa (defaults to `/usr/local/bin/python3.7` and indirectly fixes the FreeBSD part of #6892), and `medusa_restart_time` allows the user to specify how long the daemon utility should wait to restart medusa when it stops.

### DIFF
--- a/runscripts/init.freebsd
+++ b/runscripts/init.freebsd
@@ -21,6 +21,10 @@
 #			Default: /usr/local/medusa
 # medusa_datadir:	Data directory for Medusa (DB, Logs, config)
 #			Default is same as medusa_dir
+# medusa_restart_time: The amount of time before medusa restarts in seconds. 
+#			Defaults to 3 seconds 
+# medusa_python_dir: The directory where python is installed.
+#			Default: /usr/local/bin/python3.7
 
 . /etc/rc.subr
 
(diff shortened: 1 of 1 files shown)


p0psicles:
The auto restart could potentially break installs when a user tries to update from the UI. Therefor I don't think it is a good idea to enable this by default.
Is it possible to disable this by default?

Superxpdude:
As it turns out, the method used to ensure that the child processes have closed before finishing the shutdown also detects the new process that Medusa makes when it restarts. This new commit removes the auto-restart (and the variable for it), and sends a `QUIT` command to all child processes when stopping the service.

The only downside is that when Medusa restarts itself, it leaves a `<defunct>` process that shows up when using `ps -aux`. But it still manages to close everything properly when you use `service medusa stop`.
Ex.
```
medusa 18535 284.0  0.6 243200 206812  -  SJ   12:27   0:02.15 /usr/local/bin/python3.7 /usr/local/medusa/start.py -q --nolaunch --pidfile /var/run/Pymedusa/child.pid --datadir /usr/local/medusa
medusa 18476   0.0  0.0   6388   2104  -  IsJ  12:27   0:00.00 daemon: /usr/local/bin/python3.7[18477] (daemon)
medusa 18477   0.0  0.0      0      0  -  ZJ   12:27   0:02.50 <defunct>
```

medariox:
LGTM now, I think that having a defunct process lying around for a bit is totally acceptable. Thanks for the great help!

PR #7456, opened 2019-12-06
Flag files as ignoreable when they have been processed.

m1dnight:
When a file is being processed it is not marked on the filesystem as processed. When a video file is in a rar it needs to be extracted to see if it should or should not be post-processed again.

A file with the basename and the extension `ignore` can shortcircuit this behaviour and avoid the unrar. 

The rationale behind this change is that this way you don't need to setup additional scripting to move freshly downloaded torrents from the download dir to the post-processing dir. This way the post-processing dir can just keep working in the download directory. 

This PR is a draft and is made to get some initial feedback on the idea.

### DIFF
--- a/medusa/helpers/__init__.py
+++ b/medusa/helpers/__init__.py
@@ -92,6 +92,14 @@ def indent_xml(elem, level=0):
         if level and (not elem.tail or not elem.tail.strip()):
             elem.tail = i
 
+def is_ignore_file(file_path):
+    """
+    Checks if the file marks that another file should be ignored.
+    E.g., Movie.ignore tells the post-processor to ignore Movie.mkv or Movie.avi.
+    :param filename: Path of the file to check.
+    """
+    return get_extension(file_path) == "ignore"
+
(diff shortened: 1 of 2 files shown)


p0psicles:
Okay I'm just thinking out loud here. The downside of this approach is, that you'll get allot of ignore files that need to be managed.

Normally when a user wants to clean up some of it's video files they'll go into their torrent client and remove the torrent including data. But that will leave the ignore files.

Alternative approach would be to keep track of the base filenames and dirs processed. Downside there is that we already kinda do that but at a few steps later in the processing.

Just some things to think over

m1dnight:
I was mulling it over yesterday as well. An alternative approach would be
indeed to keep it in the database. Seems like a reasonable thing to do.
I will have a look at the DB schema and see where I could potentially fit
it in.

I think the fact that having these ignore files avoids the torrent client
of cleaning up is indeed a big no-go.
Give me some more time on this, I'll get it working.

PR #7460, opened 2019-12-08
Bring back code to use season images for TVDB API

medariox:

It had been removed because theTVDB API was faulty. They fixed it now so we can keep getting the season images.


p0psicles:
Was this broken?

medariox:
Yeah, I had to remove it because sub_key was returning `graphical` instead of the season no.

Issue #7462, opened 2019-12-08
Problem after current update Master


ripper1967:
**Describe the bug**
After the current update to master when clicking on forced search for a show it redirects to a success page

**To Reproduce**
1. Go to '...' Schedule page
2. Click on 'a missed show forced search.' 
3. it redirects to a results success page.
![sucess](https://user-images.githubusercontent.com/7034303/70398018-20fedc00-19e5-11ea-9788-d0a5ed7bf24d.png)

Branch: masterCommit: 8b7f433d4dce89058a8b9b62cc9bcd7a162a3344
Version: 0.3.8Database: 44.14
Python Version: | 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)]
SSL Version: | OpenSSL 1.1.0j 20 Nov 2018
OS: | Windows-10-10.0.18362-SP0
Locale: | en_US.cp1252

*Debug logs 
https://gist.github.com/ripper1967/233acc0341139a0c25577a8e97e26b2e

jandreaucodes:
I can confirm I'm seeing this too.

It looks like the Ajax search isn't working properly and is redirecting to the JSON results, instead of handling it properly.

sharkykh:
Fixed on develop with #7512

PR #7498, opened 2019-12-15
Update AppLink to enable native mouse navigation

sharkykh:
Add a `href` attribute to enable native mouse navigation (middle click, ctrl+click, etc.)
(This is for when AppLink is a RouterLink, it's annoying that you can't open it in a new tab)


OmgImAlexis:
Looks like this broke manual searching. Links as such no longer send a background request, instead the link is followed.

http://192.168.1.21:8081/home/searchEpisode?indexername=tvdb&seriesid=351134&season=2&episode=3

themes-default/slim/src/components/helpers/app-link.vue
```diff
@@ -124,7 +124,9 @@ export default {
                 if (window.loadMainApp) {
                     return {
                         is: 'router-link',
-                        to: matchingVueRoute.fullPath
+                        to: matchingVueRoute.fullPath,
+                        // Add a `href` attribute to enable native mouse navigation (middle click, ctrl+click, etc.)
+                        href: new URL(matchingVueRoute.fullPath, base).href
```
OmgImAlexis: This is likely needed to be moved down to the below block, if it's added here it'll need to have some kind of check since not all `AppLink` components should be openable in another tab.

sharkykh:
It was already reported before this got merged #7462 so I don't think that this change caused that. Haven't had time to check though.

themes-default/slim/src/components/helpers/app-link.vue
sharkykh: Well currently it only affects converted pages' links to other converted pages (like config-general to config). It's possible there's an edge case I missed.

PR #7521, opened 2019-12-23
Update YggTorrent Domain

adaur:
YggTorrent updated its domain name once again.


medariox:
Thanks! The provider should still work with the old domain, so there is no need to hurry getting this merged to master.

PR #7040, opened 2019-08-07
Feature/qbittorrent api v2

p0psicles:
- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

Implement #6505

PR #4219, opened 2018-05-19
Encode the filename utf-8 before sending it to q-bittorrent.

p0psicles:
* Added logging to show part of the files send, for torrent client requests.

- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

Fixes #4218

Issue #4218, opened 2018-05-18
Failed to send torrent to qBittorrent if theres an accent in the name


DarkSupremo:
**Branch/Commit:** develop
**OS:** Windows 10

I'm developing an new provider (BJ-Share) and after many tests i discovered that if the torrent name include accents, i receive the telegram message just fine saying about the download, but its not added to qbittorrent. if i remove the accents from the torrent name, it download just fine.

I tested with Black Hole and the torrent was saved fine, even with accents, so must be some problem in the API Call to qBittorrent (i'm not sure if that happens to others torrent clients)

**Logs:**

That's the response from qbittorrent when i try to send an torrent with accent:

```2018-05-18 13:04:06 INFO     SNATCHQUEUE-MANUALSNATCH-331753 :: [71edd70] Beginning to manual snatch release: Black.Clover.-.S01E32.MKV./.H.264./.AAC./.HDTV./.Legendado./.Japonês./.1080p
2018-05-18 13:04:06 INFO     SNATCHQUEUE-MANUALSNATCH-331753 :: [71edd70] Downloading Black.Clover.-.S01E32.MKV./.H.264./.AAC./.HDTV./.Legendado./.Japonês./.1080p with 8 seeders and 0 leechers and size 547.08 MB from BJ-Share
2018-05-18 13:04:07 DEBUG    SNATCHQUEUE-MANUALSNATCH-331753 :: [71edd70] GET URL: https://bj-share.me/torrents.php?action=download&id=153167&authkey=a04a6d622a11a69cacc0bc6180c14c13&torrent_pass=n8y6e8lkljrp78gixy4651vutqueo2pg [Status: 200]
2018-05-18 13:04:07 DEBUG    SNATCHQUEUE-MANUALSNATCH-331753 :: [71edd70] User-Agent: Medusa/0.2.2 (Windows; 10; cdb05a61-5ab4-11e8-92f1-7cdd9083bde4)
2018-05-18 13:04:07 DEBUG    SNATCHQUEUE-MANUALSNATCH-331753 :: [71edd70] Calling qbittorrent Client
2018-05-18 13:04:07 DEBUG    SNATCHQUEUE-MANUALSNATCH-331753 :: [71edd70] GET URL: http://localhost:9496/version/api [Status: 200]
[...]

PR #7533, opened 2019-12-28
Fix for an reintroduced bug: Failed to send torrent to qBittorrent if theres an accent in the file name (#4218)

DarkSupremo:

This is an fix for an bug that was fixed in #4219 and then reintroduced in #7040
The problem happens when there's an torrent with accents in file name, it will not be sent to qbittorrent, as already discussed in the pull request #4219 the filename in the api call is not necessary (tested again in v2 and still work without it) and cause it to fail when there's accents.


sharkykh:
Sorry about that (it was my branch) and thank you!

Issue #7575, opened 2020-01-11
Post processing not seeing lots of episode 1 shows


skarekrow:
**Describe the bug**
It seems recently a handful of episode 1's that have started airing for some anime are not being seen by the post processor as having enough information to post process

**To Reproduce**
1. Go to 'Shows -> Manual Post-Processing'
2. Find the download directory of Oda Cinnamon nobunaga
3. Process

**Expected behavior**
I expected this to process this as episode 1.

**Medusa (please complete the following information):**
 - OS: docker
 - Branch: master
 - Commit: tag - 0.3.9
 - Python version: 3.8.1
 - Database version: 44.14

**Debug logs (at least 50 lines):**
General > Advanced Settings > Enable debug
<details>

```
2020-01-11 14:28:35 WARNING	Thread :: /nas/downloads/nzb/Series/[HorribleSubs] Oda Cinnamon Nobunaga - 01 [1080p]/[HorribleSubs] Oda Cinnamon Nobunaga - 01 [1080p].mkv: Processing failed: Not enough information to determine what episode this is
2020-01-11 14:28:35 WARNING	Thread :: Problem(s) during processing, failed for the following files/folders: 
2020-01-11 14:28:35 INFO	Thread :: Not deleting folder /nas/downloads/nzb/Series/[HorribleSubs] Oda Cinnamon Nobunaga - 01 [1080p] found the following files: ['[HorribleSubs] Oda Cinnamon Nobunaga - 01 [1080p].mkv']
2020-01-11 14:28:35 WARNING	Thread :: Processing failed for /nas/downloads/nzb/Series/[HorribleSubs] Oda Cinnamon Nobunaga - 01 [1080p]/[HorribleSubs] Oda Cinnamon Nobunaga - 01 [1080p].mkv: Not enough information to determine what episode this is
2020-01-11 14:28:35 DEBUG	Thread :: Episode doesn't have a season number, but this show appears to have only 1 season, setting season number to 1...
2020-01-11 14:28:35 DEBUG	Thread :: Show information parsed from file name
2020-01-11 14:28:35 DEBUG	Thread :: Parse result (release_group): HorribleSubs
2020-01-11 14:28:35 DEBUG	Thread :: Parse result (air_date): None
2020-01-11 14:28:35 DEBUG	Thread :: Parse result (ab_episode_numbers): [1]
2020-01-11 14:28:35 DEBUG	Thread :: Parse result (episode_numbers): []
[...]

PR #7577, opened 2020-01-11
Fix parse result removing episodes if they are like absolute episodes

medariox:

No idea why this was like this. Fixes https://github.com/pymedusa/Medusa/issues/7575
Also fixes https://github.com/pymedusa/Medusa/issues/7576


p0psicles:
I don't know. This breaks some of the unit tests. So that's kind of a red flag

medariox:
> Where these `if` `then` blocks purpose less before?

Yeah, they were just for logging 😕

p0psicles:
That definitely wasn't the case at first

medariox:
@p0psicles 
Should be all good now. The main issue was that we didn't account the fact that our guessit parses anime shows with season in the name correctly now (without the season). That needed a little change in the method. The tests also pass without any change.

skarekrow:
Didn't see this in the 0.3.10 release notes, but this is included I believe?

medariox:
Yes, it is.

PR #7628, opened 2020-01-25
Remove proper strings override for BTN provider

jdrouhard:
Hardly a full rewrite of the provider, but this at least enables proper/repack searches for BTN. Verified this worked with at least the REPACK of the first episode of Star Trek: Picard.

### DIFF
--- a/medusa/providers/torrent/json/btn.py
+++ b/medusa/providers/torrent/json/btn.py
@@ -70,9 +70,6 @@ class BTNProvider(TorrentProvider):
             'base_url': 'https://api.broadcasthe.net',
         }
 
-        # Proper Strings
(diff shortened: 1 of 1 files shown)


p0psicles:
You should use a real python list. Not a string with |

medariox:
@p0psicles 
Depends, if the provider accepts the OR `|` it's more efficient to use it. But at that point you can also remove the `self.proper_strings` so it uses the one from the generic class (`'PROPER|REPACK|REAL|RERIP'`).

jdrouhard:
I will update the PR to just remove it. The pipe did work to match a repack with only `REPACK` in the name.

PR #7699, opened 2020-02-05
Fix guessit parser not using scene exceptions

medariox:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Improvements
 
 #### Fixes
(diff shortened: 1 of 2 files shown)


p0psicles:
Did this fix a particular issue? Exceptions returns the same as aliases right? I agree though that's it's better to use aliases

medariox:
No, sadly it doesn't return anything at all (empty set), unless aliases is explicitly set and I don't think that's done anywhere... https://github.com/pymedusa/Medusa/blob/master/medusa/tv/series.py#L600

PR #7790, opened 2020-02-28
Fix show updater not updating shows without recent updates

medariox:

Shows will always be updated after max. 12 weeks or when the indexer has an update for them.


medusa/tv/series.py
```diff
@@ -875,77 +882,6 @@ def get_episode(self, season=None, episode=None, filepath=None, no_create=False,
 
         return ep
 
-    def should_update(self, update_date=datetime.date.today()):
```
medariox: This is not used.

PR #7862, opened 2020-03-19
Load providers dynamically

medariox:

This is the first step to allow to separate providers from the rest of the code base.
So that it will be possible to have user created / custom providers, which will be maintained separately.
Also removed TNTvillage since it's dead and updated a few provider names.


medusa/providers/__init__.py
```diff
@@ -3,78 +3,29 @@
...
+modules = [os.path.join(dirname, 'torrent', kind) for kind in torrent_kinds] + [os.path.join(dirname, 'nzb')]
+
+provider_names = []
+for (module_loader, name, ispkg) in pkgutil.iter_modules(modules):
+    base = os.path.basename(module_loader.path)
+    if base != 'nzb':
+        base = os.path.basename(os.path.dirname(module_loader.path)) + '.' + base
+    module = importlib.import_module('.' + name, __package__ + '.' + base)
+    if getattr(module, 'provider', None):
+        provider_names.append(module.provider.name.lower())
```
sharkykh: Couldn't you use the filename of the provider file (module name) instead? Using the provider name is a bit risky if it contains non-ASCII characters.
Also this way you don't need to change provider names (that users see).

medariox: Good point! I'm not even sure why I went for the provider name in the end... 🤔 The module name is definitely the better choice.

p0psicles:
Nice PR guys! Keep it up!

medariox:
Anyone wants to approve?

PR #7869, opened 2020-03-22
Load show display seasons only when needed

medariox:

This will allow to only load the seasons that are really needed in the show display page. It avoids the "background loading" that we have so many complains about. neededSeasons() calculates the seasons to load depending on the rows to display x page and the amount of episodes of each season. This is my first attempt at Vue code, would appreciate some feedback @p0psicles @sharkykh


p0psicles:
We can also move the specials to a separate table in this one. That way you only have to focus on the regular seasons?

p0psicles:
I can't find allot of bugs/drawbacks besides one. And that is that the select/filter checkboxes don't work on the second vue-good-table.
image

But that can be some work to fix, and for what? Dunno, maybe it's important?

p0psicles:
LGTM

Issue #4034, opened 2018-04-16
Search sub-folders for subtitles


flick1999:
It appears that Medusa is not seeing subtitles when they're in a sub-folder from the folder containing the media files. Should be an easy solution, just have it search all sub-folders within the folder containing the media files.

medariox:
You need to add a relative subtitles directory within your subtitle settings.

phraymond:
I appreciate the assist, thank you. Sadly though, I don't see where to make that setting?

medariox:
Gear > Subtitles Settings
![subs](https://snapr.pw/i/8e6ad6d4f5.png)

medariox:
It will be the subfolder where you're supposed to keep the subtitles for X show.

phraymond:
Ok, so I did understand it right. That's not going to rectify my issue, I'm
afraid. I need the post-processor to look in \queue\tvshowname\subs and
then move them along with the media to \tv\tvshowname\season1\. That's the
issue I'm having.

phoenix83:
Why is this closed? A lot of releases put subtitles in subfolders. Sickrage handles this fine.

p0psicles:
I closed all issues older then 6 months

phoenix83:
Ok I'm not confident enough to submit it myself, but I believe I may have found where the issue is.

The functionality appears to be implemented in the _search_files() method of post_processor.py:
> 
        files = []
        for root, _, filenames in os.walk(directory):
            for filename in fnmatch.filter(filenames, pattern):
                files.append(os.path.join(root, filename))
            if not subfolders:
                break

however in the _combined_file_operation() method, the calls to list_associated_files() do not explicitly set the 'subfolders' argument (which then defaults to false):
> 
        file_list = [file_path]
        if associated_files:
            file_list += self.list_associated_files(file_path, refine=True)
        elif subtitles:
            file_list += self.list_associated_files(file_path, subtitles_only=True, refine=True)

Ideally it needs to be implemented as a user option though.

phoenix83:
I'm surprised this issue isn't more popular, but perhaps the issue title is misleading. Something like "Post-processor not including sub-folders" would be better as it's more than just sub-titles.

papampi:
+1 on this
So many full season shows put their subtitles inside a sub directory and medusa not process them

medariox:
Added with: https://github.com/pymedusa/Medusa/pull/7889

phoenix83:
[...]

PR #7889, opened 2020-03-29
Search sub-folders for subtitles during post-processing

medariox:

Feature request https://github.com/pymedusa/Medusa/issues/4034

### DIFF
--- a/medusa/post_processor.py
+++ b/medusa/post_processor.py
@@ -206,7 +206,7 @@ class PostProcessor(object):
         :param refine: refine the associated files with additional options
         :return: A list containing all files which are associated to the given file
         """
-        files = self._search_files(file_path, subfolders=subfolders)
+        files = self._search_files(file_path, subfolders=subfolders, basename_only=True)
 
         # file path to the video file that is being processed (without extension)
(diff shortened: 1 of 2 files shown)


medusa/post_processor.py
```diff
@@ -1230,8 +1234,8 @@ def process(self):
                     if helpers.is_file_locked(self.file_path, False):
                         raise EpisodePostProcessingFailedException('File is locked for reading')
 
-                self.post_process_action(self.file_path, dest_path, new_base_name,
-                                         app.MOVE_ASSOCIATED_FILES, app.USE_SUBTITLES and ep_obj.series.subtitles)
+                self.post_process_action(self.file_path, dest_path, new_base_name, bool(app.MOVE_ASSOCIATED_FILES),
```
p0psicles: Specific reason to use a bool() here?

medariox: `app.MOVE_ASSOCIATED_FILES` and `ep_obj.series.subtitles` both use 1 and 0 as truthy values. I think we should be using booleans internally instead, like we usually do. A better change would be to make these changes now, but for now I wanted to make it consistent.

PR #7940, opened 2020-04-09
Run WindowsSelectorEventLoop event loop explicit on windows and py3.

p0psicles:
* Required from python3.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,18 @@ jobs:
       env:
         - TOXENV=py36,lint
       <<: *_backend_tests
+    # backend tests (py3.7) start here
+    - name: 'Backend tests (py3.7)'
+      python: '3.7'
+      env:
+        - TOXENV=py37
+      <<: *_backend_tests
+    # backend tests (py3.8) start here
+    - name: 'Backend tests (py3.8)'
+      python: '3.8'
+      env:
+        - TOXENV=py38
+      <<: *_backend_tests
     # dredd tests (py2.7) start here
     - name: 'Dredd tests (py2.7)'
       python: '2.7'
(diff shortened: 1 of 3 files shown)


sharkykh:
FYI this will not work for Python 3.5 and 3.6 as `WindowsSelectorEventLoopPolicy` is only available on 3.7.0 and newer:
https://docs.python.org/3.7/whatsnew/changelog.html#id98
> [bpo-33792](https://bugs.python.org/issue33792): Add `asyncio.WindowsSelectorEventLoopPolicy` and `asyncio.WindowsProactorEventLoopPolicy`.

More info from [Tornado](https://www.tornadoweb.org/en/stable/index.html#installation):
> On Windows, Tornado requires the `WindowsSelectorEventLoop`. This is
the default in Python 3.7 and older, but Python 3.8 defaults to an
event loop that is not compatible with Tornado. Applications that use
Tornado on Windows with Python 3.8 must call
`asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())`
at the beginning of their `main` file/function.

I suggest
```py
# We need to set the WindowsSelectorEventLoop event loop on python 3 (3.8 and higher) running on windows
if sys.platform == 'win32':
    try:
        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
    except AttributeError:  # Only available since Python 3.7.0
        pass
```

medariox:
@sharkykh 
Good catch! Mind opening a PR?

p0psicles:
From what is Read was that it was always available below 3.8. But from 3.8 needed to be set explicitly. 

Good to know.
Tnx

sharkykh:
@medariox Sure
@p0psicles It was the default up until 3.8. Tornado says that since 3.8 the default policy on Windows is not compatible with Tornado.

PR #7955, opened 2020-04-11
Feature/show sub original release

p0psicles:

As requested by @ratoaq2 

#### Reminders
- [x] Update changelog
- [x] Update api spec
- [x] Do some real testing


medusa/server/api/v2/episode_history.py
```diff
@@ -0,0 +1,140 @@
...
+        sql_base += ' ORDER BY date DESC'
+        results = db.DBConnection().select(sql_base, params)
+
+        def data_generator():
+            """Read history data and normalize key/value pairs."""
+            for item in results:
+                d = {}
+                d['id'] = item['rowid']
+
+                if item['indexer_id'] and item['showid']:
```
medariox, line 84: Can these really be missing?

medusa/server/api/v2/episode_history.py
```diff
@@ -0,0 +1,140 @@
...
+                provider = get_provider_class(GenericProvider.make_id(item['provider']))
+                d['provider'] = {}
+                if provider:
+                    d['provider']['id'] = provider.get_id()
+                    d['provider']['name'] = provider.name
+                    d['provider']['imageName'] = provider.image_name()
+
+                yield d
+
+        if not len(results):
```
medariox, line 107: I don't think len() is needed here

medusa/server/api/v2/history.py
```diff
@@ -0,0 +1,99 @@
...
+        def data_generator():
+            """Read log lines based on the specified criteria."""
+            start = arg_limit * (arg_page - 1) + 1
+
+            for item in results[start - 1:start - 1 + arg_limit]:
+                d = {}
+                d['id'] = item['rowid']
+
+                if item['indexer_id'] and item['showid']:
+                    d['series'] = SeriesIdentifier.from_id(item['indexer_id'], item['showid']).slug
```
medariox, line 61: same as above

medusa/server/api/v2/history.py
```diff
@@ -0,0 +1,99 @@
...
+                provider = get_provider_class(GenericProvider.make_id(item['provider']))
+                d['provider'] = {}
+                if provider:
+                    d['provider']['id'] = provider.get_id()
+                    d['provider']['name'] = provider.name
+                    d['provider']['imageName'] = provider.image_name()
+
+                yield d
+
+        if not len(results):
```
medariox, line 82: same as above

medusa/server/api/v2/history.py
p0psicles: Here it's required. As for example for the history.mako page, we just want to get all history information.

medusa/server/api/v2/episode_history.py
p0psicles: For an episode it can't.
But i've been overthinking the decistion to create the episode_history.py route. (`apiv2/history/tvdb12345/episode/s01e01`). But maybe I should just use a parameter to specify an episode?

medariox: I think it's consistent with the rest of the routes. But since they can't be missing for episode, we should raise an exception or leave the check out completely?

medusa/server/api/v2/history.py
medariox: But still indexer_id and showid must be in the DB?

p0psicles: ow yeah good point

medusa/server/api/v2/history.py
```diff
@@ -0,0 +1,100 @@
...
+
+            sql_base += ' WHERE indexer_id = ? AND showid = ?'
+            params += [series_identifier.indexer.id, series_identifier.id]
+
+        sql_base += ' ORDER BY date DESC'
+        results = db.DBConnection().select(sql_base, params)
+
+        def data_generator():
+            """Read log lines based on the specified criteria."""
+            start = arg_limit * (arg_page - 1) + 1
```
medariox, line 57: If you remove the +1 here you can remove the two -1 below

medusa/server/api/v2/history.py
```diff
@@ -0,0 +1,100 @@
...
+
+            for item in results[start - 1:start - 1 + arg_limit]:
```
medariox, line 59: What if there are less results than limit? Possible list index out of range?

p0psicles: I don't think that's possible, as it's slicing the results

medusa/server/api/v2/history.py
p0psicles: Hmm, not sure. For ex. Page 1, with a Limit of 50. Will result in 1, page 2 will result in 51

p0psicles: There was a reason I did it like this. Can't remember. I'll try it out

p0psicles: There was a reason I did it like this. Can't remember. I'll try it out

p0psicles: I'll try it out

dredd/api-description.yml
```diff
@@ -994,6 +994,33 @@ paths:
...
+            path-params:
+              seriesid: asdf
+          x-disabled: true
+        404:
+          $ref: '#/responses/error'
+          description: Series not found
+          x-request:
+            path-params:
+              seriesid: tvdb999999999
+          x-disabled: true
```
p0psicles, line 1023: @medariox @sharkykh yes i'm disabling these.
I'm not getting it to work. If some of you want to spend time on this, feel free.

Issue #7947, opened 2020-04-10
[APP SUBMITTED]: RuntimeError: dictionary changed size during iteration


XxUnkn0wnxX:
### INFO
**Python Version**: `3.7.7 (default, Mar 10 2020, 15:43:33) [Clang 11.0.0 (clang-1100.0.33.17)]`
**Operating System**: `Darwin-19.4.0-x86_64-i386-64bit`
**Locale**: `UTF-8`
**Branch**: [develop](../tree/develop)
**Database**: `44.15`
**Commit**: pymedusa/Medusa@3fc2e3f212f3451810076198150796c8ef744b97
**Link to Log**: https://gist.github.com/498d980f26e3cf8fe8ca92d8a047d5d2
### ERROR
<pre>
2020-04-11 03:56:08 ERROR    SEARCHQUEUE-DAILY-SEARCH :: [Il Corsaro Nero] :: [3fc2e3f] DailySearchQueueItem Exception, error: RuntimeError('dictionary changed size during iteration')
Traceback (most recent call last):
  File "/Users/ovidijus/Apps/Medusa/<a href="../blob/3fc2e3f212f3451810076198150796c8ef744b97/medusa/search/queue.py#L266">medusa/search/queue.py</a>", line 266, in run
    found_results = search_for_needed_episodes(self.scheduler_start_time, force=self.force)
  File "/Users/ovidijus/Apps/Medusa/<a href="../blob/3fc2e3f212f3451810076198150796c8ef744b97/medusa/search/core.py#L526">medusa/search/core.py</a>", line 526, in search_for_needed_episodes
    cur_provider.cache.update_cache(scheduler_start_time)
  File "/Users/ovidijus/Apps/Medusa/<a href="../blob/3fc2e3f212f3451810076198150796c8ef744b97/medusa/tv/cache.py#L244">medusa/tv/cache.py</a>", line 244, in update_cache
    parsed_result = NameParser().parse(item['title'])
  File "/Users/ovidijus/Apps/Medusa/<a href="../blob/3fc2e3f212f3451810076198150796c8ef744b97/medusa/name_parser/parser.py#L390">medusa/name_parser/parser.py</a>", line 390, in parse
    result = self._parse_string(name)
  File "/Users/ovidijus/Apps/Medusa/<a href="../blob/3fc2e3f212f3451810076198150796c8ef744b97/medusa/name_parser/parser.py#L297">medusa/name_parser/parser.py</a>", line 297, in _parse_string
[...]

PR #7956, opened 2020-04-12
Fix size changed during iteration.

p0psicles:
`dict.values()` is a generator. So it can cause errors when the dict is changed during iteration.

Prevent this by casting to an array, before iterating. Most py2 backwards compatible solution.

Fix for #7947


p0psicles:
ow lol, i realize I just broke py2 with my .values()

PR #7967, opened 2020-04-14
Feature/enable filesize upgrades

mlsmithjr:
Many times I want to upgrade the quality of media with larger file versions at higher bit rates. Currently Medusa prevents this as it thinks the quality (1080p) is the same.  
The new Post Processing config toggle, which defaults to Off, allows users like me to override this default logic.


medariox:
I must say I don't see the usefulness of this. If you want to upgrade to a different release, you can just use manual search and it will overwrite files with the same quality. How do you get a file with a different size and the same quality?

mlsmithjr:
I do this all the time. When a download is triggered I get whatever file it
picks as long as it's 1080p.  If that file is 1g and the quality (upon
viewing) is bad I'll issue a manual search and pick an alternate.
Affer that is downloaded Medusa post processing rejects it because it is
1080p just like the first one.
That's my use-case.
But if there's a workaround I'm happy to go with that.

PR #7508, opened 2019-12-18
Added provider EZTV

JorgMuskens:
I hope I do this correctly as i have not much experience with github and this is my first contribution.
Hopefully someone can use this as for me the EZTV provider is one of the best out there.
I started with HTML scraping but moved to JSON as this works better.
Most likely code can be cleaned up in several places but it does the job perfectly.

Please follow these guidelines before sending a pull request:

Update your fork to the latest upstream version.
-Check

Use the develop branch to base your code off of. Create a topic-branch for your work. We will not merge your 'dev' branch, or your 'master' branch, only topic branches, coming from dev are merged.
-Oh man i hope i am doing this correctly

Follow the coding conventions of the original repository. Do not change line endings of the existing file, as this will rewrite the file and loses history.
-Check

Keep your commits as autonomous as possible, i.e. create a new commit for every single bug fix or feature added.
-Check

Always add meaningful commit messages. We should not have to guess at what your code is supposed to do.
-Check used: Added provider EZTV

One pull request per feature. If you want multiple features, send multiple PR's
-Check one of many 

If i have done this wrong let me know

PR #8004, opened 2020-04-23
Feature/add provider eztv

JorgMuskens:
Hi Guys,
I am referencing to #7508 
I checked if there was any affiliation. As far as I could see and google there is none. The only thing I could find is.

> EZTV was a TV-torrent distribution group founded in May 2005 and dissolved in April 2015, after a hostile takeover of their domains and brand by “EZCLOUD LIMITED”.

> This popular torrent site continues to operate under new ownership (EZTV.io) and releases its own torrents now. Due to its controversial history, a large number of torrent sites have banned these ‘new’ releases from their sites. 

> EZTV has been blocked in many countries. however, it has managed to hold to the list of best torrent sites.

Here is a reddit thread about quote I stated
https://www.reddit.com/r/torrents/comments/40pzuf/is_it_safe_to_use_eztv_again_after_being_overtaken/

From my point of view it looks okay. Therefore i am opening a new pull request. Feel free to check it aswell. I am still using the same code for 5 months now without any problem. Hopefully this provider can be added as it is my favorite as i think for many others.
Best Regards
Jorg

Stay save


medariox:
Thanks for the PR! Looks good, can you just fix the lint errors? https://travis-ci.org/github/pymedusa/Medusa/jobs/678687560#L1937

JorgMuskens:
@medariox i am not sure where to begin. I never used travis-ci and when i open your link i see a lot of errors. Can you point me where to look? I am a newbie on this! If i now where to look i can investigate. The onyl this is that i am using python 3.7 and not 3.6.1. How would i tst the code? As in 3.7 i am not not getting errors. The only error i have is comming from subtitle finder since last master update. Please help

JorgMuskens:
I understand where to look now. will update in a moment

p0psicles:
There like three pep8 warnings left. 
```
medusa/providers/torrent/json/eztv.py:14:1: I202 Additional newline in a group of imports. 'from medusa.helper.common import convert_size' is identified as Third Party and 'from medusa import tv' is identified as Third Party.
/home/travis/build/pymedusa/Medusa/medusa/providers/torrent/json/eztv.py:152:1: E305 expected 2 blank lines after class or function definition, found 1
/home/travis/build/pymedusa/Medusa/medusa/providers/torrent/json/eztv.py:153:1: W391 blank line at end of file
```

JorgMuskens:
I managed to fix those but now dredd fails 😕  How can i submit those

medusa/providers/torrent/json/eztv.py
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        for mode in search_strings:
+            log.debug('Search mode: {0}', mode)
+
+            for search_string in search_strings[mode]:
+
+                if mode != 'RSS':
+                    imdb_id = self.series.externals.get(mappings[10])
+                    imdb_id = imdb_id[2:]  # strip two tt's of id as they are not used by eztv
```
medariox, line 67: Please put this after the if check below, as there are shows that don't have an imdb entry and it would cause an exception otherwise

medusa/providers/torrent/json/eztv.py
```diff
@@ -0,0 +1,150 @@
...
+                        search_params['imdb_id'] = imdb_id
+                        log.debug('Search string (IMDb ID): {imdb_id}',
+                                  {'imdb_id': imdb_id})
+                    else:
+                        log.debug('IMDb ID not found')
+                        continue
+
+                # Maximum requests allowed are 1req/2sec
+                # Changing to 5 because of server clock desync
+                time.sleep(5)
```
medariox, line 78: Is this really necessary?

medusa/providers/torrent/json/eztv.py
JorgMuskens: Atually this can not be moved. Every show on eztv works based on imdb. So if a show has no IMDB the show will not be on EZTV. Therefor the else of the if statement shows log.debug('IMDb ID not found'). Please verify i do not have any shows without IMDB.

medusa/providers/torrent/json/eztv.py
JorgMuskens: Actually yes. I am working back from 30sec. It is like a debounce routine which is needed else eztv will sometimes not be able to search due to the limitation of search request they added to their API.

medusa/providers/torrent/json/eztv.py
JorgMuskens: Shows without IMDB will not be able to be search through the API. This how their API works see https://eztv.io/api/

medusa/providers/torrent/json/eztv.py
JorgMuskens: Hmmz seems to be working without it. This was from their old API. I removed it but it requires testing. I will look into that and post my findings. Either way in or out will not harm something.

JorgMuskens:
I just restored everything back to version that travis found ok but now the build is not working anymore. I do not understand why this is happening. I tried several commits to revert back but all changes have no effect, Can somebody please tell me what i am doing wrong?

medusa/providers/torrent/json/eztv.py
duramato: I belive what he was saying is to move the slicing to right after the imdb_id check, so after line 68

medusa/providers/torrent/json/eztv.py
JorgMuskens: Actually i just tested without the time.sleep and manual seacrh breaks. So it is still needed for manual search

medusa/providers/torrent/json/eztv.py
JorgMuskens: Yeah because if returns null then errors out if sliced. makes sense!

JorgMuskens: Just pushed this in latest commit but travis is failing me 😕

JorgMuskens:
Can someone check the travis errors. It is i think not code related but command related ??

OmgImAlexis:
This really shouldn’t be using the name but instead another with a warning. I feel it’s misleading to our users otherwise.

p0psicles:
I think Travis is down. Just try again later

JorgMuskens:
> This really shouldn’t be using the name but instead another with a warning. I feel it’s misleading to our users otherwise.

What do you mean? Can you elaborate what name is misleading other users?

medusa/providers/torrent/json/eztv.py
medariox: Just tested and it works without the sleep()

JorgMuskens:
Cool thanks to you guys.

JorgMuskens:
my old code contains the old eztv.io link which does not work anymore. 
How can update to eztv.re??

p0psicles:
Update the base URL

JorgMuskens:
I hope i did i correctly and i opened new pr #8708

PR #8058, opened 2020-05-05
Add feature to search by indexer id.

p0psicles:
* A search by id is tried when the search term is a digit.
* Fixed searching by alias (tvdb)

image

### DIFF
--- a/medusa/classes.py
+++ b/medusa/classes.py
@@ -344,6 +344,9 @@ class AllShowsListUI(object):  # pylint: disable=too-few-public-methods
                     if 'aliases' in cur_show:
                         series_names.extend(cur_show['aliases'].split('|'))
 
+                    if search_term.isdigit():
+                        series_names.append(search_term)
+
                     for name in series_names:
                         if search_term.lower() in name.lower():
                             if 'firstaired' not in cur_show:
@@ -352,10 +355,6 @@ class AllShowsListUI(object):  # pylint: disable=too-few-public-methods
 
                             if cur_show not in search_results:
                                 search_results += [cur_show]
-
-                    # add the show if it's a digit search term
-                    if search_term.isdigit() and str(search_term) == str(cur_show['id']):
-                        search_results += [cur_show]
 
         return search_results
 
--- a/medusa/classes.py
+++ b/medusa/classes.py
@@ -336,13 +336,13 @@ class AllShowsListUI(object):  # pylint: disable=too-few-public-methods
                 search_term = self.config['searchterm']
                 # try to pick a show that's in my show list
                 for cur_show in all_series:
-                    if cur_show in search_results:
+                    if [result for result in search_results if str(cur_show['id']) == str(result['id'])]:
                         continue
 
                     if 'seriesname' in cur_show:
                         series_names.append(cur_show['seriesname'])
-                    if 'aliasnames' in cur_show:
-                        series_names.extend(cur_show['aliasnames'].split('|'))
+                    if 'aliases' in cur_show:
+                        series_names.extend(cur_show['aliases'].split('|'))
 
                     for name in series_names:
                         if search_term.lower() in name.lower():
@@ -352,6 +352,10 @@ class AllShowsListUI(object):  # pylint: disable=too-few-public-methods
 
                             if cur_show not in search_results:
                                 search_results += [cur_show]
+
+                    # add the show if it's a digit search term
+                    if search_term.isdigit() and str(search_term) == str(cur_show['id']):
+                        search_results += [cur_show]
 
         return search_results
 
--- a/medusa/indexers/tmdb/tmdb.py
+++ b/medusa/indexers/tmdb/tmdb.py
@@ -182,12 +182,17 @@ class Tmdb(BaseIndexer):
         """
         log.debug('Searching for show: {0}', series)
 
+        results = None
         # If search term is digit's only, store it and attempt to search by id.
         show_by_id = None
-        if series.isdigit():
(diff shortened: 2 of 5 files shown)


medusa/indexers/tmdb/tmdb.py
```diff
@@ -182,11 +182,27 @@ def search(self, series):
...
-        results = self._show_search(series, request_language=self.config['language'])
-        if not results:
+        results = None
+        # If search term is digit's only, store it and attempt to search by id.
+        show_by_id = None
+
+        try:
+            if series.isdigit():
+                show_by_id = self._get_show_by_id(series, request_language=self.config['language'])
+            results = self._show_search(series, request_language=self.config['language'])
```
medariox, line 192: This will trigger a double search. Intentional?

medariox, line 192: I think we should rename `_show_search` to `_get_show_by_name` at this point.

medusa/indexers/tmdb/tmdb.py
```diff
@@ -182,11 +182,27 @@ def search(self, series):
...
+        except IndexerShowNotFound:
+            pass
+
+        if not results and not show_by_id:
             return
 
-        mapped_results = self._map_results(results, self.series_map, '|')
+        mapped_results = []
+        if results:
+            mapped_results = self._map_results(results, self.series_map, '|')
```
medariox, line 201: Wouldn't it make more sense to just map the results in `_show_search` (aka `_get_show_by_name`)?

medusa/classes.py
```diff
@@ -336,13 +336,13 @@ def select_series(self, all_series):
                 search_term = self.config['searchterm']
                 # try to pick a show that's in my show list
                 for cur_show in all_series:
-                    if cur_show in search_results:
+                    if [result for result in search_results if str(cur_show['id']) == str(result['id'])]:
```
medariox, line 339: Can we really compare the ids? What if the you have the show with the same id from another indexer?

medusa/classes.py
```diff
@@ -353,6 +353,10 @@ def select_series(self, all_series):
                             if cur_show not in search_results:
                                 search_results += [cur_show]
 
+                    # add the show if it's a digit search term
+                    if search_term.isdigit() and str(search_term) == str(cur_show['id']):
+                        search_results += [cur_show]
```
medariox, line 358: I think it also needs a check if it already exists in search results (see above).

medusa/indexers/tvdbv2/tvdbv2_api.py
```diff
@@ -157,12 +157,28 @@ def search(self, series):
...
-        if not results:
+        if not results and not show_by_id:
             return
 
-        mapped_results = self._object_to_dict(results, self.series_map, '|')
-        mapped_results = [mapped_results] if not isinstance(mapped_results, list) else mapped_results
+        mapped_results = []
+        if results:
+            mapped_results = self._object_to_dict(results, self.series_map, '|')
+            mapped_results = [mapped_results] if not isinstance(mapped_results, list) else mapped_results
```
medariox, line 177: Can we standardize this to use `_map_results` like the other indexers? Possibly in `_show_search` (aka `_get_show_by_name`)

medusa/indexers/tvmaze/tvmaze_api.py
```diff
@@ -183,11 +183,29 @@ def search(self, series):
...
-        if not results:
-            return
+        results = None
+        # If search term is digit's only, store it and attempt to search by id.
+        show_by_id = None
+
+        try:
+            if series.isdigit():
+                show_by_id = self._get_show_by_id(series, request_language=self.config['language'])
+            results = self._show_search(series, request_language=self.config['language'])
```
medariox, line 193: Same as TMDB

medusa/classes.py
p0psicles: I tested this. And as you can see in the screenshot above, it will give back both shows.

medusa/classes.py
p0psicles: The check is done above. Just below the for:

medusa/indexers/tmdb/tmdb.py
p0psicles: Yes. If a show name has only numbers, it can still be found. Don't know if they exist, but this solution is prepared for that

p0psicles: Sure I can do some refactoring

medusa/indexers/tvdbv2/tvdbv2_api.py
p0psicles: I'm not sure. Ill check out the effort

medusa/indexers/tmdb/tmdb.py
p0psicles: Second thought. Get_show_by_name is different from a search. The name implies it will return one show. I think its more clear to keep the term "search" In it

medusa/classes.py
p0psicles: This is run in the context of one indexer search

medusa/classes.py
medariox: You mean
```
if cur_show not in search_results:
    search_results += [cur_show]
```
?

The problem is if that passes the show can still be added twice.

p0psicles: That's what this is for. 
https://github.com/pymedusa/Medusa/blob/2717afdd20f40526aa9aa15faed9a598928575b2/medusa/classes.py#L339

PR #8181, opened 2020-06-08
Make use of `travis_retry`

sharkykh:
https://docs.travis-ci.com/user/common-build-problems/#travis_retry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ script: false
 _node_version: &_node_version '12.13.0'
 _nvm_install: &_nvm_install nvm install v12.13.0
 _install_yarn_bin: &_install_yarn_bin >-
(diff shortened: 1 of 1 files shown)


p0psicles:
@sharkykh on the subject. What do you think about extending renovate options. Like "postUpgradeTasks"? 
https://docs.renovatebot.com/configuration-options/?ref=hackernoon.com

sharkykh:
I don't think you can use `postUpgradeTasks` because you need `trustLevel = 'high'` and I assume it can only be set on self-hosted instances of Renovate (not the official GitHub App)...

p0psicles:
That makes sense. Would be awesome though if we could use it.

p0psicles:
I had to so some digging. But in deed. Trust level High can only be used when your running your own bot

PR #8265, opened 2020-07-25
Add decorator to cache a method's return value for x seconds.

p0psicles:
* Add the decorator to Series.prev_airdate.
* Removed conditions for prev_aired that will never trigger.

If you want to play around with the decorator. Place this in a py file.

```python
import functools
from datetime import (datetime, timedelta)


class _TimeCache(object):
    """Cache the result for x hours."""

    def __init__(self, func, seconds=3600):
        """Initialize the PlexFallback decorator."""
        self.func = func
        self.seconds = seconds
        self.timer = None
        self.cached = None

    def __get__(self, obj, objtype):
        """Support instance methods."""
        return functools.partial(self.__call__, obj)

    def __call__(self, *args, **kwargs):
        """
        Return a cached value, if the timer hasn't expired yet.

        Fixed the timeout to 1 hour.
        """
        if self.timer and self.timer > datetime.now() - timedelta(seconds=self.seconds):
            # Timer not expired yet
            return self.cached
        else:
            # Timer expired
            self.cached = self.func(*args, **kwargs)
            # Set new timer
            self.timer = datetime.now()
            return self.cached


# wrap _TimeCache to allow for deferred calling
def time_cache(function=None, seconds=3600):
    """
    Wrap function for the TimeCache decorator class.

    Note! Pass the optional parameter seconds as a keyword argument and not a positional argument.
    The Decorator will not work, when passed with positional args.

    Example: @time_cache(seconds=7200) # For two hours. or @time_cache # for the default 1 hour.
    """
    if function:
        return _TimeCache(function)
    else:
        def wrapper(function):
            return _TimeCache(function, seconds)

        return wrapper


@time_cache(seconds=5)
def demofunc(arg1, arg2):
    return datetime.now()

import time

for x in range(0, 9):
    print(demofunc(1, 2))
    time.sleep(1)
```
It will return new timestamps after 5 seconds

### DIFF
--- a/medusa/helpers/utils.py
+++ b/medusa/helpers/utils.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 
 import functools
 from builtins import str
-from datetime import (datetime, timedelta)
+from datetime import datetime, timedelta
 
 from dateutil import tz
 
--- a/medusa/helpers/utils.py
+++ b/medusa/helpers/utils.py
@@ -149,7 +149,7 @@ class _TimeCache(object):
     """Cache the result for x hours."""
 
     def __init__(self, func, seconds=3600):
-        """Initialize the PlexFallback decorator."""
+        """Initialize the decorator."""
         self.func = func
         self.seconds = seconds
         self.timer = None
@@ -160,11 +160,7 @@ class _TimeCache(object):
         return functools.partial(self.__call__, obj)
 
     def __call__(self, *args, **kwargs):
-        """
-        Return a cached value, if the timer hasn't expired yet.
-
-        Fixed the timeout to 1 hour.
-        """
+        """Return a cached value, if the timer hasn't expired yet."""
         if self.timer and self.timer > datetime.now() - timedelta(seconds=self.seconds):
             # Timer not expired yet
             return self.cached
--- a/medusa/helpers/utils.py
+++ b/medusa/helpers/utils.py
@@ -3,8 +3,9 @@
 """General utility functions."""
 from __future__ import unicode_literals
 
+import functools
 from builtins import str
-from datetime import datetime
+from datetime import (datetime, timedelta)
 
 from dateutil import tz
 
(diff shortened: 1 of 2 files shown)


medusa/tv/series.py
```diff
@@ -559,22 +559,19 @@ def last_update_indexer(self):
...
         return self._prev_aired
 
     @property
     def next_aired(self):
         """Return next aired episode ordinal."""
         today = datetime.date.today().toordinal()
         if not self._next_aired or self._next_aired < today:
-            if self._last_update_indexer >= today:
-                self._next_aired = self.next_episode()
+            self._next_aired = self.next_episode()
```
medariox, line 570: Shouldn't next_episode() be cached as well? E.g. when a season is over and the next is coming in a year or so. It would constantly trigger. Maybe with a a longer TTL like 12h?

medusa/tv/series.py
```diff
@@ -559,22 +559,19 @@ def last_update_indexer(self):
...
         return self._next_aired
 
     @property
+    @time_cache
```
medariox, line 574: 1h seems a bit short?

medusa/helpers/utils.py
```diff
@@ -3,8 +3,9 @@
 """General utility functions."""
 from __future__ import unicode_literals
 
+import functools
 from builtins import str
-from datetime import datetime
+from datetime import (datetime, timedelta)
```
medariox, line 8: I'd avoid the parenthesis (no need + consistency)

medusa/helpers/utils.py
```diff
@@ -142,3 +143,53 @@ def int_default(value, default=0):
     if value is not None:
         return int(value)
     return default
+
+
+class _TimeCache(object):
+    """Cache the result for x hours."""
+
+    def __init__(self, func, seconds=3600):
+        """Initialize the PlexFallback decorator."""
```
medariox, line 152: Comment doesn't match (PlexFallback)

p0psicles: oops

medusa/tv/series.py
p0psicles: Your right. I intended to use 12 hours. Let's use 12 for both?

medusa/tv/series.py
p0psicles: I did use next_airdate in stead.

PR #8267, opened 2020-07-25
Migrating from master-slave jargon

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.

### DIFF
--- a/.github/check_version.py
+++ b/.github/check_version.py
@@ -18,7 +18,7 @@ if TRAVIS:
     TRAVIS_BUILD_DIR = os.environ['TRAVIS_BUILD_DIR']
 else:
     TRAVIS_PULL_REQUEST = '1234'
-    TRAVIS_PR_TARGET_BRANCH = 'master'
+    TRAVIS_PR_TARGET_BRANCH = 'main'
     TRAVIS_PR_SOURCE_BRANCH = 'develop'  # or 'release/release-0.2.3'
     TRAVIS_BUILD_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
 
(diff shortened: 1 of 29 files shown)


p0psicles:
Don't know how I feel about this. Also don't want to start a discussion about this here. I'll discuss with other devs to see if were bothered by this.

p0psicles:
Just so you know. This change you did will break allot. 

It will at least break updating for everyone.

p0psicles:
I'm not opposed. But there is definitely more to it then doing some renames. 

This pr would break allot. Like the CI integration. Travis and other config files that need to be in a master branch. 

Also in the code there are direct references to master branch. Like the front end. 

Also if making a change, I would prefer main/client..i think. 

So if you're committed to figure all these things out. And properly test them, sure. 
But then you have got to do them in the smallest PR's possible. Therefore this one is closed. If you want to discuss with us. Find us on discord.

p0psicles:
I'm not opposed. But there is definitely more to it then doing some renames. 

This pr would break allot. Like the CI integration. Travis and other config files that need to be in a master branch. 

Also in the code there are direct references to master branch. Like the front end. 

Also if making a change, I would prefer main/client..i think. 

So if you're committed to figure all these things out. And properly test them, sure. 
But then you have got to do them in the smallest PR's possible. Therefore this one is closed. If you want to discuss with us. Find us on discord.

Issue #5558, opened 2018-10-30
No NZB/Torrent providers shouldn't be a warning


DrSpaldo:
Hi, I do not use NZB or Torrent providers directly from Medusa. They are transferred from Seedbox and post processed via Medusa. So, I am getting the following warning:
```
2018-10-30 22:05:20 WARNING  SEARCHQUEUE-DAILY-SEARCH :: [0c0a735] No NZB/Torrent providers found or enabled in the application config for daily searches. Please check your settings
```
&
```
2018-10-30 14:36:12 WARNING  SEARCHQUEUE-BACKLOG-340348 :: [4614efc] No NZB/Torrent providers found or enabled in the application config for backlog searches. Please check your settings
```
Is it possible that this be changed to a log entry, rather than a warning?

DrSpaldo:
Maybe it could be an option that it is a log entry rather than a warning? Then you may make everyone happy?

DrSpaldo:
That would be perfect. Thank you very much. I’m enjoying Medusa:)

p0psicles:
Ill need you to checkout a branch amd test later

DrSpaldo:
> Ill need you to checkout a branch amd test later

Hi, any luck with this branch?

p0psicles:
Forgot about it. I'll check it out.

DrSpaldo:
@p0psicles I am happy with what ever option you think would work best for me and the majority of the users. I like to keep an eye on the warnings and there is no need for any with my set up. If you think that is the best way, I am happy 🥇

p0psicles:
I can't think of any scenarios where youd want to run a (daily or backlog) search, without having any providers configured. So I'll go with that solution.

DrSpaldo:
Yep, all of my data comes from post processing as the files are dumped into that folder. So as long as it doesn't interfere with that, it sounds good to me :)

p0psicles:
@DrSpaldo mind testing out the branch referenced in the pull request?

p0psicles:
You should see some additional settings for disabling daily searches and backlog searches in config - search.

DrSpaldo:
@p0psicles Thanks for the quick reply. I am semi familiar with Github but not great. Do you mean that I should try the 'develop' branch? 

Ie. URL git://github.com/pymedusa/Medusa.git then branch develop instead of master?

p0psicles:
You'll need to checkout feature/start-stop-threads-from-config

[...]

Issue #5399, opened 2018-10-13
Disable post processor warnings for certain issues


DrSpaldo:
Hi, I have recently switched from SickRage to Medusa. So far it feels nicer. However, I have noticed that in the way I have my Medusa set up that I am getting a lot more warnings than before.

I have Medusa installed on Synology with a mounted folder that has all the TV from the seedbox. This is needed as it is shared and the only way I could get it to connect properly. So the mounted folder is listed in the post processing section. Unfortunately, as there are other shows I do not have in Medusa that other users use, it is picking them up and giving me the following error/warning:

This show isn't in your list, you need to add it before post-processing an episode

Is there is way I can turn this off as a warning?

Hope this makes sense. Thanks.,

desepticon:
You might be able to write a script that symlinks all the files processed by Medusa and puts them in a folder outside the PP folder.

kezek:
+1

neo62248:
+1

My Transmission client is seeding many downloads in the complete directory. This directory is already used by MEDUSA to postprocessing. Films that don't have to deal with MEDUSA are in this directory. I need to move them to avoid warnings. But i loose the capacity of seeding my files moved.

sspencer10:
So if anyone is looking for a workaround, I changed the log level from warning to info in this file:

https://github.com/pymedusa/Medusa/blob/master/medusa/process_tv.py

Look at lines 208 and 210

p0psicles:
I'll make some logs, level adjustable. 
Almost finished

p0psicles:
As i've nearly completed this feature in branch: feature/customize-logging
Can I get some early testers? @sspencer10 @neo62248 @kezek?

YOu'll need to checkout the branch.

sspencer10:
@p0psicles I will, but am away for the weekend. Will report back sometime next week

p0psicles:
Added to develop

PR #8561, opened 2020-10-01
Feature/customize logging

p0psicles:

implements #5558 #5399 

#### TODO
- [x] Add basic implementation
- [x] db migration
- [x] ui


medusa/process_tv.py
```diff
@@ -24,7 +24,7 @@
 from six import iteritems
 
 
-log = BraceAdapter(logging.getLogger(__name__))
+log = CustomBraceAdapter(logging.getLogger(__name__))
```
p0psicles, line 27: @medariox the way this works, is. CustomBraceAdapter is a subclass of BraceAdapter. Only in CustomBraceAdapter the log message (only the message not the kwargs) is evaluated. And if found, the log level is overwritten (only if set). 0 == unset.

But because we evaluate the message (and not the parsed message), I had to refactor process_tv.py. (which I think is not a bad thing). I removed the "old" medusa.logger('lalal', logger.WARNING). And at the same time improved the self.log() method a little, so both logging and format use the **kwargs.

p0psicles: I renamed self.log to self.log_and_output, to eliminate confusing between sef.log() and log.log()

medusa/databases/main_db.py
```diff
@@ -945,3 +945,27 @@ def execute(self):
...
 
         self.inc_minor_version()
+
+
+class AddCustomLogs(AddShowLists):
+    """Create a new table custom_logs in main.db."""
+
+    def test(self):
+        """
+        Test if the version is at least 44.15
```
medariox, line 955: Should be 44.17

medusa/databases/main_db.py
```diff
@@ -945,3 +945,27 @@ def execute(self):
...
+        return self.connection.version >= (44, 17)
+
+    def execute(self):
+        utils.backup_database(self.connection.path, self.connection.version)
+
+        log.info('Creating a new table custom_logs in the main.db database.')
+
+        self.connection.action(
+            'CREATE TABLE custom_logs '
+            '(row_id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT, '
```
medariox, line 966: Maybe log_id would be better

medusa/databases/main_db.py
```diff
@@ -945,3 +945,27 @@ def execute(self):
...
+            'identifier	TEXT NOT NULL, '
```
medariox, line 967: Double space

medusa/databases/main_db.py
p0psicles: will change

medariox (approved):

PR #8823, opened 2020-12-02
decode() bytes for saving into config.ini

### LINKED ISSUE #8810: Synology Download Station not "keeping" authentication in Medusa
I have my Medus install set up on my Synology DS 1315+ NAS.

My authentication from Medusa to my Synology diskstation will not "stick" after a day or so. Medusa will suddenly stop downloading episodes and if i go to the Search Settings and "Test connection" for Torrent search, i get:

image

But if i go up to the password field and type in the same password that was alreadyt supposed to be in the field, it works and authenticates:

image

This happens every couple of days, which means the software is useless, since i have to constantly log in. I have checked the logs but can find no signs in Medusa that the authentication suddenly fails. 

What do ?

image

p0psicles:

Fix #8810

### DIFF
--- a/medusa/__main__.py
+++ b/medusa/__main__.py
@@ -452,10 +452,7 @@ class Application(object):
             app.PRIVACY_LEVEL = check_setting_str(app.CFG, 'General', 'privacy_level', 'normal')
             # Need to be before any passwords
             app.ENCRYPTION_VERSION = check_setting_int(app.CFG, 'General', 'encryption_version', 0)
-            app.ENCRYPTION_SECRET = check_setting_str(
-                app.CFG, 'General', 'encryption_secret',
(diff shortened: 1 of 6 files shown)


p0psicles:
@medariox the scecrets where stored as "binary". Like "b'sakljdflskdjflkdsjflksdfjldskjf'"
I suspect this was the issue. But should we create a config migration for this?

medariox:
Hmm, I think there is something wrong with encode() and decode()... like the secret is getting decoded and encoded again, although it shouldn't if it is already encoded.

medariox:
@p0psicles 
No need for a migration as `b'asasqqwqwe'` and `asasqqwqwe` are both valid secrets. New installs won't have the `b'` prefix tho.

PR #8917, opened 2020-12-24
Add cache for guessit

sinnlosername:

First off, I don't have much experience with python nor writing pull requests, so I hope this is according to your standards

guessit is one of the slowest parts of the name parsing. This PR adds a cache for guessit results to improve performance if called repetitive. This happens for example when scheduled post-processing is used.

Normal name parsing already has a cache, however it's only used for successful parses (which makes sense as it also includes db searches). If there are many files in the torrent/nzb download folder which aren't related to medusa (basically any other torrent download) it will take a very long time, as files without a show will cause a parse to be unsuccessful.

### DIFF
--- a/medusa/name_parser/cache.py
+++ b/medusa/name_parser/cache.py
@@ -43,16 +43,17 @@ class BaseCache(object):
         :rtype: object
         """
         with self.lock:
-            return self.thread_unsafe_get(name)
-
-    def thread_unsafe_get(self, name):
-        """Return a cache item from the cache. This function doesn't lock and is therefore not thread safe."""
-        if name in self.cache:
-            log.debug('Using cache item for {name}', {'name': name})
-            return self.cache[name]
+            if name in self.cache:
+                log.debug('Using cache item for {name}', {'name': name})
+                return self.cache[name]
 
     def remove(self, name):
         """Remove a cache item given name."""
         with self.lock:
             del self.cache[name]
             log.debug('Removed cache item for {name}', {'name': name})
+
+    def clear(self):
+        """Removes all items from the cache."""
+        with self.lock:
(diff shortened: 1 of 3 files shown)


medariox:
First of all: thank you for the nice PR! 
While it makes a lot of sense to add this, there is caveat. Think about the situation where a title is not matched to a show and needs a scene exception to be matched correctly. This would cause the file to never be matched, even after adding the scene exception. I haven't looked at the code, but I'm afraid that's what will happen and also why this hasn't been added yet.
What we need is a way to invalidate the cached items for a show, specifically when options that alter guessit are changed (anime, scene numbering, scene exceptions,...).

p0psicles:
For the invalidating of the cache. I think we should use a decorator for the properties that should trigger the invalidation.

sinnlosername:
> While it makes a lot of sense to add this, there is caveat. Think about the situation where a title is not matched to a show and needs a scene exception to be matched correctly. This would cause the file to never be matched, even after adding the scene exception. I haven't looked at the code, but I'm afraid that's what will happen and also why this hasn't been added yet.

I checked it, you're right about the problem with changing options. I didn't realize this because for me it looks like a direct import of guessit, while in reality the import is overwritten. 

> For the invalidating of the cache. I think we should use a decorator for the properties that should trigger the invalidation.

If I understand it correctly guessit is configured only by the options object passed to the guessit call, right? If so, what if we clear the cache everytime the options object changes? This could prevent any inconsistencies.

p0psicles:
That's an even better idea. Wanna join us on discord? It's easier for discussing these things

sinnlosername:
I'm there. I've tagged you.

medariox:
Great job, thanks!

PR #8948, opened 2021-01-02
Added the CHANGE_* methods from medusa.config.py

p0psicles:
Used these methods in set properties of their app variables.

@medariox  with the vueification we lost the ability to "hot-reload" the threads and their settings.
I used @property now to accomplish the same as what we had in the server/web/config/handler.py routes.


p0psicles:
@medariox  aah great. Another one of those "pyTest is running fine locally, but not on build slave"
Will have to give this another go later.

PR #8955, opened 2021-01-04
Add UI option saveMagnetFile

p0psicles:

Implement https://github.com/pymedusa/Medusa/discussions/7407


p0psicles:
this: https://github.com/pymedusa/Medusa/pull/8955/files#diff-20bd3fcb964e966806e0e242d2576e8c13f2fbf58e0bffef2ff33d1aa0f53dbfR154 will always make the magnet -> torrent downlaod fail. I put it in for testing. Needs to be fixed before merge of course

p0psicles:
I sneaked in a fix for anidex. I bypassed DDOS protection.

p0psicles:
should all be good now

medusa/providers/torrent/torrent_provider.py
```diff
@@ -137,6 +137,43 @@ def _verify_download(self, file_name=None):
 
         return False
 
+    def _verify_magnet(self, file_name=None):
```
medariox, line 140: If `file_name` is not optional, I'd remove the `=None`.
Also maybe `file_path` is a better fitting name?

medusa/providers/torrent/torrent_provider.py
```diff
@@ -137,6 +137,43 @@ def _verify_download(self, file_name=None):
...
+            magnet_uri = fp.read()
+
+        if self._get_info_from_magnet(magnet_uri):
+            return True
+        return False
+
+    @staticmethod
+    def _get_torrent_name_from_magnet(magnet_uri):
+        """Try to extract a torrent name from a Magnet URI."""
+        torrent_name = ''
```
medariox, line 162: This could return torrent_name as `''`. Do we want this? Maybe declare it directly `torrent_name = 'NO_DOWNLOAD_NAME'`

medusa/providers/torrent/torrent_provider.py
```diff
@@ -137,6 +137,43 @@ def _verify_download(self, file_name=None):
...
+            torrent_name = re.findall('dn=([^&]+)', magnet_uri)[0]
+        except Exception:
+            torrent_name = 'NO_DOWNLOAD_NAME'
+        return torrent_name
+
+    @staticmethod
+    def _get_info_from_magnet(magnet_uri):
+        """Try to extract an info_hash from a Magnet URI."""
+        info_hash = re.findall(r'urn:btih:([\w]{32,40})', magnet_uri)[0].upper()
+        if len(info_hash) == 32:
```
medariox, line 173: What happens if `info_hash` is None or not 32 long? Maybe we should handle this.

medusa/providers/generic_provider.py
```diff
@@ -111,6 +112,27 @@ def kind(cls):
...
+        Create a .magnet file containing the Magnet URI.
+
+        :param filename: base filename (without extension).
+        :param result: SearchResult object.
+        :returns: True if the magnet file is created and verified.
+        """
+        filename_ext = '{filename}.magnet'.format(filename=filename)
+        log.info('Saving magnet file {result} to {location}',
+                 {'result': result.name, 'location': filename_ext})
+        with io.open(filename_ext, 'wb') as fp:
```
medariox, line 126: Is this really needed to be written as bytes? 
Also we can use open since we don"t need Python 2 compatibility

medusa/providers/torrent/torrent_provider.py
p0psicles: It extracts a hash of 32 or 40 chars long. If 32, it needs to do another translation.

Issue #9000, opened 2021-01-12
Mass update functionality only updates first show in selection.


BeMacized:
**Describe the bug**
When trying to use the mass update functionality, only the first show from the selection shows any changes. 
Additionally, after saving changes, I end up on the route "/manage/massEditSubmit", with no UI. Just a plain gray background. 

**To Reproduce**

1. Go to "Mass Update" under "Manage"
2. Select all shows in the "Edit" column
3. Click "Edit Selected"
4. Make any change
5. Press "Save Changes"
6. Wait for the page to finish loading
7. Navigate back to the "Mass Update" page manually (As there is no longer any UI shown)
8. Note that only the first show in your selection received the changes. 

**Expected behavior**
I would expect all shows that I selected to receive the changes.

**Medusa**
- Branch: master
- Commit: Unknown (Using docker container: `linuxserver/medusa:version-v0.5.3`)
- Version: 0.5.3
- Database: 44.17
- Python Version: 3.8.5 (default, Jul 20 2020, 23:11:29)  [GCC 9.3.0]
- SSL Version: OpenSSL 1.1.1i  8 Dec 2020
- OS: Linux-5.4.65-1-pve-x86_64-with (Docker on Proxmox)
- Locale: en_US.UTF-8

**Debug logs**
```
2021-01-12 10:59:15 ERROR	Thread :: Exception generated: tuple indices must be integers or slices, not str
Traceback (most recent call last):
  File "/app/medusa/medusa/server/web/core/base.py", line 252, in async_call
    result = function(**kwargs)
  File "/app/medusa/medusa/server/web/manage/handler.py", line 691, in massEditSubmit
    errors += self.massEditShow(
  File "/app/medusa/medusa/server/web/home/handler.py", line 959, in massEditShow
    series_obj.aliases = get_scene_exceptions(series_obj)
  File "/app/medusa/medusa/tv/base.py", line 63, in __setattr__
    super(TV, self).__setattr__(key, value)
  File "/app/medusa/medusa/tv/series.py", line 669, in aliases
    update_scene_exceptions(self, exceptions)
  File "/app/medusa/medusa/scene_exceptions.py", line 233, in update_scene_exceptions
    if exception['title'] not in exceptions_cache[(series_obj.indexer, series_obj.series_id)][exception['season']]:
TypeError: tuple indices must be integers or slices, not str
```

GamerGun:
Same here, updating in batches seems to work though, might be a limit / time out or something?

[...]

PR #9016, opened 2021-01-13
Fix scene exceptions sometimes breaking mass update functionality

BeMacized:

Please note that **I am not a Python developer**: I am not sure whether this is the correct solution, please check me on this.

This PR resolves the issue in #9000. 

From what I gathered, the fields on the TitleException tuple are being accessed incorrectly, leading to the stacktrace you can find in the issue. This should be solved with these changes.


p0psicles:
I appreciate the dodiligence! But unfortunately this change will break something else.
Fixed in #9017

PR #9038, opened 2021-01-16
Feature/trakt fixes

p0psicles:

### DIFF
--- a/medusa/tv/episode.py
+++ b/medusa/tv/episode.py
@@ -424,8 +424,9 @@ class Episode(TV):
 
     @status.setter
     def status(self, value):
-        self._status = value
-        self._sync_trakt(value)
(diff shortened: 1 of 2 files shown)


medusa/tv/episode.py
```diff
@@ -424,11 +424,16 @@ def status(self):
 
     @status.setter
     def status(self, value):
-        self._status = value
-        self._sync_trakt(value)
+        if value != self._status:
```
medariox, line 427: I don't think this is needed?

p0psicles: yeah, it's not.
Well I added it when the trakt call was sstill there, to limit the times it did api calls. But in  the end, it was still spamming traktk calls

PR #9118, opened 2021-01-26
Don't use add (year) option for backend related processes

p0psicles:
- [x] Add experimental (visual) flag

When using the config - general (Append (year) to each show title) option.
Only apply this to show folder creation and front-end.


p0psicles:
The old implementation also caused a number of bugs.
As for example a lookup to IMDB used the show.name and the year. And would concatenate them. Which would result in: `My Show (2020) (2020)`. This will fix that and a number of other issues.

medusa/tv/series.py
```diff
@@ -220,7 +220,8 @@ def __init__(self, indexer, indexerid, lang='', quality=None,
             {'episodes', 'next_aired', 'release_groups', 'imdb_info'})
 
         self.show_id = None
-        self._name = ''
+        self.name = ''
+        self._title = ''
```
medariox, line 224: What's the point of having `self._title` if it's never set?

p0psicles: good catch. I used it before. But decided to go with a different set up.
It's removed

Issue #9041, opened 2021-01-17
Display of fanart in backstretch div lags behind when navigating


Giovanni-NL:
**Describe the bug**
When you navigate from one show to another, the fanart image that is displayed in the background lags behind.  When you go from show A to show B you will see show B with the background fanart of A. If you then navigate to show C it will display show C with the fanart of show B. This goes on and on.

**To Reproduce**
1. Go to main page of Medusa. A random fanart image is displayed in the background.
2. Click on the name of any show in the list. Its details will be shown, but the background will be another random background.
3. Select any other show from the dropdown list in the header. You will then see the info of that show, but the fanart will be the fanart of the previously selected show. 
4. Go back to step 3 and repeat over and over. The fanart shown will always be the one of the show before.

**Expected behavior**
I would expect the proper fanart to be shown for a show.

**Screenshots**
If applicable, add screenshots to help explain your problem
![Medusa](https://user-images.githubusercontent.com/4085882/104846025-b81caf80-58d8-11eb-9399-1d9a7e849b71.png)
This image shows the network traffic that happens when you select a show from the dropdown in the header. It first requests info on the newly selected show(which is the good id), next it requests fanart with the id of the currently displaying show (which is the wrong id), then it requests episodes with the id of the newly selected show (which is the good id), all further requests are with the proper id in the URL. Only the request for fanart uses the wrong id in the URL.

**Medusa (please complete the following information):**
 - OS: Windows 10 Home 1909 build 18363.1316
 - Branch: master
 - Commit: release 0.5.3 (latest)
 - Python version: 3.7.3 (the one in the Medusa folder)
 - Database version: I have no idea

**Debug logs (at least 50 lines):**
This is all that is logged after restarting Medusa and navigating through various shows. Debug logging has been enabled in the advanced settings. 
General > Advanced Settings > Enable debug
<details>

```
2021-01-17 15:03:10 INFO	CHECKVERSION :: [698ca1d] No update needed
2021-01-17 15:03:08 INFO	CHECKVERSION :: [698ca1d] Checking for updates using GitHub Updater
2021-01-17 15:03:07 INFO	MAIN :: [698ca1d] Building internal name cache for all shows
2021-01-17 15:03:07 INFO	MAIN :: [698ca1d] Finished processing 3504 scene exceptions.
2021-01-17 15:03:07 INFO	MAIN :: [698ca1d] Updating exception_cache and exception_season_cache
[...]

PR #9138, opened 2021-01-28
Feature/fix duplicate load poster

p0psicles:

Fixed a really annoying bug where switching from one show to the other using the show-selector component, would result in loading a poster twice (once the old and once the new). Same would apply to the fanart and banner.

Issue is because we use the show object which we get from the vuex store. And at time of the mounted hook, the show object has not yet updated to the new show object, but the components local `id` (computed) value has.

For now i'm using a hack on a hack, which will check if the $vm.show.id[show.indexer] === $vm.id, for displaying the images.
Still a downside is, that the other show attributes like the description, flags, config settings, etc. For a second also display the old show. But this is less noticeable.


p0psicles:
fix for #9041

p0psicles:
I know this just became a huge PR. But it's easier for me to do the refactor in one go. I think I also got pretty much everything. Also tested the old mako pages and gone through some old js files.

PR #9527, opened 2021-05-12
Add backstretch to schedule.vue and history.vue components.

p0psicles:


p0psicles:
This fixes an issue where the user starts up medusa and directly after tries to open the home page.
It will be in the process of getting all the prev / next airdates from the startup, and then will again get the prev / next airdates through the api call. Running both can results in db lockups. Which will result in api call timeouts.

PR #9555, opened 2021-05-18
add support for new DownloadStation API

BenjV:


p0psicles:
Ow you did a new pr. Good.

medusa/clients/torrent/downloadstation.py
```diff
@@ -39,195 +36,135 @@ def __init__(self, host=None, username=None, password=None):
...
+    def _add_torrent(self, params, torrent_file=None):
+        """Add a torrent to DownloadStation"""
+        self.url = urljoin(app.TORRENT_HOST, 'webapi/entry.cgi')
+        if not self._request(method='post', data=params, files=torrent_file):
             return False
-
-        if self.checked_destination and self.destination == torrent_path:
+        jdata = self.response.json()
+        if jdata['success']:
+            log.info(f"Torrent added as task {jdata['data']['task_id']} to {self.name}")
```
p0psicles, line 165: we don't use template literals for logs. We use a modified paramaterized string replacement. Like how python's logging module can do it by default with it's string replacement: `log.debug('my message with %s', my_param)`. But our BraceAdapter uses a dict as the parameter.

For ex: 
```
log.debug('{name}: Response to {method} request is {response}', {
            'name': self.name,
            'method': method.upper(),
            'response': self.response.text[0:1024] + '...' if len(self.response.text) > 1027 else self.response.text
        })
```

medusa/clients/torrent/generic.py
```diff
@@ -44,67 +44,30 @@ def __init__(self, name, host=None, username=None, password=None):
...
+    def _request(self, method='get', params=None, data=None, files=None, cookies=None):
 
+        self.response = self.session.request(method, self.url, params=params, data=data, files=files, timeout=60, verify=self.verify)
         if not self.response:
-            log.warning('{name}: Unable to reach torrent client', {'name': self.name})
-            return False
-
-        if self.response.status_code == 401:
-            log.error('{name}: Invalid Username or Password, check your config', {'name': self.name})
+            log.warning(f"{self.name} {method.upper()} call to {self.url} failed!")
```
p0psicles, line 64: please use single quotes unless not possible. Like when you want to wrap a string that has a single quote in it.

medusa/clients/torrent/generic.py
```diff
@@ -133,6 +97,12 @@ def _add_torrent_file(self, result):
         """
         raise NotImplementedError
 
+    def _check_path(self):
+        """Check if the destination path is correct."""
+        log.debug('dummy check path function')
```
p0psicles, line 102: I don't think the log adds value here

medusa/clients/torrent/generic.py
```diff
@@ -232,66 +202,47 @@ def send_torrent(self, result):
         :return:
         :rtype: str or bool
         """
-        r_code = False
 
-        log.debug('Calling {name} Client', {'name': self.name})
 
-        if not self.auth:
-            if not self._get_auth():
```
p0psicles, line 240: i'm pretty sure that there are clients that need this check/call for authentication. Have you verified this with all the torrent clients?

medusa/clients/torrent/generic.py
```diff
@@ -44,67 +44,30 @@ def __init__(self, name, host=None, username=None, password=None):
...
-        if self.response.status_code == 401:
-            log.error('{name}: Invalid Username or Password, check your config', {'name': self.name})
+            log.warning(f"{self.name} {method.upper()} call to {self.url} failed!")
+            self.message = f'Connect to {self.name} on "{self.host}" failed!'
             return False
-
-        code_description = http_code_description(self.response.status_code)
-        if code_description is not None:
-            log.info('{name}: {code}', {'name': self.name, 'code': code_description})
+        if self.response.status_code  >= 400:
```
p0psicles, line 67: linter complains about double space

medusa/clients/torrent/generic.py
```diff
@@ -44,67 +44,30 @@ def __init__(self, name, host=None, username=None, password=None):
...
         self.response = None
         self.auth = None
-        self.last_time = time.time()
+        self.message = None
         self.session = ClientSession()
         self.session.auth = (self.username, self.password)
-
-    def _request(self, method='get', params=None, data=None, files=None, cookies=None):
-
-        if time.time() > self.last_time + 1800 or not self.auth:
```
p0psicles, line 57: what was your reasoning to remove the authentication caching?

medusa/clients/torrent/generic.py
```diff
@@ -44,67 +44,30 @@ def __init__(self, name, host=None, username=None, password=None):
...
-                'url': self.url,
-                'params': text_params[0:99] + '...' if len(text_params) > 102 else text_params,
-                'data': text_data[0:99] + '...' if len(text_data) > 102 else text_data,
-                'files': text_files[0:99] + '...' if len(text_files) > 102 else text_files,
-            }
-        )
-
-        if not self.auth:
-            log.warning('{name}: Authentication Failed', {'name': self.name})
+        self.verify = certifi.where()
```
p0psicles, line 55: Although I understand that it's good to verify certificates, in this case we dont know what torrent clients with what possible reverse proxies a user is using. Changing from verify=False to verify=certifi.where() can be a breaking change for users. I'm kinda in the middle of this. @medariox what do you think?

Maybe then it's better to utilize the setting `app.TORRENT_VERIFY_CERT`

medusa/clients/torrent/generic.py
```diff
@@ -44,67 +44,30 @@ def __init__(self, name, host=None, username=None, password=None):
...
+        
+        if self._get_auth():
```
p0psicles, line 57: I don't understand this one? You want to initiate authentication on the constructor? but why return false?

medusa/clients/torrent/generic.py
```diff
@@ -44,67 +44,30 @@ def __init__(self, name, host=None, username=None, password=None):
...
             return False
 
-        try:
-            self.response = self.session.request(method, self.url, params=params, data=data, files=files,
-                                                 cookies=cookies, timeout=120, verify=False)
-        except (requests.exceptions.MissingSchema, requests.exceptions.InvalidURL) as error:
```
p0psicles, line 83: What is your reasoning for removing the exception handling?

medusa/clients/torrent/generic.py
```diff
@@ -232,66 +202,47 @@ def send_torrent(self, result):
...
+        # check for the hash and add it if not there
         try:
-            # Sets per provider seed ratio
-            result.ratio = result.provider.seed_ratio()
-
-            # lazy fix for now, I'm sure we already do this somewhere else too
-            result = self._get_info_hash(result)
-
             if not result.hash:
+                raise Exception()
```
p0psicles, line 213: I don't think you need to use exceptions here. Can just do
```
if not result.hash and not self._get_info_hash(result):
    return False
```

medusa/clients/torrent/generic.py
```diff
@@ -232,66 +202,47 @@ def send_torrent(self, result):
...
-            else:
-                r_code = self._add_torrent_file(result)
-
-            if not r_code:
-                log.warning('{name}: Unable to send Torrent',
-                            {'name': self.name})
-                return False
+        if result.url.startswith('magnet:'):
+            log.info(f'Adding "{result.url}" to {self.name}')
+            r_code = self._add_torrent_uri(result)
```
p0psicles, line 220: Why did you remove these from the try/except block? As you also removed all the exception handling from the _request() method, all exception handling is removed now?

medusa/clients/torrent/generic.py
```diff
@@ -301,27 +252,16 @@ def test_authentication(self):
         :return:
         :rtype: tuple(bool, str)
         """
-        try:
```
p0psicles, line 304: have you tested the test_authentication method with other clients? Like with deluge, transmission, rtorrent, etc?

p0psicles (changes requested):

p0psicles:
I did a review on generic.py first. As this is the more critical method, as it's used by all the supported torrent clients. We need to make sure that all clients work as expected with these changes. Just from reading all these code changes in it, I'm not 100% sure it will. So the only way to know is to test it.

If you want to go through with the refactor to generic, I suggest you use docker to run them. Maybe this can be of help. When I have more time I'll go through the downloadstation.py changes. Although those looked okay at first sight. Except from the code styling and linting issues.

medusa/clients/torrent/downloadstation.py
```diff
@@ -39,195 +36,135 @@ def __init__(self, host=None, username=None, password=None):
...
-            data['destination'] = torrent_path
-        log.debug('Add torrent URI with data: {0}', json.dumps(data))
-        self._request(method='post', data=data)
-        return self._check_response()
+            # parameters "type' and "destination" must be coded as a json string so we add double quotes to those parameters
+        params = {  'api'        : 'SYNO.DownloadStation2.Task',
+                    'version'    : 2,
+                    'method'     : 'create',
+                    'create_list': 'false',
+                    'destination': f'"{app.TORRENT_PATH}"',
```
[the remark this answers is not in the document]
BenjV: As explained in the comment, the Download Station api expect some values to be formatted as a a json string so it is needed to add double quotes to them.
That's what the f-tring notation is doing

medusa/clients/torrent/generic.py
BenjV: That's a matter of opinion, I needed it during testing and debugging, but feel free to remove it.

medusa/clients/torrent/generic.py
BenjV: Yes all current clients have that function available, that's why I removed it ( I don't like superfluous code), but feel free to add it in again.

medusa/clients/torrent/generic.py
BenjV: Don't know why, this is perfectly fine python code.

medusa/clients/torrent/generic.py
BenjV: Because of the fact that the class in not used as an object but as a bunch of functions, every time a function of that GenericClient class is called, the __init__ function is called first, so all the self variables are initialized initialized fresh and that makes this attempt of caching useless.
Instead I added a call to self._get_auth in the __init__ to be sure the authorisation is done before every call.
If you want caching then you have to convert it to a real object so the __ini__ is call only once and the self variables are kept alive.

medusa/clients/torrent/generic.py
BenjV: If someone wants to use a client with an ssl connection (https) and you keep it as verify=false you actually punch a hole in the ssl security because you are saying that everything is always ok.
This creates false security.
If you don't want ssl verification you should prohibit the use of https instead of saying everything is always ok by the 
verify=False statement.

Just my two cents

medusa/clients/torrent/generic.py
BenjV: Oeps that a mistake, only this line is sufficient.
`self._get_auth()`

And as I said before the __init__ is called before every call to other functions so this is to make sure that before the actual call to fore example _add_torrent the authorization is done.
And because it is not used as an object, the session to the clients is killed after leaving this function and so a fresh authorization is needed every time.

medusa/clients/torrent/generic.py
BenjV: The function makes use of the ClientSession as defined in medusa.session.core.
The exception handling is done there, so there will never be any exception and this code is useless.
The only error handling to be needed is when you get a normal response but an status code higher or equal 400, but that is not an exception

medusa/clients/torrent/generic.py
BenjV: If the result.hash is not yet defined then this exception is needed otherwise it will kill this function  by means of an unhandled exception

medusa/clients/torrent/generic.py
BenjV: see my previous comment

medusa/clients/torrent/generic.py
BenjV: Only with transmission, but there is nothing changed for them, so I do not expect any problems.
I dont' have all those clients available.

BenjV:
Code styling is a a matter of opinion, so feel free to change it.
Be careful about changing the parameter dict, because as I said they are the tricky part of the DownloadStation api.

I use a lot of f-string formatting instead of the old %s and/or .format.
My opinion is that they are much more readable then the old style, but you have to be careful with the quotes and the double quote.
It is not always possible to use the f'' or f"" because if the variables or string contains one of them, you need to switch to the other for the outer quotes
For example:
f" This is "my" text" does not  work and should be changed in f' This is "my" work', so especially working with json formatted string or dict you must be careful.

medusa/clients/torrent/generic.py
BenjV: My personal standard with f-string is using double quote (if possible) because in all other formatting I use single quotes by default so that gives the best change of being consistent.
But be my guest if you want to change it.

medusa/clients/torrent/downloadstation.py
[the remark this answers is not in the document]
p0psicles: I think it would be cleaner to have that being handled by the requests module.
To be sure you could also use json.dumps().
Like: 
```
        params = {
            'api': 'SYNO.DownloadStation2.Task',
            'version': 2,
            'method': 'create',
            'create_list': 'false',
            'destination': app.TORRENT_PATH,
            'type': 'url',
            'url': result.url
        }
        return self._add_torrent(json.dumps(params))
```
This is less error prone. And has a better readability. And should also work fine.
Normally you would use requests `json` param, to send json data in this case. That will stringify it under the hood.
like: `requests.post(webhook, json=payload, headers=headers)`. But the json param is not exposed through the _request() method.

medusa/clients/torrent/generic.py
p0psicles: Your absolutely right. Totally forgot that I used the MedusaSafeSession for this.
Thanks for noticing.

medusa/clients/torrent/generic.py
p0psicles: I'll see if I can run some tests. I still have some of the docker containers from the download_handler pr.

medusa/clients/torrent/generic.py
p0psicles: result is a SearchResult object. `https://github.com/pymedusa/Medusa/blob/master/medusa/classes.py#L79`.
So in "worst case" it's None.

medusa/clients/torrent/downloadstation.py
p0psicles: Sorry, I think I accidentele hidden this one. Can you take a look at it?
Using string templating like `%s`, `{}` or f"" in the logging module, is not best practice. Can you change it to BraceAdapter?

p0psicles:
Can you give me access to your forks branch? Then i'm okay with making some changes to it. And then you can also test them against the downloadstation's api.

Alternative is that you make changes that are required to pass the build process, and I make the other changes after merged. But currently it's not building.

medusa/clients/torrent/downloadstation.py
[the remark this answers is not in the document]
BenjV: No that will not be possible, that will not be function on Download Station.
Please do not change this, I was problematic enough to get this working.

BenjV:
I have send you an invite for my fork.
You can change anything you like, but please do not change the things that are needed for DownLoad Station like the parameter dict.
I you do that, things will not function any longer.

And by the way, if I look at the reasons why the test are failing, non of those reasons are in the  the two modules (generic.py and downloadstation.py) that I changed.

p0psicles:
You can ignore the api test. But the lint test in backend test is failing because of changes in this pr.

medusa/clients/torrent/downloadstation.py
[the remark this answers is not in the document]
p0psicles: How do you know, did you test this out?
If not, I can make the changes in a separate commit. If it works nice, if not, we just remove the commit? No harm no foul.
It would also not be that much work for you. But the current changes feel really hacky to me, so i'd like to avoid that if possible.

p0psicles:
I accidentally pushed to the wrong branch name `master-benjv`. Sorry for that.

I pushed only a commit that fixes the lining issues. Next is a small refactor to the send_torrent() exception handling part.
And then last we can test the json.dumps() method. Let me know if you willing to test that one out for me?

BenjV:
No the jason dump methode does not work, it will convert also things that must not be converted to json.
I know it is strange, but that's the way Synology created that api and changing it is not possible. 

Don't be so stubborn and just accept my correctly working solution.
I have spent a fair amount of time testing this and what I created is working fine.

p0psicles:
I'm not asking for the change because I want to be stubborn. I'm sorry if I gave you that impression.
But i'm asking for it because i'd prefer to add maintainable code to our project. 

For example. What if you'd loose interest in supporting the downloadstation api in 2 year.
And someone else wants to extend the functionaly, like add new downloadstation api routes. They would need to trial&error like you had to. And while you managed. It's not sure if someone else would. Wouldn't it be great if we could just use basic dicts with base types like, boolean, string, integers. And not have to worry about wrapping some strings in quotes and some not?

BenjV:
One of the first things I tested was your solution with json dumps, but that does not work.
Only some of the data fields must be json coded not the whole dict, if you do the whole dict it does not work.
The only solution is to add quotes to those fields who need it and leave the rest alone.

Of course there are more ways than f-string to add quotes but that would make in even less readable.
And precisely for the reason of maintainability I added extensive comment to the code why I did it that way.

p0psicles:
Tnx for your explanation and your patience. 
Sure, let's keep it as is. 
Just need to do a small change. And then I'll merge. 

Let's park the NZB provider for now. As that would require some more frontend changes. And I don't have much time now

p0psicles (approved):

reconman:
Since the latest `develop` commit, sending torrents to the Transmission RPC url causes HTTP 409 errors. Haven't been able to find out any details yet.

Reverting back to master fixed the problem.

p0psicles:
@reconman tnx for reporting, i'll look into it asap

p0psicles:
@reconman can you confirm that the issue is fixed for you?

reconman:
Seems fixed, torrents can be transmitted to transmission again. Thank you.

BenjV:
@p0psicles 
I see that in de develop branch you choose to use the old generic.py instead of the one I changed.
This combination of old genric.py and new downloadstation.py will faulty, so on the develop branch downloadstation will not function.

PR #9598, opened 2021-06-01
Fixed Transmission (and prob others) authentication.

p0psicles:
* self.get_auth() was called just after self.url = None was set.

### DIFF
--- a/medusa/clients/torrent/deluge.py
+++ b/medusa/clients/torrent/deluge.py
@@ -120,6 +120,8 @@ class DelugeAPI(GenericClient):
         self.session.headers.update({'Content-Type': 'application/json'})
         self.url = urljoin(self.host, 'json')
 
+        self._get_auth()
+
     def _get_auth(self):
         post_data = json.dumps({
             'method': 'auth.login',
--- a/medusa/clients/torrent/deluged.py
+++ b/medusa/clients/torrent/deluged.py
@@ -41,6 +41,8 @@ class DelugeDAPI(GenericClient):
         super(DelugeDAPI, self).__init__('DelugeD', host, username, password)
         self.drpc = None
 
+        self._get_auth()
+
     def _get_auth(self):
         return True if self.connect() else None
 
--- a/medusa/clients/torrent/downloadstation.py
+++ b/medusa/clients/torrent/downloadstation.py
@@ -61,6 +61,8 @@ class DownloadStationAPI(GenericClient):
         self.url = self.host
         app.TORRENT_PATH = re.sub(r'^/volume\d*/', '', app.TORRENT_PATH)
 
+        self._get_auth()
+
     def _check_path(self):
         """Validate the destination."""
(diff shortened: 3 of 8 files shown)


BenjV:
@p0psicles 
The self.url is filled in the __init__ of the various torrent clients.
So if self.get_auth() is called the __init__ of the clients will populate the self.url.

p0psicles:
Yeah, but the issue was the order.. get_auth() needs to url for some clients. And get_auth was called in the GenericClient init. And at that point URL wasn't initialized yet

BenjV:
If you call self.get_auth() in the ___init__ of the generic.py it will call the get_auth() of the client module.
And then first of all the __init__ of the client module is executed before the get_auth().
In that __init__ the self.url is populated and so the get_auth() function gets a correctly populated self.url.

p0psicles:
It didn't work well with qbittorrent. 
> And then first of all the init of the client module is executed before the get_auth().

There is no implicit call to the client module init. And thats a good thing. Else you would get into a loop. As the client module init will again call the super() which will call the get_auth().. Etc. 

Only thing I could do is move the get_auth back to generic. And put the super call behind the url = ...

p0psicles:
I think it also was an issue with rtorrent. As I use that myself. And the test_authentication was failing before this pr was merged.

BenjV:
The init of a class is always called before any of the functions within the class is called.
Because the class is called directly in medusa instead of creating an object with the class as template, every time a methode of that class is called the the init of that class is first executed.

I looked at rtorrent and I saw the the self.url is not used at all in rtorrent _get_auth()

But if you still think it is not ok then you must add the self._get_auth() call to the send_torrent function of the generic.py to be sure the session is authorized.

Because GenericClient() is not used to create an object with but called directly, non of the self variables nor the session header nor cookies are kept alive so before every call to a client the self._getauth() must be called to create those fresh.

p0psicles:
I really don't know why you think we don't create objects. I've explained in this comment. https://github.com/pymedusa/Medusa/pull/9482#issuecomment-851560427

BenjV:
Because I tested it.
The class GenericClient() is used directly  and not to create an object with.

Otherwise where is the object created?
There should be somewhere a statement like this be present

`GenericTorrentClientObject = GenericClient()`

Then at that moment the GenericTorrentClientObject is created from the class template and the init code is executed only once.
After that all the self variables are kept alive as part of the object.

But is you call the GenericClient() directly is only function as a set of methodes without keeping the self data alive.

medariox:
@BenjV 
You can see that `client = torrent.get_client_class(app.TORRENT_METHOD)()` is indeed an object instance because of the ending `()`.

BenjV:
@medariox 
Yes but it is created new every time a torrent is added, so it does not act as an object.

If you want to use it as an object somewhere in the application it must be create once and only once and not every time new because all the object data is gone then and you only have an object in name not in functionality.

PR #9726, opened 2021-07-11
Run docker container as non-root

reconman:

See #9479 for an explanation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,4 +32,4 @@ EXPOSE 8081
 VOLUME /config /downloads /tv /anime
 
 WORKDIR /app/medusa
-CMD [ "python", "start.py", "--nolaunch", "--datadir", "/config" ]
+CMD [ "runscripts/init.docker" ]
--- a/runscripts/init.docker
+++ b/runscripts/init.docker
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+PUID=${PUID:-911}
+PGID=${PGID:-911}
+
+addgroup -g "$PGID" abc
+adduser -u "$PUID" --ingroup abc --disabled-password abc
+
+echo "
+User uid:    $(id -u abc)
+User gid:    $(id -g abc)
+-------------------------------------
+"
+chown -R abc:abc /app
+chown -R abc:abc /config
+exec su abc -s /bin/sh -c "python start.py --nolaunch --datadir /config"
--- a/runscripts/init.docker
+++ b/runscripts/init.docker
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-PUID=${PUID:-911}
-PGID=${PGID:-911}
+PUID=${PUID:0}
+PGID=${PGID:0}
 
 addgroup -g "$PGID" abc
 adduser -u "$PUID" --ingroup abc --disabled-password abc


p0psicles:
Lgtm. @medariox ?

medariox:
Will this cause permissions issues for people that already run as root?

reconman:
It will cause problems if the media files in the post-processing target dirs don't have write write permissions for the user "abc". But it's easily fixable by setting the GPID environment variable to the id of a group that has access.

You could also theoretically set the default user and group to root (0) to avoid any problems switching from the old image.

If you switch from the linuxserver.io container to this one, you won't have any problems as long as you also set the PUID and PGID environment variables to the same values.

Those 2 variables still need to be documented, the Dockerhub page would be the best place for that, which is just a copy of the README, but I wasn't sure how exactly to describe them.

medariox:
> It will cause problems if the media files in the post-processing target dirs don't have write write permissions for the user "abc". But it's easily fixable by setting the GPID environment variable to the id of a group that has access.

> You could also theoretically set the default user and group to root (0) to avoid any problems switching from the old image.

That's what I'm afraid of. We have to assume worst case scenario, meaning: user doesn't read the readme and existing files don't have permissions. So I'd suggest we default to `root` for now and make it very clear that in the future it will change (news/changelog) + instructions on how to change the group (wiki).

p0psicles:
@reconman @medariox what are next steps for this PR to succeed? Is there a way forward? Else i'll close it.

reconman:
The next step is probably deciding that root will be the default user for now and change the default PUID and GUID in my added file to 0 instead of 911.

Then the documentation also needs to be adjusted to inform users of the PUID and GUID environment variables.

p0psicles:
I can update the docker docs

reconman:
I changed it to root by default. You could copy most of the explanation about PUID and GUID from https://hub.docker.com/r/linuxserver/medusa, especially the "User / Group Identifiers" section.

p0psicles:
I had already copied some of the PUID and GUID env variable stuff.
But i've also added the USer / Group identifiers section.

Tnx for the PR!

p0psicles:
@reconman the develop docker image is failing to build now. Could you look in to this. As users are currently not able to run the develop tag.

reconman:
Replace :0 with :-0. When I searched for default values, I could only find examples without the hyphen.

Can't commit until I come home in 10 hours.

p0psicles:
I soved it in a different way. But this works. If you see any improvements. Be my guest. But pls do test it before PR it.

David-Woodward:
I'm not 100% certain, but I think this change broke my non-root container configuration (container created with the "--user uid:gid" option).

When the docker daemon is not running as root, the adduser / addgroup commands in these changes fail - which apparently also puts container a restart loop (at least on my Synology NAS).

For now I've just allowed the container to run as root and I _think_ the PUID/PGID variables will be sufficient for host resource permissions.  However, I already had those variables in my update/creation scripts and it has been so long since I initially set all this up that I can't remember why I had both the variables and the "--user" option in place.  I'm hoping it was out of an abundance of caution from a security perspective and not because it was required for compatibility with some other component in my setup.

Anyway.  I thought I'd let you know just in case there was something that needed to be changed to make the docker image compatible with non-root containers.

* Is the "abc" user/group required to be hard coded due to dependencies in some other component now?  If not, perhaps the user/group names could be variables and you could just capture the current user/group name/id when the docker container is not running as root and therefore unable to create new users.

reconman:
Can you please provide logs of the errors?

David-Woodward:
```
Group does not exist, adding abc
addgroup: permission denied (are you root?)
User does not exist, adding 
adduser: permission denied (are you root?)
id: unknown user abc
id: unknown user abc

  User uid:    
  User gid:    
  -------------------------------------
  
chown: unknown user/group abc:abc
chown: unknown user/group abc:abc
su: must be suid to work properly
```

And that repeats over and over with the Synology Docker interface showing the container is constantly restarting.

reconman:
Sounds like the --user parameter makes the entryscript run as non-root, which doesn't have permission to add users and groups in the container itself.

The entryscript added through this PR first adds the user and group as root before switching to the specified user.

So it seems like the options can't be used at the same time, but both will achieve the same results.

PR #9862, opened 2021-09-11
Start on changes for  changeIndexer page.

p0psicles:

- [x] frontend
- [x] backend (api)
- [x] connect the two


medusa/tv/episode.py
```diff
@@ -2084,8 +2084,14 @@ def update_status_quality(self, filepath):
...
             if old_status in (SNATCHED, SNATCHED_PROPER, SNATCHED_BEST) or (
-                    old_status == DOWNLOADED and old_location) or (
-                    old_status == WANTED and not old_location):
+                    old_status == DOWNLOADED and old_location
+            ) or (
+                old_status == WANTED and not old_location
+            ) or (
+                # For example when removing an existing show (keep files)
+                # and re-adding it. The status is SKIPPED just after adding it.
+                old_status == SKIPPED and not old_location
```
p0psicles, line 2093: @medariox do you see any issues with adding this option here?
It's this or almost copy all the code from update_status_quality and calling methods, just for changing the indexer.

As it stands now. After re-adding a show, the episodes of the show start with the status "SKIPPED". And this causes it to get the status "ARCHIVED". WHich is incorrect.

Issue #10065, opened 2021-11-10
No support for python 3.10


mlg9000:
Medusa does not run under python 3.10.  "AttributeError: module 'collections' has no attribute 'MutableMapping'"

BenjV:
As of python 3.3 this statement was depreciated.
`from collections import Mapping`

As of python 3.9  it was removed and so it is also not available in python 3.10
The **A**bstract **B**ase Classes for **C**ontainers should be used

That code should be changed in:
`from collections.abc import Mapping`

Or if backwards compatibility is important this should be used:
```
try:
    from collections.abc import Mapping
except ImportError:
    from collections import Mapping
```

medariox:
Fixed with https://github.com/pymedusa/Medusa/pull/10068 
Will be available with the upcoming version.

PR #10068, opened 2021-11-10
Feature/python 3.10

p0psicles:

two things.
1. Had to update babelfish manually. (copied over the files). mvt throws an error.
2. Had to remove the package dirtyjson. Doesn't support python 3.10 (also the from collections import Mapping issue). But that project hasn't been updated since 2017. TorrentDay used this package. So we'll need to test if we still need this. And search for an alternative.

Fix for #10065


medariox:
`guessit` added the `importlib-resources` (https://pypi.org/project/importlib-resources/) for python_version<3.9, so we need to add that as well.

PR #10092, opened 2021-11-14
Fixed provider morethantv

p0psicles:

### DIFF
--- a/medusa/providers/torrent/html/morethantv.py
+++ b/medusa/providers/torrent/html/morethantv.py
@@ -39,9 +39,9 @@ class MoreThanTVProvider(TorrentProvider):
         self.password = None
 
         # URLs
-        self.url = 'https://www.morethan.tv/'
+        self.url = 'https://www.morethantv.me/'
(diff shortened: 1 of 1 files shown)


medusa/providers/torrent/html/morethantv.py
```diff
@@ -192,19 +193,29 @@ def login(self):
...
+            log.warning('Unable to get login token')
+            return False
+
         login_params = {
             'username': self.username,
             'password': self.password,
-            'keeplogged': '1',
-            'login': 'Log in',
+            # screen_resolution (x) | screen_resolution (y) | color depth | timezone offset.
+            'cinfo': f'1536|864|24|{int(time.timezone / 60)}',
```
p0psicles, line 207: I had to set the screen resolution and color depth to fixed values (that matched my laptop's screen).
So this might divert from what your sending if your manually accessing MTV from your computer. I don't suspect they will bann you for this. As you should be allowed to access your site with multiple devices from the same ip.

PR #10126, opened 2021-11-23
Fix torrentday generating a JSONDecodeError

p0psicles:
when parsing a result with single backslash in it.


p0psicles:
@medariox this should fix it.
You can get an example with a bad release in using the following query:  `domain.com/t.json?limit=1000`

PR #10312, opened 2022-02-05
Fix addic7ed subliminal subtitle provider

p0psicles:
* credits to @nixxou


p0psicles:
@nixxou I'm going to add some info to medusa subtitle config page. So users will know that they'll have to replace username with their addic7ed user id.

For now we're overwriting subliminals sub provider with our own. But i'd appreciate it if you'd did a PR to subliminal so that we can get the code in there.

PR #10322, opened 2022-02-07
Add RSS feed client 

sawyersteven:

Overview
* Adds the ability to select RSS as a download client. When a searchresult is snatched by the search core the download url and some metadata are written to medusa.xml in a user-configured directory.

Concerns
* Serving the xml file. I've been using nginx to alias to the xml file so my server hosts it as `url.com/medusa.rss`. Some users may not want to use a similar setup and would prefer that medusa serve the file itself eg `url.com:8081/medusa.rss` or some appropriate endpoint.

* I'm not a frontend dev by any measure and it seems `yarn build` changed the formatting of many files without actually changing the content (only a few select files were changed by my commits). This would explain the massive difference in line count for this PR. If I need to do something to make yarn build the ui similar to the existing files I'll gladly rebuild and push to this branch.


p0psicles:
Can you explain what problem your solution solves.
What alternatives did you consider.
How can this feature be of help to other users?

p0psicles:
As for the build process. For feature branches we build using `yarn dev`. Or `yarn dev --watch` if you want to watch files for changes.

`yarn build` is done by us when we release to master.

sawyersteven:
I kind of forgot this PR existed, but I'm still interested in getting it merged if it is acceptable. I don't believe there are any conflicts other than the frontend stuff I already mentioned.

But to answer your questions:
> Can you explain what problem your solution solves.

> How can this feature be of help to other users?

These two are basically the same question.

I have Medusa and my torrent client running on separate machines. This allows medusa to simply find appropriate downloads and publish them for any client to read at any time. Unlike a blackhole, this doesn't require read access to a directory, only the rss feed itself. It also makes it easier to broadcast downloads to several clients without tightly coupling any of them to the medusa instance. I don't use Medusa to manage the library after episodes are downloaded.

> What alternatives did you consider.

As mentioned above, a black hole could work for my use case, but it would require exposing the black hole directory and managing who gets to clean up old files after they've been downloaded or have gotten old enough. Its not ideal and rss makes this very easy.

PR #10370, opened 2022-02-23
Fix content_path not being available in some situations

p0psicles:

### DIFF
--- a/medusa/clients/torrent/qbittorrent.py
+++ b/medusa/clients/torrent/qbittorrent.py
@@ -416,8 +416,9 @@ class QBittorrentAPI(GenericClient):
         # Store destination
         client_status.destination = torrent['save_path']
 
-        # Store resource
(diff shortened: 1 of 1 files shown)


medariox:
Can't we use `torrent['name']` instead of `torrent['content_path']` for older APIs?

p0psicles:
You sure this is an api version issue? Because I think that `torrent['content_path']` is not available in the early stage of adding the torrent to the client. So while being added it doesn't know a content_path. But that get's available when it matters.

medariox:
Hmm, no not sure. But why not use the torrent name? Seems like the other clients also use the name?

p0psicles:
Hi @medariox sorry forgot about this one.
The reason I started using content_path is that content_name can be different from the eventual finished location. And medusa then can't pickup the resulted file.

PR #10515, opened 2022-04-20
Fix docker running with and without PUID/PGID

p0psicles:

I'm not that well versed with docker. But i've tested this script within a docker container with and without providing PUID/PGID. So i think this should work


p0psicles:
@reconman this should have already fixed it.

PR #10533, opened 2022-04-26
Don't overwrite already arhived episodes when postprocessing.

p0psicles:
* Exempt manually snatched downloads.

@medariox what do you think of these rules?


p0psicles:
Argh this doesn't work. Because eps with a status of `Archived` are changed towards `Snatched` when snatched, obviously. So there isn't really a point to this.

PR #10578, opened 2022-05-18
Strip tt from Imdb ID for Eztv provider

medariox:


p0psicles:
Does it require to include the leading zeros? Like does it need '0012345' or does '12345' also work?

medariox:
Just the leading tt needs to be removed.

Issue #10761, opened 2022-06-14
Need to  RE-try to connect to Medusa service on broken connection; Refresh of home page with logo-press


rafi-d:
The medusa GUI is not showing the home page after failed connection (usually after Windows startup & Browser auto-startup)
**To emulate and Reproduce**
1. Stop medusa service and close the browser (make sure to be on /home/ page
2. re-start  the browser.  Try to connect  and fail
3. Start medusa server . Medusa will not re-connect
4. Press the medusa logo , medusa will not show the shows' list
5. refresh the browser page - everything is fine
**Expected behavior**
1. Medusa should keep trying to connect to the service, say once every 5 seconds
2. Pressing Medusa Logo when on the home page - should refresh the home page, this will also be consistent with pressing the logo from other pages and moving to home.

**Screenshots**
![image](https://user-images.githubusercontent.com/2458280/173493779-2cab8801-7bb1-443e-8e5d-448e2e0aa3a1.png)

**Medusa (please complete the following information):**
<html><body>

Medusa Info: | Branch:                 master                  Commit:                 e4870f8f4a322e67c77434597ccc09aa6524aa05                  Version:                 0.5.29                  Database:                 44.19
-- | --
Python Version: | 3.10.4 (tags/v3.10.4:9d38120, Mar 23 2022, 23:13:41) [MSC v.1929 64 bit (AMD64)]
SSL Version: | OpenSSL 1.1.1n  15 Mar 2022
OS: | Windows-10-10.0.19044-SP0

</body>
</html>Medusa Info: 	Branch: [master](https://github.com/pymedusa/Medusa/tree/master)
Commit: [e4870f8f4a322e67c77434597ccc09aa6524aa05](https://github.com/pymedusa/Medusa/commit/e4870f8f4a322e67c77434597ccc09aa6524aa05)
Version: [0.5.29](https://github.com/pymedusa/Medusa/releases/tag/v0.5.29)
Database: 44.19
Python Version:	3.10.4 (tags/v3.10.4:9d38120, Mar 23 2022, 23:13:41) [MSC v.1929 64 bit (AMD64)]
SSL Version:	OpenSSL 1.1.1n 15 Mar 2022
OS:	Windows-10-10.0.19044-SP0

p0psicles:
The medusa UI is build using a SPA (single page application). What you are asking is to implement PWA (progressive web app), or also called Service Workers. Which is something that we will do at some point. But not that high of a prio.

[...]

PR #10774, opened 2022-06-22
Feature/UI indicate when no connection

p0psicles:

Add a small indicator (red cross) to the Medusa log in the top left, for when medusa has no connection to the apiv2 or the websocket.

Modified the behavior of the icon, in that when a connection is lost, clicking on it will reload the whole page. Instead of routing to `/#`

Implement #10761


themes-default/slim/src/components/app-header.vue
```diff
@@ -9,7 +9,11 @@
                     <span class="icon-bar" />
                     <span class="icon-bar" />
                 </button>
-                <app-link class="navbar-brand" href="home/" title="Medusa"><img alt="Medusa" src="images/medusa.png" style="height: 50px;" class="img-responsive pull-left"></app-link>
+                <app-link v-if="isConnected || !socketIsConnected" class="navbar-brand" style="position: relative" title="Medusa">
```
reconman, line 12: Isn't this supposed to be `!isConnected`?

Right now I always get a disconnected icon and so clicking on the logo refreshes the page.

p0psicles: I fixed that in a PR that came after

reconman: Was broken on my end. Had the fixed version, but my Synology Reverse Proxy didn't forward it correctly without some additional config.

https://mlohr.com/websockets-for-synology-dsm/

PR #10976, opened 2022-10-19
fix(subliminal): set english language in requests header

jee-r:

Addic7ed force language based on IP location so if your medusa instance is located in a country where english is not the first language. Subliminal will failed to download subtitle because it match only on  `Completed` results. if the request is made from france the word will be "Terminé". 

set `en-US,en;q=1.0` as Accept-Language in requests header solve this issue

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

Cheers


p0psicles:
Tnx

PR #11172, opened 2023-03-11
Read correct process_methods from config on startup

cyatta:

On startup, init will overwrite previous values set for `process_method_torrent` and `process_method_nzb` because it's actually reading `process_method` for each setting.

This change fixes that by setting `PROCESS_METHOD_NZB` and `PROCESS_METHOD_TORRENT` to read from the correct string within config.

For quick reference, the config names for the specific process methods:
https://github.com/pymedusa/Medusa/blob/49ad1f901a050c79b7e8cb583cb02ee1a6abeecf/medusa/__main__.py#L1714-L1715


medariox:
Good catch, thank you!

PR #11216, opened 2023-05-11
Feature/remove deprecated legendastv subtitle website

gleepwurp:
Portugese Subtitle Website "legendas.tv" has closed. Removing Ext Python file for it.
Also just fixed a word that annoyed me.  "postproccessing" --> "postprocessing" (two 'c' where there should be one).


gleepwurp:
I'm seeing other dependencies/references to legendas.tv, will fix before pulling... Sorry.

Issue #10777, opened 2022-06-22
Exec format error when pull fresh docker image


f0xMccloud:
**Describe the bug**
Hey everyone, I'm trying to pull latest (tried with latest and master) from  dockerhub on raspberry pi 4 and I get this error:
`Starting medusa as root
Traceback (most recent call last):
  File "/app/medusa/start.py", line 6, in <module>
    from medusa.__main__ import main
  File "/app/medusa/medusa/__init__.py", line 24, in <module>
    initialize()
  File "/app/medusa/medusa/init/__init__.py", line 37, in initialize
    _configure_guessit()
  File "/app/medusa/medusa/init/__init__.py", line 202, in _configure_guessit
    from medusa.name_parser.guessit_parser import (
  File "/app/medusa/medusa/name_parser/guessit_parser.py", line 12, in <module>
    from medusa.logger.adapters.style import BraceAdapter
  File "/app/medusa/medusa/logger/__init__.py", line 53, in <module>
    import subliminal
  File "/app/medusa/ext/subliminal/__init__.py", line 11, in <module>
    from .core import (AsyncProviderPool, ProviderPool, check_video, download_best_subtitles, download_subtitles,
  File "/app/medusa/ext/subliminal/core.py", line 13, in <module>
    from rarfile import BadRarFile, NotRarFile, RarCannotExec, RarFile, Error, is_rarfile
  File "/app/medusa/ext/rarfile.py", line 3039, in <module>
    _check_unrar_tool()
  File "/app/medusa/ext/rarfile.py", line 3019, in _check_unrar_tool
    custom_check([ORIG_UNRAR_TOOL], True)
  File "/app/medusa/ext/rarfile.py", line 2911, in custom_check
    p = custom_popen(cmd)
  File "/app/medusa/ext/rarfile.py", line 2898, in custom_popen
    p = Popen(cmd, bufsize=0, stdout=PIPE, stdin=PIPE, stderr=STDOUT,
  File "/usr/local/lib/python3.10/subprocess.py", line 966, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.10/subprocess.py", line 1842, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 8] Exec format error: 'unrar'
`

**To Reproduce**
1. Run docker pull pymedusa/medusa:latest (or master)
[...]

PR #11524, opened 2023-10-20
Switch from using `unrar` to `7z`

tonyg:

This little patch experimentally replaces use of `unrar` (problematic in e.g. #10777) with `7z`, which is actually pretty close to drop-in compatible, it seems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,3 @@
-# Need this image for the /bin/start-build script.
-# Build unrar.  It has been moved to non-free since Alpine 3.15.
-# https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.15.0#unrar_moved_to_non-free
-FROM jlesage/alpine-abuild:3.15 AS unrar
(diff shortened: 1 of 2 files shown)


medariox:
Hey! Thanks for the PR. I don't understand what PySocks is required for. Could you clarify?

tonyg:
Oops. Sorry, that commit snuck in there. PySocks is required to do anything with a socks proxy at all (AFAICT). I needed the unrar-to-7z thing just to get it to run at all in docker, and then the PySocks thing was notionally separate (though I don't think I understand how socks could ever have worked if PySocks wasn't present!). From the contribution guidelines, I imagine you'd prefer me to redo the branch keeping the two changes separate; let me know if I need to do that or if you're happy to just take it as is (which would be easier for me at least!). Thanks for the software!

p0psicles:
If it's experimental and mainly added for arm. I'd prefer to have a separate module, that we can load on arm devices. Or feature flag. Would you be okay looking into such a solution?

PR #11546, opened 2023-11-01
Add try/catch block around localStorage.setItem call

dotsam:

Safari in particular isn't happy with the size of the shows localStorage item and will often throw a quota error. Catch the error to allow the show delete to continue


medariox:
Thank you for the PR! 👍

PR #11586, opened 2023-12-13
Fix bugs in rss client

sawyersteven:
Strips all non-ascii characters when writing the xml document.

Writes a new xml document if the existing document is empty or just whitespace


medariox:
Thanks for the PR! Could you please rebase it onto the current develop branch?

rogiersplayground:
It has conflicts. So the easiest to fix this, is to merge develop back into your branch. Then use your IDE to resolve the conflicts. So see what code you touched and what's touched on the remote. Often it's easiest to just accept the remote changes. And re-apply yours.

PR #11691, opened 2024-04-01
Testing for providers requiring authorization

k0mmsussert0d:

This PR demonstrates a proposed approach to testing torrent providers requiring authorization using `animebytes` as an example.

Authorization mechanism is a part of an individual provider implementation. Usually, it depends on credentials stored as attributes of a provider class. In this particular example of `animebytes` provider, `username` and `passkey` attributes store user-provided credentials.

When provider class is imported to the testing context, advanced configuration of the instance can performed by injecting custom attributes found in `_meta.attr` in the test YAML file.

To prevent credentials from getting leaked by VCR recording them in the request, VCR context configuration can be defined in `_meta.vcr`. In case of `animebytes`, credentials are supplied as `username` and `torrent_pass` query parameters, so they have to be stripped with `filter_query_parameters=['username', 'torrent_pass']` param provided to `use_cassette()`. [Reference](https://vcrpy.readthedocs.io/en/latest/advanced.html#filter-information-from-http-querystring)

To sum up, this PR introduces:
- post-construction initialization step for provider instance
- customizing vcr configuration per-case
- some test cases for animebytes


medariox:
Great job! Thank you for the exceptional effort 👍

PR #11718, opened 2024-05-17
Calendar route should write appropriate header

pidario:
- [ *] PR is based on the DEVELOP branch
- [ *] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [ *] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

As per PR title, calendar route should also add `Content-Type: text-calendar` header.
Tested with `vdirsyncer`: without this header the software can't interpret correctly the calendar. A possible (but ugly) workaround would be to `curl` the medusa calendar in order to download the body response to a calendar file and then make `vdirsyncer` read that file instead.


medusa/server/web/core/calendar.py
```diff
@@ -26,6 +26,7 @@ def get(self):
         """
         Render the iCalendar
         """
+        self.add_header("Content-Type", "text/calendar")
```
p0psicles, line 29: please change quotes to single. We reserve double for docstrings. Or to encapsulate single quotes in strings.

p0psicles (changes requested):

medusa/server/web/core/calendar.py
pidario: Done

medariox (approved):

medariox:
Thank you!

Issue #11844, opened 2024-10-01
[APP SUBMITTED]: qBittorrent: Unable to set the pause for Torrent


papampi:
### INFO
**Python Version**: `3.11.7 (main, Aug  8 2024, 12:42:55) [GCC 8.4.0]`
**Operating System**: `Linux-5.10.60-qnap-x86_64-with-glibc2.27`
**Locale**: `UTF-8`
**Branch**: [master](../tree/master)
**Database**: `44.19`
**Commit**: pymedusa/Medusa@34a67cf159acab69017dd5b95e0233ecde0de28a
**Link to Log**: **No Log available**
### ERROR
<pre>
2024-10-01 10:46:49 ERROR    SNATCHQUEUE-SNATCH-258902 :: [34a67cf] qBittorrent: Unable to set the pause for Torrent
</pre>
---
_STAFF NOTIFIED_: @pymedusa/support @pymedusa/moderators

papampi:
Hi,
QBittorrent updated to V5.0.0, seems like something is changed and Medusa keeps popping this error after every episode snatch since qbittorrent updated.

papampi:
Logs after a manual snatch if needed:


```
2024-10-01 13:11:49 INFO     SNATCHQUEUE-SNATCH-258902 :: [34a67cf] Beginning to snatch release: English.Teacher.S01E06.1080p.HEVC.x265-MeGusta
2024-10-01 13:11:49 INFO     SNATCHQUEUE-SNATCH-258902 :: [34a67cf] Downloading English.Teacher.S01E06.1080p.HEVC.x265-MeGusta with size: 0.00 B from ezrss, through a SearchType.MANUAL_SEARCH search
2024-10-01 13:11:49 DEBUG    SNATCHQUEUE-SNATCH-258902 :: [34a67cf] POST URL: http://192.168.1.250:8080/api/v2/auth/login [Status: 200]
2024-10-01 13:11:49 DEBUG    SNATCHQUEUE-SNATCH-258902 :: [34a67cf] User-Agent: Medusa/1.0.21 (Linux; 5.10.60-qnap; d7fe2f2c-7fda-11ef-b86a-f22162fbf60c)
2024-10-01 13:11:49 DEBUG    SNATCHQUEUE-SNATCH-258902 :: [34a67cf] With post data: username=**********&password=**********
[...]

PR #11857, opened 2024-10-12
Fix error added by qbittorrent 5.0 pymedusa/Medusa#11844

borntohonk:

Proposed fixes for error(s) from https://github.com/pymedusa/Medusa/issues/11844 caused by qbittorrent 5.0 api changes, which for reasons unknown to mortal man is that they changed "pause/paused" to "stop/stopped", and "resume/resumed" to "start".

Things addressed/fixed:
* Download handler is unable to process completed & "stopped" torrents (completed & uploading/seeding worked fine), which affects people who don't seed or if the download handlers schedule is slower than seeding goal is hit.

* Log-level error spawned from not being able to parse this new unknown "stopped" state.

Also tested:
qQbittorrent v4.6.7, to ensure new changes to accommodate qQbittorrent v5.0+ does not negatively affect previous version(s)


medusa/schedulers/download_handler.py
```diff
@@ -45,6 +45,7 @@ class ClientStatusEnum(Enum):
 
     SNATCHED = 0  # 0
     PAUSED = 1  # 1
+    STOPPED = 1
```
p0psicles, line 48: the only thing i'm not sure here, is if we need another download handler state. What benefit brings this?

borntohonk: Semantics to go along with qbittorent deciding to rebrand pause to stop, and the other parts adding "stop" wouldn't work was my interpretation.

borntohonk: Also a quick effort to not affect other torrent programs that don't conform to whatever qbittorrent maintainers silly decision, and retain backwards compatibility with pre 5.0.0 qbittorrent versions.

at least that was my intention

borntohonk: @p0psicles i remember now
https://github.com/pymedusa/Medusa/pull/11857/files#diff-28c840b0e4fc446765e43a85e5cb6aed6a874eee15753e980007e12c86fa6096R243
This line would not work unless a new state of stopped was defined, and i didn't want to cause any unwanted effects to existing ecosystem, only support qbittorrent 5.0+

borntohonk: (i am first now registering i made a typo in that function too)

borntohonk: i summarized this better here: https://github.com/pymedusa/Medusa/issues/11844#issuecomment-2395316482

borntohonk: I had time to deploy this on a physical machine again, and removed the changes to download_handler.py

XxUnkn0wnxX:
Is this still being worked on, or is it stuck in limbo?

borntohonk:
> Is this still being worked on, or is it stuck in limbo?

I'm not a maintainer so that's not up to me to decide, but I did remove the downloader_handler state per comment, and the "fix" itself does do what it's supposed to do.

borntohonk:
(ideally, one would go nag qbittorrent folks to make it backwards compatible instead of this new nonsense https://github.com/qbittorrent/qBittorrent/blob/master/src/webui/www/private/scripts/mocha-init.js#L523-L549 )

medariox (dismissed):

medariox:
LGTM! Thank you 🙏

PR #11893, opened 2024-12-14
Remove deprecated distutils for python 3.12 compatability

borntohonk:

As seen in the medusa discord, distutils causes crashes for end users because of lines 6 and 45 in medusa/updater/update_manager.py

due to references to distutils, which is deprecated in python 3.12

I've also attempted to update helper libraries which are not needed to address the crash itself.

### DIFF
--- a/.github/workflows/api-tests.yml
+++ b/.github/workflows/api-tests.yml
@@ -25,6 +25,7 @@ jobs:
         pip install dredd_hooks
         pip install 'PyYAML>=5.1'
         pip install 'six>=1.13.0'
+        pip install packaging
     - name: Install Node.js dependencies
       run: yarn install --ignore-scripts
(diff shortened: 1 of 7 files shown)


borntohonk:
(i apologize for messy PR, but I wanted I wanted to address tox after realizing the tests were failing due to test method being deprecated)

borntohonk:
My understanding is that https://github.com/pymedusa/Medusa/blob/master/.github/workflows/api-tests.yml
This errors per the action logs:
ModuleNotFoundError: No module named 'packaging'
I'm not sure if this is addressed with:
pip install packaging

borntohonk:
after a lot of confusion on my end, it seems the api test is using dredd, which is made by apriary, which is owned by oracle since 8 years ago and deprecated last month

PR #11926, opened 2025-02-05
Removed dependency on unmaintained docker container

### LINKED ISSUE #11924: Remove dependency on jlesage/alpine-abuild:3.15
- jlesage/alpine-abuild:3.15 is unmaintained and hasn't been updated in years
- jlesage/alpine-abuild:3.15 is only built for AMD64

I was looking at linuxserver.io's dockerfile for medusa and they simply build unrar directly from source. This gives direct control of the unrar version and also makes it build on all of the platforms alpine supports (risc-v, ppc64le, arm64, etc).

bobpaul:
Closes #11924
This has allowed me to build the docker container for ppc64le and should allow building for arm and risc as well.

The one thing I'm not sure that I like is I took this from a multi-stage build to a single stage. I don't think that should impact the image size, as I installed and uninstalled the build deps for unrar all in a single step. But it wouldn't be hard to switch this to a multi-stage build if that's desired.


medariox:
Thanks for the nice PR!

Question: aren't the flags:
 ```
 LDFLAGS=-pthread|LDFLAGS=-pthread -static|' makefile && \
 sed -i 's|CXXFLAGS=-march=native|CXXFLAGS=-mcpu=power8
 ```
 going to be an issue when the goal is a broad architecture support?
 Can the be removed / reduced?

bobpaul:
OK, so turns out leaving it as `native` is bad, since that would build a binary that only works on the same CPU sub-architecture as the build machine (so you might end up with something that runs on a recent xeon but doesn't work on an ivy bridge or ryzen).

So I followed the suggestions in the [multiplatform build](https://docs.docker.com/build/building/multi-platform/) guide and used the `TARGETARCH` and `TARGETPLATFORM` args that docker buildx populates.

On the native platforms (building amd64 on amd64, building arm64 on arm64, building arm32 on arm32) this works with `docker . build` just as before. 

And for unsupported platforms, one can override the CXXFLAGS entirely, such as `docker build . --buildargs='CXXFLAGS=-mcpu=power8'.

bobpaul:
OK, I'm done making changes.

Where it's at right now is it has build flags for `amd64`, `arm64`, and `arm` that I think are good defaults (I looked at what archlinux arm uses for their flags to confirm arm and arm64 were safe). Everywhere else it just leave it blank, which the GCC manpage says will produce a binary that should "run well everywhere". But a user doing their own build can easily override the cflags if they'd like.

locally (and without providing any `--build-arg`) I was able to build for `ppc64le`, `arm`, `arm64`, `riscv64`, and `amd64`. (I don't have `riscv64` or `arm` handy, so I had to use emulation via [aptman/qus](https://github.com/dbhi/qus) for those.

I'm not sure if you'd be interested in pushing images for additional architectures beyond what you currently support (amd64, armv7, arm64), but if you do I could help select/test more optimized build flags for those platforms.

medariox:
Thanks!

Issue #11906, opened 2025-01-02
Required words: AND/OR


LethalCorpse:
### Discussed in https://github.com/pymedusa/Medusa/discussions/11868

<div type='discussions-op-text'>

<sup>Originally posted by **LethalCorpse** October 31, 2024</sup>
Say you've got a pair of shows consisting of a UK original and a US remake. Medusa has a very hard time telling the difference, and will frequently grab episodes from one as episodes from the other. We can try working around this by adding, for example "US" and "2021" to the required words list for the show. This almost works, in that it should only grab releases containing at least one of those words. There's a problem if you also have a global list requiring, say, x265.

Unfortunately, the show list is just appended to global list to make an effective list for that show, with an OR condition, not an AND condition. So Medusa will then grab any release which contains "US" OR "2021" OR "x265". This breaks the functionality of both of the lists - you'll get x265 releases of the wrong show, or x264 releases of the right show.

For this to work properly there needs to be control over whether each required word is AND (in addition to), or "OR" (instead of) other required words.</div>

PR #12167, opened 2026-02-27
Changing found_required_word filtering from OR to AND

KagamiKuno:
Trying to download some specific multi-audio anime i run into incosistent results so i made this change to significantly improve the search precision by requiring all keywords to be present as distinct words, eliminating inconsistent matches.

Strict Matching: Replaced contains_at_least_one_word with an all() condition to ensure every required word is present in the result name before validation.

Word Boundaries: Implemented regex with \b anchors to ensure we match whole words only, preventing false positives from partial string matches (e.g. "cat" matching "category").

Normalization & Safety: Added case-insensitive matching via .lower() and used re.escape() to safely handle any special characters within the required words list.

### DIFF
--- a/medusa/search/core.py
+++ b/medusa/search/core.py
@@ -11,6 +11,7 @@ import operator
 import os
 import threading
 import time
+import re
 
 from medusa import (
     app,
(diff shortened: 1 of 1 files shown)


KagamiKuno:
Related issue [pymedusa/Medusa#11906](https://github.com/pymedusa/Medusa/issues/11906)

Also sorry for the mess on the workflow, first pull :)

PR #12204, opened 2026-05-16
Fix Plex .plexmatch episode paths to be relative

flick1999:
Plex expects paths in a series-level .plexmatch file to be relative to the show folder. (Ref: https://support.plex.tv/articles/plexmatch/)

Before this change, Medusa generated entries like:

ep: s01e01: /Season 1/file.mkv

That leading slash prevents Plex from matching the file correctly in some cases.

This changes the generated path to use os.path.relpath(), then normalizes separators to forward slashes for Plex compatibility.

Example output after this fix:

ep: s01e01: Season 1/file.mkv


medariox:
Thanks!

PR #12218, opened 2026-06-12
fix: allow multiple custom search templates per title

### LINKED ISSUE #12217: Only one custom search template per scene exception title is persisted
**Describe the bug**
Medusa allows users to add multiple custom search templates for the same show title, alias, or scene exception, but only one custom template per title, season, and search type is persisted.

Users should be able to configure any number of distinct custom search templates for the same title. There should be no fixed numeric limit.

For example, all of these episode search templates should be able to coexist for the same title or scene exception:

```text
%SN
%SN %EN
%SN 1x%0E
%SN S%0SE%0E
````

These are only examples. Custom templates may contain any supported combination of variables, fixed text, separators, and numbering styles.

Currently, each additional custom template with the same title, season, and episode/season search type overwrites the previously persisted template.

The UI may temporarily display several custom templates, but only one remains after saving and reloading the page.

This issue is separate from #12214. In this case, custom templates are not removed by `_clean()`. They are collapsed while being saved because the custom-template upsert key does not include the template pattern.

**To Reproduce**

1. Add any show with at least one alias or scene exception.
2. Open the show configuration.
3. Go to **Search Templates**.
4. Scroll down to **Custom Templates**.
5. Select the same title, alias, or scene exception.
6. Add and enable several distinct episode templates for that title, for example:

   ```text
   %SN
   %SN %EN
   %SN 1x%0E
   %SN S%0SE%0E
   ```
7. Save the show configuration.
8. Reload the page or restart Medusa.
9. See that only one custom template remains for that title and search type.

**Expected behavior**
All distinct custom templates should remain stored and enabled for the same title, alias, or scene exception.

There should be no fixed numeric limit on the number of custom templates.

Medusa should use every enabled custom template as a separate search strategy.

Only exact duplicates should be rejected. A custom template should be uniquely identified by:

```text
title + season + template + seasonSearch
```

**Screenshots**
If applicable, add screenshots showing:

* several custom templates visible before saving;
* only one custom template remaining after reloading;
* the corresponding rows in the `search_templates` database table.

* OS: Windows 11
* Branch: develop
* Commit: 7c88625
* Python version: `<Python version>`
* Database version: `<Database version>`

**Debug logs (at least 50 lines):**
General > Advanced Settings > Enable debug

This issue does not appear to generate an exception or explicit error in the logs. The custom templates are silently overwritten while the show configuration is persisted.

<details>

```text
No relevant error is logged when reproducing the issue.
```

</details>

**Additional context**
`SearchTemplates.update()` first removes all custom templates:

```python
self.remove_custom()
```

It then saves every submitted template again through `SearchTemplates.save()`.

For a new custom template without an ID, `save()` uses the following values as the `upsert()` key:

```python
control_values = {
    'indexer': self.show_obj.indexer,
    'series_id': self.show_obj.series_id,
    'title': template['title'],
    'season': template['season'],
    '`default`': template_default,
    'season_search': season_search,
}
```

The template pattern itself is not included in the key.

As a result, any number of distinct custom templates with the same title, season, and search type produce the same upsert key. Each template after the first updates the same database row instead of inserting a separate row.

Because `SearchTemplates.update()` removes all custom rows before saving them again, only the last matching template survives.

The provider code already supports multiple templates. `GenericProvider._get_episode_search_strings()` iterates over all enabled episode templates and adds each formatted template to the search query list.

Persistence therefore prevents behavior that the provider search implementation already supports.

There is also a related frontend issue in:

```text
themes-default/slim/src/store/modules/shows.js
```

The `ADD_SHOW_CONFIG_TEMPLATE` mutation checks duplicates using:

```javascript
template.pattern
```

However, stored template objects use:

```javascript
template.template
```

The duplicate check is also not scoped by title, season, or search type. It should compare:

```text
title + season + template + seasonSearch
```

A possible backend fix would be to include the pattern in the upsert key for new custom templates:

```python
control_values = {
    'indexer': self.show_obj.indexer,
    'series_id': self.show_obj.series_id,
    'title': template['title'],
    'season': template['season'],
    '`default`': template_default,
    'season_search': season_search,
    'template': template['template'],
}
```

Alternatively, new custom templates could always be inserted, while existing templates continue to be updated using `search_template_id`.

Suggested regression tests:

1. Save several custom episode templates using the same title and season.
2. Verify that all templates persist after `update()` followed by `read_from_db()`.
3. Save several custom season templates using the same title and season.
4. Verify that the same pattern can be used for different aliases or scene exceptions.
5. Verify that every enabled template generates a separate provider search string.

This issue is related to, but separate from, #12214:

* #12214 concerns custom templates being removed during cleanup or after restart.
* This issue concerns multiple distinct custom templates for the same title being collapsed during save.

Mika3578:

Fixes #12217

Custom search templates using the same title and season currently share the same database upsert key. As a result, adding another pattern for the same show title, alias, or scene exception overwrites the previously stored custom template.

This change allows multiple distinct custom search templates to coexist for the same title.

### Backend

Include the template pattern in the upsert key used for new custom search templates.

Templates are now uniquely identified by their:

- indexer
- series
- title
- season
- default status
- search type
- template pattern

### Frontend

Update custom-template duplicate detection to compare the title, season, and stored template pattern.

This also fixes the check using `template.pattern` instead of the stored `template.template` property.

### Tests

Add `test_multiple_custom_templates_same_title_persist` to verify that multiple patterns using the same title survive an update and database reload.

## Validation

- `11/11` search-template tests pass
- Six custom templates persisted successfully:
  - two patterns
  - three aliases
- All six rows remained after restarting Medusa
- Manual search generated all six expected queries:
  - `%SN %EN`
  - `%SN 1x%0E`
  - for each configured alias

No database migration is required.


Mika3578:

It fixes #12217 by allowing multiple custom search-template patterns to coexist for the same title and season instead of overwriting one another.

The change is limited to the database upsert key, frontend duplicate detection, and regression coverage. No database migration is required.

Validation completed:

all 11 search-template tests pass;
six templates using two patterns and three aliases persisted after restarting Medusa;
manual search generated all six expected queries;
all upstream CI workflows are passing.

medariox:
Thanks!

PR #12220, opened 2026-06-12
fix: match non-standard manual search releases by episode title

### LINKED ISSUE #12219: Contextual release matching for manual search when numbering is missing or unreliable
## Describe the problem

Manual episode searches can find relevant provider results, but Medusa discards them when their release names do not follow standard scene naming conventions.

Affected release names may:

* contain the series and episode titles but no `SxxExx` or `1x01` numbering;
* contain prefixes, segment names, uploader names, performer names, or other text before the useful title;
* contain suffixes or website names after the useful title;
* contain unrelated numbers that GuessIt incorrectly interprets as episode numbers;
* produce a GuessIt alias that is longer than the actual series title and cannot be matched against Medusa's name cache.

A typical non-standard structure may look like:

```text
<prefix> - <series title> - <segment> - <episode title> - <suffix>.mkv
```

or:

```text
<series title> - <episode title> - <website>-1--<release group>.avi
```

The provider search itself succeeds, but every result is rejected during release-name parsing.

## Current behavior

The current flow is approximately:

```text
Provider returns relevant results
    ↓
NameParser parses each release name
    ↓
Series cannot be resolved, numbering is missing,
or an unrelated number is parsed as the episode
    ↓
InvalidShowException / InvalidNameException
    ↓
result_wanted = False
    ↓
Manual search displays no results
```

During provider searches, Medusa already knows which series and episode are being searched. However, that context is not currently used to safely resolve non-standard release names.

Simply passing the current series to `NameParser` would not fully solve the problem. A recognized series with no parsed episode number can currently be interpreted as a season pack, which could incorrectly associate a single release with an entire season.

GuessIt is treated as the single source of truth for series/episode association, even though Medusa already knows which series and episode are being searched.

## Expected behavior

For a manual search targeting one specific episode, Medusa should be able to use the existing search context as a controlled fallback when standard parsing fails.

A result should be accepted when:

* the release contains the normalized series title or a known alias;
* the requested episode title appears as an exact or otherwise unambiguous normalized match;
* there is no reliable explicit numbering that contradicts the requested episode;
* the result appears to represent media content rather than an explicitly unsupported file type.

Standard scene-numbered releases must continue to use the existing strict parsing behavior.

## Suggested implementation

Add a contextual release matcher in the search layer rather than making the global series lookup or `NameParser` broadly permissive.

For example:

```text
medusa/search/release_matcher.py
```

The matcher could receive:

```python
match_search_result(
    release_name=search_result.name,
    series=series,
    requested_episodes=episodes,
    parsed_result=parsed_result,
    manual_search=manual_search,
    search_mode=search_mode,
)
```

### Matching sequence

1. Run `NameParser` for extraction (with `series=` passed during manual search).
2. Classify any parsed numbering as either reliable or weak.
3. Let `ReleaseMatcher` make the final association using search context when strict parsing fails or only weak numbering was found.
4. Normalize and compare:

   * the original release name;
   * the series title and known aliases;
   * the requested episode title.
5. Assign the requested season and episode only when the title match is unique and unambiguous.
6. Never convert an unnumbered individual result into a season pack unless the release name explicitly identifies itself as a season pack.

### Integration point

In `generic_provider.find_search_results()`, for single-episode manual search:

* Pass `series=` to `NameParser`
* Disable parser cache for contextual parsing (`cache_result=False`, `use_cache=False`) until cache keys include series context
* Run `ReleaseMatcher` after parsing; apply its season/episode assignment as the final decision

Automatic/backlog search should keep existing strict behavior until the matcher is validated more broadly.

## Numbering confidence

Explicit patterns should be treated as reliable, for example:

```text
S01E02
S01.E02
1x02
Season 1 Episode 2
Saison 1 Episode 2
```

A number returned by GuessIt without one of these explicit structures should not automatically be trusted.

Examples of weak or unrelated numbers include:

```text
<website>-1--<release group>
<channel>-2--<release group>
1080
720
1995
```

Suggested decision rules:

| Parsed information           | Requested episode title  | Decision                      |
| ---------------------------- | ------------------------ | ----------------------------- |
| Explicit numbering matches   | Any                      | Accept using numbering        |
| Explicit numbering conflicts | Title matches            | Reject the result             |
| No numbering                 | Exact unique title match | Accept using search context   |
| Weak numbering               | Exact unique title match | Ignore weak number and accept |
| No numbering                 | No title match           | Reject                        |
| Ambiguous title match        | Any                      | Reject                        |
| Explicit season-pack marker  | Not applicable           | Process as season pack        |

## Title normalization

Matching should tolerate harmless formatting differences:

* upper/lower case;
* accents;
* straight and typographic apostrophes;
* dots, underscores, hyphens and repeated spaces;
* brackets and parentheses;
* prefixes and suffixes surrounding the useful title.

The original release name should still be passed unchanged to GuessIt. A separate normalized value should be used only for contextual matching.

The first implementation should use strict normalized title containment or equality rather than broad fuzzy matching, to reduce false positives.

## Media validation

The contextual fallback should reject results with explicitly unsupported file extensions such as image, text, metadata, or executable files.

Known video extensions can be treated as a positive signal:

```text
.mkv
.mp4
.avi
.mpg
.mpeg
.m4v
.mov
.wmv
.ts
.m2ts
.webm
```

A torrent or magnet result should not be rejected only because its displayed name has no extension, since the contained file list may not yet be available.

Inspecting the actual files contained in torrents can be handled in a follow-up PR.

### First PR (minimal, safe)

* New `medusa/search/release_matcher.py`
* Integration in `generic_provider.py` for single-episode manual search only
* Unit tests for normalization, strong/weak numbering, title fallback, conflicts, season packs
* Match diagnostics on `SearchResult` (`match_method`, `match_confidence`, `match_reason`, `numbering_confidence`)

### Follow-up PRs

* Stop assuming season 1 / season pack in `NameParser._parse_series()` when numbering is unresolved
* Context-aware parser cache keys
* Torrent file inspection before snatch (video extension / file list validation)
* Optional per-series setting to expose contextual matching beyond manual search

The initial implementation should be limited to:

* manual searches;
* searches targeting one specific episode;
* results for which standard parsing failed or produced only weak numbering;
* exact or unambiguous normalized episode-title matches.

## Logging

Add explicit diagnostic messages such as:

```text
Matched release using explicit season and episode numbering
Matched release using the requested episode title
Ignored weak episode number extracted from release suffix
Rejected release because explicit numbering conflicts with the requested episode
Rejected release because no unique episode-title match was found
Rejected unnumbered release instead of treating it as a season pack
```

## Acceptance criteria

* Standard `SxxExx` and `1x01` releases continue to work unchanged.
* A manual result containing the series title and requested episode title can be matched without episode numbering.
* Prefixes and suffixes do not prevent a valid contextual match.
* Unrelated numeric suffixes are not trusted as explicit episode numbering.
* Explicit numbering that conflicts with the requested episode causes rejection.
* A release without numbering or a recognizable episode title remains rejected.
* Ambiguous episode-title matches remain rejected.
* An unnumbered individual release is never silently converted into a season pack.
* Explicit non-video filenames are rejected.
* Existing backlog/automatic search behavior is unchanged in the first PR.
* Unit tests cover successful, conflicting, ambiguous and rejected matching cases.

## Notes

Scene exceptions should remain **series aliases**, not per-episode naming patterns. Episode association should come from the matcher and episode metadata, not from growing the exception list for every release format.

Mika3578:
## Summary

Add contextual release matching for single-episode manual searches when provider results use non-standard release names.

Some valid results contain the series and episode titles but do not include reliable SxxExx or 1x01 numbering. They may also contain prefixes, suffixes, or unrelated numeric tokens that GuessIt interprets as episode numbers.

Previously, these results were rejected during strict release parsing and never appeared in the manual search results.

## Changes

- Add a contextual `ReleaseMatcher` for manual single-episode searches.
- Keep the existing strict `NameParser` path as the primary matching method.
- Fall back to normalized series and episode-title matching when strict parsing fails or only weak numbering is detected.
- Distinguish explicit TV numbering from unrelated numeric suffixes.
- Reject releases when explicit numbering conflicts with the requested episode.
- Prevent unresolved individual releases from being treated as implicit season packs.
- Keep automatic, backlog, daily, proper, season, post-processing, and library-scan behavior unchanged.
- Add focused tests for normalization, explicit and weak numbering, title matching, conflicts, ambiguity, media validation, and contextual-matching gates.

## Matching rules

The contextual fallback is limited to:

- manual searches;
- one requested episode;
- episode search mode;
- manual search type episode.

Reliable explicit numbering remains authoritative.

When reliable numbering is absent, a result may be associated with the requested episode only when:

- the current series name or a known scene alias is identified;
- the requested episode title has a unique normalized match;
- no reliable explicit numbering contradicts the requested episode.

## Tests

51 passed

Executed:

```
python -m pytest \
  tests/search/test_release_matcher.py \
  tests/providers/test_generic_provider.py \
  tests/name_parser/test_parse_series.py \
  -q
```

Additional validation:

```
python -m compileall \
  medusa/search/release_matcher.py \
  medusa/providers/generic_provider.py
```

```
git diff --check
```

Manual runtime validation confirmed that:

- non-standard provider results are displayed in manual search;
- weak numeric suffixes are ignored when the episode-title match is unique;
- the selected result is associated only with the requested episode;
- explicit conflicting numbering is rejected.

## Scope

This PR intentionally does not change:

- global `NameParser._parse_series()` behavior;
- automatic or backlog matching;
- parser cache keys;
- torrent file-content inspection;
- the secondary `update_cache_manual_search` logging behavior.

Fixes #12219

### DIFF
--- a/medusa/providers/generic_provider.py
+++ b/medusa/providers/generic_provider.py
@@ -15,6 +15,8 @@ from os.path import join
 
 from dateutil import parser, tz
 
+import guessit
+
 from medusa import (
     app,
     config,
@@ -44,6 +46,11 @@ from medusa.name_parser.parser import (
     NameParser,
 )
 from medusa.search import FORCED_SEARCH, PROPER_SEARCH
+from medusa.search.release_matcher import (
+    ReleaseMatcher,
+    extract_strong_numbering,
+    has_explicit_non_video_extension,
+)
 from medusa.session.core import ProviderSession
 from medusa.show.show import Show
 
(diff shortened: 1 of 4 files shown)


Mika3578:
Closing draft PR to validate fixes on fork first. Will reopen upstream PR after internal review and CI on Mika3578/Medusa.

PR #12221, opened 2026-06-12
fix: match non-standard manual search releases by episode title

### LINKED ISSUE #12219: Contextual release matching for manual search when numbering is missing or unreliable

Mika3578:
## Summary

Allow single-episode manual searches to match valid provider results whose release names do not contain reliable `SxxExx` or `1x01` numbering.

GuessIt 4.1.0 / Rebulk 6.x still parse these names as title-only results with no season or episode. This PR does not add GuessIt rules. It uses the series and requested-episode context already available during a manual search.

## Changes

- Add a contextual matcher for manual searches targeting one episode.
- Keep strict parsing and reliable explicit numbering authoritative.
- Fall back to normalized series and episode-title matching when reliable numbering is absent.
- Treat episode numbers extracted without an explicit TV-numbering pattern as weak.
- Reject releases whose explicit numbering conflicts with the requested episode.
- Reject releases whose explicit bare `Season N` / `Saison N` conflicts with the requested season (`explicit_season_conflict`). Episode title and matching dates never override that veto.
- Support scene numbering and relevant scene aliases.
- Reject ambiguous episode-title matches and releases belonging to another series.
- Prevent unnumbered individual releases from being treated as implicit season packs.
- Reject explicit non-media filenames while preserving processable RAR releases.
- Disable the global NameParser cache in the contextual path to prevent cross-series cache contamination.

## Scope

The contextual fallback is enabled only for:

- manual searches
- exactly one requested episode
- `eponly` search mode
- manual search type `episode`

Unchanged:

- automatic, daily and backlog searches
- proper searches
- season searches
- post-processing and library scanning
- global series lookup and general NameParser behavior
- GuessIt rules

The Post-Processing → Guessit → Test Release Name UI is out of scope. It calls `NameParser().parse(release)` without series or requested-episode context, so an unnumbered name such as `Example.Documentary.The.Hidden.Story.mp4` can still fail there. This PR covers the manual episode-search path, not that diagnostic screen.

## Matching rules

- Reliable explicit numbering matching the requested episode is accepted.
- Reliable explicit numbering conflicting with the requested episode is rejected. The episode title never overrides that conflict.
- Explicit bare `Season N` / `Saison N` is a season constraint only: it is never enough to accept a release by itself, and a conflicting value rejects even when the requested episode title is present.
- Title matching is used only when the series and requested episode title are unambiguous.
- Matching uses normalized contiguous token sequences, not fuzzy similarity (`all(token in release_tokens)` is not used).
- Series-before-episode and episode-before-series orders are both accepted.
- Prefixes, suffixes, Unicode/diacritic folding, apostrophes, and punctuation are normalized.
- Wrong series, wrong episode title, interleaved series tokens, series-only, and episode-only names are rejected.
- Weak unrelated numeric suffixes may be ignored only after a unique title match.
- Shared recording/airdate stamps are not treated as episode identity.
- An unnumbered result is never converted into a season pack without an explicit pack marker.

## Validation

Generic fixtures only. No production series names, episode titles, filenames, or magnets.

Coverage includes:

- unnumbered `series + episode title`
- reversed episode-title / series order
- Unicode, combining marks, apostrophes, and punctuation
- wrong series / wrong episode rejection
- interleaved token rejection
- series-only and episode-only rejection
- explicit numbering match and explicit numbering conflict
- bare `Season`/`Saison` match with requested title
- bare `Season`/`Saison` conflict not overridden by episode title
- same-date different-season rejection
- prefix/suffix around `Saison` + title
- ambiguous duplicate episode titles
- parser-cache isolation
- provider integration
- preservation of the non-manual strict parsing path

```bash
python -m pytest \
  tests/search/test_release_matcher.py \
  tests/providers/test_generic_provider.py \
  tests/name_parser \
  -q

python -m compileall \
  medusa/search/release_matcher.py \
  medusa/providers/generic_provider.py

git diff --check
```

Fixes #12219


Mika3578:

It fixes #12219 by adding a narrowly scoped fallback for single-episode manual searches when a valid release has recognizable series and episode titles but no reliable TV numbering.

Strict parsing remains authoritative whenever explicit numbering is present. Automatic searches, backlog searches, season searches, proper searches, post-processing, and the general parser behavior are unchanged.

Validation covers conflicting and ambiguous matches, scene numbering and aliases, Unicode normalization, parser-cache isolation, provider integration, non-media files, and processable RAR releases. Runtime testing also confirmed that valid non-standard releases are assigned to the requested episode.

All upstream CI workflows are passing.

medariox:
Guessit just released a new version that fixes a bunch of issues. Would you like to update the guessit verison first, to see if it solves the issue? I'm a bit afraid of adding more guessit rules, as it makes the parser slower.

Mika3578:
Thanks for the feedback.

Yes, I can test the latest GuessIt first and compare it against the failing manual-search cases from #12219.

One thing I noticed is that Medusa currently vendors GuessIt 3.4.2 and pins rebulk 3.2.0, while GuessIt 4.x appears to require Python >= 3.10 and rebulk 6. Since the current backend workflow still includes Python 3.9, the GuessIt upgrade may be a broader dependency-compatibility change rather than a drop-in update.

I’ll test the affected release names with GuessIt 4.1.0 separately and report whether it fully fixes the issue. If it does, we can probably prefer a dedicated GuessIt upgrade PR. If it only fixes part of the problem, or if the dependency upgrade is too broad, this PR can remain as a narrowly scoped fallback for single-episode manual searches only.

Also, this PR does not add global GuessIt rules; the fallback is limited to manual single-episode searches. But I agree that we should verify whether the new GuessIt release makes some or all of this unnecessary before merging it.

Mika3578:
## Follow-up on GuessIt 4, and a missing test case

@medariox GuessIt 4 does **not** fix this class of names.

The failing releases have no `SxxExx` / `1x01` at all. GuessIt (3.x vendored and 4.x) parses them as a movie/title-only result. Example from a current manual search:

```
Ovnis.dossiers.reveles.ovnis.vs.avions.FR.mp4
title: Ovnis dossiers reveles ovnis vs avions
language: fr
season: None
episode: []
```

The provider search succeeded because the series uses a custom `%SN %EN` template. Medusa then discarded the hit in `NameParser` (`Unable to match ... to a series in your database`). Passing `series=` into `NameParser` would still be wrong: `_parse_series()` would treat the unnumbered file as a season pack.

This PR’s contextual matcher is still the right fix: it is not extra GuessIt rules, and it is limited to single-episode manual search.

### Gap in the current tests

The matcher’s NFKD token rules already accept this release (series `Ovnis : Dossiers Révélés` + episode `Ovnis vs avions`) and reject `... autre episode ...`. That behaviour is not pinned by a test. Suggested additions to `tests/search/test_release_matcher.py`:

1. Accented series title vs ASCII release: `Révélés` ↔ `reveles`.
2. Colon / punctuation: `Ovnis : Dossiers Révélés` vs `Ovnis.dossiers.reveles...`.
3. Unique episode title present, no numbering → accept.
4. Same series, different episode title → reject.
5. Apostrophes: `l'immigration` vs `l’immigration`.
6. `À` vs `A` in the episode title.

The PR branch is also behind `develop` (`mergeStateStatus: BEHIND`) and will need a rebase before merge. No review of the matcher design itself from this follow-up; the original scope still looks correct.

Mika3578:
## GuessIt 4 is now on `develop`

https://github.com/pymedusa/Medusa/pull/12239 merged today (GuessIt 4.1.0). That does **not** replace this matcher: unnumbered `%SN %EN` releases still have no `SxxExx` / `1x01` for GuessIt to recover.

This branch is behind current `develop` and needs a rebase/retest against that tree. The matcher should stay limited to manual single-episode search, keep strict parsing when explicit numbering is present, and gain generic fixtures only (series/episode/release names that are not production titles). I will not open a second matching PR.

Mika3578:
Update is on this branch (fast-forward, no force-push):

- merged current `develop` including GuessIt 4.1.0, no conflicts
- added generic matcher fixtures only (`Example Show`, `À la découverte`, apostrophes, colon punctuation, explicit numbering conflict, weak numbers, no-title / no implicit season pack)

`python -m pytest tests/search/test_release_matcher.py -q`: **40 passed** locally. CI on this new HEAD still needs to run against the GuessIt 4 tree.

Mika3578:
Synced with current `develop` (GuessIt 4.1.0 / Rebulk 6.x) and retested.

GuessIt 4.1.0 still parses unnumbered `series + episode title` names as title-only results without season/episode, so this contextual matcher remains necessary. It is still limited to single-episode manual search. GuessIt rules are unchanged.

Added generic regression coverage for:

- series + episode title without numbering
- reversed title/series order
- Unicode/diacritics, combining marks, and apostrophe normalization
- wrong series / wrong episode rejection
- interleaved tokens, series-only, and episode-only rejection
- explicit numbering conflicts (title cannot override)
- parser-cache isolation
- non-manual strict behavior

No user-specific series, episode, or release names are included in the tests.

The Post-Processing → Guessit → Test Release Name UI is out of scope: it parses without series/episode context, so unnumbered names can still fail there.

Local tests: `tests/search/test_release_matcher.py` 56 passed; `tests/providers/test_generic_provider.py` 29 passed; `tests/name_parser` 45 passed. GitHub CI on this HEAD is running.

Mika3578:
@medariox Follow-up after rebasing this branch onto current `develop` (GuessIt 4.1.0 / Rebulk 6) and running a broader generic test matrix.

I tested **75 generic release-name variants** (no production/user filenames in the tests) across raw GuessIt, Medusa `NameParser`, the contextual `ReleaseMatcher`, and the provider integration path.

### What the matrix confirms

- Raw GuessIt identified season+episode in **3/75** cases: only the explicit-numbering controls (`SxxExx` / `1x05`).
- `NameParser` identified the same **3/75** cases.
- For the **72 unnumbered** names, GuessIt still returns a title-only/movie-style parse; Medusa's wrapper can enrich the type to `episode`, but `season` remains `None` and `episode` remains empty.
- Passing `series=` to `NameParser` is still not sufficient: an unnumbered name can fall into the season-pack path rather than identifying the requested episode.
- The contextual matcher accepted **59** cases and rejected **16**, with **0 mismatches** against the expected accept/reject outcomes in this matrix.
- `_apply_contextual_manual_parse()` agreed with `ReleaseMatcher` on all **75/75** cases.

### Generic forms verified

The contextual path correctly handles, without numbering:

- `Example.Documentary.The.Hidden.Story.mp4`
- `The.Hidden.Story.Example.Documentary.mp4` (episode title before series title)
- arbitrary prefixes/suffixes such as archive/channel/language/quality tokens
- punctuation differences
- Unicode NFC/NFD and diacritic folding (`Série` / `Serie`, `À` / `A`)
- straight vs typographic apostrophes

The matcher still rejects the important false-positive cases:

- wrong series + correct episode title
- correct series + wrong episode title
- interleaved/shuffled series tokens (not bag-of-words matching)
- series title only
- episode title only
- explicit numbering that conflicts with the requested episode
- duplicate episode titles in the library (`ambiguous_episode_title`)

Explicit correct numbering remains authoritative and stays on the strict path.

### `Test Release Name` UI

I also checked the `Post-Processing -> Guessit -> Test Release Name` path. It calls `NameParser().parse(release)` without a requested series/episode context, so it is expected to remain unable to resolve these unnumbered names.

That is separate from this PR's manual single-episode search fallback, where the target episode is already known.

### Remaining non-blocking hardening ideas

Two permissive cases are worth considering separately, but I don't think they should expand this PR's scope:

1. very short/common episode titles (`One`, `Home`) can still be accepted;
2. a filename containing two different known episode titles can match the requested one instead of being treated as a potential multi-episode/ambiguous release.

Those are hardening questions rather than failures of the original `series title + episode title without numbering` case.

Current PR HEAD is `ceccee3e0`; it is mergeable, and the current GitHub workflows (`Backend tests`, `API tests`, `Frontend tests`, and `Check app version`) are all green.

Mika3578:
## Broadcast/DVR case confirming the matcher

Audited against the contextual `ReleaseMatcher` with generic fixtures only.

### Target
Known series / `S19E01` / distinctive requested episode title

### Positive (MATCH → S19E01)
```text
Example.Show.-.Saison.19.-.Requested.Episode.Title_Channel.5_2025_07_18_21_00.mp4
```
`matched=True`, `method=episode_title`

### Negative (REJECT — same date family, wrong season/title)
```text
Example.Show.-.Saison.17.-.Other.Episode.Title_Channel.5_2025_07_18_21_50.mp4
```
`matched=False`

These two belong together: they show why date alone cannot identify the episode.

### Follow-up landed in `5cd3b8562`
Bare `Season N` / `Saison N` is now an explicit season constraint/veto (`explicit_season_conflict`). A matching episode title no longer overrides a conflicting season, e.g.:

```text
Example.Show.-.Saison.17.-.Requested.Episode.Title_Channel.5_2025_07_18_21_00.mp4
```
→ `REJECT` / `explicit_season_conflict`

Covered by the new generic broadcast fixtures in `tests/search/test_release_matcher.py`. No second matcher PR.

Mika3578:
@medariox Could you please review and merge this when you have a moment?

### Why this is still needed after GuessIt 4
GuessIt 4.1.0 (already on `develop`) still parses unnumbered `series + episode title` (+ optional airdate) releases as title-only results with no season/episode. Manual search finds the provider hit, then `NameParser` rejects it with "Unable to match … to a series".

Confirmed again today on a live air-by-date show after correcting the search template to `%SN %A-D`:
- provider returned a valid `Show.Name.Episode.Title.Words.DD.MM.YYYY.…` result
- Medusa discarded it on parse (GuessIt folded the episode title into `title`)
- this PR's contextual single-episode manual matcher is the path that would accept it

### Status
- Ready for review (no longer draft)
- CI green on the latest push
- Branch is up to date with `develop` (not behind)
- Scope unchanged: manual / single episode / `eponly` only; no new GuessIt rules; automatic/backlog/season/proper/post-process paths untouched

Thanks!

PR #12222, opened 2026-06-12
fix: evict expired TTL cache entries

### LINKED ISSUE #12187: Massive memory leaks when using torrent automated download handling
**Describe the bug**
When using a torrent client, either qbittorrent or utorrent, and then enabling the download handler, the memory used by Medusa is leaking massively over time. 

**To Reproduce**
1. Use a qbittorrent or utorrent to handle torrents
2. Enable the `Enable download handler` option in post-processing
3. ...
4. Memory leak massively

**Expected behavior**
No memory leaks. 

**Screenshots**
I've profiled the app using the Datadog continuous profiler, and it clearly identified the QBittorrentAPI as an offender.

Image

 - OS: ubuntu 24.04 x64
 - Branch: 1.0.25
 - Commit:  27e13769e83197d8e4ba7145f964384058592d4a

**Debug logs (at least 50 lines):**
Not relevant here.

**Additional context**
I asked claude about why it is leaking, here is its answer:

<details>
Root cause

The leak is in the interaction between @ttl_cache(60.0) (ext/ttl_cache.py) and the fact that _get_torrents is an instance method of QBittorrentAPI.

How ttl_cache works (ext/ttl_cache.py:30-55):
- The decorator creates a single dict _tmp = {} in its closure at class-definition time.
- That dict lives for the entire process lifetime — entries are never evicted; the TTL only controls whether a cached value is returned as fresh.
- Cache key is built from _hash(args) + _hash(kwargs). For methods, args[0] is self.

Why that leaks:
1. QBittorrentAPI has no custom __hash__/__eq__, so it uses identity-based hashing. hash(self) succeeds, and self is stored as-is in the cache key tuple.
2. That means the module-level _tmp dict holds a strong reference to every QBittorrentAPI instance ever created, via the key.
3. Each cached value is self.response.json() — the full torrent list (potentially thousands of dicts), plus the dict holds self, which transitively keeps the requests.Session, cookies, and self.response alive.
4. A fresh QBittorrentAPI() is instantiated on every scheduler tick — see medusa/schedulers/download_handler.py:404, medusa/search/core.py:173, medusa/process_tv.py:1047, medusa/server/web/home/handler.py:131. None of them
reuse an instance.
5. The download handler runs on a repeating schedule, so every tick adds a new (self, filter, category, sort) → (deadline, full_torrent_list) entry to _tmp, and nothing is ever removed.

Bonus pathology: because self is part of the key and changes every tick, the cache never produces a hit across ticks — it only helps within a single instance's lifetime (e.g., when _torrent_properties is called repeatedly for multiple hashes in one scheduler run). So the cache is simultaneously leaking and failing at its one job.

Proposed fix

Make the cache per-instance instead of module-level. The current cross-instance behavior has no benefit (different self → different key → miss), so we aren't losing anything, and each instance is naturally GC'd when the scheduler tick returns.

This preserves intra-tick caching (the only caching that was actually working), and the _torrents_cache dict dies with the instance at the end of the scheduler tick.

Alternative (broader fix)

If you'd rather keep the decorator in use across the codebase, the other option is to fix ext/ttl_cache.py itself — but that's more invasive because the cache is used elsewhere, and making it weakref-keyed on self is fiddly (you'd need to special-case the first arg of bound methods). The per-instance fix above is scoped to the one function known to leak and is the minimal-risk change.

utorrent.py:251 has the same pattern on an instance method — almost certainly leaks identically. series.py:624/633 with a 6-hour TTL on Series instance methods is the most worrying: Series objects are heavy, and a 6h TTL in a long-running process will pin many of them. The nzbget/sab ones are decorated on module-level functions, so those are fine.
</details>

Mika3578:
Fixes #12187

## Summary

The TTL cache retained every expired key indefinitely. When used on instance
methods, each key held a strong reference to the client instance and its cached
response.

Expired entries are now evicted on access so temporary torrent-client instances
and their torrent lists can be garbage-collected.

## Scope

- `ext/ttl_cache.py`
  - evict expired entries on access;
  - preserve `ignore_error=True` by restoring the stale entry only when a
    refresh attempt fails.
- `tests/test_ttl_cache.py`
  - add coverage for normal cache behavior;
  - verify expired-value eviction and garbage collection;
  - cover repeated `ignore_error` refresh failures;
  - verify qBittorrent `_get_torrents` cache-key isolation.

No scheduler or torrent-client architecture changes are included.

## Validation

- `python -m pytest tests/test_ttl_cache.py -v` — 10 passed
- `python -m pytest tests/clients -q` — 18 passed
- `python -m compileall ext/ttl_cache.py` — passed
- `git diff --check` — passed

Manual smoke test against a real qBittorrent WebAPI:

- authentication returned HTTP 204;
- `torrents/info` returned HTTP 200.

The garbage-collection tests reproduce the retention pattern with temporary
qBittorrent client instances and verify that expired instances and cached
responses are released after a subsequent cache access.

## Risk

Low. The behavior of valid, unexpired cache entries is unchanged.

With `ignore_error=True`, an expired value is restored after a failed refresh.
This intentionally preserves the previous behavior so repeated failures continue
returning the last known good value.

## Rollback

Revert this PR. No database migration or configuration change is required.

### DIFF
--- a/ext/ttl_cache.py
+++ b/ext/ttl_cache.py
@@ -31,22 +31,34 @@ def cache(ttl, typed=False, ignore_error=False, random_base=0):
 
         @functools.wraps(fn)
         def fn_wrapped(*args, **kwargs):
-            result = _tmp  # fake identifier
             now = monotonic()
+
+            def _evict_expired():
+                for expired_key, (expired_cd, _) in list(_tmp.items()):
+                    if expired_cd <= now:
+                        del _tmp[expired_key]
+
             key = _hash(args) + _hash(kwargs)
             if typed:
                 key += tuple(map(type, args))
-            if key in _tmp:
-                cd, result = _tmp[key]
+
+            stale_entry = _tmp.get(key)
+
+            if stale_entry:
+                cd, result = stale_entry
                 if cd > now:
+                    _evict_expired()
                     return result
+
+            _evict_expired()
 
             try:
                 result = fn(*args, **kwargs)
             except Exception:
-                if ignore_error and result is not _tmp:
-                    return result
-                raise  # no previous result in _tmp
+                if ignore_error and stale_entry is not None:
+                    _tmp[key] = stale_entry
+                    return stale_entry[1]
+                raise
 
             cd = now + ttl + random.random() * random_base
             _tmp[key] = cd, result
--- a/tests/test_ttl_cache.py
+++ b/tests/test_ttl_cache.py
@@ -1,0 +1,247 @@
+# coding=utf-8
+"""Tests for ext.ttl_cache."""
+from __future__ import unicode_literals
+
+import gc
+import time
+import weakref
+
+from medusa.clients.torrent.qbittorrent import QBittorrentAPI
+
+import pytest
+import ttl_cache
+
+
+class CacheValue(object):
+    """Simple object that supports weak references in cache tests."""
+
+
+def test_identical_calls_within_ttl_execute_once():
+    calls = []
+
+    @ttl_cache(60.0)
+    def cached(value):
+        calls.append(value)
+        return value
+
+    assert cached(1) == 1
+    assert cached(1) == 1
+    assert calls == [1]
+
+
+def test_call_after_expiration_recalculates():
+    calls = []
+
+    @ttl_cache(0.05)
+    def cached(value):
+        calls.append(value)
+        return value
+
+    assert cached(1) == 1
+    time.sleep(0.1)
+    assert cached(1) == 1
+    assert calls == [1, 1]
+
+
+def test_expired_cached_values_are_evicted():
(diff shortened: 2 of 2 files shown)


Mika3578:

It fixes #12187 by evicting expired TTL-cache entries so temporary torrent-client instances and their cached responses can be garbage-collected.

The change is limited to ext/ttl_cache.py and its regression tests. It preserves the existing ignore_error=True behavior by restoring the last valid value only when a refresh fails. No scheduler or torrent-client architecture changes are included.

Validation completed:

10 TTL-cache tests pass;
18 torrent-client tests pass;
garbage-collection tests verify that expired clients and responses are released;
a qBittorrent Web API smoke test returned successful authentication and torrent-list responses;
all upstream CI workflows are passing.

PR #12271, opened 2026-08-12
Fix: episode_updater flips network-less shows to WANTED before their airdate

### LINKED ISSUE #12270: EpisodeUpdater flips network-less shows to WANTED up to 2 days before their airdate
**Describe the bug**

For shows with no `network` value set (e.g. shows indexed via IMDb, which never
supplies a network name), `EpisodeUpdater.run()` in
`medusa/schedulers/episode_updater.py` can flip an episode's status to
`WANTED` up to two days before it actually airs, causing backlog/daily/manual
searches to hunt for a release that doesn't exist yet.

**Root cause**

1. `EpisodeUpdater.run()` widens its DB lookahead window from 1 day to 2 days
   whenever the module-level `network_dict` (from `medusa.network_timezones`)
   is falsy:
   ```python
   cur_date = (
       date.today() + timedelta(days=1 if network_dict else 2) + ...
   ).toordinal()
   ```
   `network_dict` starts as `None` and is only populated by
   `load_network_dict()`, which `update_network_dict()` only calls when the
   remote reference list differs from the local `cache.db` copy. Once the
   cache matches the remote list (steady state), `load_network_dict()` is
   never called, so `network_dict` can stay `None` for the life of the
   process — silently using the wider 2-day window on every run.

2. This 2-day window is normally harmless because a second, per-episode check
   further down uses the show's real `airs`/`network` values to compute the
   actual air time and veto premature flips:
   ```python
   if series_obj.airs and series_obj.network:
       ...
       if end_time + timedelta(hours=series_obj.airdate_offset) > cur_time:
           continue
   ```
   But when `series_obj.network` is empty/unset (true for shows indexed via
   IMDb), this whole check is skipped — there's no fallback validation, so
   any episode whose airdate falls inside the (already too-wide) window gets
   flipped straight to `WANTED`, regardless of how far in the future it
   actually is.

**To Reproduce**

1. Add a show indexed via IMDb (no `network` set in `tv_shows`).
2. Have an upcoming episode whose airdate falls within 2 days from today.
3. Let `EpisodeUpdater` run (daily scheduler) — the episode is set to
   `WANTED` even though it hasn't aired yet, and searches start immediately.

Observed in the wild: "Stuart Fails to Save the Universe" S01E04 (airdate
2026-08-13) was flipped to `WANTED` at 2026-08-11 00:11:49, two full days
early.

**Expected behavior**

Episodes should not be marked `WANTED` before their actual calendar airdate
has arrived, regardless of whether precise `airs`/`network` timing data is
available.

 - OS: Linux
 - Branch/Commit: develop

DavidHunterJS:
## Summary

Fixes #12270 — shows with no `network` value set (e.g. IMDb-indexed shows)
could have episodes flipped to `WANTED` up to two days before their actual
airdate, because `EpisodeUpdater`'s per-episode "hasn't finished airing yet"
safety check is entirely skipped when `series_obj.airs`/`series_obj.network`
aren't both set — there was no fallback validation at all in that branch.

## Fix

Add a fallback in `medusa/schedulers/episode_updater.py`: when precise
air-time data isn't available, only allow the WANTED flip once the episode's
plain calendar airdate has actually arrived, instead of relying solely on the
scheduler's intentionally forward-looking (1-2 day) DB lookahead window.

## Test plan

- [x] Reproduced live: "Stuart Fails to Save the Universe" S01E04 (airdate
      two days out) was set to `WANTED` by `EpisodeUpdater`, confirmed via
      logs and DB state.
- [x] Applied the fix, restarted the service, confirmed clean startup with no
      errors.
- [x] Confirmed the affected episode's status stays `UNAIRED` after the
      restart with the fix in place.

### DIFF
--- a/medusa/schedulers/episode_updater.py
+++ b/medusa/schedulers/episode_updater.py
@@ -110,6 +110,12 @@ class EpisodeUpdater(object):
                 # filter out any episodes that haven't finished airing yet
                 if end_time + timedelta(hours=series_obj.airdate_offset) > cur_time:
                     continue
+            else:
+                # Without airs/network info (e.g. IMDb-indexed shows) we can't compute a precise
+                # air time, so fall back to a plain calendar-date check instead of relying solely
+                # on the (intentionally forward-looking) cur_date window used for the DB query.
+                if db_episode['airdate'] > date.today().toordinal():
+                    continue
 
             with cur_ep.lock:
                 cur_ep.status = series_obj.default_ep_status if cur_ep.season else common.SKIPPED


medariox:
Thanks!

PR #12113, opened 2025-11-16
Fix unable to determine image type for jpegs

samuelrobyr:
- [X] PR is based on the DEVELOP branch
- [X] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [X] Read the [contribution guide](https://github.com/pymedusa/Medusa/blob/master/.github/CONTRIBUTING.md)

Fix "Unable to determine image type for JPEG IMAGE" (issue #12111)

PR #12274, opened 2026-08-17
Fix JPEG image size parsing and dimension unpacking

### LINKED ISSUE #12111: Unable to determine image type for JPG image
**Describe the bug**
Medusa v1.0.25 does not properly retrieve JPEG image size on some files.

**To Reproduce**
1. Add the attached image (banner.jpg) to a show
2. Open the show page
3. Click on "Re-scan files"
4. WARNING will appear in the log

```
2025-11-14 18:27:50 WARNING	SHOWQUEUE-REFRESH :: [27e1376] Unable to determine image type for /path/to/Sheriff.Country/banner.jpg
2025-11-14 18:27:50 WARNING	SHOWQUEUE-REFRESH :: [27e1376] Aspect ratio (31.142857142857142) does not match any known types.
```

After some digging, I found two bugs:

**BUG 1**: The function **get_image_size(image_path)** in **medusa/helpers/__init__.py** does not properly parse the JPEG. After finding the SOF (Start of Frame) marker, the code incorrectly skips bytes, causing it to read the segment length field as part of the image dimensions:

```
# JPEG check
elif head.startswith(b'\xff\xd8') or img_ext in ('jpg', 'jpeg'):
    f.seek(0)
    size = 2
    ftype = 0
    while True:
        f.seek(size, 1)
        byte = f.read(1)
        if not byte:
            return None  # EOF reached unexpectedly
        while ord(byte) == 0xff:
            byte = f.read(1)
        ftype = ord(byte)
        if 0xc0 <= ftype <= 0xcf and ftype != 0xc4 and ftype != 0xc8 and ftype != 0xcc:
            break # Found SOF marker
        size_bytes = f.read(2)
        if len(size_bytes) != 2:
            return None
        size = struct.unpack('>H', size_bytes)[0] - 2
    
    # --- THIS IS THE FIX ---
    # We are currently positioned after the SOF marker (e.g., 0xC0).
    # We must skip the 2-byte segment length and 1-byte precision.
    f.seek(3, 1)  # Skip 3 bytes (Length[2] + Precision[1])
    
    # Now we can read the height and width
    height, width = struct.unpack('>HH', f.read(4))
    return width, height
````

**BUG 2**:  The function **def which_type(path)** in **/medusa/image_cache.py** invert height and width.

```
image_dimension = get_image_size(path)
if not image_dimension:
    log.debug('Skipping image. Unable to get metadata from {0}', path)
    return

# --- THIS IS THE FIX ---
# get_image_size returns (width, height), not (height, width)
width, height = image_dimension 
```

Image

arrrghhh:
## Description

Fix two bugs in the image type detection pipeline that cause all local JPEG files to fail aspect ratio checks, making Medusa fall back to the default logo for banners/posters/fanart.

### Bug 1: `get_image_size()` JPEG SOF parser

After finding the SOF marker, `f.seek(1, 1)` skips only 1 byte (precision), but the 2-byte segment length field sits between the marker type byte and the precision byte. This causes the length to be misinterpreted as part of the image dimensions.

For a typical 3-component baseline JPEG, the SOF length is 17 (0x0011), which gets combined with the precision byte (0x08) to produce a bogus dimension of 4360 instead of the actual height/width.

**Fix:** `f.seek(1, 1)` -> `f.seek(3, 1)` to correctly skip the 2-byte length + 1-byte precision.

### Bug 2: `which_type()` dimension unpacking

`get_image_size()` returns `(width, height)` (as documented in its docstring and implemented for PNG/GIF), but `which_type()` unpacks it as `height, width`, swapping the values.

**Fix:** `height, width = image_dimension` -> `width, height = image_dimension`

## Reproducing

1. Add a show with a local `banner.jpg` in the show directory
2. Trigger a show refresh
3. Observe: `WARNING Unable to determine image type for .../banner.jpg` and `WARNING Aspect ratio (31.142857142857142) does not match any known types.`

The bogus aspect ratio 31.14 = 4360 / 140 where 4360 is (0x11 << 8) | 0x08 (SOF length + precision).

## Fix Verification

Tested with a synthetic 758x140 JPEG:
- **Before fix:** `get_image_size()` returns (140, 4360) -> aspect ratio 0.03
- **After fix:** `get_image_size()` returns (758, 140) -> aspect ratio 5.41 (correct for banner)

## Related Issue

Fixes #12111

### DIFF
--- a/medusa/helpers/__init__.py
+++ b/medusa/helpers/__init__.py
@@ -1523,7 +1523,7 @@ def get_image_size(image_path):
                 if len(size_bytes) != 2:
                     return None
                 size = struct.unpack('>H', size_bytes)[0] - 2
-            f.seek(1, 1)  # skip precision byte
+            f.seek(3, 1)  # skip 2-byte length and 1-byte precision
             height, width = struct.unpack('>HH', f.read(4))
             return width, height
         else:
--- a/medusa/image_cache.py
+++ b/medusa/image_cache.py
@@ -141,7 +141,7 @@ def which_type(path):
         log.debug('Skipping image. Unable to get metadata from {0}', path)
         return
 
-    height, width = image_dimension
+    width, height = image_dimension
     if not width or not height:
         log.debug('Skipping image. zero width or height {0}', path)
         return


arrrghhh:
Duplicate of #12113 which was already merged. Closing.

PR #12305, opened 2026-09-05
Throttle backlog searches into bounded, resumable batches

Mika3578:
## Summary

A scheduled full backlog currently queues all wanted work at once. On large
libraries this can create a burst of provider searches that slow or
single-slot providers cannot absorb.

This adds optional event-driven backlog batching while keeping the existing
behavior by default (`backlog_batch_size = 0`).

## Design

* bounded batches of deterministic `(indexer, series_id, season)` segments
* event-driven refill when pending backlog work reaches the configured threshold
* completion-based persisted cursor
* at-least-once restart semantics so interrupted work is repeated, never skipped
* final queue drain before a backlog pass is considered complete
* locked pending-work accounting
* consistent configuration normalization across config load, API v2 and web UI
* DB migration `44.20` for `info.backlog_cursor`

Daily searches, targeted per-show backlog searches and manual search queues
remain immediate.

Limited/forced backlog passes may use batching but do not persist the resume
cursor.

## Configuration

```text
backlog_batch_size
```

Default:

```text
0
```

`0` preserves the current unlimited behavior.

```text
backlog_batch_refill_threshold
```

Controls when the next batch is queued.

When batching is enabled:

```text
0 <= threshold < batch_size
```

## Validation

```text
python -m pytest tests -q --ignore=tests/legacy
1230 passed, 2 skipped, 1 xfailed
```

Focused backlog / API / web-config / search-core / show-queue:

```text
96 passed
```

## Risk / compatibility

* disabled by default
* no behavior change for existing installations unless batching is enabled
* DB migration adds one text cursor field on `info`
* interrupted work uses conservative at-least-once replay

## Follow-up

Out of scope for this PR: the pre-existing `BacklogSearcher.forced` flag is
never cleared after a forced run (upstream behavior unchanged).


Mika3578:
@medariox Ready for review.

Fork PR: https://github.com/Mika3578/Medusa/pull/40
Upstream branch is the squashed counterpart of the fork work. CI was green on the fork HEAD; no open review threads.

